PR #11, opened 2015-09-03
Qualimap integration - initial PR

guillermo-carrasco:
So, this adds the module for qualimap and the coverage histogram plot, it looks like this now:

screen shot 2015-09-03 at 15 52 52

More plots need to be added, but let's iterate on this PR by now :)

### DIFF
--- a/multiqc/qualimap/__init__.py
+++ b/multiqc/qualimap/__init__.py
@@ -1,0 +1,26 @@
+#!/usr/bin/env python
+
+""" MultiQC module to parse output from QualiMap """
+
+from __future__ import print_function
+import logging
+import os
+
+import multiqc
+
+class MultiqcModule(multiqc.BaseMultiqcModule):
+
+    def __init__(self, report):
+
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__()
+
+        # Static variables
+        self.name = "QualiMap"
+        self.anchor = "qualimap"
+        self.intro = '<p><a href="http://qualimap.bioinfo.cipf.es/" target="_blank">QualiMap</a> \
+             is a platform-independent application written in Java and R that provides both \
+			 a Graphical User Inteface (GUI) and a command-line interface to facilitate \
+			 the quality control of alignment sequencing data and its derivatives like feature counts. .</p>'
+        self.analysis_dir = report['analysis_dir']
+        self.output_dir = report['output_dir']
--- a/multiqc/qualimap/__init__.py
+++ b/multiqc/qualimap/__init__.py
@@ -3,6 +3,7 @@
 """ MultiQC module to parse output from QualiMap """
 
 from __future__ import print_function
+import json
 import logging
 import os
 
@@ -24,7 +25,6 @@ class MultiqcModule(multiqc.BaseMultiqcModule):
              the quality control of alignment sequencing data and its derivatives like feature counts.</p>'
         self.analysis_dir = report['analysis_dir']
         self.output_dir = report['output_dir']
-        self.data_dir = os.path.join(self.output_dir, 'report_data', 'qualimap')
 
         # Find QualiMap reports
         qualimap_raw_data = {}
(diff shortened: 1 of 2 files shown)


multiqc/qualimap/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__()
+
+        # Static variables
+        self.name = "QualiMap"
+        self.anchor = "qualimap"
+        self.intro = '<p><a href="http://qualimap.bioinfo.cipf.es/" target="_blank">QualiMap</a> \
+             is a platform-independent application written in Java and R that provides both \
+             a Graphical User Inteface (GUI) and a command-line interface to facilitate \
```
ewels: You can probably trim this down a bit if you want - GUI stuff is a bit irrelevant :sleeping:

multiqc/qualimap/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+             the quality control of alignment sequencing data and its derivatives like feature counts.</p>'
+        self.analysis_dir = report['analysis_dir']
+        self.output_dir = report['output_dir']
+
+        # Find QualiMap reports
+        qualimap_raw_data = {}
+        for root, dirnames, filenames in os.walk(self.analysis_dir, followlinks=True):
+            if 'genome_results.txt' in filenames and 'raw_data_qualimapReport' in dirnames:
+                with open(os.path.join(root, 'genome_results.txt'), 'r') as gr:
```
ewels: I've been converting all of the other modules to use the `io` module for better encoding / python 3 support.

_eg:_
```python
with io.open(os.path.join(root, 'genome_results.txt'), 'r', encoding='utf-8') as gr:
```

multiqc/qualimap/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+        logging.info("Found {} QualiMap reports".format(len(qualimap_raw_data)))
+
+        self.sections = list()
+
+        # Section 1 - Coverage Histogram
+        histogram_data = self.qualimap_cov_his(qualimap_raw_data)
+        if len(histogram_data) > 0:
+            self.sections.append({
+                'name': 'Coverage Histogram',
+                'anchor': 'coverage-histogram',
```
ewels: This is for the HTML anchor id, so the more unique the better. Maybe prefix with `qualimap-`, just in case some other tool also generates a coverage histogram?

multiqc/qualimap/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+
+                except IOError as e:
+                    logging.error("Could not load input file: {}".format(cov_report))
+                    raise
+                parsed_data[sn] = counts
+
+        return parsed_data
+
+
+    def qualimap_cov_his_plot(self, parsed_data):
```
ewels: I wonder if I should make a generalised Python plotting function in the base class? Might make sense as these are starting to conform to a common pattern now and would make life easier for simple plots such as this.

multiqc/qualimap/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+        html = '<div id="qualimap_cov_hist" class="hc-plot"></div> \n\
+                <script type="text/javascript"> \
+                    var cov_pconfig = {{ \n\
+                        "title": "Coverage Histogram",\n\
+                        "ylab": "Genome Bin Counts",\n\
+                        "xlab": "Coverage (X)",\n\
+                        "ymin": 0,\n\
+                        "xmin": 0,\n\
+                        "tt_label": "<b>{{point.x}}-X coverage </b>",\n\
+                        "use_legend": false,\n\
```
ewels: I should totally make this the default. I don't think that there are any plots that have a legend so far. Maybe even not make an option.

multiqc/qualimap/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+            for k, p in iter(sorted(parsed_data[s].items())):
+                pairs.append([int(k), p])
+            data.append({
+                'name': s,
+                'data': pairs
+            })
+
+        html = '<div id="qualimap_cov_hist" class="hc-plot"></div> \n\
+                <script type="text/javascript"> \
+                    var cov_pconfig = {{ \n\
```
ewels: Also, again maybe some kind of `qualimap-` prefix here wouldn't go amiss.

multiqc/qualimap/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+                        "title": "Coverage Histogram",\n\
+                        "ylab": "Genome Bin Counts",\n\
+                        "xlab": "Coverage (X)",\n\
+                        "ymin": 0,\n\
+                        "xmin": 0,\n\
+                        "tt_label": "<b>{{point.x}}-X coverage </b>",\n\
+                        "use_legend": false,\n\
+                    }}; \n\
+                    $(function () {{ \
+                        plot_xy_line_graph("#qualimap_cov_hist", {d}, cov_pconfig); \
```
ewels: I wish I could figure out a nice way to get the raw data out of the HTML :disappointed:

Not that it really matters I guess, but it just feels wrong!

multiqc/qualimap/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+            if cov_report:
+                counts={}
+                try:
+                    with open(cov_report, 'r') as fh:
+                        next(fh) # skip the header
+                        for line in fh:
+                            (coverage, count) = line.split(None, 1)
+                            coverage = int(round(float(coverage)))
+                            count = float(count)
+                            counts[coverage] = count
```
ewels: Could be nice to truncate this data maybe, try to reduce the long tail of the plot. Could chew back until you hit 1% of the total count or something?

Maybe not worth the effort though as the plot can be zoomed.

ewels:
Yeah man, a few minor comments - half of which are more out loud thoughts to myself really. Looks brilliant!

multiqc/qualimap/__init__.py
guillermo-carrasco: that's a good one :)

multiqc/qualimap/__init__.py
guillermo-carrasco: :+1:

multiqc/qualimap/__init__.py
guillermo-carrasco: yup, maybe we (you? :-P) should think about this, not only for the plotting, but also all the html code embedded in the code. Is not that I have a better solution right now, but seems like something that could maybe be templated?

ewels: Yeah, I mean all of the HTML code. I'll do this now after I merge your PR.

guillermo-carrasco:
oki! I pushed one more commit fixing the ID's and the `io.open` thing. Haven't worked on the plot. I also thought about that, so if you want me to work on that... will have to be next week :-D. Thanks for the review!

ewels:
Nice, looks good - thanks!

Issue #41, opened 2015-09-26
Allow multiple analysis directories to be given


ewels:
Allow multiple analysis directories to be given (positional, not flag arguments - see [this](http://stackoverflow.com/questions/7588620/os-walk-multiple-directories-at-once))

PR #44, opened 2015-09-28
Multiple directories and More logging

moonso:
- Set default logging to INFO
- Added ``-q/--quiet`` for less logging
- Allowed user to specify multiple directories to scan
- Changed way of importing version


scripts/multiqc
```diff
@@ -25,6 +25,7 @@ from multiqc.log import init_log, LEVELS
 @click.command()
 @click.argument('analysis_dir',
                     type=click.Path(exists=True),
+                    nargs=-1,
```
ewels: Nice! Could be good to add `required=True` (I was looking for the click alternative to `nargs=+`, requiring one or more arguments. Apparently this is it).

> **Note on Non-Empty Variadic Arguments**
> If you come from argparse, you might be missing support for setting nargs to + to indicate that at least one argument is required.
> This is supported by setting `required=True`. However, this should not be used if you can avoid it as we believe scripts should gracefully degrade into becoming noops if a variadic argument is empty. The reason for this is that very often, scripts are invoked with wildcard inputs from the command line and they should not error out if the wildcard is empty.

moonso:
Ah, forgot that, it's added now!

ewels:
Great! Thanks again..

ewels:
Closes #41

PR #148, opened 2016-02-17
Skewer

### LINKED ISSUE #147: Add skewer support
Skewer is a multithreaded tools for adapter trimming, https://github.com/relipmoc/skewer . 

Example output is available at: https://gist.github.com/dakl/e2e6dd0146a027c2834c

dakl:
This should close #147.

### DIFF
--- a/multiqc/modules/cutadapt/cutadapt.py
+++ b/multiqc/modules/cutadapt/cutadapt.py
@@ -117,7 +117,7 @@ class MultiqcModule(BaseMultiqcModule):
 
         headers = {}
         headers['percent_trimmed'] = {
-            'title': 'Trimmed',
+            'title': '% Trimmed',
             'description': '% Total Base Pairs trimmed',
             'max': 30,
             'min': 0,
--- a/multiqc/modules/skewer/skewer.py
+++ b/multiqc/modules/skewer/skewer.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-""" MultiQC module to parse output from QualiMap """
+""" MultiQC module to parse logs from Skewer """
 
 from __future__ import print_function
 
@@ -21,8 +21,8 @@ class MultiqcModule(BaseMultiqcModule):
     def __init__(self):
         # Initialise the parent object
         super(MultiqcModule, self).__init__(name='Skewer', anchor='skewer',
-                                            href="https://github.com/relipmoc/skewer.git",
-                                            info="is a tools to trim adapters off of reads")
+                                            href="https://github.com/relipmoc/skewer",
+                                            info="is an adapter trimming tool specially designed for processing next-generation sequencing (NGS) paired-end sequences.")
 
         self.sections = list()
         self.skewer_data = dict()
@@ -33,20 +33,22 @@ class MultiqcModule(BaseMultiqcModule):
 
         headers = OrderedDict()
         headers['r_processed'] = {
-            'title': 'Read Pairs',
-            'description': 'Total number of read pairs processed',
+            'title': 'M Read Pairs',
+            'description': 'Million read pairs processed',
             'modify': lambda x: x / 1000000.0,
             'min': 0,
             'scale': 'PuBu',
-            'format': '{:.0f}'
+            'format': '{:.0f}',
+            'shared_key': 'read_count'
         }
         headers['r_avail'] = {
-            'title': 'Kept',
-            'description': 'Read pairs available after trimming',
+            'title': 'M Kept',
+            'description': 'Million read pairs available after trimming',
             'modify': lambda x: x / 1000000.0,
             'min': 0,
             'scale': 'PuBu',
-            'format': '{:.0f}'
+            'format': '{:.0f}',
+            'shared_key': 'read_count'
         }
         headers['pct_trimmed'] = {
             'title': 'Pct Trimmed',
--- a/multiqc/modules/skewer/skewer.py
+++ b/multiqc/modules/skewer/skewer.py
@@ -72,10 +72,12 @@ class MultiqcModule(BaseMultiqcModule):
     def add_readlen_dist_plot(self):
         pconfig = {
             'id': 'skewer_read_length_histogram',
-            'title': 'Read Length Distribution',
-            'categories': True,
+            'title': 'Read Length Distribution after trimming with Skewer',
+            'categories': False,
+            'xDecimals': False,
             'ylab': '% of Reads',
             'xlab': 'Read Length',
+            'xmin': '0',
             'ymax': 100,
             'ymin': 0,
             'tt_label': '<b>{point.x}</b>: {point.y:.1f}%',
@@ -119,7 +121,7 @@ class MultiqcModule(BaseMultiqcModule):
 
             match = re.search(regex_hist, l)
             if match:
-                read_length = str(match.group(1))
+                read_length = match.group(1)
                 pct_at_rl = float(match.group(3))
                 readlen_dist[read_length] = pct_at_rl
 
--- a/multiqc/modules/skewer/skewer.py
+++ b/multiqc/modules/skewer/skewer.py
@@ -61,7 +61,7 @@ class MultiqcModule(BaseMultiqcModule):
             log.debug("Could not find any data in {}".format(config.analysis_dir))
             raise UserWarning
 
-        log.debug("Found {} reports".format(len(self.skewer_data)))
+        log.info("Found {} reports".format(len(self.skewer_data)))
 
     def parse_skewer_log(self, f):
         """ Go through log file looking for skewer output """
--- a/multiqc/modules/skewer/skewer.py
+++ b/multiqc/modules/skewer/skewer.py
@@ -4,7 +4,6 @@
 
 from __future__ import print_function
 
-import json
 import os
 from collections import OrderedDict
 import logging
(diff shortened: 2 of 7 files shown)


multiqc/modules/skewer/skewer.py
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env python
+
+""" MultiQC module to parse output from QualiMap """
```
ewels: Evidence of copy and pasting ;)

multiqc/modules/skewer/skewer.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class MultiqcModule(BaseMultiqcModule):
+    """ Skewer """
+
+    def __init__(self):
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__(name='Skewer', anchor='skewer',
+                                            href="https://github.com/relipmoc/skewer.git",
+                                            info="is a tools to trim adapters off of reads")
```
ewels: I usually take some form of the description used on the author's web page for this. So would be something like:

> is an adapter trimming tool specially designed for processing next-generation sequencing (NGS) paired-end sequences.

multiqc/modules/skewer/skewer.py
```diff
@@ -0,0 +1,112 @@
...
+log = logging.getLogger(__name__)
+
+
+class MultiqcModule(BaseMultiqcModule):
+    """ Skewer """
+
+    def __init__(self):
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__(name='Skewer', anchor='skewer',
+                                            href="https://github.com/relipmoc/skewer.git",
```
ewels: This should be a human readable URL, so https://github.com/relipmoc/skewer

multiqc/modules/skewer/skewer.py
```diff
@@ -0,0 +1,112 @@
...
+                                            info="is a tools to trim adapters off of reads")
+
+        self.skewer_data = dict()
+        for f in self.find_log_files(config.sp['skewer'], filehandles=True):
+            self.parse_skewer_log(f)
+
+        headers = OrderedDict()
+        headers['r_processed'] = {
+            'title': 'Read Pairs',
```
ewels: Should probably be `M Read Pairs` if we're dividing by one milliom

multiqc/modules/skewer/skewer.py
```diff
@@ -0,0 +1,112 @@
...
+            'description': 'Total number of read pairs processed',
```
ewels: As above, `Number of read pairs processed (Millions)`

multiqc/modules/skewer/skewer.py
```diff
@@ -0,0 +1,112 @@
...
+            'modify': lambda x: x / 1000000,
+            'min': 0,
+            'scale': 'RdYlGn',
+            'format': '{:.0f}'
```
ewels: We can add the `shared_key` attribute here as it's a read count. Described in the footer of [this section of the docs](http://multiqc.info/docs/#step-4-adding-to-the-general-statistics-table)

See the equivalent read count section in the STAR module [here](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/star/star.py#L116-L123)

dakl:
Good comments, thanks. I think my latest commit fixes them all.

multiqc/modules/cutadapt/cutadapt.py
```diff
@@ -117,7 +117,7 @@ def cutadapt_general_stats_table(self):
 
         headers = {}
         headers['percent_trimmed'] = {
-            'title': 'Trimmed',
+            'title': '% Trimmed',
```
ewels: :+1:

multiqc/modules/skewer/skewer.py
```diff
@@ -0,0 +1,148 @@
...
+            'scale': 'RdYlGn-rev',
+            'max': 100,
+            'min': 0
+        }
+
+        self.general_stats_addcols(self.skewer_data, headers)
+
+        # Write parsed report data to a file
+        self.write_data_file(self.skewer_data, 'multiqc_skewer')
+        self.write_data_file(self.skewer_readlen_dist, 'multiqc_skewer_readlen_dist')
```
ewels: So, interesting discussion point. In the cutadapt module (and others), I _don't_ export the continuous data that is parsed for the plots. This is vaguely because I was worried about it making very large files if 1000s of samples are processed. But if I'm honest I haven't thought about it a great deal.

On the one hand it's good to make the parsed data as accessible as possible for downstream applications, but on the other we don't want to bloat the reports too much for the 99% of people who aren't interested in this data.

Any thoughts?

multiqc/modules/skewer/skewer.py
```diff
@@ -0,0 +1,148 @@
...
+            'title': 'Read Length Distribution',
+            'categories': True,
+            'ylab': '% of Reads',
+            'xlab': 'Read Length',
+            'ymax': 100,
+            'ymin': 0,
+            'tt_label': '<b>{point.x}</b>: {point.y:.1f}%',
+        }
+
+        html_content = self.plot_xy_data(self.skewer_readlen_dist, pconfig)
```
ewels: Might be nice to add a sentence here explaining what the plot shows. Nothing fancy, eg:

> Length of reads remaining after trimming.

multiqc/modules/skewer/skewer.py
```diff
@@ -0,0 +1,148 @@
...
+            log.debug("Could not find any data in {}".format(config.analysis_dir))
+            raise UserWarning
+
+        log.info("Found {} reports".format(len(self.skewer_data)))
+
+    def add_readlen_dist_plot(self):
+        pconfig = {
+            'id': 'skewer_read_length_histogram',
+            'title': 'Read Length Distribution',
+            'categories': True,
```
ewels: I don't think you want this? The `x` axis is numeric, not categorical..

multiqc/modules/skewer/skewer.py
```diff
@@ -0,0 +1,148 @@
...
+            'ylab': '% of Reads',
+            'xlab': 'Read Length',
+            'ymax': 100,
```
ewels: Would it make sense to have the x axis start at 0? I only say this because most of the other plots (eg. FastQC) do this. Might look weird here I guess.

multiqc/modules/skewer/skewer.py
ewels: You'll also want `'xDecimals': False,` as you can't have half a base pair..

ewels:
Sorry for all of the comments, plus the delays in between them!

Overall this is awesome - all minor stuff, with no _"I want to rewrite this myself"_ urge on my side which is amazing as I'm a bit of a control freak :grin:

multiqc/modules/skewer/skewer.py
dakl: I think it makes sense to make it accessible to users to do other plots that we didn't think of here. Thats how why I wrote it. 

I see multiqc as two things in one: The first is a nice reporting tool which is great, but the second is a general collector of stats to enable other analyses downstream. I think the second is as important as the first. 

Even if a user has 1000s of samples, the amount of data written won't be that much.

ewels: :+1: I'm sure you're right. Will bear this in mind with other modules, though I think it is worth keeping an eye on the resulting file sizes. I regularly see MultiQC HTML reports which are 100mb or more with this same data (but less of an issue for the parsed data directory).

multiqc/modules/skewer/skewer.py
dakl: For some reason, setting the following doesn't actually change xmin in the graph. Adding a fixed `xmax` does changes the graph.

```python
        pconfig = {
            'id': 'skewer_read_length_histogram',
            'title': 'Read Length Distribution after trimming with Skewer',
            'categories': False,
            'xDecimals': False,
            'ylab': '% of Reads',
            'xlab': 'Read Length',
            'xmin': '0',
            'ymax': 100,
            'ymin': 0,
            'tt_label': '<b>{point.x}</b>: {point.y:.1f}%',
        }
```

ewels: Could it be because the `'0'` is quoted so behaving as a string instead of an `int`? I'll have a play myself soon..

ewels: Also, `False` is the default value for `categories`, so you can just delete that line..

multiqc/modules/skewer/skewer.py
dakl: Maybe a flag `--write-all-data` so it's user settable? And in the API something like `write_data_file(data, fn, is_extra=False)`? An example could then be

```        
        # Write parsed report data to a file
        self.write_data_file(self.skewer_data, 'multiqc_skewer')
        self.write_data_file(self.skewer_readlen_dist, 'multiqc_skewer_readlen_dist', is_extra=True)
```

(note `, is_extra=True` in the second call)

ewels: In fairness there's already the `--no-data-dir` if people want to disable it. I think it'll be fine. Can come back to this later if it becomes a problem..

dakl:
Glad to hear! :+1: :)

multiqc/modules/skewer/skewer.py
dakl: Good point!

multiqc/modules/skewer/skewer.py
```diff
@@ -0,0 +1,150 @@
...
+            for k, r in regexes.items():
+                match = re.search(r, l)
+                if match:
+                    data[k] = match.group(1).replace(',', '')
+
+            match = re.search(regex_hist, l)
+            if match:
+                read_length = match.group(1)
+                pct_at_rl = float(match.group(3))
+                readlen_dist[read_length] = pct_at_rl
```
ewels: This line needs to be:
```python
readlen_dist[int(read_length)] = pct_at_rl
```
Then the plot should work..

ewels:
Fantastic :clap:

PR #149, opened 2016-02-17
Samblaster

dakl:


ewels:
Yeah, all looks good! Only comment is that a plot might be a nice addition - we currently have the section header but nothing below it. I know it would be duplicating the general stats table a bit as we don't have much more info, but could allow switching between counts and percentages. Sometimes it's nice to have a standalone plot to share which isn't surrounded by other data in the table anyway.

Alternatively I could add a new option to not print the section if we don't want it, but then there's no reference to the original tool which I'd feel a bit uncomfortable about.

What do you think?

dakl:
What do you think of this?

ewels:
Looks great! Thanks again..

PR #156, opened 2016-03-14
Adding Samtools

lpantano:
Hi @ewels,

this is a super great, great, awesome package. Have i said super useful awesome great? :)

I am one of the developer of bcbio-nextgen (https://github.com/chapmanb/bcbio-nextgen). We put together tools, based on community knowledge that works the best with Variant calling, RNAseq, sRNAseq and chipseq data. 

I added multiqc now, but I would like to add `samtools stats` output here since if would be more independent of the alignment we used. 

As well, I modified feature count output to the default output (last version at least). Hope this is right, I can do it in out end, but I thought it would be better to have as recognition pattern the default output name.

Last thing, I modified was the cleaning function, because when used, it was keeping the whole path even it `-d` wasn't present. I think is best to leave the name file, since allows to have clean names as an option.

Let me know what do you think. I can split the pull request as well. I just thought it was easy with one. 

Thanks a lot in advance!


ewels:
Hi @lpantano,

This looks brilliant - thanks! I know bcbio well (a friend of mine, @guillermo-carrasco, also works on it). It's great news that you're integrating MultiQC! All of your changes sound good, I'll have a proper go through the code and do some testing tomorrow if that's ok.

Phil

ewels:
ps. If you have any test data for samtools which you could PR to the [MultiQC_TestData](https://github.com/ewels/MultiQC_TestData) repo, that would be really helpful..

lpantano:
Yep, sure. I saw some tag related to that, but didn't find the test data. Will do it! thanks a lot!

multiqc/modules/base_module.py
```diff
@@ -162,7 +162,7 @@ def clean_s_name(self, s_name, root):
             root = ''
         # Split then take first section to remove everything after these matches
         for ext in config.fn_clean_exts:
-            s_name = s_name.split(ext ,1)[0]
+            s_name = os.path.basename(s_name.split(ext ,1)[0])
```
ewels: I don't have any problems with this, but I'm surprised that it makes a difference.. The `s_name` usually comes from the `find_log_files` function which should supply the base filename. I've not seen directory paths without the `-d` option myself before, but if you were getting that then I guess this addition doesn't hurt.

multiqc/modules/featureCounts/feature_counts.py
```diff
@@ -29,10 +29,11 @@ def __init__(self):
         for f in self.find_log_files(config.sp['featurecounts']):
             parsed_data = self.parse_featurecounts_report(f['f'])
             if parsed_data is not None:
-                if f['s_name'] in self.featurecounts_data:
+                s_name = f['s_name'].replace(".counts.summary", "")
```
ewels: Might be nicer to add `.counts` to `fn_clean_exts` in `config.py` rather than hardcoding here? (I'm probably being a hypocrite, I'm sure I've done the same in other modules)

multiqc/utils/search_patterns.yaml
```diff
@@ -30,7 +30,7 @@ fastqc:
     zip:
         fn: _fastqc.zip
 featurecounts:
-    fn: _counts.txt.summary
+    fn: counts.summary
```
ewels: Instead of replacing the old version, maybe just keep both as an array instead?
```yaml
featurecounts:
    fn:
        - _counts.txt.summary
        - counts.summary
```
I think this should work, and keep backwards compatibility.

multiqc/utils/search_patterns.yaml
```diff
@@ -66,3 +66,5 @@ skewer:
     contents: 'COMMAND LINE:	skewer'
 samblaster:
     contents: 'samblaster: Version'
+samtools:
+    contents: 'samtools stats'
```
ewels: Can we try to make this a bit more specific? I have a feeling that lots of other files might end up having `samtools stats` in.. Maybe something like `# This file was produced by samtools stats` instead?

multiqc/modules/samtools/samtools.py
```diff
@@ -0,0 +1,134 @@
...
+
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__(name='samtools',
+        anchor='samtools', target='samtools',
+        href='http://www.htslib.org',
+        info="Samtools is a suite of programs for interacting"\
+        " with high-throughput sequencing data.")
+        self.samtools_data = dict()
+        for f in self.find_log_files(config.sp['samtools']):
+            s_name = f['s_name'].split(".txt")[0]
```
ewels: As above, might be nicer to just add `.txt` to `fn_clean_exts` in the config (I'm surprised it's not there already!)

multiqc/modules/samtools/samtools.py
```diff
@@ -0,0 +1,134 @@
...
+        # Basic Stats Table
+        # Report table is immutable, so just updating it works
+        self.samtools_stats_table()
+
+
+    def parse_samtools_report (self, raw_data):
+        """ Parse the samtools log file. """
+
+        regexes = {
+                'raw_total_seqs': r"raw\stotal\ssequences:\s+(\d+)",
```
ewels: Any reason to use `\s` here instead of just having a normal space? (not counting the `\s+`)

multiqc/modules/samtools/samtools.py
```diff
@@ -0,0 +1,134 @@
...
+            'reads_mapped': r"reads\smapped:\s+(\d+)",
+            'non_primary': r"non-primary\salignments:\s+(\d+)",
+            'reads_paired': r"reads\smapped\sand\spaired:\s+(\d+)",
+            'reads_duplicated': r"reads\sduplicated:\s+(\d+)",
+            'error_rate': r"error\srate:\s+(\d+\.\d+e-\d+)\s+#",
+            'pairs_on_diff_chrom': r"pairs\son\sdifferent\schromosomes:\s+(\d+)"
+        }
```
ewels: The test data file you uploaded looks really nice - loads of good numbers in there! This is totally up to you, but it could be good to parse _all_ of the stats in the `# Summary Numbers` section here, even if they're not used in the MultiQC report. They'll go into the MultiQC parsed data files and could help people with downstream analysis...

multiqc/modules/samtools/samtools.py
```diff
@@ -0,0 +1,134 @@
...
+        parsed_data = {}
+        for k, r in regexes.items():
+            r_search = re.search(r, raw_data, re.MULTILINE)
+            if r_search:
+                value = float(r_search.group(1))
+                if value > 0:
```
ewels: Why only if > 0?

multiqc/modules/samtools/samtools.py
```diff
@@ -0,0 +1,134 @@
...
+        return parsed_data
+
+
+    def samtools_stats_table(self):
+        """ Take the parsed stats from the samtools report and add them to the
+        basic stats table at the top of the report """
+
+        headers = OrderedDict()
+        headers['raw_total_seqs'] = {
+            'title': 'Total seqs',
```
ewels: Please could you add a `M` prefix if it's millions? `M Total Seqs`..

multiqc/modules/samtools/samtools.py
```diff
@@ -0,0 +1,134 @@
...
+            'description': 'Total sequences in the bam file',
```
ewels: As above, add on a `(millions)` suffix here if that's ok.

multiqc/modules/samtools/samtools.py
```diff
@@ -0,0 +1,134 @@
...
+            'min': 0,
+            'modify': lambda x: x / 1000000,
+            'shared_key': 'read_count'
+        }
+        headers['error_rate'] = {
+            'title': 'Error rate',
+            'description': 'Error rate using CIGAR',
+            'min': 0,
+            'max': 1,
```
ewels: I'm not sure how this will work with the `lambda` below.. I _think_ that it's used in the front end javascript for the colouring, in which case you'll want `100` instead.

multiqc/modules/samtools/samtools.py
```diff
@@ -0,0 +1,134 @@
...
+            'scale': 'OrRd',
+            'format': '{:.4f}%',
```
ewels: 4 decimal places seems a little excessive for a percentage? `83.2859%`?

multiqc/modules/samtools/samtools.py
```diff
@@ -0,0 +1,134 @@
...
+        headers['non_primary'] = {
+            'title': 'Non primary alignment',
+            'description': 'Non primary alignment (millions)',
+            'min': 0,
+            'scale': 'PuBu',
+            'modify': lambda x: x / 1000000,
+            'shared_key': 'read_count'
+            }
+        headers['reads_duplicated'] = {
+            'title': 'Duplicatd reads',
```
ewels: Typo: `Duplicated`

multiqc/modules/samtools/samtools.py
```diff
@@ -0,0 +1,134 @@
...
+            'title': 'Error rate',
+            'description': 'Error rate using CIGAR',
+            'min': 0,
+            'max': 1,
+            'scale': 'OrRd',
+            'format': '{:.4f}%',
+            'modify': lambda x: x * 100.0
+        }
+        headers['non_primary'] = {
+            'title': 'Non primary alignment',
```
ewels: Pl. `M Non-Primary Alignments`

ewels:
Hey @lpantano - just looked through the code, it looks brilliant! All my comments are really minor things, I can't think of any part of MultiQC that you're not using fully. Thanks for the contribution! I'll have a go at running it my end while you go through my comments, then merge.

ewels:
Was just thinking - it's nice to keep the number of columns to a minimum in the general stats table to prevent it from being too bloated (it can get very wide when there are several modules). However, you could get more of the stats into the report by adding a few bar graphs with the less important numbers. For example:
* Read alignments
  * `reads mapped and paired`
  * `reads mapped` ( - `reads mapped and paired`)
  * `reads unmapped`
* Base pairs plot
  * `bases duplicated`
  * `bases mapped (cigar)` ( - `bases duplicated`)
  * `bases trimmed`
  * `total length` ( - `bases mapped (cigar)`, `bases trimmed` & `bases duplicated`)
* Bad pairs plot
  * `inward oriented pairs`
  * `outward oriented pairs`
  * `pairs with other orientation`
  * `pairs on different chromosomes`
  * `reads paired` - above numbers?

..and so on. The tricky bit is getting the numbers to add up to the correct total when using a stacked bar, something that needs a bit of testing. These plots can be easier to read and gives a way to add in extra information to the report if present though.

Only an idea! Feel free to ignore.

ewels:
ps. Sorry for the mess with branching - I put myself onto a branch a few days ago as the stuff I was pushing to master was a bit unstable and I didn't want anyone to fork it. But I didn't check and see that you had already forked it, some seriously unlucky timing! Hope you haven't had any grief from it. I just merged my branch back to master so should all be simple now.

lpantano:
Hi!

thanks for the feedback. I agree on all your comments.

About the sample name, this happened with cutadapt for now. In that case, the s_name is got from the content file, exactly from the command line, getting the file name. If the file has the path, then it will appear. I could fix the cutadapt function, but i thought to add it in the general function in case it happens with others.

About the plots and tables, yes, I thought about that. Related to that, I wanted to ask if it would be easy to add tables instead of figures to the sections. Some times, tables are better, if you can color more, that is this case. But, I know that this could be a pain in the ass if the implementation wasn't thought for that. Just an idea.

I will work on the comments in the next two days.

Thanks!

ewels:
So in theory you can add whatever HTML you like to the report, including tables. I'm pretty sure that the code is fairly flexible with this formatting and that stuff should just work if you set the correct data attributes in the table.

Sorting [should work](https://github.com/ewels/MultiQC/blob/master/multiqc/templates/default/assets/js/multiqc_generalstats.js#L20-L26) if you give your table the class `.mqc_table` and colouring [should work](https://github.com/ewels/MultiQC/blob/master/multiqc/templates/default/assets/js/multiqc_generalstats.js#L93-L94) on any table.

Here is a simplified version of the current general statistics table markup to give you an idea of what is required:
```html
<table class="table table-condensed mqc_table">
  <thead>
    <tr>
      <th class="rowheader header">Sample Name</th>
      <th class="chroma-col header" data-chroma-scale="PuRd" data-chroma-max="134" data-chroma-min="0.0" data-shared-key="read_count">
        <span data-toggle="tooltip" title="Long descriptive text shown on hover">Short Header</span>
      </th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th class="rowheader">my_sample_name</th>
      <td class="data-coloured">
        <div class="wrapper">
          <span class="bar" style="width: 81.7901%;"></span>
          <span class="val">109.6</span>
        </div>
      </td>
    </tr>
  </tbody>
</table>
```

Not super simple sorry, but hopefully fairly self explanatory. The code that generates this is [here](https://github.com/ewels/MultiQC/blob/master/multiqc/utils/report.py#L36). The names of the ColorBrewer scales are [here](https://github.com/gka/chroma.js/wiki/Predefined-Colors) (add `-rev` to the name to reverse it).

Note that I would like to move the general stats colouring into the Python code instead of relying on JavaScript at some point, as this locks the browser up in big reports. So some of this JavaScript may change / disappear in the future. I just need a python package that works as well as [chroma.js](https://github.com/gka/chroma.js/).

multiqc/modules/base_module.py
ewels: As per your comment below - I can see how this can happen, great addition! Part of me wants you to move it out of the loop but it shouldn't make any difference :stuck_out_tongue: Nice!

multiqc/modules/samtools/samtools.py
lpantano: was to avoid big tables. there are parameters that won't affect if single end or paired ends. Mainly that. But true, that it could be mis-leading

lpantano:
So I updated the branch with your last master, Is now better? I implemented some of the comments. Working on the rest. :)

ewels:
Hey @lpantano,

I've pushed a few changes today, one probably affects you (and causes the merge conflict I suspect) - I made the filename matching use glob patterns instead of just `if str in otherstr`. See https://github.com/ewels/MultiQC/commit/53a31476cc7b1812c97130d214d2fe6dab50bd53

New plotting code looks good! You just taught me a new python function, I hadn't come across `any()` before..

Phil

lpantano:
@ewels  :) I think I updated successfully?

ewels:
Looks good - github isn't complaining about a merge conflict any more at least..

lpantano:
nice. ok, I think for me is ready. Maybe I work more in other figures for the other data is there, but will need more time for that, and really excited to have this for the next bcbio running to look at the nice reports we have now :)

ewels:
Awesome - merged! Thanks for your work with this, looks great!

lpantano:
Nice! I will work more in your repo, I think everybody should use it!!! I will add other QC metrics for the small RNA analysis we have, yuhuuuuuu

ewels:
Awesome! Yay! :smile:

ewels:
Just pushed a small commit above with a handful of minor changes, hope that's ok! Main one is I got it to ignore this line in one plot:
```
SN      bases mapped:   XXXX    # ignores clipping
```

That was because it's pretty much the same number as `bases mapped (cigar)`, so kind of doubles the number of mapped bases in the stacked bar plot. Hope that's ok!

Phil

lpantano:
hey, a question? 

when do you plan to get another stable release? Just to coordinate with bcbio and remove the one report that was merging all that by ourself, now we are using multiqc. And as well, want to plan in adding more modules. 

We use bioconda, so I think we are good in term of what version to use.

ewels:
Sorry, I was hoping to release it last Friday but didn't quite get things finished off. I'm on holiday this week but will definitely release v0.5 next week.

lpantano:
woo! thanks!

ewels:
Hi @lpantano,

I'd like to add support for some other samtools output soon, so I will probably refactor this module a little if that's ok - mostly putting your code into a separate file alongside `samtools.py` to keep things separated.

I was looking at the bargraphs you wrote and thinking that I might replace these with a dotplot of all of the discovered fields instead, much like the new bamtools stats module:

image

This is quite a bit more space efficient and so can show a lot more fields. Is that ok?

Phil

ewels:
A couple of other changes:

* Replace spaces in field names with underscores
* Calculate some percentages and use these instead of / as well as counts (eg. `% aligned`?)

ewels:
I just pushed these proposed changes to a branch called [stats_refactor](https://github.com/ewels/MultiQC/tree/stats_refactor) - let me know what you think!

Note that I changed where the search pattern is found - now `sp['samtools']['stats']` instead of just `sp['samtools']`

Would it be interesting to parse and plot insert sizes and indel distributions from these files as well?

Phil

ewels:
Example output screengrab:
image

lpantano:
I think all this is great, seriously! I like everything you proposed!

ewels:
Ok cool, then consider it merged :)

Need to head home now, but could be cool to look into plotting coverage and some of the other nice stuff in there another time.

PR #176, opened 2016-04-17
Qualimap: handle outputs with infinity

chapmanb:
Phil -- thanks for MultiQC. It's brilliant and looking forward to moving to it in bcbio. This is a small fix for an issue I ran into when testing.

For 5'/3' ratios where the 3' number is zero, Qualimap reports a
unicode infinity character. This occurs on small test cases, but
breaks the current qualimap parsing due to looking for an integer.
This fix leaves it unset in this case.


ewels:
Nice, thanks very much! These rare logging events are really difficult to find, so fixes are much appreciated..

Phil

PR #185, opened 2016-04-21
Add pattern param to clean name function.

lpantano:
Some times finel names have some naming that needs to be
clean to get the name of the samples the same across QC metrics.
This helps to remove keywords as needed.

If you think is not a general interest, I can add it to the bcbio plugin alone.

thanks!

### DIFF
--- a/multiqc/modules/base_module.py
+++ b/multiqc/modules/base_module.py
@@ -98,26 +98,29 @@ class BaseMultiqcModule(object):
                 except (IOError, OSError, ValueError, UnicodeDecodeError):
                     if config.report_readerrors:
                         logger.debug("Couldn't read file when looking for output: {}".format(fn))
-    
-    
-    def clean_s_name(self, s_name, root):
+
+    def clean_s_name(self, s_name, root, pattern=None):
         """ Helper function to take a long file name and strip it
         back to a clean sample name. Somewhat arbitrary.
         :param s_name: The sample name to clean
         :param root: The directory path that this file is within
+        :param pattern: Remove specific patterns from file name
         :config.prepend_dirs: boolean, whether to prepend dir name to s_name
         :return: The cleaned sample name, ready to be used
         """
         if root is None:
             root = ''
+        if pattern is not None:
+            s_name = s_name.replace(pattern, "")
+
         # Split then take first section to remove everything after these matches
         for ext in config.fn_clean_exts:
             s_name = os.path.basename(s_name.split(ext ,1)[0])
         if config.prepend_dirs:
             s_name = "{} | {}".format(root.replace(os.sep, ' | '), s_name).lstrip('. | ')
         return s_name
-    
-    
+
+
     def general_stats_addcols(self, data, headers={}, namespace=None):
         """ Helper function to add to the General Statistics variable.
         Adds to report.general_stats and does not return anything. Fills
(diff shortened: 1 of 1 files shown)


ewels:
Hah, brilliant - I was thinking of writing this myself about three hours ago :)

I was more thinking along the lines of having better handling of the strings that the function uses though. _eg._ checking for different variable types - strings, regular expression objects, glob patterns and functions should cover everything.

Hadn't quite figured out the specifics of this yet though. Maybe something like this?

##### `config.py`:
```python
def example_cleaning_func(str):
  return 'AWESOME: {}'.format(str)

fn_clean_exts = [ '.gz', '.fastq', '.fq', '.bam', '.sam', '.sra', '_tophat', '_star_aligned', '_fastqc', '.hicup', '.counts', '_counts',
  {'type': 'glob', 'pattern': '*.txt'} ,
  {'type': 're', 'pattern': re'/[\d+]/'} ,
  {'type': 'func', 'pattern': example_cleaning_func }
]
```

##### `base_module.py`:
```python
def clean_s_name(self, s_name, root):
          """ Helper function to take a long file name and strip it
          back to a clean sample name. Somewhat arbitrary.
          :param s_name: The sample name to clean
          :param root: The directory path that this file is within
         :param pattern: Remove specific patterns from file name
          :config.prepend_dirs: boolean, whether to prepend dir name to s_name
          :return: The cleaned sample name, ready to be used
          """
          if root is None:
              root = ''
 
          # Split then take first section to remove everything after these matches
          for ext in config.fn_clean_exts:
              if type(ext) is str:
                  ext = {'type':'str', 'pattern':ext}
              if ext['type'] == 'str':
                  s_name = os.path.basename(s_name.split(ext['pattern'] ,1)[0])
              elif ext['type'] == 'glob':
                  s_name = fnmatch.fnmatch(...) # do stuff
              elif ext['type'] == 'function':
                  s_name = ext['pattern']() # needs thought
          if config.prepend_dirs:
              s_name = "{} | {}".format(root.replace(os.sep, ' | '), s_name).lstrip('. | ')
          return s_name
```

Does that make sense? Should be fairly flexible, and allows other people's configs / plugins to modify the searches still.

What do you think?

Phil

lpantano:
oh! go ahead! if you add this, I will create functions or whatever in my plugin/module to get that working! please, do!

ewels:
..maybe I'm making this way too complicated. We could drop the function thing though and then it would still be easy enough to set in a yaml file:

```yaml
fn_clean_exts:
    - .gz
    - type: regex
      pattern: ^\w+
```

ewels:
Also probably drop the glob thing as I'm not sure how MultiQC could know which part of the match to remove. Should be easy to remove a `string + stuff to right` or a regex match though.

I'll push an update in a couple of minutes.

ewels:
Ok, added in 3e669379f266e372d8072f3138d90e07bf42d69b

Let me know if this works for you..

lpantano:
it worked, thanks! can you put `root=None`, just to use the parameter only when you want to specify something? if not I will pass None always, no problem.

ewels:
`root` is only used for working out the directory path if `-d` / `--dirs` is specified (to prevent filename collisions when identical filenames are kept in different directories). So it's good to always specify that if it's possible.

But yeah, if it's not possible then set it to `None` or an empty string, either works. I should change it to `def clean_s_name(self, s_name, root=None):` really.

Phil

lpantano:
hear that. I understand. In bcbio we take care of that, so collision won’t happen. But I get the point to have it like that. So whatever you want is fine with me.
thanks!

PR #196, opened 2016-05-03
fixed a bug which caused extra line breaks in picard summary files

dakl:
Before this fix, the trailing newline in the last element of each row would cause an extra line break if the keys we're rearranged. This broke the hsmetrics summary output and added an extra line break between every line in the insert size metrics file.


ewels:
Nice spot, thanks!

PR #201, opened 2016-05-05
Allow a file with a list of files as input

lpantano:
Hi!

this is a suggestion PR, since probably you have a better idea how to integrate it.

But we had some problems with a person with so many arguments (files) in the command line that it cannot run. (https://github.com/chapmanb/bcbio-nextgen/issues/1380)

In bcbio, we are moving to CWL, what has many advantages, but one of the inconvenient is that we cannot use commands assuming that we are in the same machine or in a standard file system. So we can not use `multiqc qc/*` etc. Instead we are passing the specific path.

So, @chapmanb thought that maybe it would be good to have an argument to allow a file that contains a list of files to be used by multiqc.

As I said, this is a suggestion, because I like to offer some kind of solution instead of just asking, but feel free to adapt however you think is better.

thanks! 

PS: another problem it is if 700 samples make any sense inside multiqc, but I think we can change things to allow a proper report in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,5 @@ script:
   - multiqc data -f --flat
   - multiqc data -o tests/multiqc_report_dev -t default_dev -k json
   - multiqc data -o tests/multiqc_report_geo -t geo -k yaml -v
+  - multiqc --is-file-list data/special_cases/file_list.txt 
   
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,5 +17,5 @@ script:
   - multiqc data -f --flat
   - multiqc data -o tests/multiqc_report_dev -t default_dev -k json
   - multiqc data -o tests/multiqc_report_geo -t geo -k yaml -v
-  - multiqc --is-file-list data/special_cases/file_list.txt 
+  - multiqc --is-file-list data/special_cases/file_list.txt -f
   
--- a/multiqc/utils/config.py
+++ b/multiqc/utils/config.py
@@ -27,6 +27,7 @@ modules_dir = os.path.join(MULTIQC_DIR, 'modules')
 creation_date = datetime.now().strftime("%Y-%m-%d, %H:%m")
 working_dir = os.getcwd()
 analysis_dir = [os.getcwd()]
+is_file_list = False
 output_dir = os.path.realpath(os.getcwd())
 output_fn_name = 'multiqc_report.html'
 data_dir_name = 'multiqc_data'
--- a/scripts/multiqc
+++ b/scripts/multiqc
@@ -44,6 +44,12 @@ plugin_hooks.mqc_trigger('config_loaded')
                     required = True,
                     metavar = "<analysis directory>"
 )
+
+@click.option('-l', '--is-file-list',
+                    is_flag = True,
+                    help = "The file given as analysis_dir has a list of file paths "
+                           "to be searched, one per row"
+)
 @click.option('-f', '--force',
                     is_flag = True,
                     help = "Overwrite any existing reports"
@@ -125,7 +131,7 @@ plugin_hooks.mqc_trigger('config_loaded')
 )
 @click.version_option(__version__)
 
-def multiqc(analysis_dir, dirs, no_clean_sname, title, template, module, exclude, outdir, ignore, filename, 
+def multiqc(analysis_dir, dirs, is_file_list, no_clean_sname, title, template, module, exclude, outdir, ignore, filename, 
 make_data_dir, data_format, zip_data_dir, force, plots_flat, plots_interactive, config_file, verbose, quiet, **kwargs):
     """MultiQC aggregates results from bioinformatics analyses across many samples into a single report.
     
@@ -181,6 +187,22 @@ make_data_dir, data_format, zip_data_dir, force, plots_flat, plots_interactive, 
     config.plots_force_flat = plots_flat
     config.plots_force_interactive = plots_interactive
     config.kwargs = kwargs # Plugin command line options
+
+    # Add files if --is-file-list option is given
+    if is_file_list:
+        if len(analysis_dir) > 1:
+            raise ValueError("If --is-file-list is giving, analysis_dir should have only one plain text file.")
+        if os.path.isdir(analysis_dir[0]):
+            raise ValueError("If --is-file-list is giving, analysis_dir should be a plain text file.")
+        config.analysis_dir = ()
+        with (open(analysis_dir[0])) as in_handle:
+            for line in in_handle:
+                if os.path.exists(line.strip()):
+                    config.analysis_dir += (os.path.abspath(line.strip()),)
+        if len(config.analysis_dir) == 0:
+            logger.warning("No files were added from %s using --is-file-list option." % analysis_dir[0])
+            logger.watning("Please, check that %s contains correct file paths." % analysis_dir[0])
+
     if no_clean_sname:
         config.fn_clean_exts = []
         logger.info("Not cleaning sample names")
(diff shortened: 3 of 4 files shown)


ewels:
Aha, yes I can see how that could be a problem. I can't say that I ever expected anyone to specify every single log file :)

This should be fairly easy to implement and your changes look good. I might suggest a couple of minor changes in strategy in inline comments though if that's ok 😉 

Note that I'm probably going to rewrite this code fairly soon. You just prompted me to write down my intentions, so now described in #202. Shouldn't affect this change though.

Phil

ewels:
Overarching code suggestions in one place here first:

* Rename `list_files` to `file_list` for clarity
* Make it a boolean flag instead of string
* Keep the `analysis_dir` field required and use this to get the path to the file list
  * Check that only one path specified if `file_list` is `true`.

Will follow these up with inline comments.

multiqc/utils/config.py
```diff
@@ -27,6 +27,7 @@
 creation_date = datetime.now().strftime("%Y-%m-%d, %H:%m")
 working_dir = os.getcwd()
 analysis_dir = [os.getcwd()]
+list_files = None
```
ewels:
```python
file_list = False
```

scripts/multiqc
```diff
@@ -41,9 +41,13 @@ plugin_hooks.mqc_trigger('config_loaded')
 @click.argument('analysis_dir',
                     type = click.Path(exists=True),
                     nargs = -1,
-                    required = True,
+                    required = False,
```
ewels: revert

scripts/multiqc
```diff
@@ -41,9 +41,13 @@ plugin_hooks.mqc_trigger('config_loaded')
...
                     metavar = "<analysis directory>"
 )
+
+@click.option('-l', '--list_files',
```
ewels:
```python
@click.option('-l', '--file_list',
                    is_flag = True,
                    help = "Specify a file describing file paths to be searched, one per row"
```

scripts/multiqc
```diff
@@ -181,6 +185,22 @@ make_data_dir, data_format, zip_data_dir, force, plots_flat, plots_interactive,
     config.plots_force_flat = plots_flat
     config.plots_force_interactive = plots_interactive
     config.kwargs = kwargs # Plugin command line options
+
+    # Check if some of the parameters giving input exists
```
ewels: Not required any more

scripts/multiqc
```diff
@@ -181,6 +185,22 @@ make_data_dir, data_format, zip_data_dir, force, plots_flat, plots_interactive,
...
+    if not analysis_dir and not list_files:
+        logger.warning("You need to give a list of files or folders as arguments, "
+                     "or use the --list_files argument pointing to a file that includes "
+                     "files in each row.")
+        raise ValueError("Missing both, analysis_dir and list_files arguments.")
+
+    # Add files if --list option is given
+    print(analysis_dir)
```
ewels: Debugging statement? Can presumably remove.

scripts/multiqc
```diff
@@ -181,6 +185,22 @@ make_data_dir, data_format, zip_data_dir, force, plots_flat, plots_interactive,
...
+    if list_files:
```
ewels:
```python
if file_list:
  assert(len(analysis_dir) == 1)
  with (open(analysis_dir[0]) as in_handle:
```

scripts/multiqc
```diff
@@ -181,6 +185,22 @@ make_data_dir, data_format, zip_data_dir, force, plots_flat, plots_interactive,
...
+        with (open(list_files)) as in_handle:
+            for line in in_handle:
+                if os.path.exists(line.strip()):
```
ewels: Maybe throw a debug log if not?

ewels:
Oh, and please also add the parameter to the example config file and add a bit to the docs about it if that's ok. Thanks!

Phil

ewels:
Actually, perhaps my suggestion about using `analysis_dir` is dump, not sure. You could keep `file_list` as specifying the path to the file, maybe more intuitive. Click [has an option](http://click.pocoo.org/5/arguments/#file-path-arguments) for file where it checks that it exists though, so some functionality for free there.

lpantano:
I followed your first comment to convert file_list in a flag. I liked more, but is not perfect either of them. Let me know what you think now. Maybe we can leave it as a flag, and change it later depending of the user suggestions ...

Thanks!!!!

scripts/multiqc
```diff
@@ -181,6 +187,22 @@ make_data_dir, data_format, zip_data_dir, force, plots_flat, plots_interactive,
     config.plots_force_flat = plots_flat
     config.plots_force_interactive = plots_interactive
     config.kwargs = kwargs # Plugin command line options
+
+    # Add files if --is-file-list option is given
+    if is_file_list:
+        if len(analysis_dir) > 1:
+            raise ValueError("If --is-file-list is giving, analysis_dir should have only one plain text file.")
+        if os.path.isdir(analysis_dir[0]):
```
ewels: You can probably skip this bit - the `open` command will throw an exception if it's a directory anyway I think.

scripts/multiqc
```diff
@@ -181,6 +187,22 @@ make_data_dir, data_format, zip_data_dir, force, plots_flat, plots_interactive,
...
+            raise ValueError("If --is-file-list is giving, analysis_dir should be a plain text file.")
+        config.analysis_dir = ()
+        with (open(analysis_dir[0])) as in_handle:
+            for line in in_handle:
+                if os.path.exists(line.strip()):
+                    config.analysis_dir += (os.path.abspath(line.strip()),)
+        if len(config.analysis_dir) == 0:
+            logger.warning("No files were added from %s using --is-file-list option." % analysis_dir[0])
```
ewels: Small thing, but I prefer using `format` syntax as in the rest of the program:
```python
logger.error("No files were added from {} using --is-file-list option.".format(analysis_dir[0]))
```
Also, this should probably be a `Raise` as the program can't continue without any files.

scripts/multiqc
```diff
@@ -181,6 +187,22 @@ make_data_dir, data_format, zip_data_dir, force, plots_flat, plots_interactive,
...
+            logger.watning("Please, check that %s contains correct file paths." % analysis_dir[0])
```
ewels: Typo: `watning`

ewels:
Looks good! Couple of minor comments but then will merge. We also need to add a demo file list to MultiQC_TestData and update `.travis.yaml` to add a test for this new feature.

ewels:
Ok, added a [test file](https://github.com/ewels/MultiQC_TestData/blob/master/data/special_cases/file_list.txt) so you should now be able to add a test command to the travis config:

```
multiqc --file_list data/special_cases/file_list.txt
```

scripts/multiqc
lpantano: ups, i forgot to change that. Now you see this: 

[INFO   ]         multiqc : This is MultiQC v0.6.dev0
[WARNING]         multiqc : No files were added from Test1_bcbio.txt using --is-file-list option.
[WARNING]         multiqc : Please, check that Test1_bcbio.txt contains correct file paths.
[WARNING]         multiqc : No analysis results found. Cleaning up..

I added the warning to follow with the logic I saw. Like the last warning was there already, so I thought to add more info.  But yeah, probably should be an error instead.

.travis.yml
```diff
@@ -17,4 +17,5 @@ script:
   - multiqc data -f --flat
   - multiqc data -o tests/multiqc_report_dev -t default_dev -k json
   - multiqc data -o tests/multiqc_report_geo -t geo -k yaml -v
-  
\ No newline at end of file
+  - multiqc --is-file-list data/special_cases/file_list.txt
```
ewels: Sorry, you'll need `-f` here to get it to overwrite the existing report.

ewels:
Sorry, just saw this. Looks good!

ewels:
Thanks!

ewels:
Hi @lpantano,

I just changed `--is-file-list` to the simpler `--file-list`, hope this is ok!

Phil

ewels:
Sorry, shouldn't push changes when I'm tired! New commit with changes actually working now.

lpantano:
thanks!

PR #211, opened 2016-05-17
Add full table for HsMetrics

lpantano:
Hi,

We have some users that would like to have the full table from hsmetric as a section in the report. 
How do you feel about it? 
Right now, I just parse almost all columns, because all are kind if important. 
Do you have any concern about just adding the table as it is in the output?

thanks!

### DIFF
--- a/multiqc/modules/picard/HsMetrics.py
+++ b/multiqc/modules/picard/HsMetrics.py
@@ -159,7 +159,9 @@ def parse_reports(self):
                 'anchor': 'picard_hsmetrics',
                 'content': table_html})
         self.sections.append(_add_target_bases(data))
-        self.sections.append(_add_hs_penalty(data))
+        penalty_html = _add_hs_penalty(data)
+        if penalty_html:
+            self.sections.append(penalty_html)
         # Return the number of detected samples to the parent module
     return len(self.picard_HsMetrics_data)
 
@@ -247,5 +249,5 @@ def _add_hs_penalty(data):
     'content': subtitle + plots.linegraph.plot(data_clean, pconfig)}
 
     if not any_non_zero:
-        return ""
+        return None
     return section
--- a/multiqc/modules/picard/HsMetrics.py
+++ b/multiqc/modules/picard/HsMetrics.py
@@ -13,6 +13,45 @@ from multiqc import plots
 # Initialise the logger
 log = logging.getLogger(__name__)
 
+DESC = {
+        'BAIT_SET': 'The name of the bait set used in the hybrid selection.',
+        'BAIT_TERRITORY': 'The number of bases which have one or more baits on top of them.',
+        'BAIT_DESIGN_EFFICIENCY': 'Target terrirtoy / bait territory. 1 == perfectly efficient, 0.5 = half of baited bases are not target.',
+        'TOTAL_READS': 'The total number of reads in the SAM or BAM file examine.',
+        'PF_READS': 'The number of reads that pass the vendor\'s filter.',
+        'PF_UNIQUE_READS': 'The number of PF reads that are not marked as duplicates.',
+        'PCT_PF_READS': 'PF reads / total reads. The percent of reads passing filter.',
+        'PCT_PF_UQ_READS': 'PF Unique Reads / Total Reads.',
+        'PF_UQ_READS_ALIGNED': 'The number of PF unique reads that are aligned with mapping score > 0 to the reference genome.',
+        'PCT_PF_UQ_READS_ALIGNED': 'PF Reads Aligned / PF Reads.',
+        'PF_BASES_ALIGNED': 'The number of PF unique bases that are aligned with mapping score > 0 to the reference genome.',
+        'PF_UQ_BASES_ALIGNED': 'The number of bases in the PF aligned reads that are mapped to a reference base. Accounts for clipping and gaps.',
+        'ON_BAIT_BASES': 'The number of PF aligned bases that mapped to a baited region of the genome.',
+        'NEAR_BAIT_BASES': 'The number of PF aligned bases that mapped to within a fixed interval of a baited region, but not on a baited region.',
+        'OFF_BAIT_BASES': 'The number of PF aligned bases that mapped to neither on or near a bait.',
+        'ON_TARGET_BASES': 'The number of PF aligned bases that mapped to a targeted region of the genome.',
+        'PCT_SELECTED_BASES': 'On+Near Bait Bases / PF Bases Aligned.',
+        'PCT_OFF_BAIT': 'The percentage of aligned PF bases that mapped neither on or near a bait.',
+        'ON_BAIT_VS_SELECTED': 'The percentage of on+near bait bases that are on as opposed to near.',
+        'MEAN_BAIT_COVERAGE': 'The mean coverage of all baits in the experiment.',
+        'MEAN_TARGET_COVERAGE': 'The mean coverage of targets.',
+        'MEDIAN_TARGET_COVERAGE': 'The median coverage of targets.',
+        'PCT_USABLE_BASES_ON_BAIT': 'The number of aligned, de-duped, on-bait bases out of the PF bases available.',
+        'PCT_USABLE_BASES_ON_TARGET': 'The number of aligned, de-duped, on-target bases out of the PF bases available.',
+        'FOLD_ENRICHMENT': 'The fold by which the baited region has been amplified above genomic background.',
+        'ZERO_CVG_TARGETS_PCT': 'The fraction of targets that did not reach coverage=1 over any base.',
+        'PCT_EXC_DUPE': 'The fraction of aligned bases that were filtered out because they were in reads marked as duplicates.',
+        'PCT_EXC_MAPQ': 'The fraction of aligned bases that were filtered out because they were in reads with low mapping quality.',
+        'PCT_EXC_BASEQ': 'The fraction of aligned bases that were filtered out because they were of low base quality.',
+        'PCT_EXC_OVERLAP': 'The fraction of aligned bases that were filtered out because they were the second observation from an insert with overlapping reads.',
+        'PCT_EXC_OFF_TARGET': 'The fraction of aligned bases that were filtered out because they did not align over a target base.',
+        'OLD_80_BASE_PENALTY': 'The fold over-coverage necessary to raise 80% of bases in "non-zero-cvg" targets to the mean coverage level in those targets.',
+        'AT_DROPOUT': 'A measure of how undercovered <= 50% GC regions are relative to the mean.',
+        'GC_DROPOUT': 'A measure of how undercovered >= 50% GC regions are relative to the mean.',
+        'HET_SNP_SENSITIVITY': 'The theoretical HET SNP sensitivity.',
+        'HET_SNP_Q': 'The Phred Scaled Q Score of the theoretical HET SNP sensitivity.'
+        }
+
 
 def parse_reports(self):
     """ Find Picard HsMetrics reports and parse their data """
@@ -109,17 +148,17 @@ def parse_reports(self):
             if s_name not in self.general_stats_data:
                 self.general_stats_data[s_name] = dict()
             self.general_stats_data[s_name].update( data[s_name] )
-    data_table = _clean_table(data)
-    table_html = plots.table.plot(data_table, _get_headers(data_table))
-    if not isinstance(self.sections, list):
-        self.sections = list()
-    self.sections.append({
-            'name': 'HSMetrics',
-            'anchor': 'picard_hsmetrics',
-            'content': table_html})
-    self.sections.append(_add_target_bases(data))
-    self.sections.append(_add_hs_penalty(data))
-    # Return the number of detected samples to the parent module
+        data_table = _clean_table(data)
+        table_html = plots.table.plot(data_table, _get_headers(data_table))
+        if not isinstance(self.sections, list):
+            self.sections = list()
+        self.sections.append({
+                'name': 'HSMetrics',
+                'anchor': 'picard_hsmetrics',
+                'content': table_html})
+        self.sections.append(_add_target_bases(data))
+        self.sections.append(_add_hs_penalty(data))
+        # Return the number of detected samples to the parent module
     return len(self.picard_HsMetrics_data)
 
 def _clean_table(data):
@@ -146,6 +185,7 @@ def _get_headers(data):
             if h not in header:
                 this = {
                 'title': h.replace("_", " ").lower().capitalize(),
+                'description' : DESC[h] if h in DESC else "",
                 'scale': 'RdYlGn'
                 }
                 if h.startswith("PCT"):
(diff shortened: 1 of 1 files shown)


ewels:
Cool! How does it look? Do you have an example report?

lpantano:
I am using the examples in multiqc_data, they are ok for that, they have all the information

ewels:
Ok cool, I pulled your PR and had a go.. wow, that's a big table!

So there are a few things I think we can do to make this a bit more usable before it goes in..

1. Sentence case the headers so that they are not shouting 😄 
2. Sort the headers alphabetically so that similar columns are next to each other
3. Use `'shared_key': 'read_count'` and  `'shared_key': 'base_count'` to make scaling consistent

This gives us the following headers, which is already a bit easier to digest:

1. At Dropout
1. Bait Design Efficiency
1. Bait Territory
1. Fold 80 Base Penalty
1. Fold Enrichment
1. Gc Dropout
1. Genome Size
1. Hs Penalty 100X
1. Hs Penalty 10X
1. Hs Penalty 20X
1. Hs Penalty 30X
1. Hs Penalty 40X
1. Hs Penalty 50X
1. Mean Bait Coverage
1. Mean Target Coverage
1. Near Bait Bases
1. Off Bait Bases
1. On Bait Bases
1. On Bait Vs Selected
1. On Target Bases
1. Pct Off Bait
1. Pct Pf Reads
1. Pct Pf Uq Reads
1. Pct Pf Uq Reads Aligned
1. Pct Selected Bases
1. Pct Target Bases 100X
1. Pct Target Bases 10X
1. Pct Target Bases 20X
1. Pct Target Bases 2X
1. Pct Target Bases 30X
1. Pct Target Bases 40X
1. Pct Target Bases 50X
1. Pct Usable Bases On Bait
1. Pct Usable Bases On Target
1. Pf Reads
1. Pf Unique Reads
1. Pf Uq Bases Aligned
1. Pf Uq Reads Aligned
1. Target Territory
1. Total Reads
1. Zero Cvg Targets Pct

..but 41 columns is still kind of excessive? It looks like we could ignore quite a few:

* HS PENALTY ..X
  * These are all 0 in the test data and make up 7 columns
* PCT TARGET BASES ..X
  * This is probably better represented in a plot (eg. the Qualimap BamQC plot) rather than 7 columns
* GENOME SIZE
  * Should always be constant for all samples, so probably not informative

Can we remove these? And any others? If not removing, I think a lot we can make hidden by default?

@dakl - any thoughts? I know you have been playing with this module as well and were thinking about using tables here.

Phil

lpantano:
i think we can put hs penalty and pct target  in figures and removed the cols you mentioned, that it should clean a lot. 

sorry for my paused replies, i am sick and not all the time online.

thanks!

sent not from my computer

ewels:
Sounds good! No problem, I'm about to be away for a week also. Get well soon!

dakl:
Nice! 😄 

As @lpantano suggests, I think pct on target (Pct Selected Bases) is important, but I'd also like to see Total reads and Mean Target Coverage as these are important metrics for capture experiments. As for Hs Penalty, I commonly keep track of Fold 80 base penalty rather the the hs penalties. Depending on the samples analysed, different hs penalties will be interesting, so choosing one to print is probably not the best thing to do. 

I'm assuming that all columns are captured and printed in the data files, and what we are talking about is hiding rather than removing columns, correct?

lpantano:
Nice!, I will put those set of columns in plots, and then remove a few more that doesn't make sense, and make non-shouting headers! thanks both!

ewels:
Sounds great!

lpantano:
Hi,

I changed some of the stuff, if you both can look at it and give me some feedback, it would be great!

thanks!

ewels:
Nice! This is looking better :) Will add some comments in the code..

multiqc/modules/picard/HsMetrics.py
```diff
@@ -108,7 +109,80 @@ def parse_reports(self):
...
+
+def _add_target_bases(data):
+    data_clean = defaultdict(dict)
+    for s in data:
+        for h in data[s]:
+            if h.startswith("PCT_TARGET"):
+                data_clean[s][h] = data[s][h] * 100.0
+    return {'name': 'Pct target bases',
+    'anchor': 'picard_hsmetrics_target_bases',
+    'content': plots.beeswarm.plot(data_clean, _get_headers(data_clean))}
```
ewels: Might be nice to add a line of text here explaining what this is showing?
```python
'content': '<p>The percentage of target bases reaching a specific level of coverage for each sample.</p>'
              + plots.beeswarm.plot(data_clean, _get_headers(data_clean))
```

multiqc/modules/picard/HsMetrics.py
```diff
@@ -108,7 +109,80 @@ def parse_reports(self):
...
+    header = {}
+    for s in data:
+        for h in data[s]:
+            try:
+                float(data[s][h])
+            except:
+                continue
+            if h not in header:
+                this = {
+                'title': h.replace("_", " ").lower().capitalize(),
```
ewels: These titles are a bit too long on my screen and overlap the plot legend. Maybe something like this instead?
```python
'title': h.replace('PCT_TARGET_BASES_', '% Target '),
'description': h.replace('PCT_TARGET_BASES_', 'Percent of target bases with >= '),
```

multiqc/modules/picard/HsMetrics.py
```diff
@@ -108,7 +109,80 @@ def parse_reports(self):
...
+                header.update({h : this})
+    return OrderedDict(sorted(header.items(), key=lambda t: t[1]['title']))
+
+def _add_target_bases(data):
+    data_clean = defaultdict(dict)
+    for s in data:
+        for h in data[s]:
+            if h.startswith("PCT_TARGET"):
+                data_clean[s][h] = data[s][h] * 100.0
+    return {'name': 'Pct target bases',
```
ewels: Hmm, maybe `Target Region Coverage` or something? It wasn't very intuitive to me what this was saying at first..

multiqc/modules/picard/HsMetrics.py
ewels: General comment - would it make more sense to show this data as a line plot instead of a beeswarm plot? As the data is connected.

Also, for me it sorts alphabetically which is a bit confusing:

1. 100X
2. 10X
3. 20X

_..etc_

multiqc/modules/picard/HsMetrics.py
```diff
@@ -108,7 +109,80 @@ def parse_reports(self):
...
+    any_non_zero = False
+    for s in data:
+        for h in data[s]:
+            if h.startswith("HS_PENALTY"):
+                data_clean[s][h.replace("_", " ").lower().capitalize()] = data[s][h]
+                if data[s][h] > 0:
+                    any_non_zero = True
+    section =  {'name': 'HS penalty',
+    'anchor': 'picard_hsmetrics_hs_penalty',
+    'content': plots.beeswarm.plot(data_clean, _get_headers(data_clean))}
```
ewels: Same comment as above - maybe a line plot instead of beeswarm plot as connected? I'm not really sure what this metric shows though. Any thoughts though @dakl? Current demo report shows me this:

image

ewels:
Yeah, table is still pretty massive, but far easier to read now I think - nice work! Sorry for being so pedantic :flushed:

ewels:
Ooh, also the Picard Website has a nice table with long descriptions of what every metric means:
https://broadinstitute.github.io/picard/picard-metric-definitions.html#HsMetrics

It would be brilliant if this could be used for the `description` hover text on table column headers, and also in a `<p></p>` tag before the plots..

lpantano:
Hi @ewels, I changed quite a lot following your recommendation, so I guess need again one of your good revisions. We are closer.

Thanks for your good ideas!

ewels:
Hi @lpantano - sorry for being slow with this, I haven't forgotten! Will get to it soon I promise :)

ewels:
ok, looking good! The plots could look a bit nicer with a few extra config options. See for example the percentage line plots in [qualimap bamqc](https://github.com/ewels/MultiQC/blob/4f9f60d9b3b2166601f622c76a934d3d9956bb9b/multiqc/modules/qualimap/QM_BamQC.py#L217-L223):

```python
config = {
  'title': 'Genome Fraction Coverage',
  'ylab': 'Fraction of reference (%)',
  'xlab': 'Coverage (X)',
  'ymax': 100,
  'ymin': 0,
  'xmin': 0,
  'tt_label': '<b>{point.x}X</b>: {point.y:.2f}%',
}
```

Couple of minor comments I'll put inline, but basically done I think :)

multiqc/modules/picard/HsMetrics.py
```diff
@@ -108,7 +148,87 @@ def parse_reports(self):
...
+                    this.update({'shared_key': 'bases_count',
+                                 'modify': lambda x: x / 1000000.0})
+                    this['title'] = "Mb {0}".format(this['title'])
+                if h in ["BAIT_TERRITORY", "TOTAL_READS", "TARGET_TERRITORY", "AT_DROPOUT", "GC_DROPOUT"]:
+                    this.update({'hidden': True})
+                header.update({h : this})
+    return OrderedDict(sorted(header.items(), key=lambda t: t[1]['title']))
+
+def _add_target_bases(data):
+    subtitle = "<p>The percentage of all target bases achieving `NUMBER`X or greater coverage.</p>"
```
ewels:
```html
<p>The percentage of all target bases with at least <code>x</code> fold coverage.</p>
```

multiqc/modules/picard/HsMetrics.py
```diff
@@ -108,7 +148,87 @@ def parse_reports(self):
...
+    data_clean = defaultdict(dict)
+    for s in data:
+        for h in data[s]:
+            if h.startswith("PCT_TARGET"):
+                data_clean[s][int(h.replace("PCT_TARGET_BASES_", "")[:-1])] = data[s][h] * 100.0
+    pconfig = { 'xlab': 'Number of reads',
```
ewels: Maybe `Fold Coverage` instead of `'Number of reads'`?

multiqc/modules/picard/HsMetrics.py
```diff
@@ -108,7 +148,87 @@ def parse_reports(self):
...
+                'ylab': 'Pct of bases'}
+    return {'name': 'Target Region Coverage',
+    'anchor': 'picard_hsmetrics_target_bases',
+    'content': subtitle + plots.linegraph.plot(data_clean, pconfig)}
+
+def _add_hs_penalty(data):
+    subtitle = "<p>The hybrid selection penalty incurred to get 80% of target bases to 10X. This metric should be interpreted as: if I have a design with 10 megabases of target, and want to get 10X coverage I need to sequence until PF_ALIGNED_BASES = 10^7 * 10 * HS_PENALTY_`NUMBER`X.</p>"
```
ewels: Wow this is a confusing description from the Picard authors. How about this?

```html
<p>The "hybrid selection penalty" incurred to get 80% of target bases to a given coverage. Can be used with the formula <code>required_aligned_bases = bait_size_bp * desired_coverage * hs_penalty</code>.</p>
```

multiqc/modules/picard/HsMetrics.py
```diff
@@ -108,7 +148,87 @@ def parse_reports(self):
...
+    data_clean = defaultdict(dict)
+    any_non_zero = False
+    for s in data:
+        for h in data[s]:
+            if h.startswith("HS_PENALTY"):
+                data_clean[s][(h.replace("HS_PENALTY_", " ")[:-1])] = data[s][h]
+                if data[s][h] > 0:
+                    any_non_zero = True
+    pconfig = { 'xlab': 'Number of reads',
```
ewels: `Fold Coverage` again instead of `Number of reads` I think.

multiqc/modules/picard/HsMetrics.py
```diff
@@ -108,7 +148,87 @@ def parse_reports(self):
...
+                'ylab': 'Pct of bases'}
+    return {'name': 'Target Region Coverage',
+    'anchor': 'picard_hsmetrics_target_bases',
+    'content': subtitle + plots.linegraph.plot(data_clean, pconfig)}
+
+def _add_hs_penalty(data):
+    subtitle = "<p>The hybrid selection penalty incurred to get 80% of target bases to 10X. This metric should be interpreted as: if I have a design with 10 megabases of target, and want to get 10X coverage I need to sequence until PF_ALIGNED_BASES = 10^7 * 10 * HS_PENALTY_`NUMBER`X.</p>"
+    data_clean = defaultdict(dict)
+    any_non_zero = False
```
ewels: Is this used? Would be nice - I have all zeroes in my demo data but I get the plot still.

multiqc/modules/picard/HsMetrics.py
```diff
@@ -108,7 +148,87 @@ def parse_reports(self):
...
+    for s in data:
+        for h in data[s]:
+            try:
+                float(data[s][h])
+            except:
+                continue
+            if h not in header:
+                this = {
+                'title': h.replace("_", " ").lower().capitalize(),
+                'description' : DESC[h] if h in DESC else "",
```
ewels: I think `None` instead of `""`. If it's `None` I think it uses the title instead or something sensible? That's off the top of my head though, may make no difference.

multiqc/modules/picard/HsMetrics.py
```diff
@@ -108,7 +148,87 @@ def parse_reports(self):
...
+                'scale': 'RdYlGn'
+                }
+                if h.startswith("PCT"):
```
ewels: Don't think this is used any more?

multiqc/modules/picard/HsMetrics.py
```diff
@@ -108,7 +148,87 @@ def parse_reports(self):
...
+def _get_headers(data):
+    header = {}
+    for s in data:
+        for h in data[s]:
+            try:
+                float(data[s][h])
+            except:
+                continue
+            if h not in header:
+                this = {
```
ewels: I think most (if not all?) columns should probably have a `min: 0`?

ewels: Should probably also have `'namespace': 'HsMetrics'` too. Sorry, not much in the way of docs about this.

multiqc/modules/picard/HsMetrics.py
```diff
@@ -2,16 +2,56 @@
...
+        'PCT_EXC_MAPQ': 'The fraction of aligned bases that were filtered out because they were in reads with low mapping quality.',
+        'PCT_EXC_BASEQ': 'The fraction of aligned bases that were filtered out because they were of low base quality.',
+        'PCT_EXC_OVERLAP': 'The fraction of aligned bases that were filtered out because they were the second observation from an insert with overlapping reads.',
+        'PCT_EXC_OFF_TARGET': 'The fraction of aligned bases that were filtered out because they did not align over a target base.',
+        'OLD_80_BASE_PENALTY': 'The fold over-coverage necessary to raise 80% of bases in "non-zero-cvg" targets to the mean coverage level in those targets.',
+        'AT_DROPOUT': 'A measure of how undercovered <= 50% GC regions are relative to the mean.',
+        'GC_DROPOUT': 'A measure of how undercovered >= 50% GC regions are relative to the mean.',
+        'HET_SNP_SENSITIVITY': 'The theoretical HET SNP sensitivity.',
+        'HET_SNP_Q': 'The Phred Scaled Q Score of the theoretical HET SNP sensitivity.'
```
ewels: Missing in my test data:
```
'FOLD_80_BASE_PENALTY': 'The fold over-coverage necessary to raise 80% of bases in "non-zero-cvg" targets to the mean coverage level in those targets.',
'TARGET_TERRITORY': 'The unique number of target bases in the experiment where target is usually exons etc.'
```

ewels:
This is super cool :) I love the automatic full descriptions that you get when hovering on the column headers. Nice work! 👍

lpantano:
Thanks for the comments. I will work during the weekend!

lpantano:
I think I addressed all the comment, but feel free to check and give more feedback.
cheers

multiqc/modules/picard/HsMetrics.py
```diff
@@ -108,7 +150,102 @@ def parse_reports(self):
...
+        data_table = _clean_table(data)
+        table_html = plots.table.plot(data_table, _get_headers(data_table))
+        if not isinstance(self.sections, list):
+            self.sections = list()
+        self.sections.append({
+                'name': 'HSMetrics',
+                'anchor': 'picard_hsmetrics',
+                'content': table_html})
+        self.sections.append(_add_target_bases(data))
+        self.sections.append(_add_hs_penalty(data))
```
ewels: Only problem with this is that we still get a section (though empty) if no results:
image

Maybe instead return `None` and then do:
```python
penalty_html = _add_hs_penalty(data)
if penalty_html is not None:
    self.sections.append(penalty_html)
```
Not as pretty code, but gives a cleaner report..

ewels:
Brilliant, this looks super tidy now! I have one tiny request (see above), but then good to merge I think :)

ewels:
Awesome - I can't think of anything else to suggest!

🎉  😂  ✨ 

Thank you for your patience 😁 

it's done

lpantano:
jaja, awesome picture! thanks for all your feedback!

dakl:
😂👍🍻
tors 9 juni 2016 kl. 17:08 skrev Lorena Pantano <notifications@github.com>:

PR #222, opened 2016-05-27
adding qualimap QM_BAMQC 50x 10x 5x and 1x as default hidden

avilella:
Thx


ewels:
This is great! What do you think about catching every percentage point and then plotting the data instead of putting it into the table? Much like the Qualimap [Genome Fraction Coverage plot](http://multiqc.info/examples/wgs/multiqc_report.html#qualimap-genome-fraction-coverage)..

avilella:
For what I see, people want both: plots and tables.

Thx for accepting the PR

ewels:
Hah, I'm sorry - for some reason I read this as affecting the MethylQA module, not Qualimap. Ok, this makes more sense now..

PR #225, opened 2016-05-31
use `ID` to get `s_name`

dakl:
`SM` can be shared between libraries if multiple libraries are generated from a single biological sample. Duplicates should be reported as fine-grained as possible.


dakl:
The problem with this is that there's no single right way do get the sample name. It depends on how the user sets the `@RG` tag when aligning. Therefore I think that reporting duplicates as fine grained as possible rather than as coarse-grained (is that a word) makes more sense.

ewels:
ok, fine by me!

PR #235, opened 2016-06-10
picard 2.3.0 has renamed `CalculateHsMetrics` to `c`

dakl:
...to `CollectHsMetrics`.

... therefore in order to support both, the module needs to check for both those strings. AFAIK, the output data is the same in both `CalculateHsMetrics ` and `CollectHsMetrics `.


ewels:
Great, thanks!

PR #252, opened 2016-07-05
cutadapt: fix version detection

lpantano:
If using numbers to compare versions:
1.10 < 1.6 is true, because 1.10 == 1.1.
Using Versions package to get the right value.


ewels:
Nice, good spot - thanks!

Issue #251, opened 2016-07-04
Samtools stats / samtools idxstats


ewels:
Most people seem to run `samtools stats` for a consistent view of the alignment scores for a dataset. It would be nice to have a simple bar plot showing this information; the current beeswarm plot shows a lot of information and it's not super quick to understand what the alignment rate is.

Need to cherry pick categories and remove them from beeswarm. Would be good to write in a check to make sure that we get the correct percentages too.

See #250 for some discussion.

ewels:
Done in #257

Only mapped and unmapped at the moment, but that's probably fine.

tantrev:
Just a request: it would be quite nice if MultiQC had support for samtools' "idxstats" and "flagstat" commands. :P

ewels:
Hi @tantrev,

There is already support for `stats` and `flagstat` in v0.8dev. Just grab a development copy of the code, though I'm hoping to release v0.8 relatively soon too.

I have previously thought about adding support for `idxstats` but I'm not sure how to visualise it properly. It's the number of aligned reads per chromosome, right?

PHil

tantrev:
Awesome, thank you!  I'll grab a copy right now.

And yep, you got it.  I was more just being lazy and wanted to use `idxstats `for parsing but I'd imagine a parallel coordinates plot could do the job?

ewels:
Hmm, interesting idea. I could probably trick a regular MultiQC line graph into looking like a parallel coordinates plot - could look good!

ewels:
Damn, next problem is finding output from `idxstats`. Goes to standard out, so no common file name. Doesn't really have _any_ identifying information within the file:
```
1       249250621       860208  1
2       243199373       1106007 2
3       198022430       545145  1
4       191154276       359862  0

[...]

GL000225.1      211173  0       0
GL000192.1      547496  0       0
NC_007605       171823  9       0
hs37d5  35477943        817     0
*       0       0       4
```

Can't really just parse any files that contain four columns. Might have to just rely on users putting the phrase `idxstats` somewhere in the file path in this case 😞

tantrev:
A small price to pay for standardized convenience. 👍   And absolutely, a hacked line graph could definitely be a winner.

ewels:
So what's the number we want here - number of mapped reads divided by chromosome length? Or percentage aligned? I'm guessing that reporting just the number of mapped reads will just plot the chromosome lengths...

Also, how can unaligned reads be assigned to a chromosome?

[...]

PR #255, opened 2016-07-06
Samtools improvements

robinandeer:
Add a bargraph plot for mapped/unmapped reads similar to that of "star"-module

See #251


robinandeer:
- [ ] Use mixins for Picard and

multiqc/modules/samtools/stats.py
```diff
@@ -10,125 +10,158 @@
...
-                        parsed_data['{}_percent'.format(k)] = (parsed_data[k] / parsed_data['raw_total_sequences']) * 100
+class ParseReportMixin():
+
+    def parse_reports(self):
+        """ Find Samtools stats logs and parse their data """
+
+        self.samtools_stats = dict()
+        for f in self.find_log_files(config.sp['samtools']['stats']):
+            parsed_data = dict()
+            relevant_lines = [line for line in f['f'].splitlines()
```
ewels: Any reason why doing it this way is better?

robinandeer: oh not really but I find it easier to read when one block does one thing :)

robinandeer:
I've skipped a couple modules for now bc they have a lot more stuff in them that I'm sure if this is the right approach for them :P

ewels:
This has suddenly become really huge quite quickly. Probably a bit late now, but any chance that you could remove the mixin stuff from this PR so that I can see what else has changed? Can do the mixin thing in a separate PR afterwards..

robinandeer:
yeah sry about that - noticed it grew quite quickly - I'll separate it out somehow
maybe open a couple new PRs :P

ewels:
Awesome, thanks 👍

PR #256, opened 2016-07-06
add very basic gatk module with CompOverlap support

robinandeer:
Adds a table summarising the results from the CompOverlap section

I need some help with:

- what to add to the general table
- what to add from the other output sections

GATK outputs lots of stuff so I'm not sure how much to put in the module but I guess I'll let it grow organically for now

### DIFF
--- a/multiqc/modules/gatk/__init__.py
+++ b/multiqc/modules/gatk/__init__.py
@@ -1,0 +1,2 @@
+# -*- coding: utf-8 -*-
+from .core import MultiqcModule
--- a/multiqc/modules/gatk/core.py
+++ b/multiqc/modules/gatk/core.py
@@ -1,0 +1,113 @@
+# -*- coding: utf-8 -*-
+from collections import OrderedDict
+import csv
+
+from multiqc import config, BaseMultiqcModule, plots
+
+
+class MultiqcModule(BaseMultiqcModule):
+
+    """Parse out information from GATK Variant Eval."""
+
+    def __init__(self):
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__(
+            name='GATK',
+            anchor='gatk',
+            target='GATK',
+            href='https://www.broadinstitute.org/gatk/',
+            info=(" is a toolkit offering a wide variety of tools with a "
+                  "primary focus on variant discovery and genotyping."))
+
+        # Parse output for all the files
+        samples = {}
+        for log_file in self.find_log_files(config.sp['gatk']):
+            lines = log_file['f'].splitlines()
+            compoverlap_output = compoverlap_parse(lines)
+            compoverlap_data = compoverlap_values(compoverlap_output)
+            samples[log_file['s_name']] = compoverlap_data
+
+        self.sections = [{
+            'name': 'GATK CompOverlap',
+            'anchor': 'compoverlap',
+            'content': compoverlap_table(samples)
+        }]
+
+
+def compoverlap_values(novelties):
+    """Parse data per sample for compoverlap output."""
+    data = {}
+    for novelty in novelties:
+        if novelty['Novelty'] == 'all':
+            data['reference'] = novelty['CompRod']
+            data['comp_rate'] = float(novelty['compRate'])
+            data['concordant_rate'] = float(novelty['concordantRate'])
+            data['eval_variants'] = int(novelty['nEvalVariants'])
+            data['novel_sites'] = int(novelty['novelSites'])
+        elif novelty['Novelty'] == 'known':
+            data['known_sites'] = int(novelty['nEvalVariants'])
+    return data
+
+
+def compoverlap_parse(lines):
+    """Parse out data from CompOverlap tool."""
+    all_lines = iter(lines)
+    # scan until CompOverlap header
+    for line in all_lines:
+        if line.startswith('#:GATKTable:CompOverlap'):
+            break
+
+    # collect lines until empty lines
+    relevant_lines = []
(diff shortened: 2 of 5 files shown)


robinandeer:
I'll add some sample data to the test data repo

robinandeer:
Also it might be an idea to be more specific than to name the module "gatk" - what do you think?

ewels:
See the usage docs: https://www.broadinstitute.org/gatk/gatkdocs/org_broadinstitute_gatk_tools_walkers_varianteval_VariantEval.php

1. No standard filenames. Need to find these logs based on their content
2. Content varies according to the flags used. Need to focus on one section only and find by content.

I think it's good to have a module `GATK`, but make this a submodule to that called `VariantEval`.

For knowing which bits are interesting, really it's best to ask people who use this tool the most. My guess would be that the `concordantRate` is interesting. Also maybe `nSNPs` / `nMNPs` / `nInsertions` / `nDeletions` / `nComplex` counts? Lots of info though, so need to be careful about visualisation.

I guess doing just `nEvalVariants` for `known` and `novel` in the General stats would be good. Nice to quickly see which samples have lots of novel variants for example. I'd probably start with those two columns, plus `concordantRate` for all variants.

robinandeer:
Thanks @ewels! Yeah I noticed I need to find by contents

Do you mean have a `GATK` module like `Picard` with submodules? I think that's good!

I was thinking of doing a bargraph for the different  types of variants - should be pretty good

I've focus mostly on solving the technical problems and then I can try to talk to plp about what to show :)

ewels:
Yeah exactly - GATK has a load of different outputs so I expect that more will be added in the future.

Bargraph sounds good, though note that SNPs will vastly outnumber the others.

robinandeer:
If you use the percentage option it seems to work pretty well even if the SNP outnumber the rest :)

robinandeer:
Updated version of the output:

[multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/352634/multiqc_report.html.zip)

robinandeer:
I'll await some comments, I feel like it's a good first draft

ewels:
Cool! Some quick minor things, will have a proper look later if that's ok.

1. Add a percentage suffix to the table for comp rate: `'format': '{:.1f}%'`
2. Make the numbers floats in the table otherwise the colours / bars in the background don't work
3. Set the namespace for the table: `'namespace': 'GATK'` (done by default for the General Stats but not other tables it would seem)
4. Navigation subtitles don't need to include `GATK` in the name as well, as already clear. Maybe make them more readable too? _"Compare Overlap"_ / _"Count Variants"
5. Add `.vcf` to the default file name cleaning in [the config](https://github.com/ewels/MultiQC/blob/908d8273b0739bcc9073cd5958d7be79d4ee4cb9/multiqc/utils/config.py#L43-L44)
6. Would be nice to add longer descriptions about what the columns mean in the second table, if possible
7. Maybe better to have the variant counts above the comparisons?
8. CompOverlap table is a bit empty, so probably room to add a few more fields.
9. _..was about to mention adding to the docs but I just saw a commit come in_ 😀

robinandeer:
2. Do we need colors? What does the colors actually mean here?
3. What do you mean? Do I set the namespace per "header" section?
8. Should I move TiTv ratios there or keep it in the general header?

robinandeer:
Not quite sure why it's failing...

ewels:
It's the final test command that uses @lpantano's `--file-list` parameter:
```
multiqc -f --file-list data/special_cases/file_list.txt
```
```python
[ERROR  ]         multiqc : Oops! The 'gatk' MultiQC module broke... 
                    Please copy the following traceback and report it at https://github.com/ewels/MultiQC/issues 
                    (if possible, include a log file that triggers the error) 
============================================================
Module gatk raised an exception: Traceback (most recent call last):
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/multiqc-0.8.dev0-py2.7.egg/EGG-INFO/scripts/multiqc", line 297, in multiqc
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/multiqc-0.8.dev0-py2.7.egg/multiqc/modules/gatk/core.py", line 49, in __init__
    general_headers = titv.general_headers(reference)
UnboundLocalError: local variable 'reference' referenced before assignment
============================================================
```

Looks like it's coming from your new GATK code?

Phil

ewels:
ps.

2/  Yes! The answer is always yes! 😛  Same as the General Stats table at the top, the colours and bars help to visualise the data for better / faster interpretation.

8/  Probably fine to keep it in the header? I think this is a fairly key statistic. The Qualimap module also puts that into the general stats table.

Phil

ewels:
To be more specific - I think it's because your code doesn't raise a `UserWarning` if it finds no logs. You won't have seen this until the automated tests because I guess that you will have always tested your code with an example log file. So when the automated test uses `--file-list` and specifies which input files to use (not including any GATK logs), it dies.

Simple fix, you just need something like this after the parsing step:

```python
if len(samples) == 0:
  log.debug("Could not find any GATK Variant Eval reports")
  raise UserWarning
```

ewels:
Ok, rewritten and merged 😀 

Sorry about being really OCD and rewriting your code, I feel bad about it. But I thought it would save time in the long run to use the same coding patterns as elsewhere in MultiQC. I kept all of the stuff that actually takes me longer to do - the parsing logic, values to parse and the output style so the resulting report looks pretty much identical (and great!).

Thanks again for your work on this, I think this module will be used a lot..

PR #257, opened 2016-07-07
Samtools stats update

robinandeer:
Oki, here's a single purpose PR for updating the samtools stats module.

It moves mapped and unmapped reads to a bargraph 

PS. add `&w=1` to the URL when you go over the diff 😉 

And you can close #255

### DIFF
--- a/multiqc/modules/samtools/stats.py
+++ b/multiqc/modules/samtools/stats.py
@@ -26,8 +26,8 @@ class ParseReportMixin():
                 field = field.replace(' ','_')
                 value = float(sections[2].strip())
                 parsed_data[field] = value
-            if len(parsed_data) > 0:
-                
+
+            if len(parsed_data) > 0:                
                 # Work out some percentages
                 if 'raw_total_sequences' in parsed_data:
                     for k in list(parsed_data.keys()):
@@ -89,7 +89,10 @@ class ParseReportMixin():
             for s_name in self.samtools_stats:
                 if s_name not in self.general_stats_data:
                     self.general_stats_data[s_name] = dict()
-                self.general_stats_data[s_name].update( self.samtools_stats[s_name] )
+                self.general_stats_data[s_name].update( self.samtools_stats[s_name])
+
+            # Make bargraph plot of mapped/unmapped reads
+            self.sections.append(alignment_section(self.samtools_stats))
             
             # Make dot plot of counts
             keys = OrderedDict()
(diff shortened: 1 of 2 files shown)


multiqc/modules/samtools/stats.py
```diff
@@ -10,125 +10,163 @@
...
+            for line in f['f'].splitlines():
+                if not line.startswith("SN"):
+                    continue
+                sections = line.split("\t")
+                field = sections[1].strip()[:-1]
+                field = field.replace(' ','_')
+                value = float(sections[2].strip())
+                parsed_data[field] = value
+
+            if len(parsed_data) > 0:
```
ewels: Excess trailing whitespace! OCD alert! :trollface:

multiqc/modules/samtools/stats.py
```diff
@@ -10,125 +10,163 @@
...
+    bedgraph_data = {}
+    for sample_id, data in samples_data.items():
+        expected_total = data['raw_total_sequences']
+        read_sum = (data['reads_mapped'] + data['reads_unmapped'])
+        if read_sum == expected_total:
+            bedgraph_data[sample_id] = data
+        else:
+            log.warn("sum of mapped/unmapped reads not matching total, "
+                     "skipping samtools plot for: {}".format(sample_id))
+    bargraph = alignment_chart(bedgraph_data)
```
ewels: Any reason to put this into its own function? I confess that I generally prefer to have nice linear code, if a function is only used once I find it kind of confusing to have to keep jumping around the source to find what's going on. Is there any advantage to it?

multiqc/modules/samtools/stats.py
```diff
@@ -10,125 +10,163 @@
...
+    section = {
+        'name': 'Samtools Stats: general',
+        'anchor': 'samtools-stats-general',
+        'content': ("<p>This modules parses mapping metrics from Samtools. "
+                    "All numbers are in millions.</p> {}".format(bargraph))
```
ewels: Are they? I don't think they are? Note the config dicts above are being merged with previously defined dicts that have common variables. Specifically, this line: `'modify': lambda x: float(x) / 1000000.0,`

multiqc/modules/samtools/stats.py
```diff
@@ -10,125 +10,163 @@
...
+        expected_total = data['raw_total_sequences']
+        read_sum = (data['reads_mapped'] + data['reads_unmapped'])
+        if read_sum == expected_total:
+            bedgraph_data[sample_id] = data
+        else:
+            log.warn("sum of mapped/unmapped reads not matching total, "
+                     "skipping samtools plot for: {}".format(sample_id))
+    bargraph = alignment_chart(bedgraph_data)
+    section = {
+        'name': 'Samtools Stats: general',
```
ewels: If this contains just an alignment bargraph, maybe `Alignments` or something would be a better name?

multiqc/modules/samtools/stats.py
robinandeer: my stupid linter stopped working 😝

multiqc/modules/samtools/stats.py
robinandeer: oh, it's just a preference of mine

every function I try to keep to doing one thing. I know testing is a ways a way but if I'd write tests for this code it would make it easier to have this separate since it doesn't interact with the rest of the class :)

But it's good to unify code structure across a project so tell me if you want me to update!

multiqc/modules/samtools/stats.py
robinandeer: good call! lazy me just copy pasted - sry

multiqc/modules/samtools/stats.py
robinandeer: hah! just what I went for - great minds... 🍞

PR #258, opened 2016-07-07
Added flagstat submodule to samtools

HLWiencko:
Creates a beeswarm report for samtools flagstat output. 

It lists each category as pass/fail -- it might make more sense to have pass/fail on the same line in the plot, just with different colours. Could also represent as percentages, but I'm not sure how useful that is. 

Next to do: fix it so output labels at the bottom of the table don't run into the plot, combine pass/fail into the same line. 

multiqc_report

### DIFF
--- a/multiqc/modules/samtools/flagstat.py
+++ b/multiqc/modules/samtools/flagstat.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python
+
+""" MultiQC submodule to parse output from Samtools flagstat """
+
+import logging
+from collections import OrderedDict
+from multiqc import config, plots
+
+# Initialise the logger
+log = logging.getLogger(__name__)
+
+def parse_reports(self):
+    """ Find Samtools flagstat logs and parse their data """
+
+    self.samtools_flagstat = dict()
+    for f in self.find_log_files(config.sp['samtools']['flagstat']):
+        parsed_data = dict()
+        for line in f['f'].splitlines():
+            if not line.startswith("SN"):
+                continue
+            sections = line.split("\t")
+            field = sections[1].strip()[:-1]
+            field = field.replace(' ','_')
+            value = float(sections[2].strip())
+            parsed_data[field] = value
+        if len(parsed_data) > 0:
+            
+            # Work out some percentages
+            if 'raw_total_sequences' in parsed_data:
+                for k in list(parsed_data.keys()):
+                    if k.startswith('reads_') and k != 'raw_total_sequences':
+                        parsed_data['{}_percent'.format(k)] = (parsed_data[k] / parsed_data['raw_total_sequences']) * 100
+            
+            if f['s_name'] in self.samtools_flagstat:
+                log.debug("Duplicate sample name found! Overwriting: {}".format(f['s_name']))
+            self.add_data_source(f)
+            self.samtools_flagstat[f['s_name']] = parsed_data
+
+    if len(self.samtools_flagstat) > 0:
+
+        # Write parsed report data to a file
+        self.write_data_file(self.samtools_flagstat, 'multiqc_samtools_flagstat')
+
+        # General Stats Table
+        self.general_flagstat_headers['error_rate'] = {
+            'title': 'Error rate',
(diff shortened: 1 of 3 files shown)


ewels:
Hi @HLWiencko - this looks great!

Whilst we can use different colours, I'd prefer not to as this conflicts with the sample highlighting.

Could we perhaps just _only_ show passing numbers, but set the max value to be the total number of reads? We could have the % of total reads in the `description` which would show when hovered over..

This plot type is fairly new and immature. I'm afraid that there's no way to stop text running over other than making it shorter. I usually truncate it but keep the full name in `description` so it can be seen on hover.

Phil

ewels:
Just looked at the code - looks good generally, though I'm not keen on parsing based on line number sorry. Two things:

1. People like to concatenate files or prepend their own things and every other weirdness you can imagine (all of which will break this)
2. Tool authors often change outputs slightly, for example adding extra categories. Any minor change will break this again.

I think it's probably better to use regexes instead - then we're sure that we're pulling out the correct number. Worst case scenario, nothing will be found which is better than incorrectly presenting a number as something else.

It should be a relatively simple change, for an example see the [STAR module](https://github.com/ewels/MultiQC/blob/908d8273b0739bcc9073cd5958d7be79d4ee4cb9/multiqc/modules/star/star.py#L53-L82).

ewels:
Example to get you started (apologies if this is teaching you to suck eggs):
```python
re_groups = ['passed', 'failed', 'passed_pct', 'failed_pct']
regexes = {
  'properly_paired'; r"(\d+) \+ (\d+) properly paired \(([\d\.]+|nan)%:([\d\.]+|nan)%\)"
}

for k, r in regexes.items():
  r_search = re.search(r, raw_data, re.MULTILINE)
  if r_search:
    for i,j in enumerate(re_groups):
      try:
        key = "{}_{}".format(k, j)
        parsed_data[key] = float(r_search.group(i+1))
      except:
        pass
```

..or something. Written in GitHub and untested so probably something silly there.

ewels:
..one more thing - you have a merge conflict because Robin [updated the main samtools file](https://github.com/ewels/MultiQC/pull/257). So you'll need to pull in these new changes and resolve the merge conflict before this PR can be resolved.

I guess you're still sat next to Robin, so should hopefully be easy to ask him what you need to change for this.

HLWiencko:
I've changed the parser to use regexps, as suggested, and I resolved the conflicts in samtools.py. It runs, but the report html doesn't render (the flagstat box hangs at "loading"). Hopefully this is a simple enough fix.

ewels:
Hi @HLWiencko, this looks great - thanks! As it's not breaking the tests, I'll merge now and see if I can fix the report html. I'll post here with any changes.

Phil

ewels:
Ok, working now @HLWiencko 😀 

Problem was very simple - you had the key names wrong for the plot. The parsing function was adding `_passed` and `_failed` to each variable. For the plot you had `_pass` and `_fail`.

Fixed in a600a652dc69555eb0c732c8722a8b28da9152c6 and now working:

image

I might remove some of the categories, it's kind of a huge plot!

Phil

ewels:
Ok, I updated the plot a bit - the module now calculates the total read count (adding together QC passed and QC failed) and puts this at the top. As all rows have the `shared_key` as `read_count`, they all have the same scale as this top row. This means that there's not really any need to show those which _aren't_ properly paired, that _aren't_ mapped and so on. So I could get rid of all of those rows. Plot now looks like this:

image

I also fixed a couple of regexes - looks like files can have `nan` and also `-nan`.

ewels:
Ok, last changes - I added the number of mapped reads to the general stats table. I think your new module is pretty much done for now! 👍 

(Sorry for spamming your inbox 😉 )

HLWiencko:
Awesome! Those changes are pretty much what I'd have done if I'd had more time. Thanks again for accepting the contribution, and I'm hoping to get a chance to come back and add more some time in the future.

ewels:
Great! Glad you're happy with the changes :)

PR #261, opened 2016-07-07
Catch errors if not pconfig or keys given to bargraph

lpantano:
If `cat` is `None` It will get a `TypeError` exception, because `None` has not `_get_item` (by default, `cat` is `None`)

If I don't give `pconfig`, then only the first bargraph works and the rest are blank. If I re-define config inside the function again, then it works.

I don't fully understand what is happening with pconfig, but it is happening something. 

Not sure if this is the best of the solutions.

### DIFF
--- a/multiqc/plots/bargraph.py
+++ b/multiqc/plots/bargraph.py
@@ -44,6 +44,9 @@ def plot (data, cats=None, pconfig={}):
     :return: HTML and JS, ready to be inserted into the page
     """
     
+    if not pconfig:
+        pconfig = {}
+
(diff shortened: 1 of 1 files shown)


ewels:
If someone omits the `pconfig` when calling the function, it should default to an empty dict, no?
```python
def plot (data, cats=None, pconfig={}):
```

I guess these changes don't hurt though, and would help if someone explicitly pass `False` to the function or something else weird.

No problem with adding `TypeError` catch. I'm surprised that it didn't already do this from when I wrote the function and originally tested it, but whatever.

Phil

PR #262, opened 2016-07-07
added BaseDistributionByCycle report parsing to picard module

mlusignan:


ewels:
Nice! Thanks @mlusignan - will merge then go through and reference here.

ewels:
Couple of minor changes here:

* Removed the average base content from the General Stats table. Four columns is quite a lot, and I'm not sure that this info will be used _that_ much..
* Wrote the summary stats to a file, so people can easily get them if they want them.
* Added a `ymax` of 100% to the plots. This is so that when you're switching between different bases the y axis doesn't change, makes it easier to compare.

ewels:
(sorry, forgot to reference the issue in the second commit - stats written to a file in ff1cb7d9621ee06d58f1a0f4e68937974e4c14d0 )

ewels:
Just added cases to catch `AssertionErrors` triggered by your checks. Whilst these are nice to have, they crash the entire Picard module if one bad file is found. Now they effectively just skip that sample, so all others (including in other submodules) will still run.

mlusignan:
Thanks for cleaning up my commits!  The assertions were ugly, but both cases seemed indicative of a format change and so important to flag.  I wasn't sure what the appropriate way to identify them in the MultiQC framework was, so I left the checks as assertions.

Also, thank you for all your time at the hackathon!  I had a great time, and I am grateful to have had the opportunity to contribute.

ewels:
Yup, absolutely - they made sense. But you'd be surprised at how often MultiQC encounters truncated / broken files, so I try to make it handle these things fairly quietly. If the Picard output format changes hopefully someone will report it as a bug quite quickly 😉 

Thanks for helping out with MultiQC! Glad you enjoyed it, sorry I couldn't be there in person!

PR #267, opened 2016-07-15
Closing logger handle before attempting to remove it.

brainstorm:
This is a followup on PR #265 since we are still seeing stale log handlers not closing up.

PR #268, opened 2016-07-18
Use logging.shutdown() to cleanly shutdown the logging facilities

brainstorm:
This is related to PR #267.


brainstorm:
Or rather, that `logger.shutdown()` could be in `scripts/multiqc` towards the end as the python docs suggest: "Put this at the end of your application"... but let's try it out and see if it helps via @ohofmann ;)

ewels:
Ok nice, thanks! Yes as it's only a few lines it probably doesn't need to be outside of the main script. Can move it another time though.

ohofmann:
That did the trick! Thanks Roman and Phil!

PR #285, opened 2016-08-23
QUALIMAP:allow `,` as thousands separator

lpantano:
We have a problem if only `.` are allowed to search for the number. We got only the first numbers as the numbers of reads, i.e:

`exonic =  5,055,857` would be `5`, and that's wrong :)

thanks @roryk and @marypiper

### DIFF
--- a/multiqc/modules/qualimap/QM_RNASeq.py
+++ b/multiqc/modules/qualimap/QM_RNASeq.py
@@ -25,10 +25,10 @@ def parse_reports(self):
         'ambiguous_alignments': r"ambiguous alignments\s*=\s*([\d,]+)",
         'not_aligned': r"not aligned\s*=\s*([\d,]+)",
         '5_3_bias': r"5'-3' bias\s*=\s*([\d\.]+)",
-        'reads_aligned_exonic': r"exonic\s*=\s*([\d\.]+)",
-        'reads_aligned_intronic': r"intronic\s*=\s*([\d\.]+)",
-        'reads_aligned_intergenic': r"intergenic\s*=\s*([\d\.]+)",
-        'reads_aligned_overlapping_exon': r"overlapping exon\s*=\s*([\d\.]+)",
+        'reads_aligned_exonic': r"exonic\s*=\s*([\d\.\,]+)",
+        'reads_aligned_intronic': r"intronic\s*=\s*([\d\.\,]+)",
+        'reads_aligned_intergenic': r"intergenic\s*=\s*([\d\.\,]+)",
+        'reads_aligned_overlapping_exon': r"overlapping exon\s*=\s*([\d\.\,]+)",
     }
     for f in self.find_log_files(sp['rnaseq_results']):
         d = dict()


ewels:
Hah, what are the chances?! I had an e-mail in my inbox reporting this bug this morning and I fixed it myself about twenty minutes ago (was still testing, hadn't pushed yet).

I've made exactly the same change, plus I have made a bunch of other improvements, so I'll reject this PR if that's ok. Note I changed the regex to `[\d,]` instead of the `[\d\.\,]` that you've used here as the counts should always be integers. I don't think commas need to be escaped?

Great minds think alike.. 😉  (apologies for missing the bug in the first place though!)

Phil

ewels:
In fact, perhaps you guys can help me with the question that I'm facing before I push @lpantano @roryk @marypiper

The example Kidney data has the following stats in the report:
```
>>>>>>> Reads alignment

    reads aligned  = 10,135,485
    total alignments = 10,135,485
    secondary alignments = 0
    non-unique alignments = 4,464,055
    aligned to genes  = 4,255,142
    ambiguous alignments = 85,895
    no feature assigned = 1,330,393
    not aligned = 0


>>>>>>> Reads genomic origin

    exonic =  4,255,142 (76.18%)
    intronic = 980,432 (17.55%)
    intergenic = 349,961 (6.27%)
```

I changed the General Stats column from `aligned to genes` (a count) to `percentage aligned to genes`.  `aligned to genes` seems to always be the same number as `exonic`. To get the percentages to look right, I figured that the `Reads genomic origin` section must be based on a total count of unique, unambiguous alignments (`total alignments` - `non-unique alignments` - `ambiguous alignments`).

However, this breaks when I look at data from a paired-end run:
```
>>>>>>> Reads alignment

    reads aligned (left/right) = 3,295,794 / 3,295,797
    read pairs aligned  = 3,295,792
    total alignments = 7,758,876
    secondary alignments = 1,167,285
    non-unique alignments = 1,511,158
    aligned to genes  = 1,329,862
    ambiguous alignments = 28,642
    no feature assigned = 1,650,052
    not aligned = 2,596,101


>>>>>>> Reads genomic origin

    exonic =  1,329,862 (44.63%)
    intronic = 1,032,186 (34.64%)
    intergenic = 617,866 (20.73%)
    overlapping exon = 156,843 (5.26%)
```

Here, the percentages in `Reads genomic origin` seem to ignore `overlapping exon` and give a total count of `2,979,914`. I cannot recreate this number from the `Reads alignment` section using any combination, so I'm now confused. Calculating the percentage in genes as above gives `21.4%` instead of the `44.63%` reported to be in exons with the same read count.

I'm stumped. Any ideas how the Qualimap authors are getting to the total count for genomic origin of reads? I could just report the `% exonic` in the General Stats table I guess, but this seems like a bit of a cop-out - I'm reporting read pairs aligned alongside so it seems strange to have the percentage in genes calculated from a different number.

ewels:
Also - anyone else think that using exon overlaps is dubious for `aligned to genes`? Personally I would count an intronic alignment as being within a gene..

ewels:
Ok, I have just dropped that column from the table. Avoids any confusion. See commit 2284e64012e85ab760ca8fd766bcfac981b4362f

Thanks all!

lpantano:
Thanks Phil! awesome you was working on that today! good timing! :) thanks for the fixes :)

PR #288, opened 2016-08-31
Improve descriptions of section headers

chapmanb:
Some of the output section headers we're a bit too technical and require
knowledge of the individual programs (ie you need to know samtools stats
deals with alignments). This tries to make some small improvements to
make MultiQC reports more accessible to biologists with less tool experience.

### DIFF
--- a/multiqc/modules/samtools/stats.py
+++ b/multiqc/modules/samtools/stats.py
@@ -128,7 +128,7 @@ class StatsReportMixin():
             plot_html = plots.beeswarm.plot(self.samtools_stats, keys,
                                             {'id': 'samtools-stats-dp'})
             self.sections.append({
-                'name': 'Alignment metrics from samtools stats',
+                'name': 'Alignment metrics',
                 'anchor': 'samtools-stats',
                 'content': "<p>This module parses the output from <code>samtools stats</code>. All numbers in millions.</p> {}".format(plot_html)
             })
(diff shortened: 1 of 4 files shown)


ewels:
This is great, thanks! In fact, you've touched on something that I'd like to do across the board for all modules - improve the names and especially the descriptions above each plots to explain what they're showing. They are all a bit minimal currently, fine if you're already familiar with the program, but not ideal if you're not a bioinformatician and someone sends you a report.

I might make some comments with a few alternative suggestions if that's ok, as some of your names are quite long and I prefer having short titles in the side navigation if possible:

image

Thanks for kickstarting this, I'll make a new issue for the general all-modules task.

multiqc/modules/snpeff/snpeff.py
```diff
@@ -44,23 +44,23 @@ def __init__(self):
         # Report sections
         self.sections = list()
         self.sections.append({
-            'name': 'Counts by Genomic Region',
+            'name': 'Variant Counts by Genomic Region',
```
ewels: Perhaps just `Variants by Genomic Region`?

multiqc/modules/samtools/stats.py
```diff
@@ -128,7 +128,7 @@ def parse_samtools_stats(self):
             plot_html = plots.beeswarm.plot(self.samtools_stats, keys,
                                             {'id': 'samtools-stats-dp'})
             self.sections.append({
-                'name': 'Samtools Stats',
+                'name': 'Alignment metrics from samtools stats',
```
ewels: Maybe just `Alignment metrics`?

ewels: Maybe also the other section in this module as `Percent Mapped` instead of `Alignment`?

ewels:
See #289 for the general issue.

chapmanb:
Phil -- thanks so much for the helpful feedback. I shortened these based on your suggestions and happy to help with the more long term project of improving explanations for each tool.

ewels:
Great, thanks!

PR #312, opened 2016-10-11
samtools rmdup

avilella:
initial version of parsing STDERR from running samtools rmdup -- took code ideas from the Samblaster module


ewels:
Great! Will review the code properly lately, but I can see one missing step straight away - you need to add code to the main `samtools.py` file so that this script runs. For example, [here](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/samtools/samtools.py#L12) and [here](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/samtools/samtools.py#L49-L51).

PR #313, opened 2016-10-11
Samtools rmdup module

avilella:
ta

### DIFF
--- a/multiqc/modules/samtools/rmdup.py
+++ b/multiqc/modules/samtools/rmdup.py
@@ -14,47 +14,68 @@ log = logging.getLogger(__name__)
 class RmdupReportMixin():
 
     def parse_samtools_rmdup(self):
-        """ Go through log file looking for rmdup output.
-        If the
-        Grab the name from the RG tag of the preceding bwa command """
+        """ Find Samtools rmdup logs and parse their data """
 
-        # Example below:
-        # [bam_rmdupse_core] 26602816 / 103563641 = 0.2569 in library '   '
+        self.samtools_rmdup = dict()
+        for f in self.find_log_files(config.sp['samtools']['rmdup']):
+            parsed_data = parse_single_report(f['f'])
+            if len(parsed_data) > 0:
+                if f['s_name'] in self.samtools_rmdup:
+                    log.debug("Duplicate sample name found! Overwriting: {}".format(f['s_name']))
+                self.add_data_source(f, section='rmdup')
+                self.samtools_rmdup[f['s_name']] = parsed_data
+
+        if len(self.samtools_rmdup) > 0:
+
+            # Write parsed report data to a file (restructure first)
+            self.write_data_file(self.samtools_rmdup, 'multiqc_samtools_rmdup')
+
+        return len(self.samtools_idxstats)
+
+def parse_single_report(f):
+    """ Parse a samtools rmdup """
+
+    """ Go through log file looking for rmdup output.
+    If the
+    Grab the name from the RG tag of the preceding bwa command """
+
+    # Example below:
+    # [bam_rmdupse_core] 26602816 / 103563641 = 0.2569 in library '   '
         
-        dups_regex = "\[bam_\w+_core\] (\d+) / (\d+) = \((\d+.\d+)\) in library"
-        # dups_regex = "rmdup: (Removed|Marked) (\d+) of (\d+) \((\d+.\d+)%\) read ids as duplicates"
-        input_file_regex = "rmdup: Opening (\S+) for read."
-        # rgtag_name_regex = "\\\\tID:(\S*?)\\\\t"
-        data = {}
-        s_name = None
-        fh = f['f']
-        for l in fh:
-            # try to find name from RG-tag. If bwa mem is used upstream rmdup with pipes, then the bwa mem command
-            # including the read group will be written in the log
+    dups_regex = "\[bam_\w+_core\] (\d+) / (\d+) = \((\d+.\d+)\) in library"
+    # dups_regex = "rmdup: (Removed|Marked) (\d+) of (\d+) \((\d+.\d+)%\) read ids as duplicates"
+    input_file_regex = "rmdup: Opening (\S+) for read."
+    # rgtag_name_regex = "\\\\tID:(\S*?)\\\\t"
+    parsed_data = {}
+    s_name = None
+    for l in f.splitlines():
+        # try to find name from RG-tag. If bwa mem is used upstream rmdup with pipes, then the bwa mem command
+        # including the read group will be written in the log
 
-            # match = re.search(rgtag_name_regex, l)
-            # if match:
-            #     s_name = match.group(1)
+        # match = re.search(rgtag_name_regex, l)
+        # if match:
+        #     s_name = match.group(1)
 
-            # try to find name from the input file name, if used
-            match = re.search(input_file_regex, l)
-            if match:
-                basefn = os.path.basename(match.group(1))
-                fname, ext = os.path.splitext(basefn)
-                # if it's stdin, then try bwa RG-tag instead
-                if fname != 'stdin':
-                    s_name = fname
+        # try to find name from the input file name, if used
+        match = re.search(input_file_regex, l)
+        if match:
+            basefn = os.path.basename(match.group(1))
+            fname, ext = os.path.splitext(basefn)
+            # if it's stdin, then try bwa RG-tag instead
+            if fname != 'stdin':
+                s_name = fname
 
-            match = re.search(dups_regex, l)
-            if match:
-                data['n_dups'] = int(match.group(1))
-                data['n_tot'] = int(match.group(2))
-                fract_dups = float(match.group(3))
-                data['pct_dups'] = fract_dups*100
+        match = re.search(dups_regex, l)
+        if match:
+            parsed_data['n_dups'] = int(match.group(1))
+            parsed_data['n_tot'] = int(match.group(2))
+            fract_dups = float(match.group(3))
+            parsed_data['pct_dups'] = fract_dups*100
 
-        if s_name is not None:
-            s_name = self.clean_s_name(s_name, f['root'])
-            if s_name in self.rmdup_data:
-                log.debug("Duplicate sample name found in {}! Overwriting: {}".format(f['fn'], s_name))
-            self.add_data_source(f, s_name)
-            self.rmdup_data[s_name] = data
+    if s_name is not None:
+        s_name = self.clean_s_name(s_name, f['root'])
+        if s_name in self.rmdup_data:
+            log.debug("Duplicate sample name found in {}! Overwriting: {}".format(f['fn'], s_name))
+        # self.add_data_source(f, s_name)
+        # self.rmdup_data[s_name] = parsed_data
+    return parsed_data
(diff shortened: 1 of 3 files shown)


ewels:
Please don't close and reopen pull requests - any commits you push to your repository will automatically be appended to this PR..

avilella:
Uops, apologies...

ewels:
Bad news - now that your new code is actually being run, your build is failing 😉 
https://travis-ci.org/ewels/MultiQC/builds/166744380

```python
Module samtools raised an exception: Traceback (most recent call last):
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/multiqc-0.9.dev0-py2.7.egg/EGG-INFO/scripts/multiqc", line 345, in multiqc
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/multiqc-0.9.dev0-py2.7.egg/multiqc/modules/samtools/samtools.py", line 54, in __init__
    n['rmdup'] = self.parse_samtools_rmdup()
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/multiqc-0.9.dev0-py2.7.egg/multiqc/modules/samtools/rmdup.py", line 30, in parse_samtools_rmdup
    fh = f['f']
NameError: global name 'f' is not defined
```

Try running this locally with example data and getting it to work. Let me know when it's ready to review, or if you need any help..

Phil

avilella:
Ah, I see. I must have left something out from the copy+pasting...

avilella:
I see what I did. I saved the file mid-way through editing it and chopped
half the code off...

I'll fix it on my end...

avilella:
Did another git push. It compiles now, and I think I am testing correctly
(should it show version v0.9.dev0?) but it does not find my *rmdup* files
in the directory:

```
install=/bi/home/rocks-av/avilella/multiqc_install && export
PYTHONPATH=$PYTHONPATH:$install/lib/python2.7/site-packages/ &&
/bi/home/rocks-av/avilella/multiqc_install/bin/multiqc --force --filename
report_$(basename $PWD)_multiqc $PWD/
[INFO   ]         multiqc : This is MultiQC v0.9.dev0
[INFO   ]         multiqc : Searching
'/bi/group/cegx/CEGX_Run215/CEGX_Run215-12345678/'
[WARNING]         multiqc : No analysis results found. Cleaning up..
```

avilella:
Still clueless as to why it's not finding the file pattern...

ewels:
I'm pretty busy for the next couple of weeks, but I'll take a look when I get a chance..

ewels:
Hi @avilella,

Do you have some example data available so that I can test your code? If you could submit that in a pull request to https://github.com/ewels/MultiQC_TestData that would be brilliant..

I'll have a read through your code now too to see if I can see anything.

Phil

multiqc/modules/samtools/rmdup.py
```diff
@@ -0,0 +1,81 @@
...
+
+            # Write parsed report data to a file (restructure first)
+            self.write_data_file(self.samtools_rmdup, 'multiqc_samtools_rmdup')
+
+        return len(self.samtools_idxstats)
+
+def parse_single_report(f):
+    """ Parse a samtools rmdup """
+
+    """ Go through log file looking for rmdup output.
```
ewels: Better to have this as a single docstring rather than splitting into two.

multiqc/modules/samtools/rmdup.py
```diff
@@ -0,0 +1,81 @@
...
+    If the
```
ewels: "If the"..?

multiqc/modules/samtools/rmdup.py
```diff
@@ -0,0 +1,81 @@
...
+    Grab the name from the RG tag of the preceding bwa command """
+
+    # Example below:
+    # [bam_rmdupse_core] 26602816 / 103563641 = 0.2569 in library '   '
+        
+    dups_regex = "\[bam_\w+_core\] (\d+) / (\d+) = \((\d+.\d+)\) in library"
+    # dups_regex = "rmdup: (Removed|Marked) (\d+) of (\d+) \((\d+.\d+)%\) read ids as duplicates"
```
ewels: Please delete commented out lines if they're not comments..

multiqc/modules/samtools/rmdup.py
```diff
@@ -0,0 +1,81 @@
...
+    input_file_regex = "rmdup: Opening (\S+) for read."
+    # rgtag_name_regex = "\\\\tID:(\S*?)\\\\t"
+    parsed_data = {}
+    s_name = None
+    for l in f.splitlines():
+        # try to find name from RG-tag. If bwa mem is used upstream rmdup with pipes, then the bwa mem command
+        # including the read group will be written in the log
+
+        # match = re.search(rgtag_name_regex, l)
```
ewels: As above with deleting unused code.. It's all available in the version control history 😁

multiqc/modules/samtools/rmdup.py
```diff
@@ -0,0 +1,81 @@
...
+        # if match:
+        #     s_name = match.group(1)
+
+        # try to find name from the input file name, if used
+        match = re.search(input_file_regex, l)
```
ewels: All of your code in this function to find the sample name from the log is not used, as the function doesn't do anything with it - it's not returned.

I think what's happening here is that you're mixing chunks of copied code from multiple modules which have different strategies. If you're planning to try to set the sample name from the file contents you probably want to add the parsed data directly to the main dictionary in this function, rather than returning it as a dict.

multiqc/modules/samtools/rmdup.py
```diff
@@ -0,0 +1,81 @@
...
+                self.samtools_rmdup[f['s_name']] = parsed_data
+
+        if len(self.samtools_rmdup) > 0:
+
+            # Write parsed report data to a file (restructure first)
+            self.write_data_file(self.samtools_rmdup, 'multiqc_samtools_rmdup')
+
+        return len(self.samtools_idxstats)
+
+def parse_single_report(f):
```
ewels: If you want to access the `self` variables in this object, you need to change this to:
```python
def parse_single_report(self, f):
```

(doesn't change how you call the function - it will still accept a single argument).

multiqc/modules/samtools/rmdup.py
```diff
@@ -0,0 +1,81 @@
...
+            parsed_data['n_dups'] = int(match.group(1))
+            parsed_data['n_tot'] = int(match.group(2))
+            fract_dups = float(match.group(3))
+            parsed_data['pct_dups'] = fract_dups*100
+
+    if s_name is not None:
+        s_name = self.clean_s_name(s_name, f['root'])
+        if s_name in self.rmdup_data:
+            log.debug("Duplicate sample name found in {}! Overwriting: {}".format(f['fn'], s_name))
+        # self.add_data_source(f, s_name)
```
ewels: As described above, you want to keep these lines here but delete them from the main class `__init__` function.

multiqc/modules/samtools/rmdup.py
```diff
@@ -0,0 +1,81 @@
...
+        # self.rmdup_data[s_name] = parsed_data
+    return parsed_data
```
ewels: Replace `parsed_data` with `self.samtools_rmdup` everywhere in this function, no need to return anything.

multiqc/modules/samtools/rmdup.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class RmdupReportMixin():
+
+    def parse_samtools_rmdup(self):
+        """ Find Samtools rmdup logs and parse their data """
+
+        self.samtools_rmdup = dict()
+        for f in self.find_log_files(config.sp['samtools']['rmdup']):
+            parsed_data = parse_single_report(f['f'])
```
ewels: See comments below, but I don't think that you want this function to return a parsed dictionary. As you're setting the s_name within the function, better to just save the data to `self.samtools_rmdup` directly in the function.

multiqc/modules/samtools/rmdup.py
```diff
@@ -0,0 +1,81 @@
...
+            if len(parsed_data) > 0:
```
ewels: You initialise an empty dictionary in the parsing function, which is always returned. So `len(parsed_data)` is always `1` even if not log files are found currently.

Doesn't really matter if you're going to delete this chunk of code anyway..

multiqc/modules/samtools/rmdup.py
```diff
@@ -0,0 +1,81 @@
...
+                if f['s_name'] in self.samtools_rmdup:
+                    log.debug("Duplicate sample name found! Overwriting: {}".format(f['s_name']))
+                self.add_data_source(f, section='rmdup')
+                self.samtools_rmdup[f['s_name']] = parsed_data
+
+        if len(self.samtools_rmdup) > 0:
+
+            # Write parsed report data to a file (restructure first)
+            self.write_data_file(self.samtools_rmdup, 'multiqc_samtools_rmdup')
```
ewels: Do you not want to create a plot or add anything to the HTML report?

ewels:
I hope all of these review comments make sense - they might be a bit scattered, I got a better picture of what you were trying to do as I went through the code. Shout if you have any questions!

If you submit some example data then I'd be happy to turn this into a working module for you.. :)

avilella:
Thanks, I'll have a look. I'll submit some example data, but it's basically
txt files with 1 single line to parse, which is commented-out before the
parsing line in the code as an example.

ewels:
Ok, I'll try to use that to rewrite your module to parse that properly.

If you could still submit a PR then that would be great - I assume that there are other lines in there? Your comments / regexes suggest at least three other lines.

ewels:
Also, I take it that this comes from streamed `stdout`, rather than samtools saving any files? In which case these can be concatenated, which needs to be taken into account.

This is a pretty extensive rewrite - I will merge your PR into a branch on the repo and then I'll continue working on it. I'll reference this PR in those commits, so we can continue to discuss the new module here.

avilella:
Sounds good to me. The way it usually runs is:

samtools rmdup input.bam 1>out 2>err

And the line that needs parsing goes to err

Otherwise, if people properly label their bamfiles, the library name is
also captured in the single-quoted string... But that would require people
properly producing bam files, and we know that does not happen 100% of the
time.

A.

ewels:
What about these two regexez you wrote though:
```
rmdup: (Removed|Marked) (\d+) of (\d+) \((\d+.\d+)%\) read ids as duplicates
rmdup: Opening (\S+) for read.
```
Where do these lines end up?

ewels:
Ok, I just tested it myself. You weren't joking about it not being very verbose! There actually isn't any stdout, the command takes a second parameter for output file (for me anyway).

So now I'm even more confused as to where those two regexes come from - I can't see output like that anywhere?

avilella:
I think those two regexes are the ones from the samblaster.py code I used
as template for rmdup. So they can be deleted...

ewels:
Ok, I removed the code that should have been deleted, updated the regex to be more specific, changed the file search pattern so that it is no longer dependent on the user's filename and I got the code to parse the sample name from the label if it is present. All in 6e6c733be7663f13c14eed1c8d9f294c7f606f7f

Issue #300, opened 2016-09-18
Generic module for custom output


ewels:
MultiQC relies on standardised output from tools to be able to parse and plot data. This unfortunately excludes results from any scripts that have non-standard output (eg. custom R scripts).

At the moment, we get around this problem by writing [a plugin](https://github.com/ewels/MultiQC_NGI) with modules specific to our pipelines, but I guess this is probably a little intimidating to others.

Instead, I could write a core MultiQC module that uses new MultiQC config options to define how to parse (super simple) input files:

`multiqc_config.yaml`:

``` yaml
custom_data:
  - my_custom_data_type:
      # fn: '*multiqc_scatter.csv'  # Put this in `search_patterns.yaml` / `sp`
      format: csv                   # Default: detect from extension
      section: Analysis Type        # Default: custom_data key / 'Script Output'
      title: My Results             # Default: None
      plot_type: scatter            # Default: Guess from data type (see below)
      pconfig:                      # Default: None
        title: My Scatter Plot
        xlab: X Title
        ylab: Y Title
```

`my_data_multiqc_scatter.csv`:

```
Sample-1, 1.24, 0.34
Sample-2, 8.13, 6.19
Sample-3, 2.48, 0.12
```

Input files would have to follow pretty strict formatting guidelines, but I could put examples of these into the docs.

Any thoughts from anyone? Would this be useful? Are there better ways to implement it?

avilella:
Definitely useful. An example yaml with the different plot types would also
help.

ewels:
More ideas:

* Keep search patterns with everything else, don't move into `custom_data` dict as above
* Have generic filename matching (`*_mqc.(yaml|yml|txt|csv|tsv|log|out)`) and put the config in the header of the file
  * Allows files to work without user-specific configuration, on any installation
  * Add config as yaml to top of file. Comment out with `#` for `.csv` files?
* Allow other formats in addition to csv? `YAML` / `JSON` / `TSV`?
* Have docs with three sets of example files across all plot types:
  * MultiQC config + raw data file
  * Minimal generic file
  * Generic file with all possible config
* Need to think about defaults for minimal generic files.
  * Where to get section ID from? Could call _Script Output_ or something?

Will update top comment example with some comments / changes.

ewels:
I think we can probably do a fairly good job of guessing plot type based on the file structure..

* `num`:`num` = line
* `str`:`num` = bar plot
* `str`:`num`:`num` = scatter plot
[...]

PR #322, opened 2016-10-22
Fix keeping input image files if config.report_imgskips is set

vladsaveliev:
Am I correct that config.report_imgskips parameter is supposed to control whether the images are to be skipped or not during the file look up? In this case this PR fixes it.

### DIFF
--- a/multiqc/utils/report.py
+++ b/multiqc/utils/report.py
@@ -53,7 +53,7 @@ def get_filelist():
         if ftype is not None and ftype.startswith('image'):
             if config.report_imgskips:
                 logger.debug("Ignoring file as has filetype '{}': {}".format(ftype, fn))
-            return None
+                return None
         
         # Limit search to files under 5MB to avoid 30GB FastQ files etc.
         try:


ewels:
Hi @vladsaveliev,

I had to do a bit of detective work to try to remember what my intention was with this one. I think that `config.report_imgskips` was simply to turn on/off the _reporting_ of the skipped images. I wanted to turn this debug statement off by default because sometimes 90% of the verbose log would be the same message (not very helpful). But I added the config option to turn it back on in case anyone wanted to use it for debugging.

The [comment](https://github.com/ewels/MultiQC/blob/master/multiqc_config_example.yaml#L65-L68) in the example config file kind of says this:

> MultiQC skips a couple of debug messages when searching files as the log can get very verbose otherwise. Re-enable here to help debugging.

To be honest I could probably remove this config option and the debug statement, I can't imagine that many people would find or ever use this option.

Is there a reason that you were looking into this particular part of the code?

Phil

vladsaveliev:
Hi Phil,

Thanks for the information, that makes sense.

I have a QC tool (pretty specific and internal, so I'm making a plugin) that outputs a bunch of static PNG images. Basically the tool is a Rmd template, and one the of the plots it generates is PCA, that I wanted to add into my MultiQC page. I was struggling for a while figuring out why the module does not work until I discovered this clause that ignores all images. I know that it's probably not a good idea to add static pictures into a MultiQC report, and it was intended as a temporary solution until I redesign the QC tool to output text data instead, and make JS-based plots in the MultiQC module.

I realise now that the images are ignored because it might affect the performance when checking the "contents" patterns, like in bcftools. Am I correct? So probably specific option to control ignoring images is not a good idea.

ewels:
Hi @vladsaveliev,

Images are ignored because MultiQC can't parse any textual information from them, which is the core of what MultiQC does. In a very early version of MultiQC I used to include the original graphics from FastQC reports, but reports started to get huge file sizes very quickly. I'm keen to keep MultiQC based on numbers rather than blindly importing images if possible so this isn't something that I'm keen to change in the core code. Also it would slow MultiQC down as you say, but this isn't my main concern.

Creating a custom module for a PCA plot is entirely possible though. We have done exactly that with our own plugin for RNA data (see the [edgeR code for the MDS plot](https://github.com/SciLifeLab/NGI-RNAseq/blob/master/main.nf#L669-L680) and the [associated MultiQC module](https://github.com/ewels/MultiQC_NGI/blob/master/multiqc_ngi/modules/ngi_rnaseq/mds_plot.py)). I'd recommend getting your tool to output numeric results as you suggest, if possible.

If that's not going to happen, then it should be possible to get MultiQC to include image files by writing your own file search mechanism. `config.analysis_dir` contains the paths to be searched, so it's a case of finding your file, loading it, base64 encoding it and saving that to the HTML. To be honest, this is probably more work than reworking your PCA script though.

Potentially of interest to you is an idea I've been kicking around for a "generic module" able to parse custom output. See #300 for details. I've spoken to a few people who may be interested in this recently so I may push it up my list of priorities.

Phil

PR #325, opened 2016-10-29
Upgrades for Qualimap BamQC module 

vladsaveliev:
Two proposed improvements in 2 commits:

1). Add a line for human reference to the GC content distribution plot.
screen shot 2016-10-30 at 01 18 07

2). Draw cumulative coverage ("Genome Fraction Coverage") plot for values above 50x, and align it vertically with the coverage histogram.

Currently both plots' X axes are cut independently. The histogram is cut at values below 1% to prevent long flat tail. "Genome Fraction" is cut at 50x because Qualimap doesn't provide more data for it:

screen shot 2016-10-30 at 01 22 13

I propose to calculate cumulative coverage from the histogram data, so it will contain values above 50x; and cut it at exact same threshold:

screen shot 2016-10-30 at 01 19 31

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ dist/
 .Trashes
 ehthumbs.db
 Thumbs.db
+
+.idea/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@
   * Fixed swapped axis labels in the Variant Quality plot
 * **STAR**
   * Fixed crash when there are 0 unmapped reads.
+* **Qualimap BamQC**
+  * Add a line for pre-calculated reference genome GC content
+  * Plot cumulative coverage for values above 50x, align with the coverage histogram.
 
 Core Updates:
 * Fixed bar plot bug where missing categories could shift data between samples
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ Code contributions from:
 [@mlusignan](https://github.com/mlusignan),
 [@HLWiencko](https://github.com/HLWiencko),
 [@guillermo-carrasco](https://github.com/guillermo-carrasco),
-[@avilella](https://github.com/avilella)
+[@avilella](https://github.com/avilella),
+[@vladsaveliev](https://github.com/vladsaveliev)
 and many others. Thanks for your support!
 
--- a/multiqc/modules/qualimap/QM_BamQC.py
+++ b/multiqc/modules/qualimap/QM_BamQC.py
@@ -34,7 +34,7 @@ def parse_reports(self):
     
     # GC distribution - mapped_reads_gc-content_distribution.txt
     self.qualimap_bamqc_gc_content_dist = dict()
-    self.qualimap_bamqc_human_gc_content_dist = dict()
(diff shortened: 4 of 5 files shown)


ewels:
Hi @vladsaveliev,

This is great, thanks! The absence of data above 50X has always annoyed me a little, I'd not realised that we have the data available to recalculate this ourselves - super nice! Also it's always good to reduce the number of files we need to parse.

A few quick questions from my first impressions:

1. Human reference to the GC content distribution plot
  * Is this data _always_ present in the file? I have a vague memory of it being an optional parameter to calculate this in Qualimap meaning that the column isn't always present. I may be wrong though.
  * What happens when the data is non-Human? Does Qualimap still generate this curve for the reference?
2. No comments - looks great!

I'm hoping to have some time to work with MultiQC next week all being well. I'll have a proper go through your code then.

Phil

vladsaveliev:
Thanks Phil!

You are right, the reference GC data is actually only reported if Qualimap was run with the `-gd` parameter, and the supported parameters are only HUMAN and MOUSE. I'm adding a fix so it will only reported if the reference GC column is present.

ewels:
Great! Thanks for fixing the Py3 errors too.

So... What happens if people run MultiQC on a bunch of Qualimap reports that are both Mouse and Human? 😉  (the more popular this tool becomes, the more paranoid about edge cases I become!)

vladsaveliev:
Good point, I'll address this special case as well. It's going to be 2 extra plots. Luckily, mapped_reads_gc-content_distribution.txt reports a header specifying if this is the data for human or mouse.

If you have more weird edge cases in mind, please don't hesitate to share :)

Would you mind if I squash commits once I'm finished? Keeping one for GC and one for coverage.

ewels:
@vladsaveliev - I've read through your code and given it a quick test and it all looks great to me. I haven't been able to test the GC content line as https://github.com/ewels/MultiQC_TestData doesn't have any files with that data, but I'm happy to trust you that it works (if you fancy submitting a PR there with such example data, that would be even better).

I'll leave you to do any commit squashing that you want to do, just let me know when you're ready for me to merge. Also, it would be great if you could add to the changelog and add yourself to the list of contributors at the bottom of the readme.. :)

Thanks again, I think we'll find your changes very helpful!

ewels:
Oh, one minor thing - it would be nice to add a short bit of text above the GC plot when a dotted line is present explaining what it shows.

vladsaveliev:
Okay, I'm finished with my changes. I added a bit explaining the GC dotted line, added test data to MultiQC_TestData, also fixed a bug where QualiMap strips the sample name out of the base folder (`sname.rstrip('.qc')` removes any letter from the list, so `"with_human_gc".rstrip('.qc') == "with_human_g"`). I also put all BamQC titles to a common standard with regards to the uppercase/lowercase; that bothered my perfectionism a bit, so let me know if you are not okay with that.

I rebased my commits and put all of them above yours, so the PR is ready to be merged now without any conflicts. 

Thanks for accepting my improvements and especially for your suggestions and review!

ewels:
Sounds great, thanks for your work!

PR #336, opened 2016-11-04
Added Prokka module

boulund:
I hope this can be useful for someone else as well :).

### DIFF
--- a/multiqc/modules/prokka/prokka.py
+++ b/multiqc/modules/prokka/prokka.py
@@ -93,8 +93,8 @@ class MultiqcModule(BaseMultiqcModule):
         
         Prokka summary files are difficult to identify as there are practically
         no distinct prokka identifiers in the filenames or file contents. This
-        parser makes an attempt using the organism line, expected to be the
-        first line of the file. 
+        parser makes an attempt using the first three lines, expected to contain
+        organism, contigs, and bases statistics. 
         """
 
         s_name = None
--- a/multiqc/modules/prokka/prokka.py
+++ b/multiqc/modules/prokka/prokka.py
@@ -8,7 +8,6 @@ import json
 import logging
 import os
 import re
-from operator import itemgetter
 
 from multiqc import config, BaseMultiqcModule, plots
 
--- a/multiqc/modules/prokka/prokka.py
+++ b/multiqc/modules/prokka/prokka.py
@@ -35,7 +35,7 @@ class MultiqcModule(BaseMultiqcModule):
         log.info("Found {} logs".format(len(self.prokka)))
         self.write_data_file(self.prokka, 'multiqc_prokka')
         
-        # Add Prokka's annotation stats to the general stats table
+        # Add most important Prokka annotation stats to the general table
         headers = OrderedDict()
         headers['organism'] = {
             'title': 'Organism',
@@ -58,44 +58,16 @@ class MultiqcModule(BaseMultiqcModule):
             'description': 'Number of CDS',
             'min': 0,
             'format': '{:i}%',
-            'shared_key': 'prokka',
-        }
-        headers['tRNA'] = {
-            'title': 'tRNA',
-            'description': 'Number of tRNA',
-            'min': 0,
-            'format': '{:i}%',
-            'shared_key': 'prokka',
-            'hidden': True
-        }
-        headers['rRNA'] = {
-            'title': 'rRNA',
-            'description': 'Number of rRNA',
-            'min': 0,
-            'format': '{:i}%',
-            'shared_key': 'prokka',
-            'hidden': True
-        }
-        headers['tmRNA'] = {
-            'title': 'tmRNA',
-            'description': 'Number of tmRNA',
-            'min': 0,
-            'format': '{:i}%',
-            'shared_key': 'prokka',
-            'hidden': True
-        }
-        headers['sig_peptide'] = {
-            'title': 'sig_peptide',
-            'description': 'Number of sig_peptide',
-            'min': 0,
-            'format': '{:i}%',
-            'shared_key': 'prokka',
-            'hidden': True
         }
         self.general_stats_addcols(self.prokka, headers)
         
-        # Make a summary table  
-        self.intro += self.prokka_table()
+        # User can set a configuration attribute, 'prokka_table', to specify
+        # whether to include a table or a barplot. Default is to make a plot.
+        if getattr(config, 'prokka_table', False):
+            self.intro += self.prokka_table()
+        else:
+            self.intro += self.prokka_barplot()
+
         
     def parse_prokka(self, f):
         """ Parse prokka txt summary files. 
@@ -119,6 +91,9 @@ class MultiqcModule(BaseMultiqcModule):
             return
 
         # Get organism and sample name from the first line
+        # Assumes organism name only consists of two words, 
+        # i.e. 'Genusname speciesname', and that the remaining
+        # text on the organism line is the sample name.
         organism = " ".join(first_line.strip().split(":", 1)[1].split()[:2])
         s_name = " ".join(first_line.split()[3:])
         self.prokka[s_name] = dict()
@@ -138,15 +113,18 @@ class MultiqcModule(BaseMultiqcModule):
 
     
     def prokka_table(self):
-        """ Make basic plot of the annotation stats """
+        """ Make basic table of the annotation stats """
         
         # Specify the order of the different possible categories
         headers = OrderedDict()
+        headers['organism'] = { 
+                'title': 'Organism', 
+                'description': 'Organism name',
+        }
         headers['contigs'] = { 
                 'title': '# contigs', 
                 'description': 'Number of contigs in assembly',
                 'format': '{:i}',
-                'colour': 'red',
         }
         headers['bases'] = { 
                 'title': '# bases', 
@@ -184,10 +162,30 @@ class MultiqcModule(BaseMultiqcModule):
                 'format': '{:i}',
         }
         
-        # Config for the plot
         table_config = {
             'namespace': 'prokka',
             'min': 0,
         }
         
         return plots.table.plot(self.prokka, headers, table_config)
+    
+    def prokka_barplot(self):
+        """ Make a basic plot of the annotation stats """
+
+        # Specify the order of the different categories
+        keys = OrderedDict()
+        keys['CDS'] =           { 'name': 'CDS' }
+        keys['rRNA'] =          { 'name': 'rRNA' }
+        keys['tRNA'] =          { 'name': 'tRNA' }
+        keys['tmRNA'] =         { 'name': 'tmRNA' }
+        keys['misc_RNA'] =      { 'name': 'misc RNA' }
+        keys['sig_peptide'] =   { 'name': 'Signal peptides' }
+
+        plot_config = {
+            'id': 'prokka_plot',
+            'title': 'Prokka',
+            'ylab': '# Counts',
+            'cpswitch_counts_label': 'Features'
+        }
+
+        return plots.bargraph.plot(self.prokka, keys, plot_config)
--- a/multiqc/modules/prokka/prokka.py
+++ b/multiqc/modules/prokka/prokka.py
@@ -61,11 +61,12 @@ class MultiqcModule(BaseMultiqcModule):
         }
         self.general_stats_addcols(self.prokka, headers)
         
-        # User can set a configuration attribute, 'prokka_table', to specify
-        # whether to include a table or a barplot. Default is to make a plot.
+        # User can set configuration attributes, 'prokka_table', and
+        # 'prokka_barplot', to specify whether to include a table or a barplot, or both.
+        # Default is to make a plot only.
         if getattr(config, 'prokka_table', False):
             self.intro += self.prokka_table()
-        else:
+        if getattr(config, 'prokka_barplot', True):
             self.intro += self.prokka_barplot()
 
         
--- a/multiqc/modules/prokka/prokka.py
+++ b/multiqc/modules/prokka/prokka.py
@@ -41,52 +41,61 @@ class MultiqcModule(BaseMultiqcModule):
             'title': 'Organism',
             'description': 'Organism',
         }
-        #headers['contigs'] = {
-        #    'title': 'Contigs',
-        #    'description': 'Number of contigs',
-        #    "min": 0,
-        #}
+        headers['contigs'] = {
+            'title': 'Contigs',
+            'description': 'Number of contigs',
+            'min': 0,
+        }
         headers['bases'] = {
             'title': 'Bases',
             'description': 'Number of bases',
-            "min": 0,
-            'format': '{:i}%'
+            'min': 0,
+            'format': '{:i}%',
+            'hidden': True
         }
         headers['CDS'] = {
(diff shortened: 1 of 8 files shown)


multiqc/modules/prokka/prokka.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+
+        s_name = None
+
+        # Look at the first three lines, they are always the same
+        first_line = f['f'].readline()
+        contigs_line = f['f'].readline()
+        bases_line = f['f'].readline()
+        # If any of these fail, it's probably not a prokka summary file
+        if not all((first_line.startswith("organism:"),
```
ewels: Learn something new every day (or every PR). Not come across `all` before.. 👍

multiqc/modules/prokka/prokka.py
```diff
@@ -0,0 +1,153 @@
...
+        # Look at the first three lines, they are always the same
+        first_line = f['f'].readline()
+        contigs_line = f['f'].readline()
+        bases_line = f['f'].readline()
+        # If any of these fail, it's probably not a prokka summary file
+        if not all((first_line.startswith("organism:"), 
+                   contigs_line.startswith("contigs:"),
+                   bases_line.startswith("bases:"))):
+            return
+        log.debug("%s appears to be a prokka summary file.", f['f'].name)
```
ewels: This is probably not necessary.. Instead add the following after you're sure that the file has parsed:
```python
self.add_data_source(f, s_name)
```

This will add a line to `multiqc_data/multiqc_sources.txt` describing the file. See the [docs here](http://multiqc.info/docs/#printing-to-the-sources-file).

multiqc/modules/prokka/prokka.py
```diff
@@ -0,0 +1,153 @@
...
+
+        # Get organism and sample name from the first line
+        organism = " ".join(first_line.strip().split(":", maxsplit=1)[1].split()[:2])
+        s_name = " ".join(first_line.split()[3:])
+        self.prokka[s_name] = dict()
+        self.prokka[s_name]['organism'] = organism
+        log.debug("Found sample name '%s' for organism '%s'", s_name, organism)
```
ewels: This should be in the exported data file? Probably nice to skip it to not to make the logfile too big..

multiqc/modules/prokka/prokka.py
```diff
@@ -0,0 +1,153 @@
...
+        self.prokka[s_name]['contigs'] = int(contigs_line.split(":")[1])
+        self.prokka[s_name]['bases'] = int(bases_line.split(":")[1])
+
+        # Get additional info from remaining lines
+        for line in f['f']:
+            description, value = line.split(":")
+            try:
+                self.prokka[s_name][description] = int(value)
+            except ValueError:
+                log.debug("Unable to parse line: '%s'", line)
```
ewels: or just save as a string?
```python
self.prokka[s_name][description] = value
```

multiqc/modules/prokka/prokka.py
```diff
@@ -0,0 +1,153 @@
...
+        log.info("Found {} logs".format(len(self.prokka)))
+        self.write_data_file(self.prokka, 'multiqc_prokka')
+        
+        # Add Prokka's annotation stats to the general stats table
+        headers = OrderedDict()
+        headers['organism'] = {
+            'title': 'Organism',
+            'description': 'Organism',
+        }
+        #headers['contigs'] = {
```
ewels: Please remove commented out code if it's going to be used..

ewels: (You can include the data but set the column to be hidden by default if you're unsure about how useful it is..)

multiqc/modules/prokka/prokka.py
```diff
@@ -0,0 +1,153 @@
...
+            'title': 'tRNA',
+            'description': 'Number of tRNA',
+            "min": 0,
+            'format': '{:i}%'
+        }
+        headers['rRNA'] = {
+            'title': 'rRNA',
+            'description': 'Number of rRNA',
+            "min": 0,
+            'format': '{:i}%'
```
ewels: Personally I like to colour all of these differently so that they're easier to scan when you're familiar with the module results. Also if something is bad (like rRNA contamination?) then you can set it to reds, or whatever. See the [chroma.js docs](https://github.com/gka/chroma.js/wiki/Predefined-Colors) for the available colour schemes.

multiqc/modules/prokka/prokka.py
```diff
@@ -0,0 +1,153 @@
...
+            'title': 'CDS',
+            'description': 'Number of CDS',
+            "min": 0,
+            'format': '{:i}%'
+        }
+        headers['tRNA'] = {
+            'title': 'tRNA',
+            'description': 'Number of tRNA',
+            "min": 0,
+            'format': '{:i}%'
```
ewels: Could be nice to add `'shared_key':'prokka'` or something to all of these, so that they have the same scale..

multiqc/modules/prokka/prokka.py
```diff
@@ -0,0 +1,153 @@
...
+        }
+        headers['sig_peptide'] = {
+            'title': 'sig_peptide',
+            'description': 'Number of sig_peptide',
+            "min": 0,
+            'format': '{:i}%'
+        }
+        self.general_stats_addcols(self.prokka, headers)
```
ewels: Six columns is bordering on a few too many for the General Stats. Might be worth considering removing a few if they're non-essential, as the table will get very wide and need scrolling if there are virtually any other modules included in the report currently.

multiqc/modules/prokka/prokka.py
```diff
@@ -0,0 +1,153 @@
...
+            except ValueError:
+                log.debug("Unable to parse line: '%s'", line)
+
+    
+    def prokka_barplot (self):
+        """ Make basic plot of the annotation stats """
+        
+        # Specify the order of the different possible categories
+        keys = OrderedDict()
+        #keys['contigs'] =       { 'color': '#437bb1', 'name': 'Contigs' }
```
ewels: Ditch these again if they're not being used..

multiqc/modules/prokka/prokka.py
```diff
@@ -0,0 +1,153 @@
...
+        keys['tmRNA'] =         { 'name': 'tmRNA' }
+        keys['misc_RNA'] =      { 'name': 'misc RNA' }
+        keys['sig_peptide'] =   { 'name': 'Signaling peptides' }
+        
+        # Config for the plot
+        pconfig = {
+            'id': 'prokka_plot',
+            'title': 'Prokka',
+            'ylab': '# Counts',
+            'cpswitch_counts_label': 'Features'
```
ewels: Maybe also `'yDecimals': False`?

multiqc/utils/search_patterns.yaml
```diff
@@ -67,6 +67,8 @@ preseq:
     contents: 'EXPECTED_DISTINCT'
     # contents: 'TOTAL_READS	EXPECTED_DISTINCT'
     # contents: 'TOTAL_BASES	EXPECTED_DISTINCT'
+prokka:
+    contents: 'organism:'
```
ewels: This is horribly unspecific as you say, but I can't think of a better solution.

multiqc/modules/prokka/prokka.py
boulund: Yeah, I also thinks it's better to hide the column then. Didn't even know that was an option (I really should learn to read documentation more thouroughly)....

multiqc/modules/prokka/prokka.py
boulund: Sure, I'll try that.

multiqc/modules/prokka/prokka.py
boulund: Oops! I intended to change them when I was pleased with the overall look of the plot, but then I forgot. I'll see what I can do here.

multiqc/modules/prokka/prokka.py
boulund: Hmm. I totally see what you're getting at. I'll discuss a bit with Kaisa to see what she thinks about this issue (she asked me to write a Prokka module in the first place).

multiqc/modules/prokka/prokka.py
boulund: It has a nice friend called `any` as well. Very useful!

multiqc/modules/prokka/prokka.py
boulund: Wilco!

multiqc/modules/prokka/prokka.py
boulund: Sure thing. I just love extensive debug logs ;)

multiqc/modules/prokka/prokka.py
boulund: Hmm... Not sure about this one. I guess this is just defensive programming. Never seen any malformed files from Prokka, so I don't know if this can ever happen. It would be if we accidentally find a similar file (with the first three lines and all) that isn't actually a Prokka file. 
If it wouldn't hurt storing it as a string instead, I guess we could go with that. What's your gut feeling about this?

multiqc/utils/search_patterns.yaml
boulund: Prokka doesn't have any standard for what these files are called, so I guess this is as close as we can come unless you find something better someday. At least the prokka parser I wrote makes a few additional checks to make sure, which is better than nothing...

multiqc/modules/prokka/prokka.py
ewels: The docs are long.. 😉  Just a case of adding `'hidden': True` to this config dict.

In some places (eg. [FastQC](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/fastqc/fastqc.py#L214)) I check whether all values are identical and hide if so.

multiqc/modules/prokka/prokka.py
ewels: Up to you really - if you think that it should never happen then that's fine. I figured that it wouldn't break anything else in the module and would just add more data to the saved parsed data file if anything is found. But as you say, I guess this module has a higher chance than most of accidentally finding the wrong input file, so maybe leave it as it is..

multiqc/utils/search_patterns.yaml
ewels: Yup, I think it will have to do. Only option for me is to add extra filtering options here I guess (eg. I thought that it supported regex matching until I checked, but that was elsewhere). But probably not worth it, the module should reject the file quickly enough. It's only really a speed / paranoia thing.

multiqc/modules/prokka/prokka.py
```diff
@@ -0,0 +1,153 @@
...
+        bases_line = f['f'].readline()
+        # If any of these fail, it's probably not a prokka summary file
+        if not all((first_line.startswith("organism:"), 
+                   contigs_line.startswith("contigs:"),
+                   bases_line.startswith("bases:"))):
+            return
+        log.debug("%s appears to be a prokka summary file.", f['f'].name)
+
+        # Get organism and sample name from the first line
+        organism = " ".join(first_line.strip().split(":", maxsplit=1)[1].split()[:2])
```
boulund: The code assumes it will be at least :). I'm not 100% here. But I've yet to see any other situations occur.

multiqc/modules/prokka/prokka.py
```diff
@@ -0,0 +1,153 @@
...
+        s_name = " ".join(first_line.split()[3:])
```
boulund: I agree, very annoying. Perhaps I should write a PR to Prokka about this...

multiqc/modules/prokka/prokka.py
boulund: We've cut it down to three columns (contigs, bases, CDS). 

Putting a full table in the prokka section of the report instead.

multiqc/modules/prokka/prokka.py
boulund: We decided to skip the barplot in the end anyway. Not sure what I can do about the colours in the table format? Nothing I do seems to produce any visible change.

multiqc/modules/prokka/prokka.py
ewels: Sounds good!

multiqc/modules/prokka/prokka.py
ewels: Ah.. It could be that if you use the `shared_key` value then it overrides the colour scheme and shares it across all columns with the same shared key? Maybe? It was a while ago that I wrote that code.

ewels:
Ok, so I just ran the new version of the code. If I'm honest, I like the resulting report less than the first version.. A few thoughts:

* If you show all of the columns in the General Stats table, it's essentially identical to the bottom table
  * I generally try to avoid duplicating data in reports. Partly ideological, but also if someone runs this with thousands of samples then the report filesize can get huge surprisingly quickly. So I try to be as lean as possible.
  * It's really easy to get at these numbers if people want to - exporting from plots, the data directory, just looking at tooltips etc. So personally I'd just pick one location or the other.
  * Maybe `Organism`, `Contigs`, `Bases` (hidden) and `CDS` in the top table, then the others in the bottom? Then they're together in their groups which seems more intuitive to me.
* I know I suggested using the shared key thing, and this seemed like a good idea when reading the code. However, looking at the report I think it's a bad idea - the numbers vary so much that it essentially removes the quantitation bar for the smaller numbers, which is not helpful. Also I think you'll be able to colour them without it.
* Do you really find that table easier to interpret than the bar graph? You have to really read the numbers to understand it and it'd be easy to miss that one had an extra digit in comparison to another (especially if they're several rows apart). I definitely found the bar plot a lot easier to interpret.

Apologies, I realise that you made many of these changes in response to my suggestions! I'm probably never going to look at these reports again so I'll leave these design decisions in your hands and merge when you request (the code is fine).

ewels:
ps. I should note that I have a particular hatred of the Comic Sans font, pie charts and large difficult to read tables. So I tend to get carried away more than I should on these topics 😛

boulund:
No need to apologise. To be honest I also like the previous version better. I agree with you that the bar charts are much easier to interpret at a glance, and I understand the potentially negative impact of including duplicates of the same data in the report. 
I'm thinking I'll have another go, producing a new draft of what the report could look like. Hopefully somewhat better suited to the general MultiQC philosophy and look and feel. 
I'll have to confer with Kaisa on what arguments she has for the table representation, perhaps we can find a middle ground where we can run our own plugin version of the module to give her the tables she wants for her analysis pipeline, while keeping the official MultiQC version leaner and cleaner. Does that sound ok to you?

ewels:
Sounds good 👍  Probably no need to write a separate module if you're fixed on the table - the data is all there, so you could wrap the bargraph and the table code in conditionals that check for a config setting (bargraph on by default, table off by default). Then you guys can just add the config options to your personal config files instead of writing a whole new module (plus others can do the same if they want to).

Should just be a case of `if getattr(config, 'prokka_table', False):` - anything found in config files is loaded automatically, so that's all you need in the core code.

ewels:
Great! Only minor thing is that I'd prefer it as two config options `prokka_table` _and_ `prokka_barplot` so that people can have both / neither if they want. But then happy to merge!

ewels:
Perfect, thanks!

PR #343, opened 2016-11-21
Smaller fixes to RseQC module

Hammarn:
This small change fixes the first point on issue #330

Issue #330, opened 2016-10-31
RSeQC module improvements


ewels:
Several minor tweaks for rseqc:

- [x] _Inner Distance_ shows up in general statistics for SE data (0bp)
- [x] _Proper Pairs_ shows up in general stats for SE data (0%)
- [x] Bunch of irrelevant stats in _bam stats_ also show up. Can be hidden.

ewels:
@Hammarn - is all of this already merged?

Hammarn:
Yes, as of PR #345

PR #345, opened 2016-11-23
Rseqc fixes

Hammarn:
This is the same as #343 but with only the changes related to #330

### DIFF
--- a/multiqc/modules/rseqc/bam_stat.py
+++ b/multiqc/modules/rseqc/bam_stat.py
@@ -36,7 +36,7 @@ def parse_reports(self):
         'proper-paired_reads_map_to_different_chrom': r"Proper-paired reads map to different chrom:\s*(\d+)",
     }
     #intiate PE check
-    PE=False
+    is_paired_end=False
     # Go through files and parse data using regexes
     for f in self.find_log_files(config.sp['rseqc']['bam_stat']):
         d = dict()
@@ -59,8 +59,8 @@ def parse_reports(self):
             self.add_data_source(f, section='bam_stat')
             #Check if SE or PE
             
-            if d['read_1'] and d['read_2'] and d['reads_mapped_in_proper_pairs'] and d['proper-paired_reads_map_to_different_chrom'] != 0:
-                PE = True
+            if d['read_2'] != 0:
+                is_paired_end = True
             self.bam_stat_data[f['s_name']] = d
     
     if len(self.bam_stat_data) > 0:
@@ -82,8 +82,8 @@ def parse_reports(self):
         for s_name in self.bam_stat_data:
             if s_name not in self.general_stats_data:
                 self.general_stats_data[s_name] = dict()
-                #Don't write to general stat if nothing to write..
-                if self.bam_stat_data[s_name]['reads_mapped_in_proper_pairs'] != 0:
+                #Only write if PE, i.e. there is something to write
+                if is_paired_end: 
                     self.general_stats_data[s_name].update( self.bam_stat_data[s_name] )
         
         # Make dot plot of counts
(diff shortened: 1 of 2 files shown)


multiqc/modules/rseqc/bam_stat.py
```diff
@@ -35,7 +35,8 @@ def parse_reports(self):
         'reads_mapped_in_proper_pairs': r"Reads mapped in proper pairs:\s*(\d+)",
         'proper-paired_reads_map_to_different_chrom': r"Proper-paired reads map to different chrom:\s*(\d+)",
     }
-    
+    #intiate PE check
+    PE=False
```
ewels: I'd prefer to use slightly more descriptive variable names generally (eg. `is_paired_end`)

multiqc/modules/rseqc/bam_stat.py
```diff
@@ -51,15 +52,20 @@ def parse_reports(self):
...
                 d['proper_pairs_percent'] = (float(d['reads_mapped_in_proper_pairs']) / t)*100.0
-        
+
         if len(d) > 0:
             if f['s_name'] in self.bam_stat_data:
                 log.debug("Duplicate sample name found! Overwriting: {}".format(f['s_name']))
             self.add_data_source(f, section='bam_stat')
+            #Check if SE or PE
+            
+            if d['read_1'] and d['read_2'] and d['reads_mapped_in_proper_pairs'] and d['proper-paired_reads_map_to_different_chrom'] != 0:
```
ewels: I don't think this does what you think it does. Consider this line:
```python
if a and b and c and d != 0:
```
This is directly equivalent to this:
```python
if (a is True) and (b is True) and (c is True) and (d != 0):
```
It is _not_ equivalent to this:
```python
if (a != 0) and (b != 0) and (c != 0) and (d != 0):
```

I would probably write the above like this instead:
```python
paired_end_keys = ['read_1', 'read_2', 'reads_mapped_in_proper_pairs', 'proper-paired_reads_map_to_different_chrom']
if all(d[k] != 0 for k in paired_end_keys):
```

ewels: I don't think this does what you think it does. Consider this line:
```python
if a and b and c and d != 0:
```
This is directly equivalent to this:
```python
if (a is True) and (b is True) and (c is True) and (d != 0):
```
It is _not_ equivalent to this:
```python
if (a != 0) and (b != 0) and (c != 0) and (d != 0):
```

I would probably write the above like this instead:
```python
paired_end_keys = ['read_1', 'read_2', 'reads_mapped_in_proper_pairs', 'proper-paired_reads_map_to_different_chrom']
if all(d[k] != 0 for k in paired_end_keys):
```

multiqc/modules/rseqc/bam_stat.py
```diff
@@ -76,7 +82,9 @@ def parse_reports(self):
         for s_name in self.bam_stat_data:
             if s_name not in self.general_stats_data:
                 self.general_stats_data[s_name] = dict()
-            self.general_stats_data[s_name].update( self.bam_stat_data[s_name] )
+                #Don't write to general stat if nothing to write..
```
ewels: This is a little too harsh a test. It's possible to have perfectly valid paired end data which has 0 reads mapped in proper_pairs. Can we not use the same test as above for paired-end data?

multiqc/modules/rseqc/bam_stat.py
ewels: And in fact, thinking more about it - we probably only need to check the value of `read_2`... I think that should always be `> 0` for paired end data.

multiqc/modules/rseqc/bam_stat.py
```diff
@@ -94,14 +102,16 @@ def parse_reports(self):
         keys['unmapped_reads'] = dict(defaults, **{'title': 'Unmapped', 'description': 'Unmapped reads' })
         keys['mapq_lt_mapq_cut_non'] = dict(defaults, **{'title': 'Non-unique', 'description': 'mapq < mapq_cut (non-unique)' })
         keys['mapq_gte_mapq_cut_unique'] = dict(defaults, **{'title': 'Unique', 'description': 'mapq >= mapq_cut (unique)' })
-        keys['read_1'] = dict(defaults, **{'title': 'Read-1' })
-        keys['read_2'] = dict(defaults, **{'title': 'Read-2' })
+        if PE is False:
```
ewels: `is True`..?

multiqc/modules/rseqc/bam_stat.py
```diff
@@ -94,14 +102,16 @@ def parse_reports(self):
...
+            keys['read_1'] = dict(defaults, **{'title': 'Read-1' }) #
+            keys['read_2'] = dict(defaults, **{'title': 'Read-2' }) #
         keys['reads_map_to_sense'] = dict(defaults, **{'title': '+ve strand', 'description': "Reads map to '+'" })
         keys['reads_map_to_antisense'] = dict(defaults, **{'title': '-ve strand', 'description': "Reads map to '-'" })
         keys['non-splice_reads'] = dict(defaults, **{'title': 'Non-splice reads' })
         keys['splice_reads'] = dict(defaults, **{'title': 'Splice reads' })
-        keys['reads_mapped_in_proper_pairs'] = dict(defaults, **{'title': 'Proper pairs', 'description':'Reads mapped in proper pairs' })
-        keys['proper-paired_reads_map_to_different_chrom'] = dict(defaults, **{'title': 'Different chrom', 'description': 'Proper-paired reads map to different chrom' })
+        if PE is False:
```
ewels: `is True`...?

multiqc/modules/rseqc/inner_distance.py
```diff
@@ -24,16 +24,20 @@ def parse_reports(self):
...
             try:
                 avg_pos = (float(s[0]) +  float(s[1])) / 2.0
-                self.inner_distance[f['s_name']][avg_pos] = float(s[2])
+                parsed_data[avg_pos] = float(s[2])
             except:
                 # Don't bother running through whole file if wrong
                 break
-    
+        # Only add if we actually found something i,e it was PE data
+        if len(parsed_data) > 0:
```
ewels: Great 👍

multiqc/modules/rseqc/bam_stat.py
Hammarn: Yeah, that might be enough, you initially told me to look for all four but I agree that `read_2` should be enough.

ewels: Yeah I know, sorry :wink:

ewels:
Looks good, thanks!

PR #351, opened 2016-11-28
Slamdunk module

t-neumann:
An implementation of a module to summarize Slamdunk reports (https://github.com/t-neumann/slamdunk).


t-neumann:
Changed summary table to match new slamdunk summary output.

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ Module updates:
 * **Qualimap BamQC**
   * Add a line for pre-calculated reference genome GC content
   * Plot cumulative coverage for values above 50x, align with the coverage histogram.
+* [**Slamdunk**](https://github.com/t-neumann/slamdunk) - new module!
+  * Slamdunk is a software tool to analyze SLAMSeq data.
```
ewels: Tiny thing, but I usually put new modules at the top of the list for additional visibility.

README.md
```diff
@@ -57,6 +57,7 @@ Currently, supported tools include:
   * [Samtools](http://www.htslib.org) (`stats`, `flagstats`, `idxstats`, `rmdup`)
   * [SnpEff](http://snpeff.sourceforge.net/)
   * [Subread featureCounts](http://bioinf.wehi.edu.au/featureCounts/)
+  * [Slamdunk](https://github.com/t-neumann/slamdunk)
```
ewels: I rewrote the readme file the other day sorry, to take up less space with the list of modules. If you could pull my changes and update that would be great. Also, I try to make the list alphabetically sorted.

docs/modules/slamdunk.md
```diff
@@ -0,0 +1,24 @@
+---
+Name: Slamdunk
+URL: https://github.com/t-neumann/slamdunk
+Description: >
+    Slamdunk is a tool to analyze SLAM-Seq data.
+---
+
+Blablabla
```
ewels: ^^ :laughing:

docs/modules/slamdunk.md
```diff
@@ -0,0 +1,24 @@
...
+[Cutadapt](https://code.google.com/p/cutadapt/),
```
ewels: Yup, this needs to be written. Would be nice to have a short primer as to what SLAM-Seq actually is..! A quick google doesn't reveal much.

multiqc/modules/slamdunk/slamdunk.py
```diff
@@ -0,0 +1,832 @@
...
+
+from __future__ import print_function
+import logging
+import re
+import numpy as np
+from distutils.version import StrictVersion
+from collections import OrderedDict
+
+from multiqc import config, BaseMultiqcModule, plots
+from curses.ascii import islower
```
ewels: I'd prefer to avoid extra dependencies if possible. Any reason not to just use the core `str.islower()` function?

multiqc/modules/slamdunk/slamdunk.py
```diff
@@ -0,0 +1,832 @@
...
+    Slamdunk module class, parses slamdunk logs.
+    """
+
+    def __init__(self):
+
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__(name='Slamdunk', anchor='slamdunk',
+        href='https://github.com/t-neumann/slamdunk',
+        info="is a tool to analyze SLAMSeq data. "\
+         "More info to come.")
```
ewels: Please add to this, or remove it.

multiqc/modules/slamdunk/slamdunk.py
```diff
@@ -0,0 +1,832 @@
...
+        self.tc_per_readpos_minus = dict()
+        
+        self.nontc_per_utrpos_plus = dict()
+        self.nontc_per_utrpos_minus = dict()
+        
+        self.tc_per_utrpos_plus = dict()
+        self.tc_per_utrpos_minus = dict()
+
+        # Switch string for plots to produce
+        produce = ""
```
ewels: Safer to use a `set` or `list` instead of a `string` for this.

multiqc/modules/slamdunk/slamdunk.py
```diff
@@ -0,0 +1,832 @@
...
+        self.tc_per_utrpos_minus = dict()
+
+        # Switch string for plots to produce
+        produce = "" 
+
+        # Check whether summary also contains number of counted reads
+        extendedSummary = True
+            
+        for f in self.find_log_files(config.sp['slamdunk']['summary'], filehandles = True):
+            extendedSummary = extendedSummary and self.parseSummary(f)
```
ewels: Presumably this means that none of these reports will be shown if a single one is missing? Is it not better to show the section if any logs of this type are found?

multiqc/modules/slamdunk/slamdunk.py
```diff
@@ -0,0 +1,832 @@
...
+            
+        for f in self.find_log_files(config.sp['slamdunk']['PCA'], filehandles = True):
+            self.parsePCA(f)
+            
+        if len(self.PCA_data) == 0:
+            log.debug("Could not find PCA reports in {}".format(config.analysis_dir))
+        else :
+            produce += "PCA"
+            
+        log.info("Parsing UTR rates reports. This can take a while...")
```
ewels: If it can take a while, I'd prefer that it's done outside of MultiQC. The core ethos of MultiQC is that it should parse and visualise processed data, not do any data processing itself. Can this functionality be put into your slamdunk package instead and be written to a file?

multiqc/modules/slamdunk/slamdunk.py
```diff
@@ -0,0 +1,832 @@
...
+        for f in self.find_log_files(config.sp['slamdunk']['utrrates'], filehandles = True):
+            self.parseUtrRates(f)
+            
+        if len(self.utrates_data) == 0:
+            log.debug("Could not find UTR rates reports in {}".format(config.analysis_dir))
+        else :
+            produce += "utrates"
+            
+
+        log.info("Parsing read rates reports.")
```
ewels: I try to have max. one `log.info` statement per module if possible, on completion. With the exception of tools that have multiple distinct submodules (eg. `samtools flagstat` / `samtools idxstats` etc).

multiqc/modules/slamdunk/slamdunk.py
```diff
@@ -0,0 +1,832 @@
...
+        for f in self.find_log_files(config.sp['slamdunk']['tcperutrpos'], filehandles = True):
+            self.parseSlamdunkTCPerUtrpos(f)
+            
+        if len(self.nontc_per_utrpos_plus) == 0:
+            log.debug("Could not find conversion per UTR position reports in {}".format(config.analysis_dir))
+        else :
+            produce += "utrpos"
+        
+        if (produce == "") :
+            log.info("No slamdunk reports found.")
```
ewels: This will show for every user who ever runs MultiQC even if they've never heard of slamdunk. The log would be way too long to read if all modules did this - please convert to a `log.debug`.

multiqc/modules/slamdunk/slamdunk.py
```diff
@@ -0,0 +1,832 @@
...
+            'data_labels': [
+                "Plus Strand +",
+                "Minus Strand -", 
+            ]
+        }
+        
+        cats = [OrderedDict(), OrderedDict()]
+        
+        cats[0]['T>C'] = {
+            'name': 'T>C',
```
ewels: If the `name` is the same as the dictionary key, you don't need to specify it as this will be the default anyway.

multiqc/modules/slamdunk/slamdunk.py
```diff
@@ -0,0 +1,832 @@
...
+            'name': 'C>T',
+            'color': '#9E9AC8'
+        }
+        cats[1]['C>G'] = {
+            'name': 'C>G',
+            'color': '#6A51A3'
+        }
+        
+        
+#         for cat in self.rates_data_plus[self.rates_data_plus.keys()[0]]:
```
ewels: Please remove commented out code.

multiqc/modules/slamdunk/slamdunk.py
t-neumann: That must have slipped in there as auto-import. No reason to use this dependency.

multiqc/modules/slamdunk/slamdunk.py
t-neumann: Nope, I need it. If I remove it, this happens:

[ERROR  ]         multiqc : Oops! The 'slamdunk' MultiQC module broke...
                    Please copy the following traceback and report it at https://github.com/ewels/MultiQC/issues
                    (if possible, include a log file that triggers the error)
============================================================
Module slamdunk raised an exception: Traceback (most recent call last):
  File "/groups/zuber/zubarchive/USERS/tobias/dev/MultiQC/scripts/multiqc", line 346, in multiqc
  File "/groups/zuber/zubarchive/USERS/tobias/dev/MultiQC/multiqc/modules/slamdunk/slamdunk.py", line 140, in __init__
    self.slamdunkOverallRatesPlot()
  File "/groups/zuber/zubarchive/USERS/tobias/dev/MultiQC/multiqc/modules/slamdunk/slamdunk.py", line 616, in slamdunkOverallRatesPlot
    plots.bargraph.plot([self.rates_data_plus,self.rates_data_minus], cats, pconfig)
  File "/groups/zuber/zubarchive/USERS/tobias/dev/MultiQC/multiqc/plots/bargraph.py", line 94, in plot
    thisdict = { 'name': cats[idx][c]['name'], 'data': thisdata }
KeyError: 'name'
============================================================

multiqc/modules/slamdunk/slamdunk.py
ewels: Ok cool. But still, if it's missing for any samples then `extendedSummary` will be `False` and it'll hide the additional column for everything?

t-neumann: That's correct. However, the standard use case is that all samples will be handled within one slamdunk run and also the summary is produced afterwards on all samples. So either you have it for all or it's missing for all. So the issue will only appear when you run MultiQC on several slamdunk runs. Don't know if that's a likely scenario

multiqc/modules/slamdunk/slamdunk.py
t-neumann: Yeah it's possible. Right now I'm producing the boxplots we discussed using this file and I would really dislike producing another extra file with the same data in another format just for MultiQC. But now that I think of it, I could produce the median values and add it to the header - maybe that's the best way to go

multiqc/modules/slamdunk/slamdunk.py
ewels: Is there a reason not to show it if any are found? I try to be defensive with this stuff :)

t-neumann: Well in that particular example it would tell the user that there were 0 reads counted in genes when in fact he just didn't run the extended summary command. Would be quite misleading don't you think?

ewels: Can it not leave the table cell blank instead of adding 0? In fact, I don't think you need this logic at all. If you give MultiQC a config section with a key that is not found in the data, it should just ignore it. So you can define `headers['counted']` and if none of the data has a key called `counted` then it will just be ignored. Otherwise, any samples that do have `counted` will have the data shown, any missing will just have an empty table cell.

t-neumann: Thanks for pointing that out - this is indeed way more convenient

multiqc/modules/slamdunk/slamdunk.py
ewels: Ok, fixed in fb7a547. Should now be fine.

t-neumann: 👍

t-neumann:
I addressed all your points.

As discussed, I left the minimal documentation as it is and deleted all nonsens + updated the website links.

Most pleasing to you probably is that I managed the median calculations to Slamdunk and print them directly to the file header. That way, I didn't need to touch the original file format, neither produce a new file only for MultiQC.

ewels:
Hi @t-neumann,

Thanks for these updates! I just tried running with your example data and it worked really well, the report looks great!

Having seen the generated report, I have a few conceptual / style suggestions, mostly about the top table:

* I try to limit the number of columns in _General Statistics_ for each module to about three or four. You have eight, which is fine when _only_ the slamdunk module is run, but means that if pretty much any other modules run then the table will be really wide. For example, I would guess that a typical user could run FastQC and perhaps Qualimap to get more stats on alignments?
  * My suggestion would be to either remove or hide-by-default the counts of filtered reads
  * If you want to keep this information, you can create a separate stand-alone table within the Slamdunk section. This then allows the General Stats to be a bare minimum of summary statistics, to line up with results from other modules.
* The ordering of columns in the _General Statistics_ table should go from unprocessed numbers on the right to processed on the left. This correlates with the order of modules there (eg. FastQC will be far right, post-processing far left).
  * This is because the table can get wide and stuff on the far right gets hidden. Best to have the final stats most visible, on the left
* You've used a `shared_key` on the General Stats table, which is great. But if it's just read counts then it's probably nicer to use `read_count` instead of `slamdunk_reads` - that way it will also be shared with counts from other modules.
* You've used the same `GnBu` colour scheme for all columns in the General Stats table. I try to use [colour scales](https://github.com/gka/chroma.js/wiki/Predefined-Colors) that give context - _e.g._ `OrRd` for stuff that is bad (filtered counts) and `YlGn` for stuff that is good (Retained, Counted).
* I'm not a huge fan of the bright red bars on the T>C plot. But perhaps it's just too early in the morning for me.
* No data is saved to `multiqc_data` - it's good to use [`write_data_file`](http://multiqc.info/docs/#step-5-writing-data-to-a-file) to do this so that people can do downstream analyses.
  * This could be intentional I guess? The stuff in the General Stats table is already saved by MultiQC, but it could be nice to have the data from the bar plots.

I have a few other minor code-style things, but perhaps I'll submit a PR to you for these instead of requesting line-by-line changes.

Phil

t-neumann:
Hi Phil,

thanks for having a thorough look at everything and your feedback - really helpful!

Regarding the General Statistics Table:

You're right, I guess it will do to put all the individual filter values into a separate table and just keep #sequenced, #mapped, #retained and #counted reads in there. I'll put all the rest into a separate table for the Slamdunk module.

How would it work to "hide" columns per default in the General Statistics table? Will these be collapsed columns which you could uncollapse?

I'll flip the column order to make the most final thing the outer left most.

Nice hint about the coloring scheme, I'll adapt this.

Regarding the red bar color - do you have any other suggestions? It's basically the most relevant bar which should be paid most attention to, that's why I had it this way.

Totally forgot to add those plots to the multiqc_data file - not intentional. I'll add those. Just had it for the general stats table since I copied the cutadapt module as template.

ewels:
So hidden columns are not shown by default, but can be unfolded through the "Configure Columns" modal. It's good for stuff which isn't typically useful, but maybe interesting if people want to dig into more details. However, I have some doubts about how many users have ever clicked this modal, so people probably won't find the data.

My only problem with the red was that it's super bright! How about something like this?
image

```python
cats[0]['T>C'] = { 'color': '#2171B5' }
cats[0]['A>T'] = { 'color': '#6BAED6' }
cats[0]['A>G'] = { 'color': '#C6DBEF' }
cats[0]['A>C'] = { 'color': '#74C476' }
cats[0]['T>A'] = { 'color': '#C7E9C0' }
cats[0]['T>G'] = { 'color': '#fdbf6f' }
cats[0]['G>A'] = { 'color': '#D9D9D9' }
cats[0]['G>T'] = { 'color': '#969696' }
cats[0]['G>C'] = { 'color': '#525252' }
cats[0]['C>A'] = { 'color': '#DADAEB' }
cats[0]['C>T'] = { 'color': '#9E9AC8' }
cats[0]['C>G'] = { 'color': '#6A51A3' }
```

ewels:
Also, the [Travis build](https://travis-ci.org/ewels/MultiQC/builds/183217246) now shows a Python 3 specific error..

ewels:
Ok, so I've traced the latest weirdness in how the tables are displaying things. It's two separate issues:

* Your `'format'` strings were broken: you had `'format': '{:.f}'` and it should have been something like `'format': '{:.2f}'`
  * This broke the code, triggering an exception catch which apparently doesn't fire very often.
  * This contained old code from an earlier version of the `BaseModule`, before I wrote the `modify` stuff. This old code overwrote the modified values, meaning that the `modify` function changes were lost.
  * Changing the format strings to a valid syntax fixed the `modify` updates and you get millions.
  * Note that your lambda functions should force `floats`, as currently you get `ints`. Something like this works: `'modify': lambda x: float(x) / 1000000.0`
* The colour problem was because I had some code in there that forced columns with the same `shared_key` to have the same colour scale.
  * This probably made sense in my head at the time, but seems like a bad feature now. I'll remove this.

I'll push a commit to the main MultiQC repo fixing these two issues now. Note that you'll want to make a few changes too - fixing the format strings and using floats in the lamda function.

I should appoint you as chief bug hunter! You're great at finding these really obscure bugs deep in the MultiQC code 😉 

Phil

t-neumann:
Wow that's really a nasty edge-case ^^

No it's all working fine. Also the Travis build runs again, I used a deprecated function.

All points you have raised have been addressed with the recent commits. Let me know if something else comes up.

t-neumann:
Merged your PR and rechecked it with my tests. Looks all fine, thanks for reviewing.

I also added the zero cat hiding = FALSE to my other barplot, otherwise no further changes.

ewels:
Great stuff - I think this is good to go! Thank you for your patience with my somewhat stringent reviewing 😁

t-neumann:
Yeah no worries - the code and report improved a lot, so thanks for your advice!

Issue #344, opened 2016-11-21
FastQC: overrepresented sequences


ewels:
Some people have requested that MultiQC reports show if there are very many over-represented sequences. For MultiQC reports, we don't really care what the sequences are - just if they are accounting for a lot of the library.

I think that a potential way of doing this is to sum the total library percentage accounted for by the top 5 overrepresented sequences, and plot this as a simple bar plot.

ewels:
Done in #353

PR #353, opened 2016-11-29
Add overrepresented sequences barplot to fastqc

Hammarn:
Due to popular request I wrote this plot. 
It sums the overrepresented sequences and generates a bar plot over them. If no sample have more than 1% then only show a text instead of the plot. 
See issue #344

### DIFF
--- a/multiqc/modules/fastqc/fastqc.py
+++ b/multiqc/modules/fastqc/fastqc.py
@@ -86,7 +86,9 @@ class MultiqcModule(BaseMultiqcModule):
         
         # Add to the general statistics table
         self.fastqc_general_stats()
-        
+       
+
+    
         # Add the statuses to the intro for multiqc_fastqc.js JavaScript to pick up
         statuses = dict()
         for s_name in self.fastqc_data:
@@ -108,8 +110,8 @@ class MultiqcModule(BaseMultiqcModule):
         self.n_content_plot()
         self.seq_length_dist_plot()
         self.seq_dup_levels_plot()
+        self.overrepresented_sequences_plot()
         self.adapter_content_plot()
-
 
     def parse_fastqc_report(self, file_contents, s_name=None, f=None):
         """ Takes contents from a fastq_data.txt file and parses out required
@@ -171,7 +173,6 @@ class MultiqcModule(BaseMultiqcModule):
                             self.dup_keys.append(float(s[0]))
                         except ValueError:
                             self.dup_keys.append(s[0])
-        
         # Tidy up the Basic Stats
         self.fastqc_data[s_name]['basic_statistics'] = {d['measure']: d['value'] for d in self.fastqc_data[s_name]['basic_statistics']}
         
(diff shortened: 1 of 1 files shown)


multiqc/modules/fastqc/fastqc.py
```diff
@@ -556,7 +557,52 @@ def seq_dup_levels_plot (self):
...
         })
         
-        
+    
+    def overrepresented_sequences (self):
+        """Sum the percentages of overrepresented sequences and display them in a bar plot"""
+
+        data = dict()
+        for s_name in self.fastqc_data:
+            data[s_name]=dict()
```
ewels: If you move this into the `try` block (after calculating `total_pcnt`), the sample key won't be created for samples that don't have this data. Then you won't need the `try`/`except` block later on in the code (L586)

multiqc/modules/fastqc/fastqc.py
```diff
@@ -556,7 +557,52 @@ def seq_dup_levels_plot (self):
...
+            try:
+                total_pcnt = sum( [ float(d['percentage']) for d in self.fastqc_data[s_name]['overrepresented_sequences']] )
+                data[s_name]['overrepresented']=total_pcnt
+            except KeyError:
+                continue
```
ewels: Better to use `log.warning` or `log.debug` with an explanation of what's gone wrong and for which sample..

multiqc/modules/fastqc/fastqc.py
```diff
@@ -556,7 +557,52 @@ def seq_dup_levels_plot (self):
...
+            except KeyError:
+                continue 
+        # Config for the plot
+        pconfig = { 
+            'id': 'overrepresented_sequences',
+            'title': 'Overrepresented sequences',
+            'ymin': 0,
+            'ymax': 100,
+            'tt_percentages': False,
+            'ylab_format': '{value}%',
```
ewels: Would be good to have an axis label here - "Percentage of all reads" or something.

multiqc/modules/fastqc/fastqc.py
```diff
@@ -556,7 +557,52 @@ def seq_dup_levels_plot (self):
...
+            'cpswitch': False,
+            'use_legend': False
+        }
+        make_plot=False
+        #Check if any samples have more than 1% overrepresented sequences, else don't make plot.
+        for x in data.values():
```
ewels: Can probably replace this with a simpler version:
```python
if max( [ x['overrepresented'] for x in data.values() ) < 1:
  plot_html = '<div class="alert alert-info">{} samples had less than 1% of reads made up of overrepresented sequences</div>'.format(len(data))
else:
  ...
```

multiqc/modules/fastqc/fastqc.py
```diff
@@ -556,7 +557,52 @@ def seq_dup_levels_plot (self):
...
+        for s_name in self.fastqc_data:
+            data[s_name]=dict()
+            try:
+                total_pcnt = sum( [ float(d['percentage']) for d in self.fastqc_data[s_name]['overrepresented_sequences']] )
+                data[s_name]['overrepresented']=total_pcnt
+            except KeyError:
+                continue 
+        # Config for the plot
+        pconfig = { 
+            'id': 'overrepresented_sequences',
```
ewels: Please prepend this with `fastqc_`

multiqc/modules/fastqc/fastqc.py
```diff
@@ -556,7 +557,52 @@ def seq_dup_levels_plot (self):
...
+                continue 
+        if make_plot:
+                   plot_html=plots.bargraph.plot(data, None, pconfig)
+        else:
+            plot_html = '<div class="alert alert-warning">No samples found with overrepresented sequences > 1%</div>'
+            
+        self.sections.append({
+            'name': 'Overrepresented sequences',
+            'anchor': 'fastqc_overrepresented_sequences',
+            'content': '<p> The total amount of overrepresented sequences found in each library. ' +
```
ewels: This is a bit of a complex module, so I'd prefer a bit more verbose a description. Maybe something like this?

> FastQC reports any sequences found which make up more than 0.1% of the total reads. This MultiQC module sums the percentages of all reported over-represented sequences. Finding sequences that are very overrepresented in your data either means that they are highly biologically significant, or indicates that the library is contaminated, or not as diverse as you expected. For more information see the ***FastQC documentation***.

Hammarn:
I think I adressed all your points with my latest commits, please have a look through again @ewels. Especially at the error messages and figure explanation

ewels:
Thanks @Hammarn - this is great! I made a few changes myself, mostly revolving around adding new core features to the barplot. See 0ec49188bcf71f7ccc2dfcc0cc314652a788a505

Main difference was that I split the total percentage into two - the top value and the remaining sum. I'm still not totally sure about this, but I feel like it gives a little more granularity and doesn't really remove anything.

ewels:
fastqc_overrepresented_sequencesi_plot

Hammarn:
Looks nice, I like your change!

PR #357, opened 2016-12-07
Loosen dependencies between modules

tbooth:
As discussed on the Gitter feed, I'm looking to add some unit tests to complement the TestData repository, and I run into the everything-imports-everything problem.
This is my attempt at a fix:
1) Modules only import the specific plots they actually use
2) BaseMultiQC module is no longer imported in the top-level __init__.py
3) Lazy loading of template_mod in two plotting modules, removal from the others
4) Removal of small circular dependency when loading config and logger

The downside that the imports have to be a little more explicit, but I think that trying to minimize the number of import statements at the cost of tight code coupling is probably a false economy here.

And my sample unit test is now able to run. The test doesn't pass - there's another patch for that coming.

Unfortunately, I don't think you can tell GitHub to ignore whitespace changes when diffing files, and my editor is set to prune all trailing whitespace, so the below patch looks a bit of a mess. Rather than trying to put all the trailing spaces back I'm just going to submit what I have for comments now. Using ```git diff --ignore-space-at-eol``` will give a much clearer view of the diff.


ewels:
Wow, @tbooth - you're made of strong stuff! Seriously impressed that you tackled all of this, I appreciate how much of a pain it must have been. I've got caught with circular import problems and other things a few times previously and sort of just dodged my way out instead of properly getting to the root of the problems. I'm sorry for these past transgressions!

You can hide whitespace changes by adding `?w=1` to the diff URL (see [this post](https://github.com/blog/967-github-secrets)). I'll try to have a read through your changes in the next few days.

Phil

tbooth:
Thanks for the tip on the whitespace.

You'll see it's mostly mechanistic changes - the same import statement replaced again and again. This will mean an API change for anyone with custom plugins but I honestly think it will help in the long run.
Actually the key change is the smallest - step 3 allows you to load graph modules without bootstrapping a valid configuration, which was the main blocker for my unit testing.

I'm not entirely sure about #4 but it looks right to me.

I'll just keep tweaking the patch until Travis is happy. Then I'm away for a long weekend so I'll go silent.

I do have some more code in waiting, including a patch to make my unit test pass, but I'll sit on that for now.

ewels:
Thanks for this again @tbooth - I've just had a look through the code and it all looks good (I ended up using your `git diff --ignore-space-at-eol` trick in the end as it was easier).

I'm a little concerned about the break in backwards compatibility, it broke our [MultiQC_NGI](https://github.com/ewels/MultiQC_NGI) plugin modules and I suspect that it could affect quite a few people. It's difficult to know how many people have their own custom MultiQC plugins though. 

Having said this, I think that these changes are worthwhile, and I guess that the problem of breaking backwards compatibility is only going to get worse as time goes on. I'm assuming that there's no way that we can keep your changes without breaking stuff? (I can't think of anything).

**Tests** - having looked at your scripts for this, I'm increasingly thinking that it makes sense to put these into [MultiQC_TestData](https://github.com/ewels/MultiQC_TestData), as that's where the test data is. There are a lot of modules, there will hopefully be a lot of tests as a result, and I'm fairly keen to try to not to bloat this main repo. What is your opinion on this?

**Docs** - there is one small section of the docs that needs updating still ([MultiqcModule Class](http://multiqc.info/docs/#multiqcmodule-class)). It would also be good to add a small section somewhere (maybe at the end of the MultiQC Plugins section) describing these changes and how to update plugin code to make it work again if that's ok.

Cheers,

Phil

ewels:
Hi Tim,

I've been thinking about this more, I will certainly merge this PR and bring in your changes, but I may leave it until _after_ I release v0.9 of MultiQC next week. Once it's released I can then merge this into the new dev version immediately.

I hope this is ok - I just feel a little nervous incorporating such a big change just before a release. This is made worse by the fact that I will be away for two months after Christmas and unable to help people with any resulting fall out.

Phil

tbooth:
Hi Phil,

I'm back at my desk. Thanks for the comments, and yes I agree that merging after 0.9 makes the most sense as these are API-breaking changes. I'll update my PR once the release is made.

Regarding compatibility, I did look to see if auto-loading tricks could transparently replace the hard imports in the ```__init__.py``` files but the code just gets too messy. I'm open to suggestions but I really think that avoiding substantial imports in these files is important because the imports get triggered for every single .py file under the directory, whether it actually uses them or not.

Regarding tests, yes I'll move the unit test to the MultiQC_TestData repo. I'll just leave the test/run_tests.sh script in place and modify it to help people find and run the tests from the data repo.

I'll also look at updating the remaining docs as you pointed out.

Finally, I'm wondering if you'd consider doing a bulk strip of trailing whitespace from the .py files after the 0.9 release? PEP8 does say "Avoid trailing whitespace anywhere." so it's not just me. If not, I'll change my editor settings to leave it in place so I'm not submitting big messy patches to you.

Cheers,

TIM

ewels:
Great stuff, sounds good. No problem about the lack of workarounds, as long as it comes in the next release (and sits in the dev version for a while before full release) I think the changes should be fine.

Once the tests are in place it would be nice to refactor the Travis checks a bit too, so that it uses them. Currently it runs MultiQC a few times, which is probably not really necessary. So this could be changed to run just once, then run all of the tests.

Regarding whitespace - I guess I can do this if you like. As you've probably noticed, I'm not a huge fan of PEP8 (especially the thing about line length) and the code violates this everywhere. My editor maintains whitespace whenever I hit return, which is why those blank lines with whitespace are all over the place. The only thing I have against removing it is that it kind of messes up the git history a bit. But I'm particularly fussy about that either, so if it makes your life easier then that's fine :)

Phil

ewels:
Hi @tbooth,

Many thanks for your patience with this (especially given the burst of activity leading up to a release). v0.9 is now released, so happy to merge this whenever you're ready.

Phil

ewels:
Hi @tbooth,

I thought I could do a little work here myself, so I just [fixed the merge conflicts](c1ac9e5530abeacdf17e66abb28ed89377b3e1fd), [updated the new modules](df667a901df5235ed19db6765e51aaa74575da52) and [made the bash script run on MacOSX](a0a5084d619faaa80848788e029073d41ec2e222). Now merged so that you don't have to keep updating this PR every time some new code is added to the core MultiQC repo.

Hope this is ok! Looks like we're waiting for a TestData PR for the test scripts?

Phil

PR #358, opened 2016-12-07
QualiMap RNAseq: fix parsing 5'-3' bias

vladsaveliev:
Hi Phil,

5'-3' bias in QualiMap RNAseq is represented as float (e.g. 0.93), but the module parses it as decimal instead which leads to having always either 0 or 1 in the report. This PR fixes it.


ewels:
Thanks @vladsaveliev, good catch!

PR #360, opened 2016-12-14
bcftools: fix parsing ts/tv

vladsaveliev:
Fixing a tiny bug in the `bcftools` module that keeps it from exposing ts/tv.

(By the way, that's why I love using IDEs like PyCharm: such mistakes are just automatically highlighted there).


ewels:
Great spot, thanks!

ewels:
I use atom as an IDE usually - you have prompted me to install a linter plugin which looks useful, though it didn't pick up this error so perhaps not as clever as PyCharm..

PR #367, opened 2016-12-20
Added "number of predicted genes" barchart to quast module

epruesse:
The barchart mirrors the existing contig length barchart. Code differences mainly because the thresholds at which quast bins the lengths of the predicted genes is configurable when running quast, so must be detected at runtime.

### DIFF
--- a/multiqc/modules/quast/quast.py
+++ b/multiqc/modules/quast/quast.py
@@ -259,4 +259,7 @@ class MultiqcModule(BaseMultiqcModule):
         cats = [ ourpat.format(low,-high if high else "")
                  for low,high in zip(all_thresholds, all_thresholds[1:]+[None]) ]
 
-        return "\n".join([html, plots.bargraph.plot(data, cats)])
+        if len(cats) > 0:
+            return "\n".join([html, plots.bargraph.plot(data, cats)])
+        else:
+            return None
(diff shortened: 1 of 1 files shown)


multiqc/modules/quast/quast.py
```diff
@@ -53,7 +53,12 @@ def __init__(self):
             'anchor': 'quast-contigs',
             'content': self.quast_contigs_barplot()
         })
-
+        # Number of genes plot
+        self.sections.append({
```
ewels: Please only add this section if the predicted genes are found in the QUAST output. When I run this code I get an empty header + nav link in the report with no content.

ewels:
Hi @epruesse,

As I'm trying to do this release I made the change myself and merged. It'd still be great if you could add some example data to the TestData repository though, so that the automated tests work properly.

Many thanks!

Phil

epruesse:
Thanks Phil @ewels !

I had wondered how to get the plot not to show if it's empty. Perhaps it'd be cleaner if the code that evaluates the sections checked the return value of the `content` function and only added it if that evaluates to `not None`? Performance isn't an issue, really, but it feels wrong to run the function twice.

ewels:
Yes, I agree that would be a sensible thing to add. Would be trivial to add in the template too (why have I not thought of doing this before..?)

ewels:
Ok, first update for v1.0dev pushed! Thanks for the suggestion @epruesse - hopefully makes for cleaner code.

PR #368, opened 2016-12-20
Added more columns to QUAST Assembly Statistics table

epruesse:


ewels:
Thanks! This table is getting pretty wide now and needs horizontal scrolling (on my screen at least). Before I merge, can I first check if it's possible to drop any of these columns? I try to avoid stuff going off-screen where possible.

Alternatively - can any of the names be shortened, and can I double check that the order goes from the most important on the left to the least important on the right (Genome Fraction seems like a fairly key statistic?).

Finally, out of curiosity - how come L50 / L75 aren't integers? Going by your excellent column descriptions it sounds like they should be a count of contigs?

Phil

ewels:
Hi @epruesse,

I'm trying to get a release out now, so I'll merge this. Feel free to continue the discussion below though :)

Thanks!

Phil

epruesse:
L50/L75 probably aren't integers because of the display as "(k)" (really "kContigs"). The numbers reported by QUAST are integers.

epruesse:
I'm working on metagenomes, so I don't have the "genome fraction" statistic in Quast. I'll add one of the reports I get to the tests in a minute.

ewels:
Ok great, thanks!

PR #377, opened 2016-12-26
multiqc: config: Catch FileNotFoundError when git is not installed.

wwood:
Without this patch, the incorrect exception is caught when `git` is not in `PATH`.

### DIFF
--- a/multiqc/utils/config.py
+++ b/multiqc/utils/config.py
@@ -31,7 +31,7 @@ try:
     git_hash = subprocess.check_output(['git', 'rev-parse', 'HEAD'], stderr=subprocess.STDOUT)
     git_hash_short = git_hash[:7]
     version = '{} ({})'.format(version, git_hash_short)
-except subprocess.CalledProcessError:
+except (subprocess.CalledProcessError, FileNotFoundError):
     pass
 os.chdir(cwd)
 


ewels:
Ah rubbish, I hadn't thought of this scenario. Many thanks @wwood. Note that the `FileNotFoundError` exception type doesn't seem to be available in Python 2 though.

Phil

wwood:
So travis says. Well, for the purposes of Guix, that doesn't matter since python3 is used for running multiqc. I'm also not sure of the best way to handle this 2/3 problem, so I'll leave it to others to fix more generally.

Thanks for the quick response.

ewels:
Thanks @wwood - I'm away on holiday now and won't be back until mid-February. A [quick google](http://stackoverflow.com/questions/28633555/how-to-handle-filenotfounderror-when-try-except-ioerror-does-not-catch-it) suggests that using the base exception class `IOError` instead of `FileNotFound` should work for both Python 2 and 3, but I can't test here unfortunately. If you update your PR with that and Travis passes I'd be happy to merge from here though.

Cheers,

Phil

ewels:
Hi @wwood,

I've just pushed a change which just handles _any_ exception triggered by this code. Not great coding practice, but as this is such an unimportant feature I don't think that it should ever halt the execution of MultiQC. As a bonus, it should work with both Python 2 and 3 😅 

Hope this is ok, let me know how you get on with the change. Thanks again for the PR..

Phil

PR #379, opened 2017-01-01
linegraph: allow extra_series to be empty

chapmanb:
Fixes an edge case bug in Qualimap parsing when the species is not set,
it will pass an empty `extra_series` to linegraph:

https://github.com/ewels/MultiQC/blob/9fd1cda139f0295a99f9fa979bc7e6156c56be89/multiqc/modules/qualimap/QM_BamQC.py#L298

resulting in an index error when attempting to access the expected first
element:

https://github.com/ewels/MultiQC/blob/9fd1cda139f0295a99f9fa979bc7e6156c56be89/multiqc/plots/linegraph.py#L95

This provides a more general check to skip this.

### DIFF
--- a/multiqc/plots/linegraph.py
+++ b/multiqc/plots/linegraph.py
@@ -88,7 +88,7 @@ def plot (data, pconfig={}):
 
     # Add on annotation data series
     try:
-        if 'extra_series' in pconfig:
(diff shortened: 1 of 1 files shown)


ewels:
Looks great, thanks!

Issue #202, opened 2016-05-05
Refactor file search mechanism


ewels:
It would be good to rewrite how the log file discovery works, in preparation for the restructuring discussed in #181. Currently, the core script first looks through the analysis directories and collects the relevant files (ignoring big files / those matching exclusion patterns etc). These are then passed on to each module, which then searches through them for logs.

What we could like to do instead is for the core script to do the searching for all modules on that first pass. A list of matching files can then be passed on to each module. This first search would go by filename first, then content, and files matched from one module could then be ignored by searches for the next (unless specified otherwise). Hopefully this will speed up execution time quite a bit, especially when there are a lot of samples. It should also pave the way for future changes if/when we remove modules from the core tool.

ewels:
Would be good to add in an option to search by regex. Also an option for regex + string to only search first `n` lines.

ewels:
Currently, the 5MB filesize limit applies to files matched by filename too. Maybe worth removing this restriction.

ewels:
Ok, _definitely_ going to do this for the next release..!

ewels:
If it doesn't already work, would be nice to be able to search by _both_ filename and content. For example, search all `*.txt` files for a string.

ewels:
Now (v1.0 release) may be as good a time as ever to make the breaking-change of forcing single-level search patterns. The current mixture is confusing and problematic. For example, if _only_ changing one Picard pattern, then _all_ Picard patterns must be specified in the user config, otherwise MultiQC breaks.

Down side is that this will force many people to change their configs. Perhaps could use underscores and try to find the new pattern in case of an Exception?

epruesse:
You use `dict.update()` to merge the yaml config files, which is non-recursive. Making it recursive is simple enough though (loop over `iteritems()`, recurse if `instanceof(item, collections.Mapping)` encountered. A recursive dictionary update function probably just doesn't exist in `dict` because there are so many variants of what you may want to do (e.g. do you want to replace or append in case a list object is encountered?)

ewels:
[...]

PR #383, opened 2017-01-05
Changed search pattern for prokka from contents : 'organism:'  to fn:…

thorellk:
… '*txt'

I changed search pattern for prokka from contents "organism:" to instead search for a file with the extension ".txt" since this is the suffix for all prokka summary statistics files. Unfortunately this is not very specific but I belive this will still be more efficient than searching all files in a folder structure such as the one prokka creates. I have also double checked with @fboulund that the prokka module is written so that it very quickly discards a file if it doesn't start like the one prokka creates, which should make the process as fast as possible.

### DIFF
--- a/multiqc/utils/search_patterns.yaml
+++ b/multiqc/utils/search_patterns.yaml
@@ -93,7 +93,7 @@ preseq:
     # contents: 'TOTAL_READS	EXPECTED_DISTINCT'
     # contents: 'TOTAL_BASES	EXPECTED_DISTINCT'
 prokka:
-    contents: 'organism:'
+    fn: '*.txt'
 qualimap:
     bamqc:
         genome_results:


ewels:
Hi @thorellk, thanks for this! I'm a bit skeptical as other users of MultiQC may have hundreds of thousands of `.txt` files, so I'd like to look into it in more detail first if that's ok. Unfortunately I'm using my phone at the airport and won't be back until mid-Feb, but will have a play then if that's ok.

Phil

ewels:
Just had a look at the prokka code and I agree that it should be fast. Main thing I'd like to check is that it won't interfere with other modules matching the same files. _Should_ be fine though I think...

ewels:
Ok, looked around a bit more. Probably fine :) Thanks!

thorellk:
Hi!

Yeah, I know that it’s a bit suboptimal, I just found it super slow as it was before. Usually I use it (and I guess others might as well) on folder trees that contains both trimmed reads, assemblies, annotations and stuff and that becomes a hell lot of files to look through for the pattern ‘organism:’. I hope this solution will be better but it is a bit annoying that there is no better pattern really…

//kaisa

ewels:
Hi @thorellk,

To follow up - I'm hoping to refactor the file search mechanism soon and I'll make sure that it works if you specify both a filename filter and a search string. Then we can put the string back into the config file here. See #202.

Cheers,

Phil

PR #384, opened 2017-01-06
Change Transcript Profile units to percent instead of bp.

roryk:
Thanks to Tim Slidel for pointing this out.


ewels:
Thanks, good spot!

PR #386, opened 2017-01-11
bcftools stats module: more metrics and depth plot

chapmanb:
Phil;
I've been working on replacing some custom code/metrics with bcftools stats output, and this updates the module to provide some additional useful metrics and a depth plot of variants.

- Add heterozygote/homozygote counts to general statistics.
- Improves metrics naming in general statistics, emphasizing variant
  counts of interest.
- Avoid normalizing to million variants, since that is only useful for
  whole genome runs.
- Add plot of variants by depth for assessing coverage of variants

### DIFF
--- a/multiqc/modules/bcftools/stats.py
+++ b/multiqc/modules/bcftools/stats.py
@@ -25,6 +25,7 @@ class StatsReportMixin():
 
         self.bcftools_stats = dict()
         self.bcftools_stats_indels = dict()
+        depth_data = dict()
         for f in self.find_log_files(config.sp['bcftools']['stats']):
             s_names = list()
             for line in f['f'].splitlines():
@@ -38,6 +39,7 @@ class StatsReportMixin():
                     self.add_data_source(f, s_name, section='stats')
                     self.bcftools_stats[s_name] = dict()
                     self.bcftools_stats_indels[s_name] = dict()
+                    depth_data[s_name] = OrderedDict()
                     self.bcftools_stats_indels[s_name][0] = None # Avoid joining line across missing 0
 
                 # Parse key stats
@@ -70,6 +72,20 @@ class StatsReportMixin():
                     length = float(s[2].strip())
                     count = float(s[3].strip())
                     self.bcftools_stats_indels[s_name][length] = count
+
+                # Per-sample counts
+                if s[0] == "PSC" and len(s_names) > 0:
+                    s_name = s_names[int(s[1])]
+                    fields = ['variations_hom', 'variations_het']
+                    for i, f in enumerate(fields):
+                        self.bcftools_stats[s_name][f] = int(s[i + 4].strip())
+
+                # Depth plots
+                if s[0] == "DP" and len(s_names) > 0:
+                    s_name = s_names[int(s[1])]
+                    bin_name = s[2].strip()
+                    percent_sites = float(s[-1].strip())
+                    depth_data[s_name][bin_name] = percent_sites
 
         if len(self.bcftools_stats) > 0:
 
@@ -109,9 +125,18 @@ class StatsReportMixin():
                 }
                 self.sections.append({
                     'name': 'Indel Distribution',
-                    'anchor': 'bcftools-stats',
+                    'anchor': 'bcftools-stats_indel_plot',
                     'content': linegraph.plot(self.bcftools_stats_indels, pconfig)
                 })
+            # Make line graph of variants per depth
+            if len(depth_data) > 0:
+                pconfig = {'id': 'bcftools_stats_depth', 'title': 'Variant depths',
+                           'ylab': 'Fraction of sites (%)', 'xlab': 'Variant depth',
+                           'ymin': 0, 'ymax': 100, 'categories': True}
+                desc = '<p>Read depth support distribution for called variants</p>'
+                self.sections.append({'name': 'Variant depths',
+                                      'anchor': 'bcftools-stats_depth_plot',
+                                      'content': desc + linegraph.plot(depth_data, pconfig)})
 
         # Return the number of logs that were found
         return len(self.bcftools_stats)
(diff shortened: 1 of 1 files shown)


ewels:
Hi @chapmanb,

Many thanks for this, looks great! Are you happy to wait until I'm back from vacation in mid Feb for a review and merge, or would you like it done sooner? Happy to try to do it on my phone if so (but it'll be less thorough as I can't view the output or test).

Phil

chapmanb:
Phil -- absolutely I think it is your duty to review my PR on your phone in the middle of vacation. Seriously, have an awesome holiday and we can pick this back up when you're back. Hope you have a great time.

multiqc/modules/bcftools/stats.py
```diff
@@ -112,45 +128,56 @@ def parse_bcftools_stats(self):
                     'anchor': 'bcftools-stats',
                     'content': linegraph.plot(self.bcftools_stats_indels, pconfig)
                 })
+            # Make line graph of variants per depth
+            if len(depth_data) > 0:
+                import pprint
+                pprint.pprint(depth_data)
```
ewels: Looks like you have some debugging lines left in here..

multiqc/modules/bcftools/stats.py
```diff
@@ -112,45 +128,56 @@ def parse_bcftools_stats(self):
...
+                pconfig = {'id': 'bcftools_stats_depth', 'title': 'Variant depths',
+                           'ylab': 'Fraction of sites (%)', 'xlab': 'Variant depth',
+                           'ymin': 0, 'ymax': 100, 'categories': True}
+                self.sections.append({'name': 'Variant depths',
+                                      'anchor': 'bcftools-stats',
```
ewels: This anchor needs to be unique and different to the plot above, otherwise the navigation links on the left of the report don't work.

I should probably implement a check for this in the code..

multiqc/modules/bcftools/stats.py
```diff
@@ -112,45 +128,56 @@ def parse_bcftools_stats(self):
...
+                                      'content': linegraph.plot(depth_data, pconfig)})
```
ewels: Would be nice to add a line of text before the plot to explain what it's showing if possible..
```python
'content': '<p>This plot shows..</p>' + linegraph.plot(depth_data, pconfig)
```

multiqc/modules/bcftools/stats.py
```diff
@@ -112,45 +128,56 @@ def parse_bcftools_stats(self):
                     'anchor': 'bcftools-stats',
                     'content': linegraph.plot(self.bcftools_stats_indels, pconfig)
                 })
+            # Make line graph of variants per depth
+            if len(depth_data) > 0:
+                import pprint
+                pprint.pprint(depth_data)
+                pconfig = {'id': 'bcftools_stats_depth', 'title': 'Variant depths',
+                           'ylab': 'Fraction of sites (%)', 'xlab': 'Variant depth',
+                           'ymin': 0, 'ymax': 100, 'categories': True}
```
ewels: Any reason for setting `'categories': True` here? Seems to be a numeric axis?

ewels: Ah ok, just had a play - if I try to make `bin_name` a float when collecting the data I get an error because of one final one called `>500`. Not sure that there's really a way to get around this nicely so best as you have it I guess (though plot isn't as pretty).

ewels:
One other thing - I don't use bcftools myself so don't really have any view on which columns should be in the General Stats table, but it might be good to get feedback from others who use this module. Ping @barrymoore, @vladsaveliev, @darogan and @brentp...

Here are the proposed changes - going from this:
image

To this:
image

Any comments from others?

multiqc/modules/bcftools/stats.py
```diff
@@ -112,45 +128,56 @@ def parse_bcftools_stats(self):
...
     def bcftools_stats_genstats_table(self):
         """ Add key statistics to the General Stats table """
         stats_headers = OrderedDict()
-        stats_headers['number_of_SNPs'] = {
-            'title': 'M SNPs',
-            'description': 'Number of SNPs (millions)',
-            'min': 0,
-            'format': '{:.2f}',
-            'modify': lambda x: x / 1000000
+        stats_headers['number_of_records'] = {
```
ewels: Generally, I try to keep numbers in the General Stats under 1000 - hence the previous columns had the `'modify': lambda x: x / 1000000` to display the figures in millions.

My reasoning is that it's pretty difficult to spot the difference between 5522770 and 552277 when scanning a report quickly. The difference between 5.5 and 0.5 is easy though.

chapmanb: Thanks, this makes good sense but the major reason I changed this is that the million variants representation is only useful for whole genome data. For exome or panels it ends up being a small decimal point which was confusing -- I got questions like why is 10,000 variants represented as 0.01M which is pretty fair. Maybe formatting with commas would solve both problems? I'm not sure if there's a straightforward way to do this within MultiQC now.

chapmanb:
Phil;
Thanks so much for the review, I addressed the issues you identified and commented on the one I'm not sure about. Let me know if you want to adjust anything else at all.

multiqc/modules/bcftools/stats.py
ewels: Ah ok, this is a good point. Fine to leave it then I guess - could check the max value for all samples, but I guess it would be pretty confusing if it varied between reports.

I try to avoid commas where possible because of the German / Swedish / some-other-European-countries habit of using commas to denote decimal places (an even more annoying problem when parsing data). But I could make it use the same formatting as HighCharts with spaces (`5 522 770`). Already have the back end to customise this too: http://multiqc.info/docs/#number-formatting

ewels: See #406

ewels:
Thanks Brad! I'll leave this open for a bit longer in case anyone else wants to comment on the change in General Stats fields, but will merge in a day or two if no-one complains :)

ewels:
Ok, that should have been enough time for people to complain 😀

amilesj:
It would great to add in a line plot of Variant Qualities, just like the one in the Whole-Genome Seq example on the MultiQC homepage that plots SNP quantity as a function of quality score, except using data from bcftools stats. 

PS I am newer to GitHub, so let me know if I should post this request elsewhere!

ewels:
Hi @amilesj,

That sounds doable - please can you open a new issue with that suggestion and it’ll be added to the list..

Phil

amilesj:
Thanks, will do!

PR #399, opened 2017-02-12
New module: goleft indexcov -- coverage plots

chapmanb:
Adds support for parsing output from goleft indexcov
(https://github.com/brentp/goleft/tree/master/indexcov), which does fast
coverage estimation from BAM indexes and scales to whole genome inputs.
Provides plots of coverage by chromosome and a ROC curve of percent
coverage by depth.

Requires pre-processing `$prefix-indexcov.bed.gz` to uncompressed tsv
file with `gunzip -c $prefix-indexcov.bed.gz > $prefix-indexcov.tsv` so
it's correctly picked up by MultiQC.

Fixes brentp/goleft#12


chapmanb:
Thanks for @brentp this now includes a diagnostic plot for indicating problems with coverage bins.

ewels:
Hi @chapmanb,

Many thanks for this! Would it be possible for you or @brentp to submit some example data to [ewels/MultiQC_TestData](https://github.com/ewels/MultiQC_TestData/tree/master/data/modules) please? This is so that I can have something to run it on for testing, and so that Travis has something to work with.

Cheers,

Phil

brentp:
here is an example ped file output: http://indexcov.s3-website-us-east-1.amazonaws.com/sample-indexcov.ped
and here is the .bed.gz: http://indexcov.s3-website-us-east-1.amazonaws.com/sample-indexcov.bed.gz

ewels:
Thanks! Rerunning the travis checks now and will have a look over this next week.

Phil

ewels:
Hmm, MultiQC didn't find any goleft data when Travis run. I guess the bed file has to be ungzipped? I'll look into this properly on Monday anyway..

chapmanb:
Phil -- that's right, the BED file needs to be ungzipped and named to `*-indexcov.tsv` so MultiQC will properly recognize it. Hopefully that gets it running.

ewels:
Thanks both. I just had another go, ungzipping and renaming as you said. It still found nothing, after a bit of poking around I realised that it was because the file is 45MB, larger than the default `config.log_filesize_limit` value (10MB).

I made a local config file and upped that limit - MultiQC now finds the file but dies with the following error:
```python
Module goleft_indexcov raised an exception: Traceback (most recent call last):
  File "/Users/philewels/GitHub/MultiQC/scripts/multiqc", line 346, in multiqc
    output = mod()
  File "/Users/philewels/GitHub/MultiQC/multiqc/modules/goleft_indexcov/goleft_indexcov.py", line 21, in __init__
    cov_plot = self.coverage_plot()
  File "/Users/philewels/GitHub/MultiQC/multiqc/modules/goleft_indexcov/goleft_indexcov.py", line 64, in coverage_plot
    for chrom, start, end, cov in (l.split() for l in fn['f']):
ValueError: too many values to unpack
```
This looks to me like the code is written expecting a file with a single sample, but the one supplied above has many columns for many samples.

Secondly (and more importantly) - although MultiQC does now find the file, it's super slow as it has to read in the entire file (a bunch of times, until I sort out #202). People have written MultiQC modules in the past that do this kind of processing and it makes me uncomfortable. MultiQC has been designed to be a lightweight tool that reads tool _summaries_. When a module starts reading in big data files and calculating summary results itself, I consider that to be scope creep. Whilst I don't deny that it's doing useful stuff, I'd prefer if goleft was doing these calculations and MultiQC was just reading in the results.

This is of course subjective, there's no concrete line where a module is doing too much and I'm open to your thoughts on this. Let me know what you think..

ewels:
Also, if I could grab a`*-indexcov.roc` input file to test that part of the module that'd be great too.. Thanks!

chapmanb:
Phil -- thanks much for the code review on this. Apologies, the code was overly focused on the bcbio implementation where we have on goleft indexcov file per sample. I generalized support to handle the multisample case and pushed the fixed version to this branch.

Regarding the weight of the module. I agree, it is slow and the HTML rendering also suffers with a lot of samples. The MultiQC module isn't really doing any processing that could be pushed into goleft, but all the work comes from parsing and then plotting individual chromosome bin positions across the genome. There are just a lot of them. This is one of the most useful plots since it's easy to see chromosome level differences. I could drop the ROC plot which might be less useful with the ped plot in place. Brent, what do you think?

Thanks ag

brentp:
I think in the context of MultiQC, the order of most useful plots would be: 1) bins 2) sex 3) ROC 4) coverage by region.
This is because 4) has a lot of data and  most of that info can be conveyed in 3) which has fewer (~70) points per sample. plots 1 and 2 have 1 point per sample.

chapmanb:
Brent -- thanks for the feedback. I'm a bit sad to lose the coverage plot as it's the most easily interpretable, but it doesn't scale well on the interactive plots and causes the MultiQC HTML page to hang on Firefox so is probably not practical. I've disabled that which should make the parsing much quicker.

Phil, let me know if this looks better resource wide. There is a ROC file for testing here:

http://indexcov.s3-website-us-east-1.amazonaws.com/sample-indexcov.roc

Thanks again for all the feedback.

brentp:
@chapmanb thanks for integrating! I agree the coverage plots are the easiest to interpret, but the ROC plots aren't far off.

does the MultiQC section for indexcov link to indexcov's index.html ? If so, then the user could easily get to the coverage plots from there.

chapmanb:
Brent -- there aren't any external links right now. One of the benefits of MultiQC is that it's a standalone HTML page you can transfer around without needing to bundle up other bits. I'm cool with sticking with the ROC + bin plots and seeing how that goes as a starting point. Happy to have this in there for people to explore with.

ewels:
Hi @chapmanb,

I've just grabbed your latest changes and the ROC file that you linked to and everything runs now. The report looks great!

### Read coverage
I re-activated the commented out code so that I could have a play with it. Having thought about it, I think it should be fine to leave the code there - the large files will be skipped by default by MultiQC and so that code just won't run. If people either have small files or change their config then it should work. I think then everyone can be happy?

Happy that is, except for the fact that the report crashes the browser. But there's an easy fix - the `smooth_points` and `smooth_points_sumcounts` config options ([see docs](http://multiqc.info/docs/#line-graphs)). These config options tell MultiQC to average out datasets with loads of points. I added this to your code and although MultiQC still takes ages to run, the final report does now load:

```python
'smooth_points': 50,
'smooth_points_sumcounts': False
```

goleft_indexcov-coverage-plot 1

As you can see, you lose fidelity however. Is this acceptable? Any thoughts?

Also, would be nice to get the x-axis to describe the chromosome name so that the plot works when exported (the above is chr1, but no way to know).

###  ROC plot
Seems great, though I have a niggling edge case fear when I see all of the different plots for different chromosomes. What happens if someone runs this with a draft genome containing ten thousand scaffolds? I also can't really see users clicking through all 23 plots for every report which kind of goes against the MultiQC ethos of being able to skim over reports quickly. What do you think about showing a single average plot for the whole genome? Possibly with additional plots for sex chromosomes (configurable, but default X / Y)..? This would also have bonus of keeping the filesize down for reports with lots of samples.

The same point goes for the Read coverage plot.

### Problem coverage bins
No complaints here, looks good! I find it a little difficult to interpret as there's not much context - if it's just for spotting outliers then that's fine, but if I have only a few files then I think I'd find it difficult to know whether the numbers were good or not. Could we plot proportion of bins instead of count?


I'll have a proper go through the code once this bigger stuff is worked out, but from my quick run over now I can't see anything that needs changing. There should be a `log.info` statement saying how many of each report have been found, but that's all I think.

Once again, many thanks (both for your work and your patience!)

Phil

brentp:
Hi Guys, just injecting my 0.02 here...

The depth plot looks great, but it might be smoothing too much since the centromere seems to be hidden now.

For limiting plots by chromosome, in indexcov, I skip plotting any chromosomes that start with GL and those that have fewer than 2 16KB regions. In practice you could remove any with fewer than, e.g. 50 16KB regions. The number of bins for each chrom can be inferred from the indexcov.bed.gz file.

Re problem coverage bins, I've found this most useful when plotting a cohort so that outliers are obvious. Different reference genomes/mapping methods will have different absolute numbers (and likely ratios).  You might be right that showing a proportion would be more easily interpreted... I'll try that out in indexcov and ping this issue when/if I make a release with that change.

ewels:
Absolutely - I chose 50 points fairly arbitrarily. Could up that number as necessary (especially if we reduce the number of plots).

What about concatenating the different chromosome plots? We could use background `PlotBands` (as in the FastQC plots) to highlight every other chromosome. Again, the resolution suffers, but the user can always zoom in / refer to the full goleft report for a more detailed look. This has the benefit of being able to get a summary of the whole dataset without loads of clicking.

Phil

chapmanb:
Phil and Brent;
Thanks for all the digging into this and suggestions.

For the chromosome coverage plots, the detail and per-chromosome digging are the most useful parts. The utility is to identify large chromosomal deletions but you'll lose those with smoothing or combining together the plots. The ROC plots do a better job of detecting gender differences and major coverage issues without breaking the page with too much detail. So, I'd vote for just keeping them out for now and revisiting later if the existing plots prove less useful.

For avoid large number of contigs, this little function is meant to do by only taking numeric chromosomes and X:

https://github.com/ewels/MultiQC/pull/399/files#diff-e10770b8c0a2a10ce4979afba17f943cR50

This will work for human-like genomes but I'm not sure how best to limit for other less finished contigs so those will get skipped in the detailed plot since they don't follow this convention. Brent, would it be possible to have a indexcov option to exclude contigs below some bin size from the ROC file? They we could tackle this directly at the creation stage and not need workarounds in MultiQC.

Thanks again for all the helpful discussion.

brentp:
Re the bin plot scaling. I changed indexcov plots to use the proportion rather than total number, but the .ped output is unaffected so you can get the denominator for that with bins.in + bins.out in the .ped file.

chapmanb:
Brent -- thanks for the example and swapping this over in goleft. I changed the bin plot to use the proportational approach, and also cleaned up some descriptions and labels for the ROC plot. I'm happy with these two cleaned up plots and ready for this to go in whenever Phil thinks it is ready. Thanks again.

docs/modules/goleft_indexcov.md
```diff
@@ -0,0 +1,13 @@
...
+URL: https://github.com/brentp/goleft/tree/master/indexcov
+Description: >
+    Quickly estimate coverage from a whole-genome bam index, providing 16KB resolution.
+    This is useful as a quick QC to get coverage values across the genome.
+---
+
+The goleft indexcov module parses results generated by 
+[goleft indexcov](https://github.com/brentp/goleft/tree/master/indexcov),
+creating plots of coverage and regions covered by scaled depth. It requires
+post-processing goleft indexcov output to rename the generated `*-indexcov.bed.gz`
```
ewels: This isn't true any more.. Also needs to mention the `ped` and `roc` files.

multiqc/modules/goleft_indexcov/goleft_indexcov.py
```diff
@@ -0,0 +1,144 @@
...
+# Initialise the logger
+log = logging.getLogger(__name__)
+
+class MultiqcModule(BaseMultiqcModule):
+    def __init__(self):
+        super(MultiqcModule, self).__init__(name='goleft indexcov', anchor='goleft_indexcov',
+                                            href='https://github.com/brentp/goleft/tree/master/indexcov',
+                                            info="quickly estimates coverage from a whole-genome bam index.")
+        # Skip coverage plot, too resource intensive with multiple samples
+        # cov_plot = self.coverage_plot()
```
ewels: If we're dropping support for the coverage plot, do you mind properly deleting the code? We can still go back to it in the version control history and it keeps the codebase cleaner..

multiqc/modules/goleft_indexcov/goleft_indexcov.py
```diff
@@ -0,0 +1,144 @@
...
+        cov_plot = None
+        roc_plot = self.roc_plot()
+        bin_plot = self.bin_plot()
+        if not roc_plot and not bin_plot:
+            log.debug("Did not find goleft indexcov outputs in {}".format(config.analysis_dir))
+            raise UserWarning
+
```
ewels: There should be a `log.info` statement somewhere around here saying that outputs have been found. Preferably how many of each. _e.g._ something like this:
```
[INFO   ]          goleft : Found 3 indexcov ROC reports
[INFO   ]          goleft : Found 3 indexcov bin reports
```

multiqc/modules/goleft_indexcov/goleft_indexcov.py
```diff
@@ -0,0 +1,144 @@
...
+        self.sections = list()
+        if cov_plot:
```
ewels: Prefer to delete this if possible, as above.

multiqc/modules/goleft_indexcov/goleft_indexcov.py
```diff
@@ -0,0 +1,144 @@
...
+                'id': 'goleft_indexcov-roc-plot',
+                'title': 'ROC: genome coverage per scaled depth by chromosome',
+                'xlab': 'Scaled coverage',
+                'ylab': 'Proportion of regions covered',
+                'ymin': 0, 'ymax': 1.0,
+                'xmin': 0, 'xmax': 1.5,
+                'data_labels': [{"name": self._short_chrom(c)} for c in chroms]}
+            return desc + linegraph.plot([data[c] for c in chroms], pconfig)
+
+    def _clean_name(self, name):
```
ewels: Should use [`self.clean_s_name`](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/base_module.py#L108-L149) instead from the main Base module. Does a load of extra things then (user config cleaning, prepending directory names etc).

multiqc/modules/goleft_indexcov/goleft_indexcov.py
```diff
@@ -0,0 +1,144 @@
...
+                'content': roc_plot
+            })
+        if bin_plot:
+            self.sections.append({
+                'name': 'Problem coverage bins',
+                'anchor': 'goleft_indexcov-roc',
+                'content': bin_plot
+            })
+
+    def _short_chrom(self, chrom):
```
ewels: I'm a bit nervous about this because of non-human organisms (_e.g._ Yeast, with roman numeral chromosomes). I think it's fine as long as you allow the user to specify a list of chromosomes:
```python
try:
  chroms = config.goleft['chromosomes']
  if chrom in set(chroms):
    return chrom
except KeyError:
  # [.. rest of your code ..]
```
then documenting how to do this in the config:
```
goleft:
  chromosomes:
    - I
    - II
    - III
    - IV
```

It would also be nice to have a `log.debug` saying how many chromosomes have been skipped.

ewels: ps. Any reason to use `in set(["X"])` instead of just `in ["X"]` out of interest?

multiqc/modules/goleft_indexcov/goleft_indexcov.py
```diff
@@ -0,0 +1,144 @@
...
+        """Plot standard chromosomes + X, sorted numerically.
+        """
+        chrom = chrom.replace("chr", "")
+        try:
+            return int(chrom)
+        except ValueError:
+            if chrom in set(["X"]):
+                return chrom
+
+    def coverage_plot(self):
```
ewels: Please delete if not being used..

multiqc/modules/goleft_indexcov/goleft_indexcov.py
```diff
@@ -0,0 +1,144 @@
...
+    def roc_plot(self):
+        desc = '<p>Coverage (ROC) plot that shows genome coverage at at given (scaled) depth. \n\
+        Lower coverage samples have shorter curves where the proportion of regions covered \n\
+        drops off more quickly. This indicates a higher fraction of low coverage regions. \n\
+        </p>'
+        data = collections.defaultdict(lambda: collections.defaultdict(dict))
+        for fn in self.find_log_files(config.sp["goleft_indexcov"]["roc"], filehandles=True):
+            header = fn['f'].readline()
+            sample_names = [self._clean_name(x) for x in header.strip().split()[2:]]
+            for parts in (l.rstrip().split() for l in fn['f']):
```
ewels: Nice compression! Very tidy 😀

multiqc/modules/goleft_indexcov/goleft_indexcov.py
```diff
@@ -0,0 +1,144 @@
...
+                chrom, cov = parts[:2]
+                sample_vals = parts[2:]
+                if self._short_chrom(chrom) is not None:
+                    for val, sample in zip(sample_vals, sample_names):
+                        data[chrom][sample][float(cov)] = float(val)
+        if data:
+            chroms = sorted(data.keys(), key=self._short_chrom)
```
ewels: Can we add a fail-safe check here to make sure that there aren't more than say, 50 chromosomes?

multiqc/modules/goleft_indexcov/goleft_indexcov.py
```diff
@@ -0,0 +1,144 @@
...
+                cur["sample_id"] = self._clean_name(cur["sample_id"])
+                total = float(cur["bins.in"]) + float(cur["bins.out"])
+                data[cur["sample_id"]] =  {"x": float(cur["bins.lo"]) / total,
+                                           "y": float(cur["bins.out"]) / total}
+        if data:
+            pconfig = {
+                'id': 'goleft_indexcov-bin-plot',
+                'title': 'Problematic low and non-uniform coverage bins',
+                'xlab': 'Proportion of bins with depth < 0.15',
+                'ylab': 'Proportion of bins with depth outside of (0.85, 1.15)'}
```
ewels: Now that we have proportions, can we set axis limits to 0 and 1?

<table>
<thead><tr><th>Without limits</th><th>With limits</th></tr></thead>
<tbody><tr><td>

</td><td>

</td></tr></tbody></table>

Failing that, could be good to use `yCeiling` and `yFloor` so that the plot automatic scaling doesn't go beyond possible values.

multiqc/utils/search_patterns.yaml
```diff
@@ -52,6 +52,15 @@ featurecounts:
 gatk:
     varianteval:
         contents: '#:GATKReport'
+goleft_indexcov:
+   # disabled due to parsing and display speed
```
ewels: Prefer to delete as above.

ewels:
Ah, also getting a Python 3 error on Travis now: https://travis-ci.org/ewels/MultiQC/jobs/204218973

```python
Module goleft_indexcov raised an exception: Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.5.2/lib/python3.5/site-packages/multiqc-1.0.dev0-py3.5.egg/EGG-INFO/scripts/multiqc", line 346, in multiqc
    output = mod()
  File "/home/travis/virtualenv/python3.5.2/lib/python3.5/site-packages/multiqc-1.0.dev0-py3.5.egg/multiqc/modules/goleft_indexcov/goleft_indexcov.py", line 24, in __init__
    roc_plot = self.roc_plot()
  File "/home/travis/virtualenv/python3.5.2/lib/python3.5/site-packages/multiqc-1.0.dev0-py3.5.egg/multiqc/modules/goleft_indexcov/goleft_indexcov.py", line 102, in roc_plot
    chroms = sorted(data.keys(), key=self._short_chrom)
TypeError: unorderable types: str() < int()
```

chapmanb:
Phil;
Thanks so much for all this great feedback. I pushed fixes for all of your suggestions. The only change I didn't make was to the scaling for the bin plot. Fixing between 0 and 1 obscures the changes on the X axis since they're small numbers. It's really the relative differences that are useful so not bounding automatically to 0/1 emphasizes those.

Other than that hopefully I've addressed everything. Let me know if anything else would be useful. Thanks again.

ewels:
Brilliant, thanks for being so patient @chapmanb!

Issue #395, opened 2017-02-08
cutadapt module fails to parse sample name when reading from stdin


tdido:
When feeding cutadapt through stdin, the input file will be replaced by a dash ("-") in the command line, and therefore all samples become "-" when processing multiple output files. Example:

`Command line parameters: --minimum-length 18 --discard-untrimmed -a TGGAATTCTCGGGTGCCAAGG -o out/clipped/sample1.clipped.fastq.gz -`

This happens because the parser expects the last element of the line to be the input file name, as seen [here](https://github.com/ewels/MultiQC/blob/24736e651f7b62585b9363e51c5c13f9cfe9c086/multiqc/modules/cutadapt/cutadapt.py#L103).

One solution could be to take the output file instead of the input, but I'm not sure whether that'd fit other people's workflows. If it did, I'd be happy to submit a patch.

ewels:
Hi @tdido - thanks for this issue, I hadn't come across this problem myself yet. It's a bit of a philosophical problem really and difficult to automate (I'm away at the moment, but I think I have another issue open somewhere where I was thinking of standardising this behaviour a little and making it configurable).

The short-term solution here is probably just to check if the parsed sample name is `-` and take the input file name as the sample name if so. We shouldn't do this by default, as these logs can often be concatenated for multiple samples (the reason for adding the current functionality).

I'll do this when I get back in a couple of weeks, or I'm happy to look at a PR. The cleaned sample name from the file name is available as `f['s_name']`..

Thanks,

Phil

PR #401, opened 2017-02-15
Fix cutadapt sample name handing when reading from stdin

tdido:
As mentioned in #395


ewels:
Many thanks @tdido!

Phil

PR #405, opened 2017-02-20
add profile option

Galithil:
relies on cprofile

### DIFF
--- a/scripts/multiqc
+++ b/scripts/multiqc
@@ -397,7 +397,7 @@ plots_flat, plots_interactive, make_pdf, config_file, verbose, quiet, profile, *
 
     if config.profile:
         pr.disable()
-        pr.dump_stats(os.path.join(config.output_dir, "profile.cprof"))
+        pr.dump_stats(os.path.join(config.data_dir, "multiqc.cprof"))
 
     # Did we find anything?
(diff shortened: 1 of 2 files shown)


scripts/multiqc
```diff
@@ -333,6 +342,10 @@ plots_flat, plots_interactive, make_pdf, config_file, verbose, quiet, **kwargs):
         pass # No subdirectory variable given
 
 
+    if config.profile:
```
ewels: Any reason to start this down here, rather than straight after parsing the command line flags?

scripts/multiqc
```diff
@@ -384,6 +397,14 @@ plots_flat, plots_interactive, make_pdf, config_file, verbose, quiet, **kwargs):
                           this_module, traceback.format_exc()) + ('='*60))
             sys_exit_code = 1
 
+    if config.profile:
+        pr.disable()
```
ewels: As above, any reason not to put this right at the end?

Galithil:
Does not provide much value over the use of `python -m cProfile -o <filename>`

ewels:
Thanks for the help anyway! Will write some docs about this now.

ewels:
http://multiqc.info/docs/#profiling-performance

PR #422, opened 2017-03-13
Feature/pytest test

percyfal:
Hi Phil

as proof of principle, I added some parametrized tests that utilize pytest and test fixtures from pytest-ngsfixtures. Our recent discussion made me realize I was unclear about my primary goal of setting up tests and creating a general-purpose dataset. The aim is *not* primarily to use the data for regression tests, but rather for test-based development; after all, when writing a new parser/module/whatever, a common approach is to generate small test data to work on. That the data can be used for regression and unit tests comes as an additional benefit. 

The added multiqc test currently just parses inputs and fails if there is an error. One could imagine more elaborate evaluations. Note that due to the parametrization there is only one test, but it (currently) runs 77 different unit tests which each can be inspected individually, if need be.

On a final note, it seems this PR is heading for your master branch, which is not my intention. Try it out as a feature and let me know what you think.


ewels:
Thanks for this @percyfal - looks exciting 😀  I'm not ignoring you, I just want time to have a proper sit down and play with this and haven't had the time yet. Next week hopefully.

percyfal:
@ewels no worries, take your time. I added some docs for legibility. I'm writing some tests now that look a lot like these ones, and I realized that it can be difficult to follows what happens without proper docs and comments.

ewels:
Ok, sounds great. Hopefully should be getting the [new RNA-seq pipeline](https://github.com/SciLifeLab/NGI-RNAseq) into production this week, then I should have a bit more time. Could be good to have a quick Skype meeting about this if that's ok? It'd be nice to get a headstart on what you've built and how it works, given my lack of experience with testing stuff.

percyfal:
Sure, no problem. I'll be away next week on Easter vacation but after that anytime should work, just let me know when suits you.

ewels:
Great! I'll ping you on slack after Easter then.

wkretzsch:
This could also be used for automated acceptance testing. @ewels what do you think should be the acceptance layer for modules?

ewels:
Ah yes, I should wrap this PR up. As I remember this sort of does the same thing as the automated testing that we already have with the [MultiQC_TestData](https://github.com/ewels/MultiQC_TestData) repository (PRs run automatically and fail if they crash).

It's difficult to do more automated testing than "does it crash or not" as most of the reviewing that I do on PRs is subjective - eg. style and design choices rather than problems in code. There was an effort started at the codefest by @tbooth to write unit tests for each module as part of the development process though - these can be seen [here](https://github.com/ewels/MultiQC_TestData/tree/master/unit_tests) and are also run as part of the automated tests.

Whilst such testing is certainly nice, I'm not rushing to spend loads of time on it. MultiQC modules are very, well, modular. So if they're not being actively worked on then they tend to behave in my experience. If they are being worked on then there (should be) plenty of manual checking going on as part of the development process. Writing tests going forward for new stuff as @tbooth suggested is a great idea if people want, but retrofitting is probably a lot of work.

There are of course exceptions, as demonstrated in #556 where I missed a Python 3 problem because the test data didn't trigger that piece of code. But these are hopefully not _too_ common.

@wkretzsch - some of the stuff you've mentioned points more towards a linting approach, checking for code style and best practices. This is slightly different. Generally I've tried to write all core MultiQC code to be very forgiving, to make it as easy as possible for people to write their own modules. I guess that this could have the unintended consequence of making modules harder to write as the process can be too flexible. But it's a little late to change this philosophy now I think.

Phil

wkretzsch:
Ah, yes, a `--lint` flag to catch the kind of thing I point out in #557 might be the right approach?

ewels:
Closing this now. Many thanks for the work involved though!

PR #425, opened 2017-03-21
Fix parsing "None" values in linegraph

vladsaveliev:
Hi Phil, 

After this commit https://github.com/ewels/MultiQC/commit/44c89b3483e8f3076b0f2de69f3e8e71b421e2fa#diff-6779b900d2a994ae35f19feecbc3a7cbR40, the `bcftools` module started to crash (at the line highlighted), while building the indel distribution plot. Because the point x=0 is set to `None` to avoid joining disjoint parts of the plot: https://github.com/ewels/MultiQC/blob/master/multiqc/modules/bcftools/stats.py#L43, the plotter failed to parse that `None` as `float`.

This can be reproduced only when the number of samples is higher than 1 (https://github.com/ewels/MultiQC/blob/master/multiqc/modules/bcftools/stats.py#L116), so the tests didn't catch this.

I'm not sure what would be the right solution, but this one seem to work.

Also, not sure if plotting the indel plot only for 2 or more samples was not a bug.

### DIFF
--- a/multiqc/plots/linegraph.py
+++ b/multiqc/plots/linegraph.py
@@ -73,14 +73,16 @@ def plot (data, pconfig={}):
                     maxval = max(maxval, d[s][k])
             else:
                 for k in sorted(d[s].keys()):
-                    if 'xmax' in pconfig and float(k) > float(pconfig['xmax']):
-                        continue
-                    if 'xmin' in pconfig and float(k) < float(pconfig['xmin']):
-                        continue
-                    if 'ymax' in pconfig and float(d[s][k]) > float(pconfig['ymax']):
-                        continue
-                    if 'ymin' in pconfig and float(d[s][k]) < float(pconfig['ymin']):
-                        continue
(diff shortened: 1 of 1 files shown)


ewels:
Aha, great spot - thanks @vladsaveliev!

I'm not clear why it's `>1` and not `>0` - the [commit where I wrote that](https://github.com/ewels/MultiQC/commit/fc3f5a8d94b03c7cb427ce729e386531d5cea16d) had the title _"Don't plot indel length if no data."_. So I'm not quite sure what I was doing at the time. I'll take a look at it again.

The same change will need be done in the scatter plot too I suspect, as I made similar changes there. I'll apply your fix there now.

Thanks!

ewels:
I couldn't see why it should be `>1` instead of `>0`, so I changed it (seems to work fine).

Phil

vladsaveliev:
Thank you Phil, that's awesome!

PR #432, opened 2017-03-29
Match files to modules in one scan

epruesse:
see #202

### DIFF
--- a/multiqc/modules/bamtools/stats.py
+++ b/multiqc/modules/bamtools/stats.py
@@ -44,7 +44,7 @@ def parse_reports(self):
     }
 
     # Go through files and parse data using regexes
-    for f in self.find_log_files(config.sp['bamtools']['stats']):
+    for f in self.find_log_files('bamtools/stats'):
         d = dict()
         for k, r in regexes.items():
             r_search = re.search(r, f['f'], re.MULTILINE)
--- a/multiqc/modules/base_module.py
+++ b/multiqc/modules/base_module.py
@@ -42,6 +42,13 @@ class BaseMultiqcModule(object):
                  and either the file contents or file handle for the current matched file.
                  As yield is used, the results can be iterated over without loading all files at once
         """
+        if isinstance(patterns, str):
+            module = patterns
+            patterns = config.sp
+            for sub in module.split('/'):
+                patterns = patterns[sub]
+        else:
+            module = None
 
         # Get the search parameters
         fn_match = None
@@ -60,6 +67,10 @@ class BaseMultiqcModule(object):
             # Set up vars
             root = f['root']
             fn = f['fn']
+            mod = f['mod']
+
+            if mod is not [] and module not in mod:
+                continue
 
             # Make a sample name from the filename
             s_name = self.clean_s_name(fn, root)
--- a/multiqc/modules/bcftools/stats.py
+++ b/multiqc/modules/bcftools/stats.py
@@ -26,7 +26,7 @@ class StatsReportMixin():
         self.bcftools_stats = dict()
         self.bcftools_stats_indels = dict()
         depth_data = dict()
-        for f in self.find_log_files(config.sp['bcftools']['stats']):
+        for f in self.find_log_files('bcftools/stats'):
             s_names = list()
             for line in f['f'].splitlines():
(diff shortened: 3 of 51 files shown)


ewels:
Hi @epruesse,

Many thanks for this highly impressive code. I confess that I've never heard of `magic` before, it looks like a really powerful tool! However, before either of us put too much time into this direction, I have a couple of questions:

1. Will this work on Windows?
    * A classic unix tool this may be, but plenty of MultiQC windows run on non-unix systems
2. Will this work with all versions of Python?
    * I noticed that the travis test fails on everything that's not v3.5 Python. I guess this is a trivial fix though.
3. You say that it's "been around for ages", but I'd like to confirm how long ages is. Some people run MultiQC on terrifyingly old systems.

Generally whilst writing MultiQC I've tried to avoid using any system commands for these reasons, instead using Python packages in their stead where I can. But if there are compelling reasons that we can't use such an approach here then I'd of course welcome it.

In short, for me compatibility is more important than speed. I think that the main strength of MultiQC is that it tends to "just work" for most people.

Phil

epruesse:
Hi Phil,

1. Windows
   - I hadn't realized that windows was a consideration. I added #433 as a separate issue, having  AppVeyor to do tests would be useful (in particular since I don't have a windows machine).
   - `file` does work on windows, but obviously isn't installed by default. It could be bundled as a ready exe though. Options are the binaries from http://gnuwin32.sourceforge.net/packages/file.htm or from http://github.com/nscaife/file-windows
   - Alternatively, pyxdg and librox (https://github.com/rox-desktop/rox-lib/blob/master/ROX-Lib2/python/rox/mime.py) have python implementations of the matching done by `libmagic`. The code is under LGPL 2.1, which would allow including it with MultiQC (would have to be modified though I believe).
2. Python versions
   - I had only tested with Python 3.5, I wanted to keep the time I invested somewhat in check. Py27 failed due to the absolute imports, so it's just missing a future import. There may be more though. Nothing that couldn't work with the other versions in principle though.
   -  Do you want to add Py36 to travis testing perhaps?
3. Around for ages
   - According to [wikipedia](https://en.wikipedia.org/wiki/File_(command)),  `file` was originally released with Unix Research Version 4 in 1973. Old enough? ;-)
  It's been rewritten since then, and I'm sure the command line isn't compatible to 1973 any more. But it's up there with "mv" and "cp" -- part of POSIX(!), so not a dependency I'd worry about on any unix. 

For ease of installation and compatibility, it'd probably be best to add pure python code to do the matching. Though the question then is whether it's worth bothering with `file` at all.

ewels:
Ok great. So what are the advantages of using `file`? Does this approach give anything that we can't achieve with a pure-python approach?

A couple more edge cases I've thought of which may or may not cause problems:

* Maximum offsets would speed up searches, but can be variable for `stdout` logs
  * Good example would be cutadapt, which has a standard output but can end up in the middle of a much longer log (eg. [here](https://github.com/ewels/MultiQC_TestData/blob/3ff387e578f5e068ef0a9b40603893c3e95217cc/data/modules/clusterflow/fastq_star_onesample/ERR458837_1.fastq.gz_fastq_star_clusterFlow.txt#L100-L230)).
* Because of the above issue of mixed logs, a single file can be required for multiple modules
  * This is the original reason that I got MultiQC to run the loop for every module I think. Wasn't a problem when I was running for 20 samples with 5 modules! 40 modules with thousands of samples gets slow though.

I've added Python 3.6 to Travis, so that should be checked in the future. I _guess_ 1973 counts as "old enough" 😆  I'm learning a lot thanks to this thread!

Phil

ewels:
Hi @epruesse,

I've just written my own version of this refactoring (see #449). I'd welcome any feedback on what I've done if you have the time! I think that the core concepts are probably similar to what you've done here, but all within pure Python code and a bit more bespoke in places.

There were a bunch of nasty bugs that I had to fix for the new approach (eg. some of the hardwired stuff in the Custom Content module especially). I notice now that these same problems also affect your PR here, as quite a few modules aren't reporting any found files in the output log in the tests.

I think it makes sense to close this PR now, but if you think that any of your techniques here could benefit my PR, that would be brilliant.

Thanks again!

Phil

Issue #434, opened 2017-03-31
Include GcBiasSummaryMetrics in multiqc_data output


ahvigil:
`picard CollectGcBiasMetrics` generates both GcBiasDetailMetrics and GcBiasSummaryMetrics output. Currently GcBiasDetailMetrics is used to generate a GC coverage bias plot, but GcBiasSummaryMetrics output is ignored. To be consistent with other picard modules, multiqc should roll up gc_bias_metrics.summary.txt files into a `multiqc_picard_gcbias.txt` file or similar.

I will have a PR addressing this shortly.

ewels:
Hi @ahvigil,

Do you have any examples of the `GcBiasSummaryMetrics` output? Picard output can be extremely heterogenous - this data doesn't seem to exist in the files that I used to write this submodule (see them [here](https://github.com/ewels/MultiQC_TestData/tree/master/data/modules/picard/GcBiasMetrics)). It would be great if we could add some to the test data repo.

Phil

ewels:
Closed in #435

PR #435, opened 2017-04-01
include picard.analysis.GcBiasSummaryMetrics in picard.gcbias module

ahvigil:
addresses #434 

Modeled after the behavior of other picard modules, this aggregates GcBiasSummaryMetrics so they can be written to `multiqc_data/multiqc_picard_gcbias.txt`

### DIFF
--- a/multiqc/modules/picard/GcBiasMetrics.py
+++ b/multiqc/modules/picard/GcBiasMetrics.py
@@ -19,6 +19,7 @@ def parse_reports(self):
 
     # Set up vars
     self.picard_GCbias_data = dict()
+    self.picard_GCbiasSummary_data = dict()
 
     # Go through logs and find Metrics
     for f in self.find_log_files(config.sp['picard']['gcbias'], filehandles=True):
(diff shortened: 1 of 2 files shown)


ewels:
Hi @ahvigil,

This is great stuff, thanks! Do you have some example testing data that I can take a look at before merging?

Phil

ewels:
Ok, I generated some test data myself to try the code on and it works great, thanks! My first thought was that we should add some of these metrics to the report (eg. in the General Statistics table), but actually I'm not sure that any of them are that helpful - at least beyond what the plot already shows. But it's nice to have them printed to a file for downstream use as you say 👍 

Thanks again!

Phil

ahvigil:
ahh I just saw this, otherwise I would have gladly added test data. I see you've done that for me in https://github.com/ewels/MultiQC_TestData/commit/77fc1945e40b8c59832106dd51c011973e81c61d, thanks!

ewels:
No problem - I was running Picard on some test data for another project anyway, so it was simple to just grab the output from this while I was at it :)

PR #438, opened 2017-04-05
Fix table rendering bug in README.md

viklund:
The table rendered incorrectly on the github page, this commit fixes that.


ewels:
Ah, good spot - thanks!

ewels:
Weird, looking back through the file history it seems like it's always been broken. But I'm sure it wasn't when I first made the table? I wonder if GitHub tweaked their markdown rendering or something..

PR #443, opened 2017-04-10
new module: Adapter Removal

alexanderscholz:
Initial status of the new module, doc, readme and changelog


README.md
```diff
@@ -36,7 +36,7 @@ Currently, supported tools include:
 |[FastQ Screen][fastq-screen] | [HiCUP][hicup]         | [HTSeq][htseq]            | [Qualimap][qualimap] |
 |[Skewer][skewer]             | [Kallisto][kallisto]   | [Picard][picard]          | [QUAST][quast]       |
 |[Trimmomatic][trimmomatic]   | [Salmon][salmon]       | [Prokka][prokka]          | [RNA-SeQC][rna_seqc] |
-|                             | [Slamdunk][slamdunk]   | [Samblaster][samblaster]  | [RSeQC][rseqc]       |
+|[Adapter Removal][adapterremoval] | [Slamdunk][slamdunk]   | [Samblaster][samblaster]  | [RSeQC][rseqc]       |
```
ewels: Minor thing, but I've been trying to make these lists alphabetically sorted.

multiqc/modules/adapterRemoval/adapterRemoval.py
```diff
@@ -0,0 +1,524 @@
...
+            data_pattern['discarded_m2'] = 5
+            data_pattern['singleton_m2'] = 6
+            if not self.__collapsed:
+                data_pattern['retained'] = 8
+            else:
+                data_pattern['full-length_cp'] = 8
+                data_pattern['truncated_cp'] = 9
+                data_pattern['retained'] = 10
+
+        for title, key in data_pattern.iteritems():
```
ewels: Please use `items()` instead for Python 3 compatibility..

docs/modules/adapterRemoval.md
```diff
@@ -0,0 +1,23 @@
+---
```
ewels: Please add a reference to this file in `docs/README.md` (required to be built into the website docs).

multiqc/modules/adapterRemoval/adapterRemoval.py
```diff
@@ -0,0 +1,524 @@
...
+
+""" MultiQC module to parse output from Adapter Removal """
+
+from __future__ import print_function
+from collections import OrderedDict
+import logging
+import copy
+
+from multiqc import config
+from multiqc.plots import bargraph
```
ewels: Can simplify to:
```python
from multiqc.plots import bargraph, linegraph
```

setup.py
```diff
@@ -93,6 +93,7 @@
             'tophat = multiqc.modules.tophat:MultiqcModule',
             'trimmomatic = multiqc.modules.trimmomatic:MultiqcModule',
             'gatk = multiqc.modules.gatk:MultiqcModule',
+			'adapterRemoval = multiqc.modules.adapterRemoval:MultiqcModule',
```
ewels: Spaces not tabs please 😉

multiqc/utils/search_patterns.yaml
```diff
@@ -2,6 +2,9 @@
 # Default configurations for how modules can find their log files.
 # Loaded by the config module so that these patterns can be overwritten in user config files.
 
+adapterRemoval:
+    fn:
```
ewels: As this is only a single string, you don't need a list. Instead:
```yaml
adapterRemoval:
    fn: '*.settings'
```

ewels:
Hi @alexanderscholz,

Thanks for writing this new module! I've added a few comments to the code above, though nothing major.

I tried running your module with the test data that you added and it ran really well. However, there are loads of sections! I know that most people will only have just single-end / paired-end and so on and so won't have most of these, but it still surprised me a little. Is there any reason which each variant of output has to have its own plot? The same goes for the underlying data and parsing - it feels very verbose to split everything up into such a large array of separated structures.

Generally I try to make modules for tools that do similar things give similar output. So in this case I may try to mimic the cutadapt output for example - have a General Statistics column for the percentage of bases trimmed, a length plot and not much else.. Also consistency in plot colours - I try to pick from the standard palette unless I have a specific reason not to.

Phil

alexanderscholz:
Hey @ewels,

thanks for your fast answer.

I built the sections and plots like we did manually. But i understand your points so i will rebuild it. I will merge the corresponding section data to three different plots "Adapter Alignments", "Retained and Discarded" and "Lenght Distribution" and i will use the default colours. Is this ok?

And of course i will implement the module with python 3.4,3.5 and 3.6 support too.

If you have further point your are welcome.

Alex

ewels:
Sounds good! What do you think about putting some numbers into the General Statistics table instead of having bar plots? For instance, could have a _% reads trimmed_ columns there instead. If it's a single number it could be more useful there, lined up against the other metrics from other tools.

Phil

alexanderscholz:
Hello @ewels,

sorry for this long break. I did what i mentioned in my last message + i followed your point with the General Statistics and the first plot ("Adapter Alignments"). I think in the second plot ("Retained and Discarded") there are too much information to putting it in the General Statistics. What do you think?

I definitely want to thank you for your constructive suggestions, now there are much less code!

ewels:
Hi @alexanderscholz,

This is great - thanks! Apologies, I've just been super annoying and merged a PR of my own with a bunch of refactoring, which has introduced a merge conflict. I've also changes a couple of minor ways in which the module should work - I'll leave a couple of comments on what you need to change with this.

Sorry for the bad timing!

Phil

multiqc/modules/adapterRemoval/adapterRemoval.py
```diff
@@ -0,0 +1,297 @@
...
+            'mate2': dict(),
+            'singleton': dict(),
+            'collapsed': dict(),
+            'collapsed_truncated': dict(),
+            'discarged': dict(),
+            'all': dict(),
+        }
+
+        parsed_data = None
+        for f in self.find_log_files(config.sp['adapterRemoval'], filehandles=True):
```
ewels: After refactoring, this now takes the search pattern string instead of the dict:

```python
for f in self.find_log_files('adapterRemoval', filehandles=True):
```

multiqc/modules/adapterRemoval/adapterRemoval.py
```diff
@@ -0,0 +1,297 @@
...
+        log.info("Found {} reports".format(len(self.adapter_removal_data)))
+
+        # Write parsed report data to a file
+        self.write_data_file(self.adapter_removal_data, 'multiqc_adapter_removal')
+
+        # add data to Basic Stats Table
+        self.adapter_removal_stats_table()
+
+        # Start the sections
+        self.sections = list()
```
ewels: This line is no longer needed

multiqc/modules/adapterRemoval/adapterRemoval.py
```diff
@@ -0,0 +1,297 @@
...
+        cats_pec = OrderedDict()
+        cats_pec['singleton_m1'] = {'name': 'singleton mate1'}
+        cats_pec['singleton_m2'] = {'name': 'singleton mate2'}
+        cats_pec['retained_reads'] = {'name': 'retained read pairs'}
+        cats_pec['full-length_cp'] = {'name': 'full-length collapsed pairs'}
+        cats_pec['truncated_cp'] = {'name': 'truncated collapsed pairs'}
+        cats_pec['discarded_m1'] = {'name': 'discarded mate1'}
+        cats_pec['discarded_m2'] = {'name': 'discarded mate2'}
+        pconfig['id'] = 'ar_retained_plot'
+        self.sections.append({
```
ewels: This syntax technically should still work, but I've written a function to replace it (see new [docs](http://multiqc.info/docs/#step-6-create-report-sections)). Syntax is very similar:

```python
self.add_section (
    name = 'Retained and Discarded Paired-End Collapsed',
    anchor = 'adapter_removal_retained_plot',
    description = 'The proportions of retained and discarded reads.',
    plot = bargraph.plot(self.adapter_removal_data, cats_pec, pconfig)
)
```

multiqc/modules/adapterRemoval/adapterRemoval.py
```diff
@@ -0,0 +1,297 @@
...
+
+        cats_pec = OrderedDict()
+        cats_pec['singleton_m1'] = {'name': 'singleton mate1'}
+        cats_pec['singleton_m2'] = {'name': 'singleton mate2'}
+        cats_pec['retained_reads'] = {'name': 'retained read pairs'}
+        cats_pec['full-length_cp'] = {'name': 'full-length collapsed pairs'}
+        cats_pec['truncated_cp'] = {'name': 'truncated collapsed pairs'}
+        cats_pec['discarded_m1'] = {'name': 'discarded mate1'}
+        cats_pec['discarded_m2'] = {'name': 'discarded mate2'}
+        pconfig['id'] = 'ar_retained_plot'
```
ewels: Can just do this in the initial dict definition above..?

multiqc/modules/adapterRemoval/adapterRemoval.py
```diff
@@ -0,0 +1,297 @@
...
+        lineplot_data = [
+            self.len_dist_plot_data['mate1'],
+            self.len_dist_plot_data['mate2'],
+            self.len_dist_plot_data['singleton'],
+            self.len_dist_plot_data['collapsed'],
+            self.len_dist_plot_data['collapsed_truncated'],
+            self.len_dist_plot_data['discarged'],
+            self.len_dist_plot_data['all']
+        ]
+        self.sections.append({
```
ewels: Update to the way that this is done, see above.

alexanderscholz:
Hey @ewels,
Fortunately it is software ;-) no problem
One note: i get this output
```
[INFO   ]         multiqc : This is MultiQC v1.0.dev0 (17ad3a4)
[INFO   ]         multiqc : Prepending directory to sample names
[INFO   ]         multiqc : Searching '.../data'
Searching 5 files..
[INFO   ]  adapterRemoval : Found 5 reports
```
The "searching" line is new.

ewels:
Correct, the "searching" is new! More of the time is now dedicated to that first step of the processing, so MultiQC can hang there for several minutes. That line now shows a progress bar (you should see it when running on a bigger folder of results). eg. from the [Travis test](https://travis-ci.org/ewels/MultiQC/jobs/224416924):

```
Searching 609 files..  [####################################]  100% 
```

Will hopefully have time to look at this tomorrow 👍

multiqc/modules/adapterRemoval/adapterRemoval.py
```diff
@@ -0,0 +1,311 @@
...
+            'id': 'ar_retained_plot',
+            'ylab': '# Reads',
+            'hide_zero_cats': False,
+            'cpswitch_counts_label': 'Number of Reads'
+        }
+
+        cats_pec = OrderedDict()
+        cats_pec['singleton_m1'] = {'name': 'singleton mate1'}
+        cats_pec['singleton_m2'] = {'name': 'singleton mate2'}
+        cats_pec['retained_reads'] = {'name': 'retained read pairs'}
```
ewels: Please put this category first, then all of the discarded reads afterwards. Standard styling is the categories should be stacked from best -> worst.

multiqc/modules/adapterRemoval/adapterRemoval.py
```diff
@@ -0,0 +1,311 @@
...
+        pconfig = {
+            'title': 'retained and discarded',
+            'id': 'ar_retained_plot',
+            'ylab': '# Reads',
+            'hide_zero_cats': False,
+            'cpswitch_counts_label': 'Number of Reads'
+        }
+
+        cats_pec = OrderedDict()
+        cats_pec['singleton_m1'] = {'name': 'singleton mate1'}
```
ewels: Please use Sentence Case for category names (`Singleton Mate 1`), or at least capitalise the first word.

multiqc/modules/adapterRemoval/adapterRemoval.py
```diff
@@ -0,0 +1,311 @@
...
+        cats_pec['singleton_m2'] = {'name': 'singleton mate2'}
+        cats_pec['retained_reads'] = {'name': 'retained read pairs'}
+        cats_pec['full-length_cp'] = {'name': 'full-length collapsed pairs'}
+        cats_pec['truncated_cp'] = {'name': 'truncated collapsed pairs'}
+        cats_pec['discarded_m1'] = {'name': 'discarded mate1'}
+        cats_pec['discarded_m2'] = {'name': 'discarded mate2'}
```
ewels: It would be good if we could make these names shorter, as they overlap onto two lines when viewed in a not-huge window. Could perhaps use `R1` and `R2` instead of `mate1` and `mate2`? Not sure that there's much else actually..

multiqc/modules/adapterRemoval/adapterRemoval.py
```diff
@@ -0,0 +1,311 @@
...
+        #    'name': 'Retained and Discarded Paired-End Collapsed',
+        #    'anchor': 'adapter_removal_retained_plot',
+        #    'content': '<p>The proportions of retained and discarded reads.</p>' +
+        #               bargraph.plot(self.adapter_removal_data, cats_pec, pconfig)
+        #})
+
+        self.add_section(
+            name='Retained and Discarded Paired-End Collapsed',
+            anchor='adapter_removal_retained_plot',
+            description='The proportions of retained and discarded reads.',
```
ewels: These are not not really proportions, more counts. Perhaps instead of _The number of retained and discarded reads._ ?

multiqc/modules/adapterRemoval/adapterRemoval.py
```diff
@@ -0,0 +1,311 @@
...
+                {'name': 'Collapsed', 'ylab': 'Count'},
+                {'name': 'Collapsed Truncated', 'ylab': 'Count'},
+                {'name': 'Discarded', 'ylab': 'Count'},
+                {'name': 'All', 'ylab': 'Count'}
+            ]
+        }
+
+        lineplot_data = [
+            self.len_dist_plot_data['mate1'],
+            self.len_dist_plot_data['mate2'],
```
ewels: Will also need to remove the `data_labels` above accordingly.

multiqc/modules/adapterRemoval/adapterRemoval.py
```diff
@@ -0,0 +1,311 @@
...
+            'title': 'retained and discarded',
+            'id': 'ar_retained_plot',
+            'ylab': '# Reads',
+            'hide_zero_cats': False,
+            'cpswitch_counts_label': 'Number of Reads'
+        }
+
+        cats_pec = OrderedDict()
+        cats_pec['singleton_m1'] = {'name': 'singleton mate1'}
+        cats_pec['singleton_m2'] = {'name': 'singleton mate2'}
```
ewels: Please remove the paired-end categories if every dataset is single-end data.

multiqc/modules/adapterRemoval/adapterRemoval.py
```diff
@@ -0,0 +1,311 @@
...
+                self.len_dist_plot_data['singleton'][self.s_name] = dict()
+                self.len_dist_plot_data['collapsed'][self.s_name] = dict()
+                self.len_dist_plot_data['collapsed_truncated'][self.s_name] = dict()
+                self.len_dist_plot_data['discarged'][self.s_name] = dict()
+                self.len_dist_plot_data['all'][self.s_name] = dict()
+
+            if self.__read_type == 'single':
+                if not self.__collapsed:
+                    self.len_dist_plot_data['mate1'][self.s_name][l_data[0]] = l_data[1]
+                    self.len_dist_plot_data['mate2'][self.s_name][l_data[0]] = 0
```
ewels: I'm slightly unsure about setting these values as `0`. They're not technically zero reads, they're actually `None` (or no data). This probably doesn't matter in practice, but can cause confusion sometimes and is something I have actively tried to avoid in other modules.

Looking at the source files, I think the way that I would have approached this code is more along the lines of the [Bismark module](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/bismark/bismark.py) - set up a [load of regexes](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/bismark/bismark.py#L20-L39) with associated keys, and then [search the file for each regex](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/bismark/bismark.py#L166-L177) and save whatever values are present in the file. Then any missing values will be just that - missing. This is also probably more robust to changes in file format than this, as it doesn't depend on the position on the file - instead directly using the associated label.

Having missing data instead of zeroes also has the advantage of not needing to do special cases in the plotting, as empty categories will be ignored by the core MultiQC plotting functions (though you would may have to handle some missing-data cases). This is relevant to my comments below about removing unused categories for SE data.

What happens if you use `None` instead of `0` here? Would be worth trying at least I think.

multiqc/modules/adapterRemoval/adapterRemoval.py
```diff
@@ -0,0 +1,311 @@
...
+                'min': 0,
+                'scale': 'PuBu',
+                'shared_key': 'aligned_total'
+        }
+        self.general_stats_addcols(self.adapter_removal_data, headers)
+
+    def adapter_removal_retained_chart(self):
+
+        pconfig = {
+            'title': 'retained and discarded',
```
ewels: Please use Sentence Case. Also, perhaps `Discarded Reads` would be a little clearer?

alexanderscholz:
hello @ewels,
thanks for your detailed review and your time. The most points were clear and modified with 1cba1022bd6aac2b10cad99616cba5263822eb37. To add/remove the paired-end specific plots i added a new var to recognize if there exists a paired-end dataset. I did the same for the collapsed specific plots because they are only interesting if there are collapsed datasets. What do you think about it?

ewels:
Brilliant stuff, many thanks for your patience and work on this @alexanderscholz! Looks great now, I have no further comments :)

Phil

PR #444, opened 2017-04-10
bcftools: fix ts/tv value rounding

vladsaveliev:
The ts/tv ratio is reported as integer, fixing to display it as decimal.


ewels:
Thanks!

Issue #421, opened 2017-03-10
Module request: SortMeRNA


lfaller:
Hi Phil,

Thanks for providing the MultiQC tool -- it really is making my life a whole lot easier!

I would like to suggest adding a SortMeRNA tool. We run it routinely to identify and remove ribosomal RNAs after sequencing RNAseq data. 

Tool webpage: http://bioinfo.lifl.fr/RNA/sortmerna/
Example of a log file: https://www.dropbox.com/s/udwi7bnmbuuwieo/sortmerna.log?dl=0

The important part is the line: `Total reads failing E-value threshold = 4394708 (98.90%)`

This tells us that 98.9% of the reads analyzed were NOT rRNA, so those can be included in downstream analyses (ie a DE expression analysis). rRNA reads in this case can be thought of as noise, because they don't add anything interesting to the analysis.

It would be good to have an overview of these values for all the samples in one plot, so we can easily identify samples that might not yield good results down the line because they had, for example, large percentages of rRNA.

Unfortunately, the name of the log file by default depends on another input argument (`--aligned`), but it would always end in `.log`:

To get the sample log file I linked to above, I ran this command:

```
[...]

PR #448, opened 2017-04-13
A basic working module to parse SortMeRNA output

bschiffthaler:
SortMeRNA identifies rRNA containing reads in FASTQ files (although any custom file can be used to create a database). It is commonly used in metagenomics analyses and in general data processing to deplete rRNA or other contaminants (e.g.: PhiX)

### DIFF
--- a/multiqc/modules/sortmerna/sortmerna.py
+++ b/multiqc/modules/sortmerna/sortmerna.py
@@ -1,0 +1,151 @@
+#!/usr/bin/env python
+
+""" MultiQC module to parse output from SortMeRNA """
+
+from __future__ import print_function
+from collections import OrderedDict
+import json
+import logging
+import os
+import re
+
+from multiqc.modules.base_module import config, BaseMultiqcModule
+from multiqc import plots
+
+# Initialise the logger
+log = logging.getLogger(__name__)
+
+class MultiqcModule(BaseMultiqcModule):
+
+    def __init__(self):
+
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__(name='SortMeRNA', anchor='sortmerna',
+        href='http://bioinfo.lifl.fr/RNA/sortmerna/',
+        info="is a program tool for filtering, mapping and OTU-picking NGS reads in metatranscriptomic and metagenomic data.")
+
+        # Parse logs
+        self.sortmerna = dict()
+        for f in self.find_log_files(config.sp['sortmerna'], filehandles=True):
+            self.parse_sortmerna(f)
+
+        if len(self.sortmerna) == 0:
+            log.debug("Could not find any SortMeRNA data in {}".format(config.analysis_dir))
+            raise UserWarning
+
+        log.info("Found {} logs".format(len(self.sortmerna)))
+        self.write_data_file(self.sortmerna, 'multiqc_sortmerna')
+        log.debug(self.sortmerna)
+        # Add rRNA rate to the general stats table
+        headers = OrderedDict()
+        headers['rRNA_pct'] = {
+            'title': '% rRNA',
+            'description': '% rRNA',
+            'max': 100,
+            'min': 0,
+            'suffix': '%',
+            'scale': 'OrRd',
+            'format': '{:.1f}%'
+        }
+        self.general_stats_addcols(self.sortmerna, headers)
+
+        # Make barplot
+        self.intro += self.sortmerna_overall_barplot()
+        self.intro += self.sortmerna_detailed_barplot()
+
+    def parse_sortmerna(self, f):
+        s_name = None
+        post_results_start = False
+        post_database_start = False
+        db_number = 0
+        err = False
+
+        for l in f["f"]:
+            if "Reads file" in l:
+                s_name = os.path.basename(l.split(" ")[-1]).strip().split(".")[0]
+                self.sortmerna[s_name] = dict()
+            if "Results:" in l and not post_results_start:
+                post_results_start = True
+            if not post_results_start:
+                continue
+            if post_results_start and not post_database_start:
+                if "Total reads =" in l:
+                    m = re.search("\d+",l)
+                    if m:
+                        self.sortmerna[s_name]["total"] = int(m.group())
+                    else:
+                        err = True
+                elif "Total reads passing" in l:
+                    m = re.search("\d+",l)
+                    if m:
+                        self.sortmerna[s_name]["rRNA"] = int(m.group())
+                        self.sortmerna[s_name]["rRNA_pct"] = float(self.sortmerna[s_name]["rRNA"]) / float(self.sortmerna[s_name]["total"]) * 100
+                    else:
+                        err = True
+                elif "Total reads failing" in l:
+                    m = re.search("\d+",l)
+                    if m:
+                        self.sortmerna[s_name]["non_rRNA"] = int(m.group())
+                        self.sortmerna[s_name]["non_rRNA_pct"] = float(self.sortmerna[s_name]["non_rRNA"]) / float(self.sortmerna[s_name]["total"]) * 100
+                    else:
+                        err = True
+            if post_database_start:
+                if not l.strip():
+                    break
+                db_number = db_number + 1
+                m = re.search("    .*\t", l)
+                if m:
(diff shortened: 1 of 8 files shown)


ewels:
Aha, brilliant stuff. This one was on my to-do list, thanks @bschiffthaler! (See #421).

I'll start a code review now..

Phil

multiqc/modules/sortmerna/sortmerna.py
```diff
@@ -0,0 +1,151 @@
...
+
+        for l in f["f"]:
+            if "Reads file" in l:
+                s_name = os.path.basename(l.split(" ")[-1]).strip().split(".")[0]
+                self.sortmerna[s_name] = dict()
+            if "Results:" in l and not post_results_start:
+                post_results_start = True
+            if not post_results_start:
+                continue
+            if post_results_start and not post_database_start:
```
ewels: This could probably be generalised / simplified a little by using nested loops:

```python
if "Results:" in l:
    while "By database:" not in l:
        for l in f["f"]:
            key, val = l.strip().split('=', 1)
            key = key.strip().replace(' ', '_') 
            try:
                self.sortmerna[s_name][ key ] = int(val)
            except ValueError:
                rcount = val.split('(', 1)[0]
                self.sortmerna[s_name][key] = int(rcount)
                self.sortmerna[s_name]["{}_pct".format(key)] = float(rcount) / float(self.sortmerna[s_name]["Total_reads"]) * 100.0
if "By database:" in l:
    while l.strip() != '':
        for l in f["f"]:
            # ...
```
(only a suggestion)

multiqc/modules/sortmerna/sortmerna.py
```diff
@@ -0,0 +1,151 @@
...
+            'title': 'SortMeRNA Other vs rRNA',
+            'ylab': 'Reads'
+        }
+        log.debug(keys)
+        return plots.bargraph.plot(self.sortmerna, keys, pconfig)
+
+    def sortmerna_detailed_barplot (self):
+        """ Make the HighCharts HTML to plot the sortmerna rates """
+
+        colors = ["#1f78b4", "#b2df8a", "#33a02c", "#fb9a99",
```
ewels: Any reason for the non-default colours? Is this matching output from SortMeRNA or something?

multiqc/modules/sortmerna/sortmerna.py
```diff
@@ -0,0 +1,151 @@
...
+                else:
+                    err = True
+            if "By database:" in l and not post_database_start:
+                post_database_start = True
+        if err:
+            log.warning("Error parsing data in: " + s_name)
+            self.sortmerna.pop(s_name, 'None')
+        s_name = None
+
+    def sortmerna_overall_barplot (self):
```
ewels: Does this plot just show rRNA? I feel that it may not be necessary as %rRNA is in the General Statistics table as well.

multiqc/utils/search_patterns.yaml
```diff
@@ -203,6 +203,8 @@ star:
     fn: '*Log.final.out'
 star_genecounts:
     fn: '*ReadsPerGene.out.tab'
+sortmerna:
+    contents: 'Command: sortmerna'
```
ewels: This doesn't work with the example data in the [MultiQC_TestData](https://github.com/ewels/MultiQC_TestData/blob/master/data/modules/SortMeRNA/sortmerna.log) repo, as the command can have the full path to `sortmerna`. The above has:
```
Command: /home/lina/sortmerna/sortmerna
```

I guess just `contents: 'sortmerna'` should be sufficiently specific?

ewels:
This is really nice work, thanks @bschiffthaler!

My only general comment / question is about people using SortMeRNA with databases that don't contain rRNA sequences (as discussed a little in #421). In this case, the results in the report will be kind of confusing, as several of the categories are hardcoded with labels for rRNA. I'm not sure that there's any way to get around this though, so I'm happy to stick with the approach that you've used here if we can't think of anything better.

Phil

multiqc/utils/search_patterns.yaml
bschiffthaler: Not entirely sure actually. E.g.: We tend to rename files based on the tools they have been through, so subsequent logs would have 'sortmerna' in them because the file would be:

`..._sortmerna.fq.gz`

Could match one of the rarer terms in there s.a. `Gumbel lambda =`

multiqc/modules/sortmerna/sortmerna.py
bschiffthaler: No particular reason. I can drop them

ewels: I was mostly just curious :) If you prefer your palette then feel free to stick with it.

multiqc/modules/sortmerna/sortmerna.py
bschiffthaler: It shows hits vs. other. See [here](http://imgur.com/a/04NJ4)

multiqc/utils/search_patterns.yaml
ewels: That's an excellent point, I hadn't thought about filenames. Ok then yes, one of those rarer terms would be ideal.

multiqc/modules/sortmerna/sortmerna.py
ewels: Yup - but is this not essentially the same as the _% rRNA_ column in the table above? Aside from showing read counts (typically also shown in the General Stats table, eg. from FastQC).

bschiffthaler: It is indeed. Personally, I'm a visual type and I do like to look at the plot, but essentially you're right. It is a bit superfluous

ewels: The General Stats table has bars in the background, that's kind of like a plot..? 😳 

I'd usually always be keen on more visual plots too, but if it's essentially a duplicate, then I'd prefer to keep reports a little shorter / smaller I think. Especially given that this tool will probably pretty much always be accompanied by sections from multiple other programs.

If you'd prefer to keep this plot and find it helpful, then feel free to do so though. You'll be reading reports with this much more than I will!

ewels:
ps. Remember to add yourself to the list of contributors on the homepage!

multiqc/modules/sortmerna/sortmerna.py
bschiffthaler: I think it's better for the general user to drop this plot then. I can always have a custom module lying around for myself 😄

bschiffthaler:
>My only general comment / question is about people using SortMeRNA with databases that don't contain rRNA sequences (as discussed a little in #421). In this case, the results in the report will be kind of confusing, as several of the categories are hardcoded with labels for rRNA. I'm not sure that there's any way to get around this though, so I'm happy to stick with the approach that you've used here if we can't think of anything better.

How about 'SortMeRNA hit %' or something along those lines?

ewels:
> How about 'SortMeRNA hit %' or something along those lines?

Yeah, that could work, though it's a little verbose and unclear to anyone reading the report who hasn't run the pipeline themselves. Another option would be to default to `rRNA` but look for a config file option first, so that people can override this behaviour if they want to. This is usually what I do when I can't think of a better fix-all solution. eg:

```python
seqname = getattr(config, 'sortmerna', {}).get('seqname', 'rRNA')
```

Then it will default to rRNA, but people can set:

```yaml
sortmerna:
    seqname: 'Something Else'
```

Phil

multiqc/modules/sortmerna/sortmerna.py
```diff
@@ -33,14 +34,14 @@ def __init__(self):
             log.debug("Could not find any SortMeRNA data in {}".format(config.analysis_dir))
             raise UserWarning
 
-        log.info("Found {} logs".format(len(self.sortmerna)))
```
ewels: Did you mean to drop these two lines? I was only referring the to the `log.debug` statement that looked like it was accidentally left over from testing..

bschiffthaler: I did not. Should not multi-task...

ewels:
Brilliant, thanks! 👍

PR #449, opened 2017-04-19
Search refactoring

ewels:
A rewrite of how the file search mechanism works. Roughly speaking, the process was / is:

* Before:
    * MultiQC generates a list of files when it begins (skipping v. large files etc)
    * Each module runs through the list looking for its files
    * Modules supply a config dictionary to the search function
    * Snazzy new progress bar whilst doing the first search
* Now:
    * MultiQC searches files for each module search pattern when it begins
    * Each module supplies it's search key (instead of a config dict) and is given back the matching files, without doing a second search
    * Backwards compatibility is currently maintained for plugins etc. that supply search dicts, though with a warning.

Because MultiQC was searching the full file list very many times before (once for each module), it could be quite slow when run with lots of files. This refactoring appears to speed up execution time by around 25-50%. Not bad really! Speed up will be most noticeable when running with lots of files which are all matched by modules by filename. Modules with these search patterns are run first and the files removed from the search list.

All feedback welcome!

Phil

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,27 @@ Version 1.0! Bumping the major version number for a couple of reasons:
    compatibility for plugins. As such, semantic versioning suggests a change in
    major version number.
 
-Most people who run MultiQC just use the core installation.
-If this is the case for you, then you have nothing to worry about. The changes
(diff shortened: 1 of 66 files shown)


ewels:
Ok, I'm happy with this now. Merging..

PR #453, opened 2017-04-25
SnpEff module: fix impacts plot title

vladsaveliev:
The "variant by effect impact" plot was incorrectly titled as "variants by region" (probably because of the copy-paste from the other plot).


ewels:
Good spot, thanks!

Issue #49, opened 2015-09-30
feature request - preseq lc_extrap curves


avilella:
For example, for a paired-end bam file, preseq lc_extrap can be run like this:
 
```
Input file(s):	SAMPLE.bam
Output file(s):   SAMPLE.preseq.lc_extrap.tsv
Tool(s):      	preseq lc_extrap
 
$ preseq lc_extrap
	-B –P  
	-o SAMPLE.preseq.lc_extrap.tsv  
	SAMPLE.bam
```
The output of preseq lc_extrap can be plotted in a curve indicating the ratio of “number of molecules (M)” and “number of distinct molecules (M)”. See below a description of three examples from the preseq website (http://smithlabresearch.org/software/preseq/challenge/)

Instructions on how to install this software and its dependencies can be found at http://smithlabresearch.org/software/preseq

ewels:
Yup, absolutely - we routinely run preseq here too, so I've been planning to write this into MultiQC as well.

I'll probably model it on our [existing plotting script for preseq](https://github.com/ewels/ngi_visualizations/tree/master/ngi_visualizations/preseq_complexity_curves)..

ewels:
@avilella - that's done. Let me know if it works for you, I wasn't super happy for the way that it finds files (currently just searches for the string `TOTAL_READS	EXPECTED_DISTINCT`

This is what the output looks like:

![image](https://cloud.githubusercontent.com/assets/465550/10454539/89e46244-71b5-11e5-879a-0286c5393305.png)

vladsaveliev:
Hi Phil, 

I'm wondering if you have MultiQC-based code for preseq with the extended features for displaying depth instead of counts and dots with current counts, like in your matplotlib-based visualization  https://github.com/ewels/ngi_visualizations/tree/master/ngi_visualizations/preseq_complexity_curves
I'm thinking of tackling it in my custom plugin for MultiQC and provide the required extra metadata though command line, but wanted to first check if you already done that.

-
Vlad

ewels:
Hi @vladsaveliev,

This isn't something I've done yet, no. It's been in the back of my mind for a while though - it would certainly be nice to have. The X coverage thing should be pretty easy and I _think_ that you should be able to plot the additional dots by using the [`extra_series`](http://multiqc.info/docs/#additional-data-series) pconfig option.

[...]

PR #461, opened 2017-05-15
Preseq: plotting coverage and real counts

vladsaveliev:
Following the discussion on #49, adding features similar to the matplotlib-based plotting script https://github.com/ewels/ngi_visualizations/blob/master/ngi_visualizations/preseq_complexity_curves/plot_complexity_curves.py:
- 2 new options in config (`preseq: genome_size` and `preseq: read_length`) that make the module show estimated depth of coverage instead of less informative molecule counts. `genome_size` can be a base pairs count or a build name (supported hg19, hg38, mm10).
- If a file named as `*preseq_real_counts*` is found, MultiQC takes it to parse out real read counts (and optionally real unique read counts) to plot corresponding dots over the line graph.
- Fixing curve trimming:
  - relying on order of the list made from `d.keys()` set - adding sorting to fix,
  - docs say that 90% of one sample of y values is enough - in fact all samples need to satisfy this.
- Fix the changelog section on curve trimming (80% -> 90%, max -> min),
- Update the Preseq docs and the changelog.


vladsaveliev:
The second commit is a small update on the new feature of configuration through command line options. Currently it overrides all values specified in other file-based YAML configs. However it would be handy to override only some specific nested parameters, for instance - set `--cl_config "preseq: {notrim: True}"` without forcing the deletion of `preseq: { genome_size: 134132, read_length: 100 }`

vladsaveliev:
This is an example of how it looks with coverage and external counts:

screen shot 2017-05-16 at 01 16 11

If the dots exceed the trimmed maximum X value, the curves are extended to catch all the dots. 

This is how it looks with `notrim: True`:
screen shot 2017-05-16 at 01 17 49

By the way, I'm thinking that the rule "until one of the datasets shows 90%" might be too aggressive for trimming. The limit is too close to the real values so the extrapolation doesn't make a lot of sense. Maybe it's better to trim when _all_ samples show 90% of y values? On the other hand, my dataset might be too specific and small, I'll take a look at bigger targets.

ewels:
Regarding the 90% cutoff - it's tricky to get a solution that stops the plots from being ridiculous but doesn't limit more normal plots. I'll e-mail you my "massive" test data so you can have a play..

ewels:
Ups, the e-mail address on your GH profile isn't co-operating..

> Your message cannot be delivered to the following recipients:
>
> Recipient address: vladsaveliev@me.com
> Reason: Over quota

vladsaveliev:
Thanks a lot for all the comments and suggestions - will sort it out and will run against the test data you've sent me.

I guess I better change my github email to the one hosted in gmail. I'm sure those are the issues with Apple's me.com. Anyway, I got the data!

multiqc/modules/preseq/preseq.py
```diff
@@ -95,22 +190,45 @@ def preseq_length_trimmed_plot (self):
...
+                    if d[x] > ymax and x > real_counts_total.get(sn, 0) and x > real_counts_unique.get(sn, 0):
+                        break
+                xmaxs.append(xmax)
+            pconfig['xmax'] = min(xmaxs)
+            description += "<p>Note that the x axis is trimmed until one of the datasets \
+                shows 90% of its maximum y-value, to avoid ridiculous scales.</p>"
+        else:
+            ymax = max(max(d.values()) for d in self.preseq_data.values())
+            
+        pconfig['extra_series'].append({  # Plot perfect library as dashed line
```
vladsaveliev: The dash line needs `ymax` to be defined first. I removed `self.total_max` and instead using `ymax` for plotting the dashed line. It's more accurate this way and also doesn't need to specifically convert `self.total_max` to coverage.

multiqc/modules/preseq/preseq.py
```diff
@@ -70,23 +74,114 @@ def parse_preseq_logs(self, f):
...
+        # Convert counts to coverage
+        genome_size = getattr(config, 'preseq', {}).get('genome_size')
+        if genome_size is not None:
+            try:
+                genome_size = float(genome_size)
+            except ValueError:
+                genome_sizes = {'hg19': 2897310462,
+                                'hg38': 3049315783,
+                                'mm10': 2652783500}
+                if genome_size in genome_size:
```
vladsaveliev: Yep, typo, thanks! I was always thinking it might be a good idea to avoid variable names that are different only in plural "-s" - this causes typos too often...

docs/modules/preseq.md
```diff
@@ -19,8 +19,45 @@ back the x axis until one of the datasets shows 90% of its maximum y-value
...
+    read_length: 300
+```
+These parameters make the script take every molecule count and divide it by 
+(genome_size / read_length).
+
+MultiQC comes with effective genome size presets for Human and Mouse, so you can 
+provide the genome build name instead, like this: `genome_size: hg38`. Only the
+following builds are supported: `hg19`, `hg38`, `mm10`.
+
+#### Plotting the "real" data
```
vladsaveliev: Well, the whole plot is read counts, both x and y axis... So _"real"_ kind of means _not extrapolated_. Maybe _"Marking actual read counts"_?

ewels: Or "Plotting externally calculated read counts"? Bit of a mouthful, but maybe clearer..

multiqc/modules/preseq/preseq.py
ewels: I thought the same 👍  `genome_presets`?

multiqc/modules/preseq/preseq.py
ewels: ok nice 👍

docs/modules/preseq.md
```diff
@@ -19,8 +19,45 @@ back the x axis until one of the datasets shows 90% of its maximum y-value
...
+preseq:
+    genome_size: 3049315783
+    read_length: 300
+```
+These parameters make the script take every molecule count and divide it by 
+(genome_size / read_length).
+
+MultiQC comes with effective genome size presets for Human and Mouse, so you can 
+provide the genome build name instead, like this: `genome_size: hg38`. Only the
+following builds are supported: `hg19`, `hg38`, `mm10`.
```
vladsaveliev: I started to think that probably human genome presets are not super useful for the Preseq's module. Human genome experiments in most cases are targeted or exomes, and `genome: hg19` works only for WGS. However it's still a good idea - there are other modules that would benefit from knowing the genome, for instance the Qualimap's module could plot GC in its histogram regardless if Qualimap provided the theoretical distribution.

ewels: Depends where you are - nearly all of our preseq runs are on WGS :) But yes, this would be in addition to other more specific things that people could set. And as you say, I envisage more "add-on" features being added in the future which could find this useful. That was kind of my thinking - better to add a generic option like this earlier rather than retrofit later.

vladsaveliev:
Alright, having a chance to get back to this. 

I'm wondering if you know what are the read counts for those big data samples?

The files a huge indeed - Preseq extrapolates up to 10 billion reads, but I'm thinking that experienced users explicitly set Preseq's command line options to set the step and the limit, so that's not a very big deal if we mis-trim. Anyway, I would change the rule "at least on sample hits 90%" (on the 1st picture) to "all samples hit 90%" (on the 2nd picture) as it looks more reasonable:

image

image

ewels:
Sounds great 👍🏻 Not sure what the read counts were (can look tomorrow), but not super huge.

multiqc/utils/config.py
```diff
@@ -185,3 +186,14 @@ def load_sample_names(snames_file):
     except (IOError, AttributeError) as e:
         logger.error("Error loading sample names file: {}".format(e))
     logger.debug("Found {} sample renaming patterns".format(len(sample_names_rename_buttons)))
+
+def update_dict(d, u):
```
vladsaveliev: Anyway it doesn't work in `util_functions` because `config` is imported there and it would cause an import loop

multiqc/modules/preseq/preseq.py
```diff
@@ -70,23 +74,114 @@ def parse_preseq_logs(self, f):
...
+                              '#f15c80', '#e4d354', '#2b908f', '#f45b5b', '#91e8e1']
+            for si, sn in enumerate(sorted(self.preseq_data.keys())):
+                if sn in real_counts_total:
+                    t_reads = float(real_counts_total[sn])
+                    point = {
+                        'color': default_colors[si],
+                        'showInLegend': False,
+                        'marker': {
+                            'enabled': True,
+                            'symbol': 'diamond'
```
vladsaveliev: A black border looks the most visible, will keep it like this:
image

multiqc/modules/preseq/preseq.py
```diff
@@ -70,23 +74,114 @@ def parse_preseq_logs(self, f):
...
+
+        real_counts_total = {}
+        real_counts_unique = {}
+        if real_counts_file_raw is not None:
+            try:
+                for line in real_counts_file_raw.splitlines():
+                    if not line.startswith('#'):
+                        line = line.strip()
+                        cols = line.split()  # Split on any whitespace
+                        for sn in [cols[0], "{}.preseq".format(cols[0]), "{}.bam".format(cols[0])]:
```
vladsaveliev: Done that

docs/modules/preseq.md
vladsaveliev: Okay, now it looks more clear I think

docs/modules/preseq.md
vladsaveliev: Added the generic `config.genome` option, using it in fastqc and preseq modules. Hope I didn't break anything, but tried to keep it as simple as possible.

vladsaveliev:
Addressed the comments, and pushed a couple of unrelated fixes (creation timestamp was off - showed month for minutes; and outdated Jinja2 got broken so specified the version range in setup.py). Looks ready to merge I hope.

And by the way I also have a pull request in MultiQC_TestData.

multiqc/utils/config.py
ewels: Ah ok fine 👍

ewels:
Just looked up the above samples: they had between 3 and 16 million reads, so the 422 million that preseq plots is probably a little excessive. These plots are generated by our automated pipeline which is fairly naïve and doesn't know about the samples or set any parameters. Anyway - your "all > 90%" sounds sensible and is totally what I should have done in the first place.

Great spot on the other stuff! Old package versions are tricky bugs to fix - I was fighting with something similar the other day and it was super confusing.

ewels:
Looks great! In a meeting now, will merge when I get back.

One thing - I kind of intended the `config.genome` thing to be in addition to everything else and only called if the other flags are missing. And without the `_genome` suffixes, as it kind of duplicates itself. So the FastQC thing you could specify `hg18_genome` or `hg18_transcriptome` etc exactly as before, but if that's not set and `genome` is set, then `genome = 'hg18'` would be interpreted as `hg18_genome` etc...

Minor style thing, so will happily merge and do this myself if you like :)

Phil

vladsaveliev:
Yep, I figured that was the original idea - to make it without the suffix. However, all modules act differently in case of a genome or a transcriptome, and I think there is no good reason to presume either of those - a user analysing RNAseq might want to just go with `genome: hg38` and not suspect that FastQC will draw an irrelevant GC distribution (by the way, what if it's an exome?). Especially if the user set the option for some other module and is not aware of FastQC using it (maybe it's a good reason to drop the idea of generic option for good?).

I agree that for some people the option named `genome` might presume genome, not transcriptome, so maybe make it `genome_build` like in bcbio? Or make 2 options: `genome: hg38` and `transcriptome: hg38`? Or allow both values `hg38` and `hg38_genome` and just add a more complex logic into modules' code? Configuration and thinking about all scenarios is always so tricky... Anyway, I won't mind if you replace my code for that with something completely different - my solution was just a quick one anyway. Let me know what you think!

ewels:
Hmm, these are good points that I hadn't thought of. Especially the thing about specifying it for one module then getting inaccurate data in another. Maybe we should drop this idea then..

vladsaveliev:
Alright, dropped the code for `config: genome`. 

I'm going to fetch upstream and rebase just to double check it works good with your more recent changes since I started the PR.

ewels:
Ok great - sorry for asking you to add that and then remove it! 😣 

Just tested the code out for myself - looks great! Are you happy for it to be merged? Only tiny things I noticed were that the coverage axes would maybe be nicer with a `X` suffix for units, and that the tooltip seems to be showing integers.

multiqc/modules/preseq/preseq.py
```diff
@@ -95,22 +193,42 @@ def preseq_length_trimmed_plot (self):
             'ymin': 0,
             'xmin': 0,
             'tt_label': '<b>{point.x:,.0f} total</b>: {point.y:,.0f} unique',
```
ewels: Here's the thing with the ints: `{point.x:,.0f}` (not usually a problem for counts, but annoying when coverage). Might be nice to add the suffix to the tooltip too (`molecules` or `X coverage`).

ewels:
ok, will merge if that's ok as I want your nice new merge function so that I can play with the new config hook :)

vladsaveliev:
Thanks Phil for merging! Sorry for delays - was busy with meetings.

I'm wondering if using floats for coverage is useful. I think in most cases the coverage numbers are rather high (especially for lc_extrap) and large floats look ugly. I might be wrong
screen shot 2017-05-17 at 18 01 15

ewels:
No problem!

Hah, for exomes (?) it does look a big ugly. I was looking at low-coverage WGS and the unique coverage was all below 1x, so everything just said `0`..

vladsaveliev:
Yep that's a targeted sequencing, but I'm thinking that even WGS usually has the average coverage of at least 10x, otherwise how would one analyse it? Sorry if I'm biased by the kind of projects I usually see.

ewels:
Quite easily I think - for example if WGS aims for 15x and gets 10x. A rubbish library could be very duplicated ending up with a final deduplicated coverage of ~ 1x.

The files I send over e-mail were from a low-input titration test and all end up with < 0.1x unique coverage. We also work with FFPE and ancient samples which can come back with very low coverage. I imagine single-cell work would be similar.

However, it should be super easy to check the maximum coverage and set dynamically if it annoys you :) I'll push this now.

Phil

vladsaveliev:
No doesn't annoy me at all, I'm just trying to make sure I'm on the same page with you :) Thanks for the examples! And thanks for the commit, that's a nice solution.

PR #464, opened 2017-05-22
Generate names with increasing report names instead of prompting for a --force

Hammarn:
This is an attempt at implementing dynamically generated report names instead of just asking  for a `-f` flag. 
This currently leaves some extra code that could be removed, but I wanted to open the PR for @ewels to have a look and leave some input before I put more work into it.


scripts/multiqc
```diff
@@ -161,6 +161,7 @@ logger = config.logger
 def multiqc(analysis_dir, dirs, dirs_depth, no_clean_sname, title, report_comment, template, module, exclude, outdir,
 ignore, ignore_samples, sample_names, file_list, filename, make_data_dir, no_data_dir, data_format, zip_data_dir, force, export_plots,
 plots_flat, plots_interactive, make_pdf, config_file, cl_config, verbose, quiet, **kwargs):
+
```
ewels: Please drop extra whitespace ;)

scripts/multiqc
```diff
@@ -459,10 +439,17 @@ plots_flat, plots_interactive, make_pdf, config_file, cl_config, verbose, quiet,
                 logger.warning("Deleting    : {}   (-f was specified)".format(os.path.relpath(config.output_fn)))
                 os.remove(config.output_fn)
             else:
-                logger.error("MultiQC Report {} already exists.".format(config.output_fn))
+                report_num = 0
                 logger.info("Use -f or --force to overwrite existing reports")
-                shutil.rmtree(tmp_dir)
-                sys.exit(1)
+                while os.path.exists(output_name):
```
ewels: You haven't defined this variable yet, so I get the following error:

```python
  File "/Users/philewels/.miniconda/envs/work/bin/multiqc", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/Users/philewels/GitHub/MultiQC/scripts/multiqc", line 629, in <module>
    multiqc()
  File "/Users/philewels/.miniconda/envs/work/lib/python2.7/site-packages/click/core.py", line 700, in __call__
    return self.main(*args, **kwargs)
  File "/Users/philewels/.miniconda/envs/work/lib/python2.7/site-packages/click/core.py", line 680, in main
    rv = self.invoke(ctx)
  File "/Users/philewels/.miniconda/envs/work/lib/python2.7/site-packages/click/core.py", line 873, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/philewels/.miniconda/envs/work/lib/python2.7/site-packages/click/core.py", line 508, in invoke
    return callback(*args, **kwargs)
  File "/Users/philewels/GitHub/MultiQC/scripts/multiqc", line 444, in multiqc
    while os.path.exists(output_name):
UnboundLocalError: local variable 'output_name' referenced before assignment
```

But anyway, I think a cleaner way to write this (without the intermediate variable `output_name` would be like this:

```python
while os.path.exists(config.output_fn):
    base, ext = os.path.splitext(config.output_fn_name)
    config.output_fn = os.path.join(config.output_dir, "{}_{}{}".format(base, report_num, ext) )
    report_num += 1
config.output_fn_name = os.path.basename(config.output_fn)
```
(`config.output_fn_name` wasn't being updated before and is used elsewhere)

scripts/multiqc
```diff
@@ -478,11 +465,12 @@ plots_flat, plots_interactive, make_pdf, config_file, cl_config, verbose, quiet,
                     logger.warning("Deleting    : {}   (-f was specified)".format(os.path.relpath(config.data_dir)))
                     shutil.rmtree(config.data_dir)
                 else:
-                    logger.error("Output directory {} already exists.".format(config.data_dir))
-                    logger.info("Use -f or --force to overwrite existing reports")
-                    shutil.rmtree(tmp_dir)
-                    sys.exit(1)
-            os.makedirs(config.data_dir)
+                    data_num = 0
```
ewels: I still think that it's possible to get numeric suffixes out of sync here. For example, if there is no `multiqc_report.html` but there is a `multiqc_data` directory, then we'll end up with `multiqc_report.html` and `multiqc_data_1`, no?

I think we'd be better off rolling these two checks into one. Something like:

```python
config.output_fn = os.path.join(config.output_dir, config.output_fn_name)
config.data_dir = os.path.join(config.output_dir, config.data_dir_name)
if os.path.exists(output_name) or os.path.exists(config.data_dir):
    # [..]
    report_num = 1
    while os.path.exists(output_name) or os.path.exists(config.data_dir):
       # [..]
```

Would need a bit more logic, for example making sure that stuff like `config.make_data_dir` still works.. Perhaps we could set `config.data_dir` to `None` in that case, and then `os.path.exists(str(config.data_dir))` should return `False` assuming that there isn't a file called `None` in the working directory..?

scripts/multiqc
```diff
@@ -459,10 +439,17 @@ plots_flat, plots_interactive, make_pdf, config_file, cl_config, verbose, quiet,
                 logger.warning("Deleting    : {}   (-f was specified)".format(os.path.relpath(config.output_fn)))
                 os.remove(config.output_fn)
             else:
-                logger.error("MultiQC Report {} already exists.".format(config.output_fn))
+                report_num = 0
```
ewels: Nitpick: I think it'd probably be nicer to start with `multiqc_report_1.html`? instead of `multiqc_report_0.html`?

scripts/multiqc
```diff
@@ -459,10 +439,17 @@ plots_flat, plots_interactive, make_pdf, config_file, cl_config, verbose, quiet,
...
-                shutil.rmtree(tmp_dir)
-                sys.exit(1)
+                while os.path.exists(output_name):
+                    output_name, file_ending = os.path.splitext(config.output_fn)
+                    output_name.append("_{}".format(report_num))
+                    output_name.append(file_ending)
+                    report_num += 1
+                #We keep going until a new name is found, set the new name as the name to use. 
+                config.output_fn=output_name
+                #shutil.rmtree(tmp_dir)
```
ewels: -- Debugging statements

scripts/multiqc
```diff
@@ -459,10 +439,17 @@ plots_flat, plots_interactive, make_pdf, config_file, cl_config, verbose, quiet,
...
                 logger.info("Use -f or --force to overwrite existing reports")
-                shutil.rmtree(tmp_dir)
-                sys.exit(1)
+                while os.path.exists(output_name):
+                    output_name, file_ending = os.path.splitext(config.output_fn)
+                    output_name.append("_{}".format(report_num))
+                    output_name.append(file_ending)
+                    report_num += 1
+                #We keep going until a new name is found, set the new name as the name to use. 
+                config.output_fn=output_name
```
ewels: Would be good to add a `logger.warning` message here saying that there is a conflicting filename, report will be saved as _x_...

Hammarn:
The code now runs properly according to the Travis test, and it produces the result that I excpect. I'll now considered this done on my end. It's probably a good idea if you take a proper look through the code @ewels in case I removed something that I shouldn't have.

ewels:
Great stuff, tested and merged! Thanks for this..

Phil

PR #466, opened 2017-05-25
Minor preseq changes, collapsing rc changes in bcftools, shorter table headers

vladsaveliev:
Hi Phil,

A few minor changes in my fork has piled up and I figured it might be useful to add them to the main code base.

1. First, a couple of minor changes and fixes in the Preseq module:
    - In the edge case of externally provided unique counts being higher then Preseq's estimation, the dashed line gets cut too early. Making it go up the highest counts number.
    - Fix rounding of "unique" depths to int.
    - Slightly changed the cut-offs for depth labels precision.
    - Add anchor and name for the plot section.

2. Slightly shortened the column headers for `qualimap bamqc` and `bcftools stats` modules, making them nicer fit in very wide tables.

3. For the Bcftools's substitutions plot, add a config option to make it collapse complementary changes like `A>G` / `T>C` which usually mirror each other in non-stranded specific experiments.


ewels:
Great stuff, thanks!

Code all looks good, only query I would have is for using `IS` instead of `Insert Size`, not sure how intuitive that one is? It's the only one that I didn't understand from its shortened name. Perhaps just `Insert` instead? Then the `bp` suffix (if it has one?) should help to make it clear (plus hover text).

vladsaveliev:
Thanks Phil!

I asked around and many find the word `insert` unclear in isolation, it might be the case that `insert size` became like an idiom that looks right only when the both words are there... But can't say for everyone of course. So maybe stick to `Ins. size`?

Yep, and it has the `bp` suffix already.

vladsaveliev:
If you are okay with that, it's good to be merged now.

ewels:
@vladsaveliev - I've changed `GC` to be `% GC` so that it's in line with FastQC / other modules. Hope this is ok!

vladsaveliev:
That's great, thanks!

PR #471, opened 2017-06-05
Jellyfish module for kmer frequency plots

vezzi:
This PR adds Jellyfish to MultiQC to plot kmer-counts that are useful to estimate the coverage in de novo project and to judge library complexity if no reference is available


multiqc/modules/jellyfish/jellyfish.py
```diff
@@ -0,0 +1,115 @@
...
+        In an ideal world (no errors in sequencing, no bias, no
+        repeated regions) this plot should be as close as
+        possible to a gaussian distribution. In reality we will
+        always see a peak for x=1 (i.e., the errors) and another
+        peak close to the expected coverage. If the genome is
+        highly heterozygous a second peak at half of the coverage
+        can be expected."""
+        
+        
+        self.sections = list()
```
ewels: The v1.0 release introduced a new syntax for this. See the [docs FAQ](http://multiqc.info/docs/#updating-for-compatibility) for more info. In short, could you please change this to:

```python
self.add_section(
  name = 'Jellyfish k-mers',
  anchor = 'jellyfish_kmer_plot',
  description = 'Jellyfish plot for k-mers with coverage between 0 and {}'.format(self.jellyfish_max_x),
  plot = self.frequencies_plot()
)
```

Note that I think `name` is used in the side navigation if there is ever a second section for this module, so should be relatively short if possible.

multiqc/modules/jellyfish/jellyfish.py
```diff
@@ -0,0 +1,115 @@
...
+        self.sections.append({
+            'name': 'Jellyfish plot for k-mers with coverage between 0 and {}'.format(self.jellyfish_max_x),
+            'anchor': 'jellyfish_kmer_plot',
+            'content': self.frequencies_plot()
+        })
+
+    def clean_jellyfish_data(self):
+        """ Avoid to plot long flat tails by loosing 0,5% of the tail """
+        max_x = 0
+        #find where the max is
```
ewels: Can you not do this whilst parsing the files, to skip a step?

setup.py
```diff
@@ -97,6 +97,7 @@
             'trimmomatic = multiqc.modules.trimmomatic:MultiqcModule',
             'gatk = multiqc.modules.gatk:MultiqcModule',
             'adapterRemoval = multiqc.modules.adapterRemoval:MultiqcModule',
+            'jellyfish = multiqc.modules.jellyfish:MultiqcModule',
```
ewels: This list was alphabetically sorted once.. :)

multiqc/utils/search_patterns.yaml
```diff
@@ -60,6 +60,8 @@ htseq:
     contents: '__too_low_aQual'
 hicup:
     fn: 'HiCUP_summary_report*'
+jellyfish:
+    fn: '*.hist'
```
ewels: This is pretty generic. Is there nothing better that we can use?

multiqc/modules/jellyfish/jellyfish.py
```diff
@@ -0,0 +1,115 @@
...
+        #clean histogram to prevent long flat tail
+        self.clean_jellyfish_data()
+        
+        if len(self.jellyfish_data) == 0:
+            log.debug("Could not find any data in {}".format(config.analysis_dir))
+            raise UserWarning
+            
+        log.info("Found {} reports".format(len(self.jellyfish_data)))
+        
+        self.intro += """Jellyfish k-mer plots for library complexity.
```
ewels: This is pretty long.. Maybe better in `self.add_section` under the `help` attr? Will display in a similar way currently, but planning to make the styling better in the future so that it's hidden by default unless the user clicks a 'Help' button..

setup.py
vezzi: back to the origin....

multiqc/utils/search_patterns.yaml
vezzi: This is the only one I am not able to address.
So Jellyfish does not create a file with a fixed extension but the default extension is `.jf`. The `.jf` file  is the database file which also contains the keyword `jellyfish` that might be used in combination with the extension to find the right file
```
jellyfish:
   fn: ´*.jf´
   contents: 'jellyfish'
```

however I do not need this file but I need the one is generated by running jellyfish command
```
jellyfish histo -o file.hist -f file.jf
```

`file.hist` does not contain anything except the data to plot. The `.hist` extension is arbitrary but should be pretty natural.

A possibility might be that I look for `.jf` and jellyfish as key, and then I use the `fn` to see if the `hist` is present. If no I can terminate the execution or skip the sample without the `.hist` file. 
Not sure that this way of proceed agrees with MultiQC behaviour...

ewels: Ah damn, these ones are the worst - no identifying file contents and a totally user-defined filename. If the `.hist` extension is totally arbitrary and we need to tell people to use a specific filename pattern anyway (or customise their search pattern config), could it be `*_jf.hist` instead? This will need a paragraph in the docs markdown file explaining how it works.

Looking for `.jf` files could be nice in addition? May be a bit of a shot in the dark though.. Not sure if it's worth it.

vezzi: Well also the `.jf` extension is only the default, a user can override it so anyway an explanation is needed.
I will add a paragraph in the docs, and search for files with extension `*_jf.hist`.... a possible feature is that a user can specify a pattern to be searched, something like `--jellyfish-extension 'fn:*.hist;contents:\'content to be searched\''`

ewels: Ah ok, then yes just describe the behaviour in the docs then. No need for a new setup I think, can just use existing config options: http://multiqc.info/docs/#module-search-patterns

Config can now also be set on the [command line](http://multiqc.info/docs/#command-line-config) so can already do the that if people don't want to mess with config files. Something like this (untested):
```
multiqc --cl_config "sp: { jellyfish: { fn: '*.whatever' } }"
```

multiqc/modules/jellyfish/jellyfish.py
```diff
@@ -0,0 +1,96 @@
...
+# Initialise the logger
+log = logging.getLogger(__name__)
+
+
+class MultiqcModule(BaseMultiqcModule):
+    def __init__(self):
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__(name='jellyfish', anchor='jellyfish',
+        href="http://www.cbcb.umd.edu/software/jellyfish/",
+        info="JELLYFISH is a tool for fast, memory-efficient counting of k-mers in DNA.")
```
ewels: For historical reasons, the first word of `info` is the `name` from above. So this reads _"jellyfish JELLYFISH is a.."_. So need to make it `info="is a tool for fast.."`.

Also maybe make the `name` sentence case? `name='Jellyfish'`

docs/modules/jellyfish.md
```diff
@@ -0,0 +1,8 @@
+---
+Name: Jellyfish
+URL: http://www.cbcb.umd.edu/software/jellyfish/
+Description: >
+    JELLYFISH is a tool for fast, memory-efficient counting of k-mers in DNA.
+---
+
+JELLYFISH is a tool for fast, memory-efficient counting of k-mers in DNA. A k-mer is a substring of length k, and counting the occurrences of all such substrings is a central step in many analyses of DNA sequence. JELLYFISH can count k-mers using an order of magnitude less memory and an order of magnitude faster than other k-mer counting packages by using an efficient encoding of a hash table and by exploiting the "compare-and-swap" CPU instruction to increase parallelism.
```
ewels: Need a description of how filenames need to be formatted here (and perhaps mention how to customise search patterns).

multiqc/modules/jellyfish/jellyfish.py
```diff
@@ -0,0 +1,96 @@
...
+            'ylab': 'Counts',
+            'xlab': 'k-mer frequency',
+            'xDecimals': False,
+            'xmin': xmin,
+            'xmax': xmax
+        }
+        
+        self.add_section(
+            anchor = 'jellyfish_kmer_plot',
+            description = description,
```
ewels: Could you split this into a shorter `description` and longer `help` text please?

docs/modules/jellyfish.md
vezzi: done

ewels:
Awesome, thanks! 👍

RvV1979:
I find it hard to identify individual libraries in the kmer plot. Would it be possible to output an interactive instead of flat plot?

ewels:
Sure - this is not specific to the module, it’s just when you have lots of samples. Use `--interactive` to force interactive plots. See https://multiqc.info/docs/#flat--interactive-plots

RvV1979:
That works. Thanks! Perhaps it is an idea to mention this in the info text associated with flat plots?

PR #481, opened 2017-06-23
Fix htseq module error when using htseq 0.8

nalcala:
Use last column of htseq-count table instead of 2nd column.
This should fix an issue that arises with htseq 0.8 when the new option "--additional-attr" is used. This new option adds a column, and thus the second column no longer contains the desired data


ewels:
Great stuff, many thanks!

PR #483, opened 2017-06-27
Add module for disambiguate.

jrderuiter:
Adds a new module for the disambiguate tool from Astrazeneca, which we use to disambiguate mouse/human reads in sequencing data from patient-derived xenografts.


multiqc/utils/config_defaults.yaml
```diff
@@ -169,6 +169,7 @@ module_order:
     - 'rseqc'
     - 'busco'
     - 'goleft_indexcov'
+    - 'disambiguate'
```
ewels: I'd class this as _'Pre-alignment QC'_ myself I think - it probably fits best alongside `fastq_screen` below, the two tools do similar tasks.

setup.py
```diff
@@ -79,6 +79,7 @@
             'htseq = multiqc.modules.htseq:MultiqcModule',
             'jellyfish = multiqc.modules.jellyfish:MultiqcModule',
             'kallisto = multiqc.modules.kallisto:MultiqcModule',
+            'disambiguate = multiqc.modules.disambiguate:MultiqcModule',
```
ewels: I usually try to keep this list alphabetically sorted (not that it actually matters).

multiqc/modules/disambiguate/disambiguate.py
```diff
@@ -0,0 +1,137 @@
...
+
+        keys = OrderedDict()
+        keys['species_a'] = {'color': '#437bb1', 'name': 'Species a'}
+        keys['species_b'] = {'color': '#b1084c', 'name': 'Species b'}
+        keys['ambiguous'] = {'color': '#333333', 'name': 'Ambiguous'}
+
+        self.add_section(
+            name='Percent ambiguous',
+            anchor='Percent ambiguous',
+            description='Disambiguation metrics, in number of reads.',
```
ewels: Perhaps instead of this, supply a `config` dict to the `bargraph.plot` function below with an axis label and title (see [docs](http://multiqc.info/docs/#bar-graphs)). For example:

```python
pconfig = {
  'id': "disambiguate_alignments",
  'title': "Disambiguate Alignments",
  'cpswitch_counts_label': "# Reads",
  'ylab': "# Reads"
}
plot=bargraph.plot(self.data, keys, pconfig)
```

multiqc/modules/disambiguate/disambiguate.py
```diff
@@ -0,0 +1,137 @@
...
+            data[sample] = {
+                'species_a': int(split[1]),
+                'species_b': int(split[2]),
+                'ambiguous': int(split[3])
+            }
+
+        return data
+
+
+    def add_stats_table(self):
```
ewels: I try to keep the number of columns in the General Stats to a minimum. As the information here seems to be the same as that in the bar graph, perhaps we could have just `% Species a` instead?

multiqc/modules/disambiguate/disambiguate.py
```diff
@@ -0,0 +1,137 @@
...
+    def add_stats_plot(self):
+        """Plots alignment stats as bargraph."""
+
+        keys = OrderedDict()
+        keys['species_a'] = {'color': '#437bb1', 'name': 'Species a'}
+        keys['species_b'] = {'color': '#b1084c', 'name': 'Species b'}
+        keys['ambiguous'] = {'color': '#333333', 'name': 'Ambiguous'}
+
+        self.add_section(
+            name='Percent ambiguous',
```
ewels: Probably no need for `name` or `description` here as the module has only a single section.

multiqc/utils/config_defaults.yaml
jrderuiter: What is your reasoning for doing so?

Disambiguate works on two aligned bam files, and produces another filtered bam file. (It doesn't work with the fastq files.) The stats are for the filtered bam file, which is why I put it in post.

multiqc/modules/disambiguate/disambiguate.py
jrderuiter: I agree!

jrderuiter:
Thanks for the feedback! This should address most of your comments. I'm not sure about the category though, I would prefer post-alignment QC as the tool operates on alignments (not on fastqs).

multiqc/utils/config_defaults.yaml
ewels: Ah, apologies - I assumed that the tool ran the alignments as well on the FastQ files. This is fine where it is then..

ewels:
Brilliant stuff, thanks!

Phil

PR #486, opened 2017-07-07
fix fastq_screen module

daler:
Hi @ewels -

1) The `fastq_screen` module was not sorting the organism dict keys, resulting in random order of the bars in the non-simple-plot version.
2) While troubleshooting that, I set `fastqscreen_simpleplot: true` in the config as indicated in the docs, but it wasn't forcing the simple plot.

Luckily the fixes are super easy, both fixed in this PR.

Thanks for such a great tool!

### DIFF
--- a/multiqc/modules/fastq_screen/fastq_screen.py
+++ b/multiqc/modules/fastq_screen/fastq_screen.py
@@ -139,7 +139,7 @@ class MultiqcModule(BaseMultiqcModule):
                 thisdata = list()
                 if len(categories) > 0:
                     getCats = False
-                for org in self.fq_screen_data[s]:
+                for org in sorted(self.fq_screen_data[s]):
                     if org == 'total_reads':
                         continue
                     try:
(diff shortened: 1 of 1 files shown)


ewels:
Great stuff, thanks for the fixes @daler!

Phil

ewels:
Hi @daler,

I just took another look at this and I'm confused about the second change relating to `fastqscreen_simpleplot`.

As I had it before, the following logic worked:

* Fancy bar plot unless:
    * lots of samples / genomes
    * using static plots
    * `config.fastqscreen_simpleplot` set to `True`
* Simple bar plot otherwise

But by changing the default value of `config.fastqscreen_simpleplot` to `True` when not set (majority of users), the module now always uses a simple bar plot unless people specifically opt in to having the fancy plot by setting `config.fastqscreen_simpleplot = False`.

I tried reverting your change and it seemed to work as expected for me. Am I missing something here?

Phil

daler:
@ewels sorry about this. Looking back at my tests I had left the `fastqscreen_simpleplot: False` in my config file when troubleshooting which made it appear OK. I didn't version control my tests, but I wonder if I had a typo when testing the `fastqscreen_simpleplot: True` entry, and the change I made to the code would have fixed the behavior under that typo condition.

Anyway, yes, please revert that commit and I'm sorry for introducing a new bug! 

(I believe the sort order in this PR is correct though)

ewels:
Yes, the sort order is good! And my fault for not properly checking before merging ;)

Phil

Issue #487, opened 2017-07-10
N75 definition and parsing QUAST report


cyklee:
Hi Phil,

Just reporting two very minor bugs.

**Description of bug:**
1. Mouseover N75 definition in QUAST output is currently `QUAST: N75 in the contig length such that using longer or equal length contigs produces half (50%) of the bases of the assembly (kilo base pairs)`. This should be 75%. (modules/quast/quast.py line 147)

2. In Linux environment, currently parsing the output directory of metagenomic QUAST (not sure about single genome mode) will result in an error message. I suspect it is because QUAST's unusual file names in the summary directory:
```
For example:
#_N's_per_100_kbp.tsv
#_operons.tsv
#_predicted_genes_(unique).tsv
Total_aligned_length.tsv
Total_length_(>=_10000_bp).tsv
```
I could mitigate this problem by pointing multiQC directly to `report.tsv` in `combined_reference` (```multiqc -d `fzf -m` ``` works really well.) This also saves a lot of time as some of the files in QUAST results are pretty big to scan through.

* **File that triggers the error:**
    * I can place the files in Dropbox if you need them.
* **Operating System:**
    * Debian Jesse in VM
* **Python Version:**
    * 3.6.1
* **Method of MultiQC installation:**
    * pip, multiqc version 1.1.dev0

Thanks and keep up the great work!

ewels:
Thanks for the bug report, I'll take a look into it.. Regarding the Linux bug - what was the error message that you get? Also, are there any fixed filenames that we could skip from the QUAST output to speed up execution? We can add these to the default list of [filenames to ignore](https://github.com/ewels/MultiQC/blob/master/multiqc/utils/config_defaults.yaml#L148-L161)..

Phil

ewels:
Part 1) fixed..

ewels:
Maybe @vladsaveliev may have some input on this issue?

vladsaveliev:
Hi guys,

`metaquast.py`'s output is different from the regular `quast.py`, and I believe that the multiqc module just wasn't designed and tested for meta. I just ran a quick test and had an error `KeyError: '# contigs (>= 0 bp)'`, which doesn't seem to be related to the file names (though I agree that having hashes and parenthesis in file names is a stupid idea - we will change that in quast too!). 

Ok, so, I'll see if I can quickly fix the module to make it work.

Regarding the files to skip, there are long outputs from alignment and large misassembly reports, which could be skipped. I'll take care of that too.

[...]

PR #491, opened 2017-07-10
Quast: processs metaquast runs properly

vladsaveliev:
Fixing https://github.com/ewels/MultiQC/issues/487

In order to support outputs from metagenomic quast runs properly, we first need to determine if we are faced with a regular or a metaquast run. The basic directory structures for those are as follows:

Regular quast run (single genome), in MultiQC we want to parse `quast.tsv`:
```
quast.log
report.tsv
```
Metaquast run (multiple reference genomes), in MultiQC we want to parse `combined_reference/report.tsv`:
```
combined_reference/
   quast.log
   report.tsv
metaquast.log
runs_per_reference/
   reference1/
      quast.log
      report.tsv
   reference2/
      quast.log
      report.tsv
   ...
not_aligned/
   quast.log
   report.tsv
```

So the trick is to make MultiQC avoid treating all those per-reference subdirectories as full quast inputs. The only way I figured is to target `metaquast.log`, and then point to `dirname(join('combined_reference', 'report.tsv')` avoiding the standard file search mechanism. Only if `metaquast.log` is not found, go and search for all occurences of 'report.tsv' and process as a regular quast. I'm not sure how to distinguish between `report.tsv` and `combined_reference/report.tsv` using `find_log_files` only.

I'm also adding test data into https://github.com/ewels/MultiQC_TestData

### DIFF
--- a/multiqc/modules/quast/quast.py
+++ b/multiqc/modules/quast/quast.py
@@ -239,30 +239,34 @@ class MultiqcModule(BaseMultiqcModule):
         """ Make a bar plot showing the number and length of contigs for each assembly """
 
         # Prep the data
+        import re
         data = dict()
+        categories = []
         for s_name, d in self.quast_data.items():
-            try:
-                p = dict()
-                p['>= 50000 bp'] = d['# contigs (>= 50000 bp)']
-                p['25000-50000 bp'] = d['# contigs (>= 25000 bp)'] - d['# contigs (>= 50000 bp)']
-                p['10000-25000 bp'] = d['# contigs (>= 10000 bp)'] - d['# contigs (>= 25000 bp)']
-                p['5000-10000 bp'] = d['# contigs (>= 5000 bp)'] - d['# contigs (>= 10000 bp)']
-                p['1000-5000 bp'] = d['# contigs (>= 1000 bp)'] - d['# contigs (>= 5000 bp)']
-                p['0-1000 bp'] = d['# contigs (>= 0 bp)'] - d['# contigs (>= 1000 bp)']
-                assert sum(p.values()) == d['# contigs (>= 0 bp)']
-                data[s_name] = p
-            except AssertionError:
-                log.warning("Contig counts didn't add up properly for {}".format(s_name))
+            nums_by_t = dict()
+            for k, v in d.items():
+                m = re.match('# contigs \(>= (\d+) bp\)', k)
+                if m:
+                    nums_by_t[int(m.groups()[0])] = v
 
-        # Define the order of the categories, small to big
-        cats = [
-            '0-1000 bp',
-            '1000-5000 bp',
-            '5000-10000 bp',
-            '10000-25000 bp',
-            '25000-50000 bp',
-            '>= 50000 bp',
-        ]
+            tresholds = sorted(nums_by_t.keys(), reverse=True)
+            p = dict()
+            cats = []
+            for i, t in enumerate(tresholds):
+                if i == 0:
+                    c = '>= ' + str(t) + ' bp'
+                    cats.append(c)
+                    p[c] = nums_by_t[t]
+                else:
+                    c = str(t) + '-' + str(tresholds[i - 1]) + ' bp'
+                    cats.append(c)
+                    p[c] = nums_by_t[t] - nums_by_t[tresholds[i - 1]]
+            if not categories:
+                categories = cats
+            elif set(cats) != set(categories):
+                log.warning("Different contig threshold categories for samples, skip plotting barplot".format(s_name))
+                continue
+            data[s_name] = p
 
         pconfig = {
             'id': 'quast_num_contigs',
@@ -271,7 +275,7 @@ class MultiqcModule(BaseMultiqcModule):
             'yDecimals': False
         }
 
-        return bargraph.plot(data, cats, pconfig)
+        return bargraph.plot(data, categories, pconfig)
 
     def quast_predicted_genes_barplot(self):
         """
--- a/multiqc/modules/quast/quast.py
+++ b/multiqc/modules/quast/quast.py
@@ -239,7 +239,6 @@ class MultiqcModule(BaseMultiqcModule):
         """ Make a bar plot showing the number and length of contigs for each assembly """
 
         # Prep the data
-        import re
         data = dict()
         categories = []
         for s_name, d in self.quast_data.items():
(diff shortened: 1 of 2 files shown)


vladsaveliev:
Phil, regarding file skipping. There are many files to skip in a quast output folder, with patterns like:
`contigs_report_*.stderr`
`contigs_report_*.stdout`
`nucmer_output/*`
`*.html`
`*.pdf`
etc.
However it would be the best to just avoid parsing all subdirectories once we detected the basic report files `report.tsv` or `metaquast.log`. Do you think there is a nice way to do that?

ewels:
Hi @vladsaveliev,

If I understand the problem correctly, is there any reason why you can't just find all Quast outputs as currently, but reject them if they contain the directory `runs_per_reference` in `f['root']`? eg:

```python
for f in self.find_log_files('quast'):
    if 'runs_per_reference' not in f['root'].split(os.path.sep):
        self.parse_quast_log(f)
```

In fact, you could just add `runs_per_reference` to `config.fn_ignore_dirs` and then those files will never make it to the `find_log_files` function. This then also solves the problem of crawling through lots of files.

Side note: if normal runs have `quast.tsv`, how did the pipeline find them before? As it's only looking for `report.tsv`?

Phil

setup.py
```diff
@@ -52,7 +52,6 @@
         'future>0.14.0',
         'lzstring',
         'jinja2>=2.9',
-        'matplotlib',
```
ewels: Any reason why you removed this? :)

vladsaveliev: Oh, my bad. I had this removed in my fork because of some issues installing matplotlib. Sorry about that Phil, I'll put it back.

vladsaveliev:
Phil, sorry for misleading - the "normal" runs had `report.tsv`, not `quast.tsv`.

vladsaveliev:
Alright, `config.fn_ignore_dirs` sounds like what is needed. Will use that!

vladsaveliev:
What's the correct way to append to `config.fn_ignore_dirs` in module level? Can't find any example of config manipulation in other modules. Is just appending to the list in quast.py before calling `self.find_log_files('quast')`, and restoring it right after, will work?

ewels:
There isn't any official way to do it at module level. As it's a core module I'd just add to [`config_defaults.yaml`](https://github.com/ewels/MultiQC/blob/45642d6f93a86774ad11acee7420e82b1fc84fa3/multiqc/utils/config_defaults.yaml#L51-L52). It's a specific enough directory name that I can't see it negatively impacting other modules. And we want it to affect the main file search loop, as that will mean that all the other modules don't have to waste time searching through those files.

If you're interested, [here is an example](https://github.com/ewels/MultiQC_NGI/blob/master/multiqc_ngi/multiqc_ngi.py#L26-L67) of how I've modified the config defaults from an external MultiQC module in the past.

Phil

vladsaveliev:
Also thinking about the case when use explicitly wants to point MultiQC to a subdirectory of a metaquast run, e.g. `runs_per_reference/reference1`. In this case we should be able to process it as a regular quast output. Or even worse, user manually names his regular quast directory `runs_per_reference` or `not_aligned`.

ewels:
In that case, the user can just tell MultiQC to search the subdirectories. For example `multiqc data/runs_per_reference/*`. I think that should still work... (not certain).

The second case can happen, but the same can be said for a lot of other MultiQC settings. But sensible defaults should win over edge cases where possible. A lot can be mitigated by just writing about this in the documentation so that when stuff doesn't work people can find out why.

Phil

vladsaveliev:
Yeah, I studied MultiQC_NGI project well and used the ideas in bcbio-nextgen and AZ plugins. In AZ, I even put the settings into yaml and reading them in a `config_loaded` hook, and use `config.update_dict` to override: https://github.com/vladsaveliev/MultiQC_az/blob/master/multiqc_az/multiqc_az.py#L28, but that's clearly overkill for quast. Anyway, I got the idea - will just add the directories `config_defaults.yaml`. Thanks for all clarifications!

vladsaveliev:
Phil, I think I'm done (hopefully).

multiqc/modules/quast/quast.py
```diff
@@ -239,30 +239,34 @@ def quast_contigs_barplot(self):
         """ Make a bar plot showing the number and length of contigs for each assembly """
 
         # Prep the data
+        import re
```
ewels: I think this is already being imported on line 8?

ewels:
Will merge and strip the extra line myself as it's so minor..

vladsaveliev:
I just done that too... Hope it will not cause merging clashes :)

ewels:
Ah great, no problem.

Just testing and I ([and travis](https://travis-ci.org/ewels/MultiQC/jobs/252517780#L662-L669)) get warnings about `Not all predicted gene thresholds available for "samplename"`.  In the report I then get the error `Error - was not able to plot data.` with no plot.

Any ideas what's going on here?

Phil

ewels:
I think the plot is failing because it's being given a bunch of empty sample dictionaries to plot. Adding a `if len(p) > 0:` check to avoid these gives a plot in the report. However, it only has a single sample (`contigs`).

vladsaveliev:
There are several test data folders for quast, and MultiQC is running on all of them at once. Some of them were run with gene prediction turned on, some of them with it turned off. The module sees that and throws a warning, because within a single run we expect to have consistent settings.

ewels:
Ok, so it's expected that these datasets will have missing data for this plot? In that case then just that additional check should be fine. But it's needed to avoid breaking the plot.

In fairness, I should probably make the core code able to skip past empty samples. I'll look into that.

Phil

ewels:
Ah, now it's working. Maybe the core code does already handle empty samples then and it was something I did. Apologies, ignore me..

Phil

vladsaveliev:
Ah, so the core works normally with empty samples? Ok!

Thanks for merging!

Also I was thinking that mentioning metaquast in docs won't hurt. Ugh, always forget about the docs. Something like that:
```
For MetaQUAST runs (`metaquast.py`), `combined_reference/report.tsv` 
is parsed, and folders `runs_per_reference` and `not_aligned` are ignored. 
If you want to run MultiQC against auxiliary MetaQUAST runs, point to 
the report.tsv explicitly, e.g.:
```
multiqc runs_per_reference/reference_1/report.tsv
```
I'll make a new PR?

ewels:
Ah good thinking, no worries - I'll add that for you now quickly..

Phil

ewels:
Done: http://multiqc.info/docs/#quast

PR #492, opened 2017-07-13
change sample name regexp in MarkDuplicates.py

jyh1:
In the original regexp `INPUT=\[?([^\\s]+)\]?`, the last character `\]?` will never be matched due to the greedy operator `+` in `[^\\s]+`. So we will get sample name with trailing `]` character, which is inconsistent with name from other tools like FastQC.

### DIFF
--- a/multiqc/modules/picard/MarkDuplicates.py
+++ b/multiqc/modules/picard/MarkDuplicates.py
@@ -28,9 +28,9 @@ def parse_reports(self):
                 s_name = None
 
                 # Pull sample name from input
-                fn_search = re.search("INPUT=\[?([^\\s]+)\]?", l)
+                fn_search = re.search((r"INPUT=(\[([^\s]+)\]|([^\s]+))", l)
                 if fn_search:
-                    s_name = os.path.basename(fn_search.group(1))
+                    s_name = os.path.basename(fn_search.group(2) or fn_search.group(3))
                     s_name = self.clean_s_name(s_name, f['root'])
 
             if s_name is not None:


ewels:
Hi @jyh1,

Thanks for this, and good spot! You're right that the regex was broken - the double backslash must have been a typo as it was only searching until a literal backslash or a literal `s`. A simpler fix is probably just to avoid any closing brackets in the match though: `INPUT=\[?([^\]\s]+)\]?`. I suspect that this is what I originally intended (`\` is next to `]` on my keyboard).

I'll submit this fix for MarkDuplicates and also all of the other Picard submodules 👍 

Phil

jyh1:
Thanks for your reply @ewels.

It seems that python doesn't recognize `\s` as an escaped character, so it is interpreted literally. `'\s'=='\\s'` actually returns `True`. Your original regex is perfectly correct in that point, there is no difference between `'[^\s]'` and `'[^\\s]'` in python. But It might become confusing. I prefer to use `r` to disable escaping when writing regex, but it's just a personal choice :).

The regex contains `\[?`, I assumed you intend to deal with two circumstances: `INPUT=[sampleName]` and `INPUT=sampleName`. Your fix will work if we assume there is no `]` in the `sampleName`. It will break when encountering sample names like `INPUT=[Galaxy22-[sample].bam]`. Although it is rare to have `]` in file names, if your data is downloaded from `Galaxy`, it always have a pair of it=_=. It might be better to explicitly express that two circumstances in regex to avoid this.

ewels:
Hah, of course. I should be used to edge cases by now I guess.. I get nervous with multiple regex match groups as they seem to often break. But we can simply catch it all with or without square brackets and trim those.

And yes, I usually define regex strings with the `r` modified, didn't notice that I'd missed that this time. New fix pushed - is this one ok?

Phil

jyh1:
Ah, that's the right fix! Thanks for your attention and this great software!

ewels:
Great! No problem and thanks for the bug report / fix 👍

PR #506, opened 2017-07-20
Column placement

tbooth:
Work from Codefest!


ewels:
Really elegant solution - thanks for this!

PR #510, opened 2017-07-20
Bugfix: Allow loading of multiple configs

DennisSchwartz:
This PR fixes a bug where you could only load one config one time. Loading a second config after having loaded the first would result in the page crashing and the highlights being added again each time "Load" was pressed.

To reproduce with current code:
- Add highlighting
- Go to config tab
- Save config
- Press load
--> Back in the highlights tab, the highlights have been duplicated and there is an error in the console.

The `hashFunction` is necessary to be able to identify the Highlights to check if they exist when loading a new config.


ewels:
Really nice stuff, thanks for looking into this 👍

PR #514, opened 2017-07-21
Group modules into tags

Hammarn:
Work from Codefest 2017 in Prague!
This allows for a user to only run a certain set of modules within a certain application, eg. RNA. The thought behind this is  a step towards futureproofing MultiQ. As the list of modules expands using all of the might become an issue for some user/systems. 
usage:
--tag RNA 
Will limits the modules that MultiQC use to:
```
rna_seqc, fastqc, star, kallisto, samtools, clusterflow, picard, rseqc, qualimap, sortmerna, featureCounts, qorts, adapterRemoval, flexbar, disambiguate, salmon, fastq_screen, htseq, preseq, cutadapt, bamtools, bowtie1, trimmomatic, bowtie2, homer, leehom, biobloomtools, hisat2, skewer, tophat, samblaster

```
All the tags are a bit arbitrary by me and they should need some revising.


ewels:
Awesome work!

Issue #411, opened 2017-03-01
Module request: vcftools


ewels:
Requested in #311 by @jcgrenier:

> One could think of some more high level QC. One that could use the output of relatedness, or relateness2 command from the newest version of vcftools that could be used to see if two RNASeq samples of different condition are the same individual or if they are from the same family. A simple triangular heatmap could do the job.

[http://vcftools.sourceforge.net/man_latest.html#OUTPUT OPTIONS](http://vcftools.sourceforge.net/man_latest.html#OUTPUT%20OPTIONS)

ewels:
Missed a bit:
> A standard heatmap could be enough to plot the pairwise values coming from the vcftools analysis.

jcgrenier:
Hi @ewels,

Thanks for coming back on this. I'm putting an example for each one of the programs actually. So one example which was done using VCFtools with the option relatedness2. The classic relatedness option is doing file of the same format.
[YP.QCheck.VQ30.GQ20.DP10.miss50.relatedness.relatedness2.txt](https://github.com/ewels/MultiQC/files/810747/YP.QCheck.VQ30.GQ20.DP10.miss50.relatedness.relatedness2.txt)

The second example is resulting from the --genome option from PLINK 1.9 (or 1.08p).
The column that interests us is the 10th one, which represents the PIHAT value, which is a combination of multiple other metrics. 

The last column could be used as well as a checkup for the samples (barplot). The ratio is supposed to be at 2. So if it's larger, the two samples are most probably related. The status you will see in the RT column will be super correlated in my example, since there's pedigree information in the hapmap dataset. Most of the time, if you are building your own, you won't have any information there. Furtheremore, the numbers look clean here since it's from a genotyping array. When you compare sequencing samples, the numbers will vary much more because of the missing data.

https://www.cog-genomics.org/plink2/ibd

[hapmap3_r2_b36_fwd.consensus.qc.poly.PIHAT.genome.txt.gz](https://github.com/ewels/MultiQC/files/810916/hapmap3_r2_b36_fwd.consensus.qc.poly.PIHAT.genome.txt.gz)


Note that I've added a ".txt" extension to the files so that I could put them in the post.
So normally, you should look for a ".relatedness2" for vcftools and a "genome" for PLINK.
Another thing that could be interesting is to only plot some barplots representing samples admixture. This value could be a QC to be sure that your sample is really coming from the good population you are studying or a result in itself. Having it plotted in the MultiQC could be pretty cool too!

[...]

PR #517, opened 2017-07-21
VCFTools relatedness2

mwhamgenomics:
Adds a module for displaying the outputs of `vcftools relatedness2` -  see #411.


multiqc/modules/vcftools/relatedness.py
```diff
@@ -0,0 +1,48 @@
+import csv
```
ewels: My standard copy and paste header looks like this:
```python
#!/usr/bin/env python

""" MultiQC module to parse output from Cutadapt """

from __future__ import print_function
import logging

from multiqc import config
from multiqc.plots import linegraph
from multiqc.modules.base_module import BaseMultiqcModule

# Initialise the logger
```

multiqc/modules/vcftools/vcftools.py
```diff
@@ -0,0 +1,14 @@
...
+class MultiqcModule(BaseMultiqcModule, Relatedness2Mixin):
+    def __init__(self):
+        super(MultiqcModule, self).__init__(
+            name='VCFTools',
+            anchor='vcftools',
+            href='https://vcftools.github.io',
+            info='is a program for working with and reporting on VCF files'
+        )
+
+        self.parse_relatedness2()
```
ewels: It's good to get this function to return the number of samples parsed (either input files or samples - up to you which). So the main thing is that it's 0 if nothing was found. Then, we can run this afterwards:

```python
# Exit if we didn't find anything
if sum(n.values()) == 0:
        log.debug("Could not find any reports in {}".format(config.analysis_dir))
        raise UserWarning
```

multiqc/modules/vcftools/relatedness.py
```diff
@@ -0,0 +1,48 @@
...
+        self.x_labels = set()
+        self.y_labels = set()
+
+        self.parse(relatedness_file['f'])
+
+    def parse(self, f):
+        rels = defaultdict(dict)
+        r = csv.DictReader(f, delimiter='\t')
+        for line in r:
+            self.x_labels.add(line['INDV1'])
```
ewels: Instead of using `self.ignore_samples` I think you need to use the internal logic of that function yourself here:

```python
# Match ignore glob patterns
glob_match = any( fnmatch.fnmatch(k, sn) for sn in config.sample_names_ignore )
re_match = any( re.match(sn, k) for sn in config.sample_names_ignore_re )
if not glob_match and not re_match:
    # continue doing processing
```

ewels: (hacky and nasty but I can't think of an easier way to get the same functionality otherwise)

multiqc/modules/vcftools/relatedness.py
```diff
@@ -0,0 +1,48 @@
...
+from multiqc.plots import heatmap
+
+
+class Relatedness2Mixin():
+    def parse_relatedness2(self):
+        for f in self.find_log_files('vcftools/relatedness2', filehandles=True):
+            m = _RelatednessMatrix(f)
+
+            self.add_section(
+                'Vcftools relatedness2 ' + m.name,
```
ewels: Please used named arguments here if you can, in case I change stuff in the future...

multiqc/modules/vcftools/relatedness.py
```diff
@@ -0,0 +1,48 @@
...
+                'vcftools_relatedness2_' + m.name,
+                plot=heatmap.plot(
```
ewels: Would be great to have a `description` and `help` text here if possible..

multiqc/modules/vcftools/relatedness.py
```diff
@@ -0,0 +1,48 @@
+import csv
+from collections import defaultdict
+from multiqc.plots import heatmap
+
+
+class Relatedness2Mixin():
+    def parse_relatedness2(self):
+        for f in self.find_log_files('vcftools/relatedness2', filehandles=True):
+            m = _RelatednessMatrix(f)
+
```
ewels: Need to add a check here to not add a section if the parsing went wrong..

ewels:
Hi @mwhamgenomics! You've got a minor merge conflict sorry, hopefully easy to fix. It's due to the config file where we now have module tags / keywords.

Question - is this output from VCFtools `relatedness` or `relatedness2`? I suspect that it's the latter given the filename. So we need to change the name throughout sorry.

Remaining tasks that I can see on the documentation side:

1. We need the `docs/modules/vcftools.md` file
    * See `gatk.md` for an example of how to describe modules with submodules).
2. Also a mention in the top YAML frontmatter in `docs/README.md`
    * The website uses this to build the modal on the homepage listing supported tools
3. Addition to the table in the main root `README.md`
    * Lots of copying and pasting sorry..!
4. Credit yourself in the `CHANGELOG.md` file 😉 

Once we have this stuff sorted I'll do some proper testing myself, but reading the code it all looks good at this point 👍 

Phil

mwhamgenomics:
Hi there,

It is indeed relatedness2, so will rename that as needed. Will pull upstream, fix conflicts and push my updates. I'm aware from the [docs](https://vcftools.github.io) that there's still _a lot_ more functionality that can be added here, so other contributors would be welcome.

Thanks,
M.

ewels:
Great work - many thanks @mwhamgenomics! I made a few minor changes (see above commit) and fixed my broken fix with the help button. Looks brilliant!

Good idea to add the to-do list to the docs page, I like it. Could also be good to add a new issue to GitHub with those suggestions too, so that they can be discussed there.

Phil

PR #520, opened 2017-07-21
Clean filenames by keeping a regex match

robinandeer:
Adds an option "regex_keep" to keep a match in a filename and discard the rest

(Comes with a PR on the test data repo)


ewels:
Super nice, thanks! 👍

PR #526, opened 2017-07-21
Allow setting a default config which will persist across page reloads

DennisSchwartz:
This PR adds two buttons to the config tab in the toolbox.
They allow to set a config as default. This will label the config with '[default]' and load it each time the page is opened.
It also allows to unset the current default config.

It uses the localstorage to save the config which means it will be lost when cookies are cleared.


ewels:
Works an absolute treat, thanks! 👍

DennisSchwartz:
Awesome, thanks :)

ewels:
Just added some minor stuff in 7d624a112f4baccb2c7ba97f0d3d1fca6ac80a24 - main thing was that after a config is auto-loaded, it's now shown as the selected value in the dropdown box as I found it kind of confusing otherwise.

PR #527, opened 2017-07-21
Bcl2fastq

### LINKED ISSUE #409: Module Request: Illumina Demultiplexing stats
Hi Phil,

I find it increasingly useful to have the entire quality control in one single report I can easily browse through and share with my colleagues. One of the few pieces I am now missing is the demultiplexing step, which converts data in BCL format to FASTQ in Illumina machines.

Following the requirements checklist:

- Description of the tool: bcl2fastq can be used to both demultiplex data and convert BCL files to FASTQ file formats for downstream analysis. There are two versions of this software: [bcl2fastq](https://support.illumina.com/downloads/bcl2fastq_conversion_software_184.html) for MiSeq and HiSeq sequencing systems running RTA versions earlier than `1.8`, and [bcl2fastq2](https://support.illumina.com/downloads/bcl2fastq-conversion-software-v2-18.html) for Illumina sequencing systems running RTA version `1.18.54` and above.
- Log files: the file I find more useful is `Demultiplex_Stats.htm`, which contains a summary of the number of reads and their quality score in every sequenced sample. There are other files that can be informative, though, like `BustardSummary.xml`.
- Naming patterns: as I said above, the file is usually called `Demultiplex_Stats.htm`.
- Most important value: probably the number of reads.
- What should be plotted: the number of reads could be easily represented as a stacked barplot, together with the percentage of those that are perfect index reads. It would be also nice to have their mean quality... perhaps as a red line overlaid on top of the barplot?

I hope it suffices. Thanks a lot! :)

iimog:
This adds support for the `Stats.json` files created by recent versions of `bcl2fastq`. This partially resolves #409 as the old `.htm` format is not yet supported. In order to avoid name conflicts the 'RunId' is prepended to both lane numbers and sample names. To not aggregate counts from files with identical information collisions of 'RunId' and 'Lane' lead to the existing information for this combination being overwritten (with a debug warning). Currently only total reads, perfect index reads and unclassified reads are reported (once per sample and once per lane). This can be extended to e.g. also include YieldQ30.


multiqc/modules/bcl2fastq/bcl2fastq.py
```diff
@@ -0,0 +1,124 @@
...
+from multiqc.plots import bargraph
+
+log = logging.getLogger(__name__)
+
+class MultiqcModule(BaseMultiqcModule):
+    def __init__(self):
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__(name='bcl2fastq', anchor='bcl2fastq',
+        href="https://support.illumina.com/downloads/bcl2fastq-conversion-software-v2-18.html",
+        info="bcl2fastq can be used to both demultiplex data and convert BCL files to FASTQ file formats for downstream analysis.")
```
ewels: The `info` block always starts with the name of the module and the hyperlink. So you need to omit it at the start here as otherwise you'll end up duplicate names like this:

> [bcl2fastq](https://support.illumina.com/downloads/bcl2fastq-conversion-software-v2-18.html) bcl2fastq can be used to both demultiplex...

multiqc/modules/bcl2fastq/bcl2fastq.py
```diff
@@ -0,0 +1,124 @@
...
+        self.write_data_file({str(k): self.bcl2fastq_bylane[k] for k in self.bcl2fastq_bylane.keys()}, 'multiqc_bcl2fastq_bylane')
+        self.write_data_file(self.bcl2fastq_bysample, 'multiqc_bcl2fastq_bysample')
+
+        # Add section for counts by lane
+        self.add_section (
+            name = 'bcl2fastq by lane',
+            anchor = 'bcl2fastq-bylane',
+            description = 'Number of reads per lane (with number of perfect index reads)',
+            helptext = "Perfect index reads are those that do not have a single mismatch. All samples of a lane are combinned. Undetermined reads are treated as a third category. To avoid conflicts the runId is prepended.",
+            plot = bargraph.plot({key: {"imperfect": self.bcl2fastq_bylane[key]["total"]-self.bcl2fastq_bylane[key]["perfectIndex"], "perfect": self.bcl2fastq_bylane[key]["perfectIndex"], "undetermined": self.bcl2fastq_bylane[key]["undetermined"]} for key in self.bcl2fastq_bylane.keys()})
```
ewels: Whilst space efficient, this is pretty difficult to read 😉  Would be good if you could break this out as a multi-line variable above.

multiqc/modules/bcl2fastq/bcl2fastq.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+
+        # Add section for counts by sample
+        self.add_section (
+            name = 'bcl2fastq by sample',
+            anchor = 'bcl2fastq-bysample',
+            description = 'Number of reads per sample (with number of perfect index reads)',
+            helptext = "Perfect index reads are those that do not have a single mismatch. All samples are aggregated across lanes combinned. Undetermined reads are ignored. Undetermined reads are treated as a separate sample. To avoid conflicts the runId is prepended.",
+            plot = bargraph.plot({key: {"imperfect": self.bcl2fastq_bysample[key]["total"]-self.bcl2fastq_bysample[key]["perfectIndex"], "perfect": self.bcl2fastq_bysample[key]["perfectIndex"]} for key in self.bcl2fastq_bysample.keys()})
```
ewels: As above.

multiqc/modules/bcl2fastq/bcl2fastq.py
```diff
@@ -0,0 +1,124 @@
...
+                            self.source_files[uniqSampleName] = []
+                        self.source_files[uniqSampleName].append(self.bcl2fastq_data[runId][lane]["samples"][sample]["filename"])
+
+    def add_general_stats(self):
+        data = {key: {"total": self.bcl2fastq_bysample[key]["total"], "perfectPercent": '{0:.1f}'.format(100*self.bcl2fastq_bysample[key]["perfectIndex"]/self.bcl2fastq_bysample[key]["total"])} for key in self.bcl2fastq_bysample.keys()}
+        headers = OrderedDict()
+        headers['total'] = {
+            'title': 'Total Reads',
+            'description': 'Total number of reads for this sample as determined by bcl2fastq demultiplexing',
+            'scale': 'Blues'
```
ewels: As this is a read count, please use the special-case formatting which allows users to configure whether to show millions / thousands / hundreds of reads etc:

```python
            'title': '{} Total Reads'.format(config.read_count_prefix),
            'description': 'Total number of reads for this sample as determined by bcl2fastq demultiplexing ({})'.format(config.read_count_desc),
            'min': 0,
            'scale': 'Blues',
            'shared_key': 'read_count'
```

multiqc/modules/bcl2fastq/bcl2fastq.py
iimog: As this contained also a lot of redundancy with the code below I extracted a method.

iimog:
Thanks a lot for the comments. I resolved the conflicts and made the requested changes.

multiqc/modules/bcl2fastq/bcl2fastq.py
```diff
@@ -0,0 +1,138 @@
...
+            self.add_data_source(s_name=s, source=",".join(list(set(self.source_files[s]))), module='bcl2fastq', section='bcl2fastq-bysample')
+
+        # Add sample counts to general stats table
+        self.add_general_stats()
+        self.write_data_file({str(k): self.bcl2fastq_bylane[k] for k in self.bcl2fastq_bylane.keys()}, 'multiqc_bcl2fastq_bylane')
+        self.write_data_file(self.bcl2fastq_bysample, 'multiqc_bcl2fastq_bysample')
+
+        # Add section for counts by lane
+        self.add_section (
+            name = 'bcl2fastq by lane',
```
ewels: I think `Reads by lane` may be better, as this sits just under the main `bcl2fastq` header.

ewels: Though perhaps `Clusters` is better to be clearer in the case of PE data?

multiqc/modules/bcl2fastq/bcl2fastq.py
```diff
@@ -0,0 +1,138 @@
...
+            anchor = 'bcl2fastq-bylane',
+            description = 'Number of reads per lane (with number of perfect index reads)',
+            helptext = "Perfect index reads are those that do not have a single mismatch. All samples of a lane are combinned. Undetermined reads are treated as a third category. To avoid conflicts the runId is prepended.",
```
ewels: `combinned` -> `combined`
`runId` -> `run ID`

multiqc/modules/bcl2fastq/bcl2fastq.py
```diff
@@ -0,0 +1,138 @@
...
+        )
+
+    def parse_file_as_json(self, myfile):
+        content = json.loads(myfile["f"])
+        runId = content["RunId"]
+        if not runId in self.bcl2fastq_data:
+            self.bcl2fastq_data[runId] = dict()
+        run_data = self.bcl2fastq_data[runId]
+        for conversionResult in content["ConversionResults"]:
+            lane = conversionResult["LaneNumber"]
```
ewels: I wonder if it's worth adding `Lane ` or just `L` here instead of just giving a number, for clarity?

multiqc/modules/bcl2fastq/bcl2fastq.py
```diff
@@ -0,0 +1,138 @@
...
+            name = 'bcl2fastq by lane',
+            anchor = 'bcl2fastq-bylane',
+            description = 'Number of reads per lane (with number of perfect index reads)',
+            helptext = "Perfect index reads are those that do not have a single mismatch. All samples of a lane are combinned. Undetermined reads are treated as a third category. To avoid conflicts the runId is prepended.",
+            plot = bargraph.plot(self.get_bar_data_from_counts(self.bcl2fastq_bylane))
+        )
+
+        # Add section for counts by sample
+        self.add_section (
+            name = 'bcl2fastq by sample',
```
ewels: As above.

multiqc/modules/bcl2fastq/bcl2fastq.py
```diff
@@ -0,0 +1,138 @@
...
+        self.write_data_file({str(k): self.bcl2fastq_bylane[k] for k in self.bcl2fastq_bylane.keys()}, 'multiqc_bcl2fastq_bylane')
+        self.write_data_file(self.bcl2fastq_bysample, 'multiqc_bcl2fastq_bysample')
+
+        # Add section for counts by lane
+        self.add_section (
+            name = 'bcl2fastq by lane',
+            anchor = 'bcl2fastq-bylane',
+            description = 'Number of reads per lane (with number of perfect index reads)',
+            helptext = "Perfect index reads are those that do not have a single mismatch. All samples of a lane are combinned. Undetermined reads are treated as a third category. To avoid conflicts the runId is prepended.",
+            plot = bargraph.plot(self.get_bar_data_from_counts(self.bcl2fastq_bylane))
```
ewels: Please manually specify the categories using an `OrderedDict` so that (a) the bar sections are always in the same order (with undetermined last) and (b) we can set `name` to have Sentence Case for my OCD 🤓

multiqc/modules/bcl2fastq/bcl2fastq.py
```diff
@@ -0,0 +1,138 @@
...
+        )
+
+        # Add section for counts by sample
+        self.add_section (
+            name = 'bcl2fastq by sample',
+            anchor = 'bcl2fastq-bysample',
+            description = 'Number of reads per sample (with number of perfect index reads)',
+            helptext = "Perfect index reads are those that do not have a single mismatch. All samples are aggregated across lanes combinned. Undetermined reads are ignored. Undetermined reads are treated as a separate sample. To avoid conflicts the runId is prepended.",
+            plot = bargraph.plot(self.get_bar_data_from_counts(self.bcl2fastq_bysample))
```
ewels: As above.

multiqc/modules/bcl2fastq/bcl2fastq.py
```diff
@@ -0,0 +1,138 @@
...
+        )
+
+    def parse_file_as_json(self, myfile):
+        content = json.loads(myfile["f"])
```
ewels: Maybe wrap this in a try / except so that if we come across an invalid JSON file we can just skip it instead of crashing the whole module..

multiqc/modules/bcl2fastq/bcl2fastq.py
```diff
@@ -0,0 +1,138 @@
...
+            self.bcl2fastq_data[runId] = dict()
+        run_data = self.bcl2fastq_data[runId]
+        for conversionResult in content["ConversionResults"]:
+            lane = conversionResult["LaneNumber"]
+            if lane in run_data:
+                log.debug("Duplicate runId/lane combination found! Overwriting: {}".format(self.prepend_runid(runId, lane)))
+            run_data[lane] = {"total": 0, "perfectIndex": 0, "samples": dict()}
+            for demuxResult in conversionResult["DemuxResults"]:
+                sample = demuxResult["SampleName"]
+                run_data[lane]["samples"][sample] = {"total": 0, "perfectIndex": 0, "filename": myfile['root']+"/"+myfile["fn"]}
```
ewels: Please use `os.path.join()` instead of `+"/"+`

multiqc/modules/bcl2fastq/bcl2fastq.py
```diff
@@ -0,0 +1,138 @@
...
+                        self.bcl2fastq_bysample[uniqSampleName] = {"total": 0, "perfectIndex": 0}
+                    self.bcl2fastq_bysample[uniqSampleName]["total"] += self.bcl2fastq_data[runId][lane]["samples"][sample]["total"]
+                    self.bcl2fastq_bysample[uniqSampleName]["perfectIndex"] += self.bcl2fastq_data[runId][lane]["samples"][sample]["perfectIndex"]
+                    if sample != "undetermined":
+                        if not uniqSampleName in self.source_files:
+                            self.source_files[uniqSampleName] = []
+                        self.source_files[uniqSampleName].append(self.bcl2fastq_data[runId][lane]["samples"][sample]["filename"])
+
+    def add_general_stats(self):
+        data = {key: {"total": self.bcl2fastq_bysample[key]["total"], "perfectPercent": '{0:.1f}'.format(100*self.bcl2fastq_bysample[key]["perfectIndex"]/self.bcl2fastq_bysample[key]["total"])} for key in self.bcl2fastq_bysample.keys()}
```
ewels: `perfectPercent` is being treated as an `int` currently so is always a whole number. Need to use `100.0` and `float()`.

ewels:
Suggestion from a conversation at our end.. Ideas open for discussion, not requests :)

### Lane information
In our sample tracking database / website we have the following information for each lane of a flow cell:

image

For the _bcl2fastq by lane_ section it could be nice to try to replicate this - basically mention more stats (eg. Q30 is maybe more important than % perfect barcode).

So suggestion would be to have one table per flow cell, then each table with one row per lane and the headers as in the screenshot above (total clusters show as millions using the same config as elsewhere). Breaking into one table per run would also make it much easier to read.

### Samples
Although I know I specifically suggested / requested than run ID should be prepended to the sample names, now that I look at the report I'm not such a fan of it. I'm thinking that maybe all sample information should be collapsed across all lanes / runs (after parsing, we want to sum unless an exact duplicate). Then the general stats table will line up with output from later on and everything is much easier to read.

Would also be good to have Q30 in the General Stats table per sample.

### Advanced suggestion
Highcharts has a nice feature called drilldown (see [example](https://www.highcharts.com/demo/column-drilldown)) which would work really nicely here. For example, the Lanes plot could then just show each lane but clicking on a bar would show the breakdown of each sample in that lane.

The down side is that this isn't supported in MultiQC, so would need quite a lot of thought and code. Probably a new type of `bargraph.plot()` call and additional JavaScript functionality. That in turn would break a load of stuff and turn into a week long sufferfest.

It's almost certainly not worth it. But I thought it was worth mentioning :)


Phil

multiqc/modules/bcl2fastq/bcl2fastq.py
iimog: Interestingly worked for me already with both python2 and 3. So not sure whether my changes fixed it. Can you please verify?

multiqc/modules/bcl2fastq/bcl2fastq.py
iimog: Used L for now. Feel free to change to whatever you like.

multiqc/modules/bcl2fastq/bcl2fastq.py
iimog: Did that. Would it make sense to log something to debug in case of failure?

ewels: Yeah, would be good to have a `log.warn` with the filename saying that it was invalid JSON.

iimog:
The suggestions sound very good.
### Lane information
I can have a look tomorrow or early next week to see if I can easily retrieve the desired metrics from the JSON (I guess so). It even looks like those values are already present in the v1.8.4 style htm file.
### Samples
I totally agree. Should not be a big deal as I'm doing the aggregation separately anyway. So just have to change a few lines of code. Important thing is mainly to keep the duplicate detection working properly.
### Advanced suggestion
Also agree here :smiley: if I find lots of time I'll give it a shot.

multiqc/modules/bcl2fastq/bcl2fastq.py
iimog: done

multiqc/modules/bcl2fastq/bcl2fastq.py
```diff
@@ -0,0 +1,147 @@
...
+                        self.bcl2fastq_bysample[uniqSampleName] = {"total": 0, "perfectIndex": 0}
+                    self.bcl2fastq_bysample[uniqSampleName]["total"] += self.bcl2fastq_data[runId][lane]["samples"][sample]["total"]
+                    self.bcl2fastq_bysample[uniqSampleName]["perfectIndex"] += self.bcl2fastq_data[runId][lane]["samples"][sample]["perfectIndex"]
+                    if sample != "undetermined":
+                        if not uniqSampleName in self.source_files:
+                            self.source_files[uniqSampleName] = []
+                        self.source_files[uniqSampleName].append(self.bcl2fastq_data[runId][lane]["samples"][sample]["filename"])
+
+    def add_general_stats(self):
+        data = {key: {"total": self.bcl2fastq_bysample[key]["total"], "perfectPercent": '{0:.1f}'.format(float(100.0*self.bcl2fastq_bysample[key]["perfectIndex"]/self.bcl2fastq_bysample[key]["total"]))} for key in self.bcl2fastq_bysample.keys()}
```
ewels: So this additional `float` won't actually help, as if they're `int`s then the calculation still comes back as an `int`. Then this call makes that int a float, but still no decimal places.

The `100.0` will force float conversion though, so I think that's fine. And I'm getting decimals now so this is fine 👍

ewels:
Great stuff, thanks for this! Merged, but yes - would be great if we could make the above changes if you're happy with that, plus anything else that you had planned with Tim 👍

PR #532, opened 2017-07-25
Fix step numbering in Writing New Modules section

boulund:
There was a missing step number in the documentation (no step 3). Renumbered existing steps so there's no more gap.


ewels:
Great stuff, thanks!

PR #536, opened 2017-07-28
Software versions

Hammarn:
Some initial work on displaying the software versions from the modules in MultiQC.
We need to go through every module eventually and add the version check to them.
Making this PR now since I'm going on vacation and someone else might want to work on it in the meanwhile.


ewels:
Finally outdated by #1927 🎉 

Not sure if you're still on vacation or not @Hammarn, but not to worry either way - we got it covered 🥳 😆

PR #538, opened 2017-07-31
Bcl2fastq

iimog:
As discussed in #527 I added some changes to the behavior of the bcl2fastq module:
 - Samples are no longer prefixed with the run ID (so same samples from different runs are combined)
 - Yield Q30 is added to the main results table
 - Yield Q30 is added to the data files (both by sample and by lane)

The rest of the suggestions is not yet implemented. I will be busy in the next weeks as I try to finish my thesis. I'm happy to continue work on this module afterwards. However, if anyone wants to take over development of this module, please feel free.

It might make sense to add another configuration option `base_count_multiplier` for yield Q30 (just like the existing `read_count_multiplier`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,8 +152,8 @@ Code contributions from:
 [@t-neumann](https://github.com/t-neumann),
 [@ahvigil](https://github.com/ahvigil),
 [@bschiffthaler](https://github.com/bschiffthaler),
-[@jrderuiter](https://github.com/jrderuiter)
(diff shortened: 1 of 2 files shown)


ewels:
Brilliant stuff, thanks @iimog! Looks great. I'll merge into a `bcl2fastq` branch now and add a few extras myself (eg. the # bases multiplier) and then open up a pull request of my own to show what I changed.

Phil

PR #544, opened 2017-08-09
bargraph doc update

Galithil:
The exemple works slightly better that way.


ewels:
Aha, thanks! Good spot. This was missed in the imports refactoring in v1.0..

PR #547, opened 2017-08-12
Add help text for Qualimap sections

tabwalsh:


ewels:
Fantastic, thanks @tabwalsh! Did you write all of this yourself? I feel like the help text alone deserves its own DOI!

tabwalsh:
Thanks! I did indeed, though I've linked to sources where I got a definition or some key info. I'm very happy that you like it anyway...and thanks for catching the non-ASCII characters.

PR #548, opened 2017-08-14
MegaQC API export

ewels:
New code to handle saving all (or _nearly_ all) data to JSON and then either dumping that to a file or pushing it to an API endpoint.


ewels:
Test failure is due to a change in the TestData repo with a bug that is fixed in `master`.

PR #552, opened 2017-08-15
Correct `--filename` argument name in docs

juliangehring:
The old ` --name` argument has been replaced by the `--filename` argument, and calling `multiqc --name NAME ...` throws "Error: no such option: --name". This updates the documentation accordingly.


ewels:
Ooh, that one has been lurking there for a long time - great spot, thanks!

Phil

PR #555, opened 2017-08-17
Initial Base Recalibrator section

wkretzsch:
This PR adds a mixin class for parsing GATK BaseRecalibrator recalibration tables.  

At this point there is only one plot that summarizes the observed quality scores, but more to follow.

This PR also moves the logging of the number of records found by each mixin class to the class itself so that verbose mode shows the origin of the log messages correctly. That is, this:
```
[2017-08-17 12:12:04,741] multiqc.modules.gatk.base_recalibrator             [INFO   ]  Found 1 BaseRecalibrator reports
[2017-08-17 12:12:04,759] multiqc.modules.gatk.varianteval                   [INFO   ]  Found 3 VariantEval reports
```

instead of this:

```
[2017-08-17 12:02:09,612] multiqc.modules.gatk.gatk                          [INFO   ]  Found 1 BaseRecalibrator reports
[2017-08-17 12:02:09,629] multiqc.modules.gatk.gatk                          [INFO   ]  Found 3 VariantEval reports
```

### DIFF
--- a/multiqc/modules/gatk/base_recalibrator.py
+++ b/multiqc/modules/gatk/base_recalibrator.py
@@ -4,11 +4,19 @@
 """ MultiQC submodule to parse output from GATK BaseRecalibrator """
 
 import logging
-from collections import OrderedDict, defaultdict
+from enum import IntEnum
 from multiqc.plots import linegraph
 
 # Initialise the logger
 log = logging.getLogger(__name__)
+
+
+class RecalTableType(IntEnum):
+    pre_recalibration = 0
+    post_recalibration = 1
+
+    def human_readable(self):
+        return self.name.capitalize().replace('_', '-')
 
 
 class BaseRecalibratorMixin():
@@ -22,28 +30,35 @@ class BaseRecalibratorMixin():
             '#:GATKTable:RecalTable1:': 'recal_table_1',
             '#:GATKTable:RecalTable2:': 'recal_table_2',
         }
-        samples_kept = set()
-        self.gatk_base_recalibrator = {table_name: {} for table_name in
-                                       report_table_headers.values()}
+        samples_kept = {rt_type: set() for rt_type in RecalTableType}
+        self.gatk_base_recalibrator = {recal_type:
+                                           {table_name: {}
+                                            for table_name
+                                            in report_table_headers.values()}
+                                       for recal_type in RecalTableType}
+
         for f in self.find_log_files('gatk/base_recalibrator', filehandles=True):
             parsed_data = self.parse_report(f['f'].readlines(), report_table_headers)
+            rt_type = determine_recal_table_type(parsed_data)
             if len(parsed_data) > 0:
-                if f['s_name'] in samples_kept:
+                if f['s_name'] in samples_kept[rt_type]:
                     log.debug("Duplicate sample name found! Overwriting: {}".format(f['s_name']))
-                samples_kept.add(f['s_name'])
+                else:
+                    samples_kept[rt_type].add(f['s_name'])
 
                 self.add_data_source(f, section='base_recalibrator')
                 for table_name, sample_tables in parsed_data.items():
-                    self.gatk_base_recalibrator[table_name][f['s_name']] = sample_tables
+                    self.gatk_base_recalibrator[rt_type][table_name][
+                        f['s_name']] = sample_tables
 
         # Filter to strip out ignored sample names
-        self.gatk_base_recalibrator = {
-            table_name: self.ignore_samples(sample_tables)
-            for table_name, sample_tables
-            in self.gatk_base_recalibrator.items()
-        }
+        for rt_type in RecalTableType:
+            for table_name, sample_tables in self.gatk_base_recalibrator[rt_type].items():
+                self.gatk_base_recalibrator[rt_type][table_name] = self.ignore_samples(
+                    sample_tables)
 
-        n_reports_found = len(samples_kept)
+        n_reports_found = sum([len(samples_kept[rt_type]) for rt_type in RecalTableType])
+
         if n_reports_found > 0:
             log.info("Found {} BaseRecalibrator reports".format(n_reports_found))
 
@@ -54,11 +69,39 @@ class BaseRecalibratorMixin():
     def add_quality_score_vs_no_of_observations_section(self):
         """ Add a section for the quality score vs number of observations line plot """
 
-        sample_tables = self.gatk_base_recalibrator['quality_quantization_map']
-        sample_data = {
-            sample: {int(x): int(y) for x, y in zip(table['QualityScore'], table['Count'])}
-            for sample, table in sample_tables.items()
-        }
+        sample_data = []
+        data_labels = []
+        for rt_type in RecalTableType:
+            sample_tables = self.gatk_base_recalibrator[rt_type]['quality_quantization_map']
+
+            sample_data.append({
+                sample: {int(x): int(y) for x, y in zip(table['QualityScore'], table['Count'])}
+                for sample, table in sample_tables.items()
+            })
+
+            sample_y_sums = {
+                sample: sum(int(y) for y in table['Count'])
+                for sample, table
+                in sample_tables.items()
+            }
+
+            sample_data.append({
+                sample: {
+                    int(x): float(y) / sample_y_sums[sample]
+                    for x, y in zip(table['QualityScore'], table['Count'])
+                }
+                for sample, table in sample_tables.items()
+            })
+
+            flat_proportions = [float(y) / sample_y_sums[sample]
+                                for sample, table in sample_tables.items()
+                                for y in table['Count']]
+            prop_ymax = max(flat_proportions)
+            data_labels.append({'name': "{} Count".format(rt_type.human_readable()),
+                                'ylab': 'Count'})
+            data_labels.append({'ymax': prop_ymax,
+                                'name': "{} Percent".format(rt_type.human_readable()),
+                                'ylab': 'Percent'})
 
         plot = linegraph.plot(
             sample_data,
@@ -67,9 +110,8 @@ class BaseRecalibratorMixin():
                 'id': 'gatk-base-recalibrator-quality-score-vs-number-of-observations',
                 'xlab': 'Observed Quality Score',
                 'ylab': 'Count',
-                'yDecimals': False,
                 'xDecimals': False,
-                'tt_label': '{point.x}: {point.y:.0f}'
+                'data_labels': data_labels,
             })
 
         # Reported vs empirical quality scores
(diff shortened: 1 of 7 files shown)


multiqc/modules/gatk/base_recalibrator.py
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+""" MultiQC submodule to parse output from GATK BaseRecalibrator """
+
+import logging
+from collections import OrderedDict, defaultdict
+from multiqc.plots import bargraph, table, linegraph
```
ewels: Minor: `defaultdict` and `bargraph` are not used

multiqc/modules/gatk/base_recalibrator.py
```diff
@@ -0,0 +1,71 @@
...
+                    '#:GATKTable:RecalTable0:': 'recal_table_0',
+                    '#:GATKTable:RecalTable1:': 'recal_table_1',
+                    '#:GATKTable:RecalTable2:': 'recal_table_2',
+                }
+            )
+            if len(parsed_data) > 0:
+                if f['s_name'] in self.gatk_base_recalibrator:
+                    log.debug("Duplicate sample name found! Overwriting: {}".format(f['s_name']))
+                self.add_data_source(f, section='base_recalibrator')
+                self.gatk_base_recalibrator[f['s_name']] = parsed_data
```
ewels: As discussed, would be good to have multiple data structures here to avoid having the multi-level `parsed_data`

multiqc/modules/gatk/base_recalibrator.py
```diff
@@ -0,0 +1,71 @@
...
+        if n_reports_found > 0:
+            log.info("Found {} BaseRecalibrator reports".format(n_reports_found))
+
+            # Write parsed report data to a file (restructure first)
+            self.write_data_file(self.gatk_base_recalibrator, 'multiqc_gatk_base_recalibrator')
+
+            # Reported vs empirical quality scores
+            self.add_section(
+                name='Observed Quality Scores',
+                plot=quality_score_vs_no_of_observations(self.gatk_base_recalibrator)
```
ewels: Would be nice to have some `helptext` and a `description` :)

multiqc/modules/gatk/base_recalibrator.py
```diff
@@ -0,0 +1,71 @@
...
+
+    sample_data = {}
+    for sample, report in data.items():
+        table = report['quality_quantization_map']
+        xy = {int(x): int(y) for x, y in zip(table['QualityScore'], table['Count'])}
+        sample_data[sample] = xy
+    return linegraph.plot(
+        sample_data,
+        pconfig={
+            'xlab': 'Observed Quality Score',
```
ewels: Please add an `id` and `title`. Without `id` one is generated randomly, meaning that when you export the plot you get something like `mqc_hcplot_gtucwirdzx.png`. And plots should always have titles ;)

multiqc/modules/gatk/base_recalibrator.py
```diff
@@ -0,0 +1,71 @@
...
+            'ylab': 'Count',
+            'yDecimals': False,
+            'xDecimals': False,
+            'tt_label': '{point.x}: {point.y:.0f}'
```
ewels: Is the range of qualities fixed? If so, it may be nice to add a background to the plot to give an indication of what is a good quality score (like the FastQC plots).

multiqc/modules/gatk/gatk.py
```diff
@@ -25,25 +26,72 @@ def __init__(self):
...
+                        col_2: [ val_1, val_2, ... ]
+                        ...
+                    }
+                table_2:
+                    ...
+            }
+        """
+
+        data = dict()
+        while True:
```
ewels: Can you not do `while line = f.readline():` instead? `while True` makes me nervous 😛

ewels: or even `while line = f.readline().rstrip() and  line != '':`

multiqc/modules/gatk/gatk.py
```diff
@@ -25,25 +26,72 @@ def __init__(self):
...
+                break
+            line = line.rstrip()
+            if line in table_names.keys():
+                data[table_names[line]] = self.parse_gatk_report_table(f)
+        return data
+
+    def parse_gatk_report_table(self, f):
+        headers = f.readline().rstrip().split()
+        data = OrderedDict([(h, []) for h in headers])
+        while True:
```
ewels: 😰

multiqc/utils/search_patterns.yaml
```diff
@@ -71,6 +71,9 @@ flexbar:
 gatk/varianteval:
     contents: '#:GATKTable:TiTvVariantEvaluator'
     shared: true
+gatk/base_recalibrator:
+    contents: '#:GATKTable:Arguments:Recalibration argument collection values used in this run'
```
ewels: If you're happy with just `'#:GATKTable:Arguments:Recalibration'` being unique I'd probably go for that. I tend to go for the shortest unique string that I can, to future proof against tool authors changing wording in the future.

multiqc/modules/gatk/gatk.py
wkretzsch: I'm not sure that's valid python syntax (see [here](https://stackoverflow.com/questions/6631128/assignment-in-while-loop-in-python) for an SO question).  However, I agree, `while True` is evil. 

I wasn't sure about reading lines from the sample file handle in different places in conjunction with the `in` keyword.  However, that appears to work.  

I'll make the changes.

wkretzsch: Oh wait, maybe not.  Are we happy with dumping the whole report into a list and parsing it line by line later?

ewels: Ok, my bad for untested code suggestions :)

Why not just this?
```python
for line in f:
    line = line.rstrip()
    # ..etc
```

If you want the whole file (as a string rather than a list) then remove the `filehandles=True` from the `find_log_files` function

wkretzsch: Ok, never mind, the `in` syntax appears to work as expected.

multiqc/utils/search_patterns.yaml
wkretzsch: I'm not sure what the shortest unique string is, but I am happy to use your suggestion.

multiqc/modules/gatk/base_recalibrator.py
wkretzsch: Yes, the quality scores are the same as the ones analyzed by FastQC.  However, all scores are valid and good scores in this context.  It's the distribution of scores that is more interesting.

In the empirical vs report qual score figure (working on that currently) a departure from the x=y line is interesting.

multiqc/modules/gatk/base_recalibrator.py
wkretzsch: Ya!  I'll work on that.

multiqc/modules/gatk/gatk.py
wkretzsch: Ah, but there is a catch. Apparently the `in` syntax does not necessarily work with a later call to `readline`.  I think I'll be setting `filehandles=False`.

wkretzsch:
Ok @ewels, see if this works for you.

multiqc/modules/gatk/base_recalibrator.py
```diff
@@ -0,0 +1,86 @@
...
+        sample_tables = self.gatk_base_recalibrator['quality_quantization_map']
+        sample_data = {
+            sample: {int(x): int(y) for x, y in zip(table['QualityScore'], table['Count'])}
+            for sample, table in sample_tables.items()
+        }
+
+        plot = linegraph.plot(
+            sample_data,
+            pconfig={
+                'xlab': 'Observed Quality Score',
```
ewels: I think this one was missed:

> Please add an `id` and `title`. Without `id` one is generated randomly, meaning that when you export the plot you get something like `mqc_hcplot_gtucwirdzx.png`. And plots should always have titles ;)

multiqc/modules/gatk/base_recalibrator.py
```diff
@@ -0,0 +1,86 @@
...
+        # Reported vs empirical quality scores
+        self.add_section(
+            name='Observed Quality Scores',
+            description=(
+                'This plot shows the distribution of base quality scores in each sample before and '
+                'after base quality score recalibration (BQSR). Applying BQSR should broaden the '
+                'distribution of base quality scores.'
+            ),
+            helptext= (
+                'For more information see <a href=https://gatkforums.broadinstitute.org/gatk/discussion/44/base-quality-score-recalibration-bqsr>'
```
ewels: FYI, this can be written in markdown if you like (I have a feeling that this isn't documented?)

multiqc/modules/gatk/base_recalibrator.py
wkretzsch: Yup.  Sorry about that.

multiqc/modules/gatk/base_recalibrator.py
wkretzsch: Hmm, I tried markdown first, but it failed for me.  I'll try that again.

I could not see any markdown mention when I searched the docs for `add_section` or `helptext`...

multiqc/modules/gatk/base_recalibrator.py
wkretzsch: The documentation does not make special mention of `id` or `title`. If this is something that is good to add, then that should probably be mentioned in the dev section.  Might be a good candidate for --lint mode... :)

ewels: Ok, special mention of this added to the docs for barplot and linegraph 👍

multiqc/modules/gatk/base_recalibrator.py
ewels: Added to the docs 👍

wkretzsch:
Right, I think that's it?

wkretzsch:
@ewels, can I get a code review and merge on this PR please?

ewels:
Aha, is it ready? Sorry - there seemed to be lots of activity so I thought I'd wait a while :) Will try to get to it asap, quite a few reviews to do currently.

wkretzsch:
The activity was mostly just rebases onto the master branch.

No pressure time-wise.

setup.py
```diff
@@ -60,6 +60,7 @@
         'simplejson',
         'spectra',
         'networkx==1.*',
+        'enum34',
```
ewels: Hey! Quick question - what is this and is it needed? ;) I'm always a little paranoid about adding dependencies (especially after the trouble we're having with a couple at the moment). Feel free to grab me in person later..

Phil

ewels:
This is awesome stuff - thanks very much for your work and your patience with my questions 😉 

Phil

PR #560, opened 2017-08-18
Fail fast: Unit tests should go first

wkretzsch:


ewels:
Does this make any difference? I think the test will keep running even if the first command fails?

wkretzsch:
Sure, but I would expect the build to turn red as soon as the first test fails? 

The reason to put unit tests first is that they tend to be very fast and give the most informative error messages.

wkretzsch:
It seemed like a no-brainer. But feel free to close this PR if you remain skeptical.

PR #561, opened 2017-08-20
Downgrade networkx dependency to v1.*

vladsaveliev:
`spectra` depends on `colormath` which requires `networkx` but incompatible with with v2: https://github.com/gtaylor/python-colormath/issues/73, which makes MultiQC installation fail (e.g. https://travis-ci.org/MultiQC/MultiQC_bcbio/jobs/266483333). This PR fixes that by downgrading `networkx` to 1.*.

Not sure why Travis can't catch that. I've activated travis for my fork - it's expectedly fails: https://travis-ci.org/vladsaveliev/MultiQC/jobs/266568649. Maybe some caching happens. What do you think about removing the `-q` flag in https://github.com/ewels/MultiQC/blob/master/.travis.yml#L9 which will help to see which versions of which libraries are installed?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,7 @@ setup(
         'requests',
         'simplejson',
         'spectra',
+        'networkx==1.*',
     ],
     entry_points = {
         'multiqc.modules.v1': [


ewels:
I think the tests are due to a timing problem - I just pushed a minor docs update and all tests failed: https://travis-ci.org/ewels/MultiQC/jobs/266720567

Thanks for the spot and fix! Annoying that the issue is just sat there waiting. I might submit a PR there later as no-one else seems to be doing anything.

Phil

vladsaveliev:
Thanks for merging and for the PR there. I'm wondering how many lazy people like me preferred to downgrade networkx rather than submit a PR there.

ewels:
I'm happy to spend thirty seconds forking and doing the PR if people such as yourself are happy to spot the problem and suggest fixes 😉 

I'm happy with both approaches - downgrading should solve the problem for MultiQC quickly and the PR is better in the long term.

How bad do you think this problem will be? I'm wondering if I need to release a hotfix version..

vladsaveliev:
Yep, agree about both approaches.

Seems like the last stable version of networkx is actually 1.11. Conda-forge also have 1.11 https://anaconda.org/conda-forge/networkx, so it shouldn't reproduce when installing from conda. However, it will be failing when installing with pip: someone uploaded the beta version of networkx this Friday https://pypi.python.org/pypi/networkx/2.0b1. That's why it seems no one complained before me: after you released 1.2, it must have worked okay even with pip.

vladsaveliev:
So I don't know. Something needs to be fixed, though I prefer it to be the networkx version in PyPI. I created an issue in their repository: https://github.com/networkx/networkx/issues/2627

ewels:
ok thanks - I'll wait a little while and see how many bug reports come in :)

ewels:
Panic over - PyPI default version restored and PR merged for future stability 😅

Issue #542, opened 2017-08-08
Group samples: merge summary stats


ewels:
To avoid having half-empty rows from broken up samples, have the option to merge the statistics given in General Statistics. This shouldn't affect plots or any data in sections, _only_ General Stats.

For example, have the following config (modelled on [sample name cleaning](http://multiqc.info/docs/#sample-name-cleaning)):

```yaml
general_stats_merge:
    - '_R1'
    - '_R2'
    - type: regex_keep
      pattern: '[A-Z]{3}[1-9]{2}'
```

Then group any samples matching these and merge the stats in a module-specific and statistic-specific manner. This would also allow situations such as multiplexed lanes etc.

Benefits:
* Sections in report maintain full original data
* Saved data in `multiqc_data` maintains full original data
* Summary table more concise and easy to skim, lines up with other modules
* Doesn't affect front-end or any major existing code infrastructure

Downsides:
* Requires specific code to be written for each module that will support this.

ewels:
To do this, have a new base module function. Pass a list of sample names to it, will return a dict of lists (same sample names, grouped into bundles that should be merged with the key as the sample basename). This result can then be iterated over for the merging process.

eg:

```python
g_stats = {
    'sample_1_R1': { #...
    'sample_1_R2': { #...
    'sample_2_R1': { #...
}
new_g_stats = {}
grouped_s_names = self.general_stats_merge( g_stats.keys() )
for basename, s_names in grouped_s_names.items():
    new_g_stats[basename] = {}
    # loop through s_names collecting stats for each key and merging them into new_g_stats
```

ewels:
Discussed with @tbooth [on gitter](https://gitter.im/ewels/MultiQC?at=599592e2a7b406262d9a93ad). Would be good to generalise the above more so that it's not tied to the General Stats table. Instead just give the ability to collapse a list of sample names to a list of sets with a shared basename. Modules can then use this to collapse rows in the General Stats if they wish, as well as splitting plots into multiple datasets.

The config for the different patterns should be split into different keys to allow this functionality to be used for different reasons. For example, it may be nice to group samples based on read pairs (likely the most common reason) and also pre- and post- trimming status.

```yaml
merge_group_patterns:
    - read_pairs:
        - type: regex
          pattern: '_R[12]$'
    - pre_post_trimming:
        - '_trimmed'
```

[...]

PR #566, opened 2017-08-28
Attempt to merge my development branch back into mainline

tbooth:
This is a little messy and will need some refining. More info and explanation to follow.
The main thing here is my attempt to address issue #542


multiqc/modules/base_module.py
```diff
@@ -209,10 +226,23 @@ def clean_s_name(self, s_name, root):
...
+
+        # Look for merge groups. Typical use case is for reads 1 and 2 where we want
+        # to prune the read number to get the sample name but we don't want to lose the
+        # information.
+        for mg, regex in config.define_merge_groups.items():
+            mo = re.search(regex, res['s_name'])
+            if not mo:
+                res[mg] = None
+            else:
+                # If there was a match, the fist group is the read name (or whatever)
```
ewels: ✊

multiqc/modules/fastqc/fastqc.py
```diff
@@ -36,20 +38,22 @@ def __init__(self):
...
-            s_name = self.clean_s_name(os.path.basename(f['root']), os.path.dirname(f['root']))
-            self.parse_fastqc_report(f['f'], s_name, f)
+            self.parse_fastqc_report(f['f'], f=f) #fffffff
 
         # Find and parse zipped FastQC reports
         for f in self.find_log_files('fastqc/zip', filecontents=False):
-            s_name = f['fn']
-            if s_name.endswith('_fastqc.zip'):
-                s_name = s_name[:-11]
+            # Do we still need this??
```
ewels: We could add `_fastqc.zip` to `config.fn_clean_exts` and then it wouldn't be necessary..

multiqc/modules/fastqc/fastqc.py
```diff
@@ -93,41 +101,62 @@ def __init__(self):
...
-        self.n_content_plot()
-        self.seq_length_dist_plot()
-        self.seq_dup_levels_plot()
-        self.overrepresented_sequences()
-        self.adapter_content_plot()
-
-    def parse_fastqc_report(self, file_contents, s_name=None, f=None):
+        # Allow plot selection in config
+        try:
+            plots_enabled = config.fastqc_plots_enabled
```
ewels:
```python
plots_enabled = getattr(config, 'fastqc_plots_enabled', [])
```

multiqc/plots/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+    - random letters work ok but it seems messy to put that code in every chart
+    module.
+    - UUID seems overkill
+    - asking the report for a UID would work but I want to sever the dependency
+    between plots and report classes
+"""
+
+_uid = 0
+
+def get_uid():
```
ewels: Please use `report.save_htmlid()` instead ([code here](https://github.com/ewels/MultiQC/blob/master/multiqc/utils/report.py#L250-L274)). This collects all IDs across the report to ensure that they're unique without using random strings.

If you want it to work with empty strings then it'd be better to modify that function to return a blank uid in this way. The approach you've used here is nice, but should _really_ be a last resort - I'm trying to avoid using random strings where possible as they are very unhelpful when exporting plots (the unique ID is used there) and in other downstream applications which use plot data (eg. the upcoming MegaQC).

multiqc/plots/bargraph.py
```diff
@@ -248,9 +246,10 @@ def matplotlib_bargraph (plotdata, plotsamples, pconfig=None):
 
     # Plot group ID
     if pconfig.get('id') is None:
-        pconfig['id'] = 'mqc_mplplot_'+''.join(random.sample(letters, 10))
+        pconfig['id'] = 'mqc_mplplot_{}'.format(id(pconfig))
 
     # Sanitise plot ID and check for duplicates
+    from multiqc.utils import report
```
ewels: Why move this down here instead of having it in the main import block?

multiqc/plots/beeswarm.py
```diff
@@ -33,9 +35,10 @@ def plot (data, headers=None, pconfig=None):
 
 def make_plot(dt):
 
-    bs_id = dt.pconfig.get('id', 'table_{}'.format(''.join(random.sample(letters, 4))) )
+    bs_id = dt.pconfig.get( 'id', 'table_{}'.format(id(dt)) )
 
     # Sanitise plot ID and check for duplicates
+    from multiqc.utils import report
```
ewels: Is this causing some kind of recursion error or something?

multiqc/utils/config.py
```diff
@@ -62,9 +59,32 @@
...
+for searchp_fn in [ os.path.join( d, 'utils', 'search_patterns.yaml') for d in module_roots ]:
+    try:
+        with open(searchp_fn) as f:
+            sp.update(yaml.load(f))
+    except OSError:
+        pass
+
+# We can't log yet, but this is a serious error.
+if not sp:
+    print("No default search patterns were loaded. Looked in locations:\n{}".format(pformat(module_roots)),
```
ewels: Would this ever happen? Surely the base config is always going to load with some search patterns?

multiqc/utils/report.py
```diff
@@ -282,3 +309,4 @@ def compress_json(data):
     x = lzstring.LZString()
     return x.compressToBase64(json_string)
 
+    return found
```
ewels: Is this a GitHub diff error or is this really here? Probably a typo if so..

scripts/multiqc
```diff
@@ -350,27 +350,32 @@ plots_flat, plots_interactive, make_pdf, config_file, cl_config, verbose, quiet,
...
+    run_modules.update( [ (n, c) for m in config.top_modules for n, c in m.items() if n in config.avail_modules ] )
+    run_modules.update( [ (n, {}) for n in config.avail_modules if n not in mod_keys and n not in run_modules ] )
+    run_modules.update( [ (n, c) for m in config.module_order for n, c in m.items() if n in config.avail_modules and n not in run_modules ] )
 
     if module:
-        run_modules = [ m for m in run_modules if list(m.keys())[0] in module ]
-        logger.info('Only using modules {}'.format(', '.join(module)))
+        #Pare down config.run_modules
+        for m in list(run_modules):
+            if m not in module: del(run_modules[m])
```
ewels: Does deleting a list item whilst looping over the same list like this work?

ewels:
Thanks Tim! This is a huge PR so it'll take a while to go through.. First comments attached in the above review.

Regarding the new file merge functionality, I'm still not sold on the approach of modifying the results from `self.find_log_files()` if I'm honest. I don't really see the advantage of taking that approach instead of introducing a new separate function that can be called when required.. As mentioned in a comment, lots of modules get their sample names from the file contents so would have to do a separate call to get this information anyway. In #542 you mention that this "avoids adding redundant code" but I don't see how that's the case - would you mind explaining your thinking about this a little?

I had already started on a branch of my own for this, so I'll try to wrap it up into a PR and we can compare approaches..

Phil

multiqc/modules/base_module.py
```diff
@@ -54,6 +54,11 @@ def __init__(self, name='base', anchor='base', target=None, href=None, info=None
 
         self.sections = list()
 
+    def gather(self):
+        """This will be called after module init.  You can choose to use it or not.
+        """
+        pass
```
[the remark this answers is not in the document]
tbooth: Yes. I know we discussed this before and didn't come to a firm conclusion, but I can't think of a good way to test the modules if everything triggers from `__init__()`. I'll fix the comment anyway.

multiqc/modules/base_module.py
```diff
@@ -162,12 +167,23 @@ def add_section(self, name=None, anchor=None, description='', comment='', helpte
         })
 
     def clean_s_name(self, s_name, root):
+        """ For backwards compatibility, returns only the s_name value from
+            clean_and_scan_s_name.
+            Most of the modules calling this directly seem to be doing so
+            redundantly as it is already called by find_log_files.
```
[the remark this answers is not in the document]
tbooth: OK, I'll fix the comment. There were definitely redundant calls in the FastQC modules when I was modifying them.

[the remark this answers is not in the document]
ewels: Cool - if you can find them again then that's even better, as they should be removed really..

multiqc/modules/fastqc/fastqc.py
tbooth: This is something I added just because I needed it myself (sorry there is a lot of conflated stuff in this PR!). Do you like the idea for inclusion in mainline? In general, should modules be allowed to have their own configuration and if so should it look like this?

multiqc/plots/__init__.py
tbooth: Agreed - I think I was actually trying to get rid of the random strings in parallel with your own efforts and just didn't do a thorough job. I'll clean this up.

multiqc/modules/base_module.py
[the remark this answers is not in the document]
ewels: ..and yes, if there are redundant calls anywhere, they'll be in the FastQC module. It's the oldest module so much of that code predates the core functions.

multiqc/plots/__init__.py
tbooth: (greps code...) As I suspected - it's imported but never called

multiqc/modules/fastqc/fastqc.py
ewels: Yup, definitely. Quite a few modules already have their own config - eg. [preseq](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/preseq/preseq.py#L78-L80), [qualimap](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/qualimap/QM_BamQC.py#L20-L30) and others.

In fact, FastQC is already doing it [here](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/fastqc/fastqc.py#L447), so maybe nest under the `fastqc_config` key?

multiqc/plots/beeswarm.py
tbooth: Yes it's more of me trying to avoid circular imports and global variables in order to maintain loose-coupling and hence enable unit testing.
Ideally, IMHO, the plot libraries should not need to see the global report object in order to generate HTML snippets. When I started the fork, they could be decoupled with the count_hc_plots()/count_mpl_plots() functions that I added in place of global counters, but since then there is more coupling added, so I moved the imports as a compromise.

multiqc/utils/config.py
tbooth: I guess this is more of an assertion failure than a runtime error since every module should come with a matching set of file patterns to scan.
If the module is defined as part of a plugin (which is what I'm working on) then the main base config has no idea about the module and what files to feed it. To me, having all the plugins listed out in a monolithic base config file goes against the modular ethos anyway, so this is my attempt to fix it both for plugins and in general by allowing config fragments to be discovered and merged.

multiqc/utils/report.py
tbooth: it's a mis-merge. surprised pyflakes didn't spot that

scripts/multiqc
tbooth: Yes because the cast to `list()` does a shallow copy. If this was only for Py3 I would have done run_modules.copy() which is a little clearer.

tbooth:
Cheers for all the comments, Phil. My plan here was to get the PR into the system as soon as I could for a combination of transparency and to make use of the CI checks (which I see have failed) so I was fully expecting a good amount of tinkering, polishing and reverting bits that don't work in mainline or else need more though before they can go in.
Which branch should I be looking at for your work on #542? I can get GitHub to generate an inter-branch comparison without needing a PR.

I think the best way to understand my thinking on the implementation of the file merge is to see how I incorporated it into the FastQC parsing code. I tried to go with your suggestion that the functionality should be added to the existing `clean_s_name()` code, and since `find_log_files()` calls this on each filename it made logical sense to me that it should therefore pick up the new information that has been extracted, so any module that cares about it has the info right there in the `f[...]` dict. I'll also try to avoid writing functions that process whole batches of sample names and return data structures - I'd prefer to have a function that processes one pattern at a time and returns a simple dict and leaves it up to the caller how to apply it to batches of input.

tbooth:
I'll also note that my main motivation for tinkering with run_modules was that I have a custom module that parses cutadapt logs, but cutadapt is being run specifically to detect adapter dimers in my case. So my module is not useful for 'generic' cutadapt output and the standard cutadapt module does not show useful graphs for the log files I have. In short, I wanted a way for a custom module to auto-blacklist a core module by writing into this data structure. Then I realised that it seemed to make more sense for it to be an OrderedDict() rather than a list of singleton dicts so I altered the code to do that.

ewels:
> [Opened the PR to] make use of the CI checks

I think that you should be able to get Travis to run on your fork of MultiQC if you enable that repository on [travis-ci.org](https://travis-ci.org/), then you get tests without requiring a PR.. (and get notifications every time you push a commit)

> Which branch should I be looking at for your work on #542?

Sorry, the code I had was a bit of a mess and not really working. I've just tidied up and pushed it to the new [group_samples branch](https://github.com/ewels/MultiQC/tree/group_samples). It adds two new core functions and has proof-of-concept code for the FastQC General Stats table and one of the FastQC plots.

> I'd prefer to have a function that processes one pattern at a time and returns a simple dict and leaves it up to the caller how to apply it to batches of input.

The problem that I have with this is that it means all modules run the code, when only a few will use it. Plus it compromises flexibility if the commands aren't called directly (different modules can't call the commands in different ways). For example, I'm not clear on your code allows for multiple different grouping schemes? (Sorry if it does, the PR was long!)

> I wanted a way for a custom module to auto-blacklist a core module by writing into this data structure.

You can already exclude modules using `--exclude` on the command line - would it not be easier to just add a `config.exclude_modules` list into this existing code? If you don't want users to have to set this themselves, you can write a plugin snippet to set this by default alongside the module by tying into the `before_config` plugin hook (eg. as done [here](https://github.com/ewels/MultiQC_NGI/blob/master/multiqc_ngi/multiqc_ngi.py#L27)).

Regarding the length of the PR.. If you're able to cherry pick related commits / changes into separate branches and open up smaller PRs for separate features that would be much easier to review and merge. I'm a bit worried that this PR will sit here for a long time as there are so many changes contained within.

Phil

multiqc/modules/fastqc/fastqc.py
tbooth: Ah - how did I miss that? Fixed (and noted that it needs documenting)

multiqc/plots/beeswarm.py
ewels: Yes, there's probably a better way of doing it. I hesitated before writing it that way, but I needed the global `report.html_ids` list to keep track of the IDs as they were being created to avoid duplicates and I couldn't think of an alternative way to do that.

multiqc/utils/config.py
ewels: Ah I see, sorry I misunderstood this change and assumed that this was an extra option to allow a `search_patterns.yaml` file for external modules.

Yes, I don't have a problem with breaking up the central `search_patterns.yaml` file into smaller files that are in each module directory. They used to be kept in the main `config_defaults.yaml` file, which in turn used to all be hardcoded into `config.py` (which in turn used to be part of `multiqc.py` a long, long time ago)...

scripts/multiqc
ewels: Gotcha 👍  Need to remember that trick! I usually end up writing iterators to do it, but they're not as clear to read.

ewels:
I think the time to close this PR has come, though having seen it in the PR list for so long it feels like an old friend now... 😥 

If ever you fancy submitting this again in slightly smaller, more atomic PRs, then I'd be happy to consider them 😊 

Phil

PR #569, opened 2017-08-31
resolved bug in special case max_x=0

vezzi:
if `max_x=0` then this expression breaks:
```  depth_range = list(range(0, max_x + 1, math.ceil(float(max_x)/400.0)))```

because the step need to be greater than 0 (`math.ceil(float(max_x)/400.0)` evaluates to 0). 
This happens if the most occurring coverage is 0, that might happen if you align capture data against the whole genome.


ewels:
Great stuff, thanks!

Issue #413, opened 2017-03-01
Module request: 10X chromium


ewels:
A few people have asked for MultiQC to support output from the 10X chromium. This has three distinct sets of software:

* [`cellranger`](https://support.10xgenomics.com/single-cell/software/pipelines/latest/output/overview) (single cell)
* [`longranger`](https://support.10xgenomics.com/genome-exome/software/pipelines/latest/output/overview) (long reads)
* [`supernova`](https://support.10xgenomics.com/de-novo-assembly/software/pipelines/latest/output/overview) (de novo)

There's also the `mkfastq` commands that each tool has, which has mostly consistent stats.

We should be able to start by parsing some basic stats from the `metrics_summary.csv` file produced by both `cellranger` and `longranger`.

ewels:
Supernova added, closing this issue now as it's pretty broad. If people are interested in the other tools, please create a new issue..

atermanini:
cellranger mkfastq output (metrics_summary.csv) can be very useful to have

PR #570, opened 2017-08-31
Supernova module

remiolsen:
Requested in #413


multiqc/modules/supernova/supernova.py
```diff
@@ -0,0 +1,412 @@
...
+                if key == 'valid_bc_perc':
+                    value = 100 - value
+                data[stats[key]] = value
+        
+        return (sid, data)
+
+
+    def parse_report(self, content):
+        # Some short-hands for converting the report numbers (pi is exactly three!)
+        exp = {
```
remiolsen: I feel as though this a crime against science, but I couldn't think of any other way to keep the `multiqc_data/multiqc_supernova.txt` consistent. Thoughts @ewels ?

ewels: What's wrong with this approach? That we're making it look like the numbers are more accurate than they really are? I think it's fine... Better to have useful numbers for downstream use than less helpful string values (where people will just do this themselves anyway).

You can always confess to your crime in the docs if you really want ;)

docs/modules/supernova.md
```diff
@@ -0,0 +1,24 @@
+---
+Name: Supernova
+URL: https://www.10xgenomics.com/
+Description: >
+    Supernova is a de novo genome assembler for 10X Genomics linked-reads.
+---
+
+## Important notes
```
ewels: Please make these h4 headings (`####`) for consistency with other module docs - thanks!

multiqc/modules/supernova/supernova.py
remiolsen: Yeah, I guess I'm just worrying too much. And yes, having something useful and not in need of interpretation / conversion of the raw data is what I was going for.

I already confessed my crime with that code cheeky comment :wink:

ewels:
Hey!

First super quick review after running the code with the test data in MultiQC_TestData.. Any reason why one of these samples has a 0.0 assembly size?

image

From the sample names it looks like it should be 1. I'm slightly suspicious about it being `1.0` and `0.0` too - sometimes this indicates that a `float` is being incorrectly cast as an `int`... Also `# Long Scaff` is different.

Other speed thoughts:

* Are the two tables the same, just with some columns hidden in General Statistics?
   * I would prefer to have far fewer columns in General Stats, maybe max 4 and max 2 hidden?
   * I don't like duplicating data where possible. So basically if everything is in the Supernova table I don't think it needs to be in General Stats as well...
* Is it worth trimming the long tail off the k-mer plot? MultiQC has built in functionality to do this if you want to.
* I'm not sure that the `b` numeric suffix is clear..
    * Without looking at the code, I guess you're doing this automatically with a function?
    * Maybe better to have nothing if just regular base pairs (`'Mb'`, `'Kb'`, `''`)
    * It's good to have the long-form of the unit in the column header tooltip
* The order of columns in the supernova table seems backwards to me. I try to have it from unprocessed on the right to most processed on the left.
* `% missing BC`, `%dup` and `%phased` - can we use these numbers to make a stacked bar plot and remove them from the table
* `N50` values - can we have these in a bar plot, with each category as a different tab above the plot?


Phil

ewels:
> Is it worth trimming the long tail off the k-mer plot? MultiQC has built in functionality to do this if you want to.

I lied about this being built-in functionality. But Qualimap BamQC and Picard WGS both do it, plus I recommended it earlier on another PR. So maybe it should be core functionality!

> No, I don't think that these are mutually exclusive sets of reads. 

Fair enough. Note that digging through supernova's guts probably isn't necessary though. I've had similar situations in previous modules and if you have a bunch of samples it's usually pretty easy to see if numbers sub to a total number reported elsewhere, or if percentages sum to 100 etc. But fair enough if this isn't the case here.

ewels:
Awesome progress, it's much nicer now!

Summary of things discussed on slack just now:

* Multiple plot datasets instead of `stacking: false` for the bar plot
   * See [docs](http://multiqc.info/docs/#switching-datasets-1), but basically just supply a list of data dicts instead of one.
* Add `description` and `helptext` to the different sections
    * See [docs](http://multiqc.info/docs/#step-5-create-report-sections) - note that they take markdown.
    * eg. It would be great to have an explanation of what the different N50s mean and what you should look for.

Hopefully minor things, then we can merge 👍 

Phil

multiqc/modules/supernova/supernova.py
```diff
@@ -0,0 +1,440 @@
...
+                'suffix': '%',
+                'scale': 'OrRd',
+        }
+        self.headers['% R2 Q30'] = {
+                'description': 'fraction of Q30 bases in read 2; ideal 75-85',
+                'suffix': '%',
+                'scale': 'OrRd',
+        }
+        self.headers['Insert size'] = {
+                'description': 'median insert size (in bases); ideal 0.35-0.40',
```
ewels: Do you mean `0.35-0.40` here?

multiqc/modules/supernova/supernova.py
```diff
@@ -0,0 +1,440 @@
...
+                'description': 'mean read length (in bases) after trimming; ideal 140',
+                'suffix': 'b',
+                'scale': 'PuBu',
+                'format': '{:,.0f}',
+                'hidden': True
+        }
+        self.headers['# Reads'] = {
+                'description': 'number of reads (in millions); ideal 800M-1200M for human',
+                'modify': lambda x: x / 1000000.0,
+                'suffix': 'M',
```
ewels: Instead of having an `M` suffix, I prefer (and most other modules do this) to have the `M` in the column name...

```python
self.headers['# Reads'] = {
    'name': '# Reads (M)',
    'description': 'number of reads (in millions); ideal 800M-1200M for human',
    'modify': lambda x: x / 1000000.0,
    'scale': 'PuBu',
}
```

multiqc/modules/supernova/supernova.py
remiolsen: Aha! This text was mostly cut 'n pasted from the `report.txt` files of supernova. But it would make more sense to change this to `350-400` now

multiqc/modules/supernova/supernova.py
remiolsen: I was merely trying to save precious horizontal screen space :smirk:
Or does giving a (longer) `name` option something special with the column widths?

remiolsen: Hmm, `'name': '# Reads (M)'` didn't really do anything. The columns still said `# Reads`

ewels: Sorry, it's `title` not `name`. I should really read my own documentation...

ewels: And no, nothing special with column widths. This is just about consistency with other modules.

remiolsen: No problem :joy:
Maybe do the same for Coverage? Change it to `Coverage (x)`. What about the (mega/kilo) bases units? Do you want me to move those units to the column title as well?

ewels: Meh, it's probably fine..

ewels:
Awesome stuff - thanks @remiolsen ! 🎉 

Phil

remiolsen:
Great! Thank you :shipit:

PR #571, opened 2017-08-31
Safer networkx version specification

vladsaveliev:
Let's make it `<2` to avoid installation errors with old setuptools. Anyways we don't care about older versions for other packages, so why worry here. But just to make sure, I tests against networkx<1 and it works smoothly.


ewels:
Perfect, thanks!

PR #572, opened 2017-08-31
samtools-stats "Percent Mapped" plot help test

vladsaveliev:
This has been sitting in my fork for quite a while, waiting when I brace myself and write help for other sections and modules. But I guess it's better to start contributing with a little thing than procrastinate forever.

### DIFF
--- a/multiqc/modules/samtools/stats.py
+++ b/multiqc/modules/samtools/stats.py
@@ -153,18 +153,17 @@ class StatsReportMixin():
             name = 'Percent Mapped',
             anchor = 'samtools-stats-alignment',
             description = "Alignment metrics from <code>samtools stats</code>; mapped vs. unmapped reads.",
-            helptext="""For a set of samples that have come from the same multiplexed library, 
-        similar numbers of reads for each sample are expected. Large differences in numbers might 
-        indicate issues during the library preparation process. Whilst large differences in read 
-        numbers may be controlled for in downstream processings (e.g. read count normalisation), 
-        you may wish to consider whether the read depths achieved have fallen below recommended 
-        levels depending on the applications (e.g. ~ 10 million reads for differential gene
-        expression analysis.)</p>
-        <p>A high-quality library should have ~95% of reads align. Low alignment rates could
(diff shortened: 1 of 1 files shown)


ewels:
Great stuff - have you tested it and does it look ok? My suspicion is that you could get some odd markdown formatting here with code blocks. MultiQC unindents stuff automatically with helptext, but that only works if you have the same indentation across the whole string. NB:

```python
        helptext = '''
        This has the
        same indentation
        ''',
        helptext = '''This has no indentation
        on the first line
        ''',
```

ewels:
Also, you don't need `</p><p>` to give spacing between paragraphs. Just do a double line break.

multiqc/modules/samtools/stats.py
```diff
@@ -153,6 +153,18 @@ def alignment_section(self, samples_data):
...
             anchor = 'samtools-stats-alignment',
             description = "Alignment metrics from <code>samtools stats</code>; mapped vs. unmapped reads.",
+            helptext="""For a set of samples that have come from the same multiplexed library, 
+        similar numbers of reads for each sample are expected. Large differences in numbers might 
+        indicate issues during the library preparation process. Whilst large differences in read 
+        numbers may be controlled for in downstream processings (e.g. read count normalisation), 
+        you may wish to consider whether the read depths achieved have fallen below recommended 
+        levels depending on the applications (e.g. ~ 10 million reads for differential gene
+        expression analysis.)</p>
+        <p>A high-quality library should have ~95% of reads align. Low alignment rates could
```
ewels: I think this sentence needs a little bit more qualification as it depends so heavily on the library type. If this is used with a bisulfite mapper, 80% alignment is a great result. If it's non-Human then 50% can be a great result etc. etc.

vladsaveliev:
The formatting actually looks good this way, but it breaks when I replace <p> with a double line break. 

But I reformatted it as you suggested, to be fully correct and consistent.

multiqc/modules/samtools/stats.py
vladsaveliev: I removed all specific stuff from there for now.

ewels:
Great stuff, thanks! 👍

PR #573, opened 2017-09-05
Fix FastQC link targets

juliangehring:
Replaces all occurrences of `target="_bkank"` with a valid `target="_blank"` in the FastQC module.


ewels:
haha, oops! Thanks for the fix :)

PR #574, opened 2017-09-05
Simplify the `general_stats_addcols` example

juliangehring:
Drops passing the `header` argument since it is not defined or discussed up to this point in the documentation (it is coming in the following section). The example here works fine without specifying the header explicitly.

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -377,7 +377,7 @@ data = {
         'second_col': '66.3%'
     }
 }
-self.general_stats_addcols(data, headers)
+self.general_stats_addcols(data)
 ```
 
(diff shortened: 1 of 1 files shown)


juliangehring:
I think this is only an issue if you read the docs from top to bottom for the first time, the meaning of the `headers` argument gets explained shortly after this part. But it may be nevertheless also nice to shown that `headers` is optional and one can get results with the default value.

ewels:
Yup, absolutely - I suspect that this was an error rather than intentional as the example wouldn't run as `headers` isn't defined in the code block. Good spot again, thanks!

Issue #380, opened 2017-01-03
BBMap Module


c-guzman:
Was hoping we could get a module for a couple of the BBMap Suite tools. They're incredibly useful for splice aware and unsplice aware NGS data. (https://sourceforge.net/projects/bbmap/)

BBMap is a suite of NGS tools for NGS. I'm specifically intrested in modules for BBMap, and BBDuk.

BBMap: Short read aligner for DNA and RNA-seq data. Capable of handling arbitrarily large genomes with millions of scaffolds. Handles Illumina, PacBio, 454, and other reads; very high sensitivity and tolerant of errors and numerous large indels. Very fast.

BBDuk: Filters, trims, or masks reads with kmer matches to an artifact/contaminant file.

ewels:
Hi @c-guzman,

Sounds great! I'm away for a while now (until mid-Feb) so won't be able to write anything until then, but I'd be happy to do so after that. Until then, if you could submit some example data to [ewels/MultiQC_TestData](https://github.com/ewels/MultiQC_TestData) then that would be really helpful - then I have something to work with when I get back. Pull requests to this repo for new modules are also welcome, I'll do whatever reviewing I can.

Phil

Phil

boulund:
I've been meaning to write some MultiQC modules for the BBTools suite for my own use. Didn't realize there was an issue on this here already. Maybe we should discuss how to best identify the relevant output files, as none of the BBTools enforce any file naming scheme? 
I'm essentially interested in presenting most of the statistics your can get out of e.g. BBMap or BBDuk, which for me means something like this (from the help output):
```
Histogram output parameters:                                 
bhist=<file>        Base composition histogram by position.  
qhist=<file>        Quality histogram by position.           
qchist=<file>       Count of bases with each quality value.  
aqhist=<file>       Histogram of average read quality.       
bqhist=<file>       Quality histogram designed for box plots.
lhist=<file>        Read length histogram.                   
gchist=<file>       Read GC content histogram.               
```
I would also be interested in summarizing these with MultiQC:
```
stats=<file>        Write statistics about which contamininants were detected.
refstats=<file>     Write statistics on a per-reference-file basis.           
```

I normally output these as compressed (gzipped) textfiles, but as BBTools doesn't enforce any naming scheme I think it can be come somewhat annoying to auto-identify these output files (especially considering that the sample/filenames are not included anywhere in the output files). Perhaps it is easiest to recommend/enforce a naming scheme from the MultiQC module's side.

[...]

PR #577, opened 2017-09-06
First attempt at BBMap module (beta) 

boulund:
Here's a first BBMap module. It produces graphs for most BBMap log types, but includes no information in the MultiQC summary table at the top of the report (I have no need for it at the moment). Massive kudos to @epruesse who wrote all of the parsing code, and designed the data structures. I refactored the code a bit, and split out a lot of the plotting code to separate modules to make it easier to get an overview of what's going on. There's now some unnecessary code duplication as a result, but that can be improved in a future minor refactor. 

There are some things that remain to be done feature-wise, but the module works decently as it is. Please let me know if there's something wrong with the code or the PR in general. 

edit: I should perhaps mention the associated Issue #380


ewels:
Aha, great stuff! Many thanks both for working on this - looks like it's been a huge job!

Unfortunately there are a couple of merge conflicts, once they're sorted and the tests are passing I'll take a look. I'm away travelling next week so it may be a little while until I get to it (plus there are quite a few PRs stacking up now) but I'll get there as soon as I can :)

Phil

boulund:
Tried to fix the merge conflicts in `README.md` and `setup.py`, they seemed easy enough. A bit unsure about the `networkx` package in `setup.py` however, not sure if it's supposed to be there or if I forgot to remove it in a previous merge sometime before. Please advise @ewels

ewels:
Thanks! The `networkx` is meant to be there, yes. See #561 and #571 for explanation (it's due to an upstream dependency problem).

Phil

boulund:
Great. Then we'll just hope the CI tests pass :)

ewels:
Hi both,

I've finally found time to sit down and try to run your code! 🎉  Sorry it's taken me so long.. Ok, first thoughts from running without reading:

1. I think that these would be better as `log.debug` messages instead.. It's kind of scary to see this and there isn't actually anything going wrong..
    ```
    [WARNING]           bbmap : Can't parse 'rpkm' -- implementation missing
    [WARNING]           bbmap : Can't parse 'rpkm' -- implementation missing
    [WARNING]           bbmap : Can't parse 'rpkm' -- implementation missing
    [WARNING]           bbmap : Can't parse 'bincov' -- implementation missing
    [WARNING]           bbmap : Can't parse 'bincov' -- implementation missing
    [WARNING]           bbmap : Can't parse 'bincov' -- implementation missing
    [WARNING]           bbmap : Can't parse 'statsfile' -- implementation missing
    [WARNING]           bbmap : Can't parse 'covstats' -- implementation missing
    [WARNING]           bbmap : Can't parse 'covstats' -- implementation missing
    [WARNING]           bbmap : Can't parse 'covstats' -- implementation missing
    ```
2. I'd prefer not to have the commands in the section headers (in brackets)
   * I think this makes the nav harder to read / use. It's easy to add this info to the description under each header.
3. It looks like the help text isn't yet written for any of the sections:
    * image
4. It'd be good to add `.idhist`, `.gchist` and other suffixes to the [`fn_clean_exts`](https://github.com/ewels/MultiQC/blob/master/multiqc/utils/config_defaults.yaml#L96) config option to have nicer sample names
5. Plots are generally missing axis labels, and where they are there they're often not very descriptive.
    * eg. Please try to avoid `Counts` where possible (I know I'm being hypocritical here).
    * Plots can be exported, so try to assume that people may view them without any context of the report.
6. Please use the [`smooth_points `](http://multiqc.info/docs/#line-graphs) plot config option to limit the number of data points in the larger line graphs
    * Having 22k + data points for lots of samples makes the report filesize very big, very quickly
    * I usually do a bit of trial and error to get it down to a minimum before distorting the shape of the plot. Usually 200-500 points is a good maximum.
7. Some histograms have very long tails making them a bit difficult to read.
   * See modules such as [Qualimap](https://github.com/ewels/MultiQC/blob/3d0ae8881957152ea0015f2b9e44042a006f82fe/multiqc/modules/qualimap/QM_BamQC.py#L225-L237) and [Picard WGS](https://github.com/ewels/MultiQC/blob/90112bc99a5a23da2e83bc08f7f7e07159e8dc95/multiqc/modules/picard/WgsMetrics.py#L141-L157) for examples of where others have also done this.
8. Tables would benefit from having better column hover-text describing what they are showing
   * eg. What is `PercentOfPairs` on the insert size table?
   * Columns should always describe units where applicable.
9. The _Base quality_ plot seems to show the wrong data for `Read 2 average scores`
10. Please try to avoid having multiple lines per sample in a line plot. This doesn't scale beyond a handful of samples
    * eg. the different quartiles in the _Base quality_ plot
    * Also the _Base composition_ plot. I have used a custom-written heatmap for this in the FastQC module, but that was before there was a core heatmap plotting function. It should be possible to emulate it using that now.
    * Through all modules in MultiQC, I sacrifice amount of data shown for easy interpretation. So if you can't summarise a trend using a single line then it shouldn't really be in the report. The initial idea for MultiQC still holds true - the idea is to highlight outliers so that they can be followed up in the original lots / plots / reports, not to emulate all data.

Generally it looks great though - the report is huge so kudos for what must have been a lot of work! Most of the above is just polishing, I'm pretty strict when it comes to this stuff as I strongly believe that it's this small stuff that makes MultiQC what it is. But hopefully none of it is too onerous a task to complete.

Please let me know if I can help or if you have any queries!

Phil

boulund:
Ok. So I think I've made some kind of progress:

1. `log.warning` messages replaced with `log.debug`
2. No more command names in section headers. Instead moved to description text.
3. Removed help text boxes for sections without help text. Plan to write help text later as I start using the different plot types. 
4. BBMap specific suffixes added to `fn_clean_exts`
5. Decent axis labels added to all plots (I think)
6. Smoothed the heavier plots
7. Improved the long tails of some extreme histograms. 
8. Added better column headers for the columns I know what they mean. There isn't a lot of documentation about these outputs from BBMap, so I actually have no clue what `PercentOfPairs` actually means :D. I hope to fix this in future updates to the module as I figure out what things are. 
9. Not sure if there is an error here, does it still look wrong to you?
10. Improved some of the plot types to reduce unecessary lines. Still some work left to do on the `bhist` plot type to get a useful reduction of data here, so I deferred to tabbed line plots until I have time to sit down and fix this properly.

Issue #581, opened 2017-09-08
Respecting sample order in barplots


dpryan79:
This is a non-module feature request, so I hope you won't mind if I dispense with the templates.

When creating a bargraph plot one can request a specific order of the categories within each sample. That's sufficient when the categories are all grouped together by sample. However, as is already the case in the fastq_screen module, when one instead wants to nest samples in the categories then it'd be preferable to either pass a `samples` list or, better yet, just use an `OrderedDict()` and have MultiQC respect that. As is, MultiQC is [sorting things](https://github.com/ewels/MultiQC/blob/master/multiqc/plots/bargraph.py#L98).

Below is an example demonstrating my particular use case:

```python
OrderedDict([
    ('X-intercept', {'sample1': 0.384324723042, 'sample2': 0.0105007515448}),
    ('Synthetic X-intercept', {'sample1': 0.0, 'sample2': 0.0}),
    ('Elbow Point', {'sample1': 0.80857178645, 'sample2': 0.80652591438}),
    ('Synthetic Elbow Point', {'sample1': 0.498885480302, 'sample2': 0.501149702022}),
    ...])
```

That's a snippet of what one gets from `plotFingerprint --outQualityMetrics` in deepTools, which is what I happen to be creating a module for at the moment. If I use that with `bargraph.plot()`, then it's no longer the case that the synthetic and observed metrics are next to each other. Rather, the categories are just sorted:

![image](https://user-images.githubusercontent.com/6399000/30212100-ba9f9530-94a3-11e7-9045-622a954a3b2d.png)

I would propose checking for an `OrderedDict()` in `bargraph.py` and not sorting the list of keys if one is used. In python 3.6 it happens that all dictionaries are ordered, but since (A) that isn't guaranteed to be the case in python 3.7 (though I would bet it will be) and (B) I'm not sure how many people really know that, it would suggest omitting any python version checks for the time being.

My PR for a deepTools module contains one possible implementation of this.

ewels:
Hi @dpryan79,

Am I right in thinking that this is now implemented and can be closed since your deepTools PR was merged?

Phil

dpryan79:
Yup, closing.

PR #582, opened 2017-09-08
Add a deepTools module

dpryan79:
This adds support for a number of outputs produced by deepTools (xref https://github.com/fidelram/deepTools/issues/554). Note that it also implements a possible solution to issue #581, so either the code related to that in this PR needs to be removed or it needs to be resolved that the solution provided here is OK before merging.

As an aside, my goal with this module is less to make deepTools' default output interactive and more to merge information from multiple runs into a single very convenient format. I know there has been work done by others to allow MultiQC to handle the output from things like `plotPCA`, but in deepTools 2.6 I've already added interactive images (via plotly), so foresee adding a multiQC module for that as adding much that's really useful. At least in our use-cases, we often run things like `plotFingerprint` multiple times with various subsets of samples so that we can get appropriate jensen-shannon divergence values, so combining all of them (and the other produced metrics) into a single interactive plot would be immensely helpful for us.

### DIFF
--- a/multiqc/modules/deeptools/bamPEFragmentSize.py
+++ b/multiqc/modules/deeptools/bamPEFragmentSize.py
@@ -17,7 +17,7 @@ class bamPEFragmentSizeMixin():
         """Find bamPEFragmentSize output. Only the output from --table is supported."""
         self.deeptools_bamPEFragmentSize = dict()
         for f in self.find_log_files('deeptools/bamPEFragmentSize'):
-            parsed_data = self.parseBamPEFile(f['f'], f['fn'])
+            parsed_data = self.parseBamPEFile(f)
             for k, v in parsed_data.items():
                 if k in self.deeptools_bamPEFragmentSize:
                     log.warning("Replacing duplicate sample {}.".format(k))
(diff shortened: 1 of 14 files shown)


dpryan79:
N.B., I'm happy to add test data to the MultiQC_TestData repository, let me know if that should be done first. Otherwise I'll make a PR for that after this is merged.

multiqc/modules/deeptools/bamPEFragmentSize.py
```diff
@@ -0,0 +1,173 @@
...
+            headersSE["Read Length 1st Quartile"] = {'title': 'Read Len. 1st Qu.', 'description': '1st quartile read length'}
+            headersSE["Read Length Mean"] = {'title': 'Read Len. Mean', 'description': 'Mean read length'}
+            headersSE["Read Length Median"] = {'title': 'Read Len. Median', 'description': 'Median read length'}
+            headersSE["Read Length 3rd Quartile"] = {'title': 'Read Len. 3rd Qu.', 'description': '3rd quartile read length'}
+            headersSE["Read Length Maximum"] = {'title': 'Read Len. Max.', 'description': 'Maximum read length'}
+            headersSE["Read Length Std. Dev."] = {'title': 'Read Len. Std. Dev.', 'description': 'read length standard deviation'}
+            headersSE["Read Length Med. Abs. Dev."] = {'title': 'Read Len. MAD', 'description': 'read length median absolute deviation'}
+            self.add_section(name="bamPEFragmentSize (read length metrics)",
+                             anchor="bamPEFragmentSize",
+                             plot=table.plot(self.deeptools_bamPEFragmentSize, headersSE))
```
ewels: It's nice to add a third config dict here with `namespace`:
```
config = {
    'namespace': 'deepTools bamPEFragmentSize',
}
```

Then the hover text displays as `deepTools bamPEFragmentSize: Reads Sampled` instead of just `: Reads Sampled` (I should probably update the core function to drop the `:` if no namespace is given).

multiqc/modules/deeptools/bamPEFragmentSize.py
```diff
@@ -0,0 +1,173 @@
...
+                if 'Fragment Length Minimum' in v:
+                    PE = True
+                    break
+            if PE:
+                self.add_section(name="bamPEFragmentSize (fragment length metrics)",
+                                 anchor="bamPEFragmentSize",
+                                 plot=table.plot(self.deeptools_bamPEFragmentSize, headersPE))
+
+            # Read length plot
+            config = {'data_labels': [dict(name="Read length distribution", title="Read length distribution", ylab="Read length (bases)"),
```
ewels: Please add an `id` (most visible when exporting the plot) and `title` to all plots.

multiqc/modules/deeptools/plotEnrichment.py
```diff
@@ -0,0 +1,79 @@
...
+                    if category not in dCounts:
+                        dCounts[category] = dict()
+                        dPercents[category] = dict()
+                    dCounts[category][sample] = v2['count']
+                    dPercents[category][sample] = v2['percent']
+            config = {'data_labels': [
+                          {'name': 'Counts in features', 'ylab': 'Counts in feature'},
+                          {'name': 'Percents in features', 'ylab': 'Percent of reads in feature'}],
+                      'ylab': 'Counts in feature',
+                      'stacking': None,
```
ewels: Cool! This is the only example of this feature being used that I can remember. However, I'm not sure that it can stay... It looks really cool in the demo data with two samples, but it won't scale well at all. Because of people like @FelixKrueger doing single-cell work / other examples, MultiQC reports are often generated with hundreds or thousands of samples and this has a habit of breaking everything. Having split bars where each bar is a sample is a prime example of this - bar plots break down at the best of times with large sample numbers and here we're effectively multiplying the number of samples by eight.

My suggestion would be to replace this with a line graph that has a categorical x axis. It's not ideal, but it would have one line per sample, which would scale well.

multiqc/modules/deeptools/plotFingerprint.py
```diff
@@ -0,0 +1,131 @@
...
+            d = OrderedDict()
+            categories = []
+            for sample, v in self.deeptools_plotFingerprintOutQualityMetrics.items():
+                sample = str(sample)
+                for category, v2 in v.items():
+                    if category not in d:
+                        d[category] = dict()
+                        categories.append(category)
+                    d[category][sample] = v2
+            config = dict(cpswitch=False, hide_zero_cats=False, ymin=0.0, ymax=1.0, ylab='Value', stacking=None, tt_percentages=False, tt_decimals=3)
```
ewels: Same `stacking=None` comment as above with the other plot.

multiqc/utils/search_patterns.yaml
```diff
@@ -55,6 +55,27 @@ cutadapt:
...
+    contents: '	Frag. Sampled	Frag. Len. Min.	Frag. Len. 1st. Qu.	Frag. Len. Mean	Frag. Len. Median	Frag. Len. 3rd Qu.'
+    num_lines: 1
+deeptools/estimateReadFiltering:
+    contents: 'Sample	Total Reads	Mapped Reads	Alignments in blacklisted regions	Estimated mapped reads'
+    num_lines: 1
+deeptools/plotCoverageStdout:
+    contents: 'sample	mean	std	min	25%	50%	75%	max'
+    num_lines: 1
+deeptools/plotCoverageOutRawCounts:
+    contents: "#'chr'	'start'	'end'	'"
```
ewels: ooh, is this really the most specific string we can use? I could imagine a lot of other outputs starting with the same...

multiqc/modules/deeptools/bamPEFragmentSize.py
```diff
@@ -0,0 +1,173 @@
...
+            cols = line.strip().split("\t")
+
+            if len(cols) != 37:
+                # This is not really the output from bamPEFragmentSize!
+                log.warning("{} was initially flagged as the tabular output from bamPEFragmentSize, but that seems to not be the case. Skipping...".format(fname))
+                return dict()
+
+            if cols[0] in d:
+                log.warning("Replacing duplicate sample {}.".format(cols[0]))
+            d[cols[0]] = dict()
```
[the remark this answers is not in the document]
dpryan79: The label in the file is either some user provided name or the relative path to a file given from where ever the tool was executed. I'm hesitant to strip the path, since I know sometimes people use sample names for folders and then have files with the same name in each.

dpryan79:
Thanks for the comments! I'll let you know when I've had a chance to go through everything and made changes.

multiqc/utils/search_patterns.yaml
dpryan79: Maybe I'll just add a more useful header line to the output real quick... :)

multiqc/modules/deeptools/bamPEFragmentSize.py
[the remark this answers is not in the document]
dpryan79: Ah, there's a `--fullnames` option. Ignore what I wrote, you've obviously considered that :)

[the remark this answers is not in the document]
ewels: Yes sorry, I should have added more context. There's `--fullnames` to prevent this and also [`--dirs`](http://multiqc.info/docs/#sample-names-prefixed-with-directories) to add back on the directory paths. That's what I was referring to above with using the trimmed directory name instead of the default `f['root']` of the log directory. See also the docs for [`clean_s_name`](http://multiqc.info/docs/#custom-sample-names).

multiqc/utils/search_patterns.yaml
ewels: Aha, now you're talking! 😉  I'm a big fan of adding tool name and version as a header line to all output file names. If you have the version name too then we can plug it into a future feature that grabs these automatically where possible (see https://github.com/ewels/MultiQC/pull/536)

ewels:
Great stuff, thanks!

multiqc/modules/deeptools/plotEnrichment.py
dpryan79: Hmmm, true, we happen to not use that particular tool for our single-cell stuff. I'll see if I can get a useful line plot.

multiqc/modules/deeptools/estimateReadFiltering.py
```diff
@@ -0,0 +1,94 @@
...
+                        '% Blacklisted': 100. * v['blacklisted'] / float(v['total']),
+                        '% Below MAPQ': 100. * v['mapq'] / float(v['total']),
+                        '% Missing Flags': 100. * v['required flags'] / float(v['total']),
+                        '% Forbidden Flags': 100. * v['excluded flags'] / float(v['total']),
+                        '% deepTools Dupes': 100. * v['internal dupes'] / float(v['total']),
+                        '% Duplication': 100. * v['dupes'] / float(v['total']),
+                        '% Singletons': 100. * v['singletons'] / float(v['total']),
+                        '% Strand Filtered': 100. * v['strand'] / float(v['total'])}
+
+            self.general_stats_addcols(d, header)
```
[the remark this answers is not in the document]
dpryan79: These end up not summing to 100%, but to some larger indeterminate number. The values show numbers/percentages for each filtering setting, so many alignments will get double counted.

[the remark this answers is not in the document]
ewels: Ah ok, that's a shame. In that case it may make sense to make a separate table it a deepTools section (`table.plot`). That way we keep the stats in a table but don't fill the General Stats table with loads of columns.

dpryan79:
@ewels Have a look at the new changes, I think I've addressed everything now. If you're happy with them then I'll fix the merge conflict so CI testing will run.

ewels:
Hi @dpryan79,

Thanks for your work on this! I've just had a play and changed a few things before merging, I hope this is ok! Namely:

* Order of sections (for me it's more natural now - follows chronological progress)
* Formatting of tables (decimal places, column titles, some colouring, order of cols in places, shared keys for shared minimum and maximum scales)

I'm happy with the code now so have merged. I think that the module could benefit from a few additions still, such as:

* Better descriptions and help text - how deepTools calculates numbers and how to interpret plots
    * For instance, I don't know what any of the _Fingerprint quality metrics_ categories mean. Sure, I can go and google this, but it's nice to include it in a helptext markdown string so that everything is contained in the report.
* Moving two or three key statistics into the _General Statistics_ table to line up alongside other modules
* Maybe removing some of the tabular data. Personally I don't find the min/quartile etc values particularly interesting usually. I don't usually bother with these for MultiQC, preferring to go for clarity and easy of interpretation over including all data (people can always go back to the deepTools results for more detail).

Anyway, all of these would be added bonuses only so can go into a future PR or whatever.

Please take a look over the changes that I made above though and make sure that you agree with them all. Shout if you disagree with anything, or make a new PR. Otherwise I hope to get it out in a release soon.

Thanks again!

Phil

dpryan79:
@ewels Great thanks, I'll make a mental note to address some of those in a future PR :)

ewels:
Hi @dpryan79,

In the [example data](https://github.com/ewels/MultiQC_TestData/blob/master/data/modules/deeptools/plotFingerprint_outRawCounts.txt#L1) that you put up for this module, the output had a header comment line which is used to find the files:

```
#plotFingerprint --outRawCounts
```

I'm just trying to get deeptools into the MultiQC report for our [ChIPseq pipeline](https://github.com/SciLifeLab/NGI-ChIPseq) and our output doesn't seem to have this header line. We're running the latest deepTools release (v2.5.4) - am I missing something?

Cheers,

Phil

ewels:
ps. I'd like to add support for `plotCorrelation` and `plotPCA` but I'm hoping that these files should also have header comments so that I can find them.. 😄

dpryan79:
@ewels I had hoped to have a version with that header released by now, but now that's all going to come in a version 3.0 release. This is mostly finished, so it should be there by the end of the year (if you need that functionality now then go ahead and install from the `develop` branch, though note that the options for `bamCoverage` and `bamCompare` have changed and are still in flux).

I'll create an issue to remind myself to add headers to `plotCorrelation` and `plotPCA` :)

ewels:
Aha, that explains it! Ok good, then at least I know why they weren't there.. No problem, I can wait until the new year 👍 

Thanks!

Phil

PR #589, opened 2017-09-18
WIP: new module: Illumina InterOp

matthdsm:


ewels:
So the data structure that the `table.plot` function expects is described in the [docs](http://multiqc.info/docs/#creating-a-table):

```python
data = {
    'sample 1': {
        'aligned': 23542,
        'not_aligned': 343,
    },
    'sample 2': {
        'aligned': 1275,
        'not_aligned': 7328,
    }
}
table_html = table.plot(data)
```

I'll try to get to reviewing your code and providing more help soon. In the mean time if you have specific questions (eg. _"I tried this and it didn't work"_) then shout and I'll try to get to those asap :)

Phil

matthdsm:
Thanks a lot, I'll keep playing around a bit in the mean time.

PS, don't worry about the stray commits and merges, I'll prepare a new PR with a clean history once everythings finished.

Cheers
M

ewels:
ok great, thanks! There should be quite a few other modules that do similar csv parsing.. eg. [snpeff](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/snpeff/snpeff.py#L119-L183).

It looks like your files are slightly tricky as they are essentially multiple files concatenated together. But a bit of custom code to look for specific strings should be fine. No worries if you struggle, just say and I can have a stab at writing something for you.

Phil

matthdsm:
Hi Phil,
I've finished most of the module (although I'm sure there's a lot room for improvement), but I'm still struggling with the input data structure for the summary.
How many levels of nesting can the table function handle? Can it handle a list of dicts? Or is it possible to put multiple tables in one section? I don't see any other way to properly structure the data..

Currently, I've got it the table parsed to something like this:
```
data:{
  readname:[
    {key1: val1, key2: val2},
    ....
  ]
}
```
If list handling is out of the question, the table could be parsed to a structure like this:
```
data:{
  readname:{
    Surfacename: {
     key1: val1,
     ....
    }
  }
}
```

What do you think is the best approach?

Thanks for the help.
M

ewels:
Hi @matthdsm,

Tables only have rows and columns, so displaying `readname` _and_ `Surfacename` _and_ `key1` all at the same time for each value isn't possible. The supported data structures support this, so MultiQC expects a two dimensions - key (sample name) and then key: value pairs. Most of the plots and other core functions work in a similar way.

Looking quickly at the `.csv` files, they seem to be several tables concatenated together. We can add multiple tables to each section, but I'm not super excited about dumping loads of tables into MultiQC reports as this sort of goes against the ethos of what MultiQC tries to do. Instead, I'd prefer to summarise this data somehow - basically reduce the number of dimensions I guess, or find key statistics to put into a plot. Does anything jump to mind along these lines?

Phil

ewels:
Hi @matthdsm,

I changed the branch for this PR to a dev branch and have started working on your code. I think it's going to be a case of reducing the amount of data used basically. I've got the first table working with just `Total` values from the summary statistics. eg: [multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/1359890/multiqc_report.html.zip)

When I get more time I'll come back to this and continue similar work with the rest of the code. What you've done already is great - the parsing is really nice, it's just the output that needs some tweaking.

If you're curious, you can see what I've done so far in f8456a4cb7b458bb2bd8560bf21d5092aa784fc2

Phil

matthdsm:
Hi Phil,

Looking good! Thanks a lot for the work you're putting into this! I was a little lost at how to proceed.
Would it be ok if I fiddle around some more with what you have, now that I've got a solid base to work from? I'll create a new PR if I manage to produce something workable.

Cheers!
M

matthdsm:
Hi again,

Just a quick question now. What is your stance on combining data in a singe cell? I suppose the biggest problem with the summary detail table can be fixed by merging the `Lane` and `Read` columns into one. That way it's easier to get all the data into the table, without nesting.

Cheers
M

matthdsm:
If we "ignore" the possibility of multiple samples, I've got this.

[multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/1362225/multiqc_report.html.zip)

ewels:
Hi @matthdsm,

Looking good! I suggest that we make a concatenated key as you've done here, but also with sample name. Then have columns for Read and Lane so that the table can be sorted by those. For example:

```json
"Sample 1 - R1L1": {"Read": "Read 1", "Lane": "Lane 1", "Aligned": 0.1},
"Sample 1 - R2L1": {"Read": "Read 2(I)", "Lane": "Lane 1", "Aligned": 0.2},
"Sample 1 - R1L2": {"Read": "Read 1", "Lane": "Lane 2", "Aligned": 0.3},
"Sample 1 - R2L2": {"Read": "Read 2(I)", "Lane": "Lane 2", "Aligned": 0.4},
```

The same can be done for _Indexing QC Metrics summary_ (without the read info).

There's a bunch of tidying that we can do on the output once everything is in place, but we can get to that later.

Phil

matthdsm:
I think this should do it then.
Only thing left is Indexing QC Metrics details, which doensn't show all the data yet.
[multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/1362423/multiqc_report.html.zip)

matthdsm:
More like this then?
[multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/1362632/multiqc_report.html.zip)

I removed the +- errors, split the Phas/Prehas and hid away some columns by default.
The only thing thats bothering me is the rounding of the values in the Phasing and Prephasing column, those should show the complete value, but don't for some reason.

1 more table to go!
M

matthdsm:
I think I've got a somewhat finished product.
I'll leave all the fancy tricks up to you. I'm sure you're waaaaaaay more proficient with all the special features of MultiQC.
[multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/1362736/multiqc_report.html.zip)

Cheers
M

ewels:
Thanks! I'm away next week but will try to take a look when I'm back..

Phil

Nitin123-4:
Can we also add here more plots like https://github.com/MultiQC/MultiQC_SAV did (This doesn't work for NovaseqXPlus)

PR #598, opened 2017-09-28
Update links for bcl2fastq modules

juliangehring:
- Fixes the broken link for bcl2fastq2
- Uses the overview page for both bcl2fastq and bcl2fastq2


ewels:
Thanks again! 👍

PR #602, opened 2017-10-02
add Dockerfile for MultiQC 1.1

MaxUlysse:
Add a Dockerfile based on `python:2.7-slim`.
Works well for MultiQC 1.1, haven't tried 1.2 or 1.3-dev, but I can see no reason why it would not work either.
I'll advise to make a tag for each version, so that people will still be able to try older versions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+FROM python:2.7-slim
+
+LABEL \
+  author="Phil Ewels" \
+  description="MultiQC" \
+  maintainer="phil.ewels@scilifelab.se"
+
+# Setup ENV variables
+ENV MULTIQC_VERSION="1.1"
+
+# Install libraries
+RUN \
+  apt-get update && apt-get install -y --no-install-recommends \
+  g++ \
+  git \
+  wget \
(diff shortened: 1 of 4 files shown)


ewels:
Brilliant, thanks @MaxUlysse!

Is it worth just installing from the master branch instead? The automatic tagging upon release on dockerhub should then give fixed versions for each version as you say. Then the `:latest` tag will always have the last commit.

It'd also be great if you could add yourself to the list of contributors on the readme and also add in a section to the documentation describing the image with a paragraph explaining (briefly) how to use it.

Phil

MaxUlysse:
OK, I'll do that right away then ;-)

MaxUlysse:
I added myself in the readme.
And I sorted the lists of contributors and documentations.
Made a small paragraph in the usage documentation (I supposed that you will be using `ewels/multiqc`).
And updatated the Dockerfile wich compiles for current `1.3-dev`.

ewels:
Thanks! 🎉 

https://hub.docker.com/r/ewels/multiqc/builds/

PR #605, opened 2017-10-06
Interop: module tweaks, now fully functional

matthdsm:
I think this should do the trick.

Thanks for all the help!
Cheers
M

### DIFF
--- a/multiqc/modules/interop/interop.py
+++ b/multiqc/modules/interop/interop.py
@@ -237,19 +237,23 @@ class MultiqcModule(BaseMultiqcModule):
             'description': 'The percentage of clusters passing filtering, +/- 1 standard deviation.',
             'suffix': '%',
         }
-        # headers['Phas/Prephas'] = {
-        #     'title': 'Phas/Prephas (%)',
-        #     'description': 'The value used by RTA for the percentage of molecules in a cluster for which sequencing falls behind (phasing) or jumps ahead (prephasing) the current cycle within a read.'
-        # }
         headers['Phased'] = {
             'title': 'Phased (%)',
             'description': 'The value used by RTA for the percentage of molecules in a cluster for which sequencing falls behind (phasing) or jumps ahead (prephasing) the current cycle within a read.',
+            'format': '{:.,2f}',
+            'min': 0,
+            'max': 100,
             'suffix': '%',
+            'scale': 'OrRd'
         }
         headers['Prephased'] = {
             'title': 'Prephased (%)',
             'description': 'The value used by RTA for the percentage of molecules in a cluster for which sequencing falls behind (phasing) or jumps ahead (prephasing) the current cycle within a read.',
+            'format': '{:.,2f}',
+            'min': 0,
+            'max': 100,
             'suffix': '%',
+            'scale': 'OrRd'
         }
         headers['Reads'] = {
             'title': 'Reads',
@@ -407,6 +411,7 @@ class MultiqcModule(BaseMultiqcModule):
             'table_title': 'Index Read Statistics Details',
             'col1_header': 'Run - Sample - Lane',
         }
+
         tdata = {}
         for s_name in data:
             for key in data[s_name]['details']:
--- a/multiqc/modules/interop/interop.py
+++ b/multiqc/modules/interop/interop.py
@@ -370,12 +370,12 @@ class MultiqcModule(BaseMultiqcModule):
         }
         headers['Min'] = {
             'title': 'Min',
-            'description': 'The lowest representation for any index.'
+            'description': 'The lowest representation for any index.',
             'format': '{:.,2f}',
         }
         headers['Max'] = {
             'title': 'Max',
-            'description': 'The highest representation for any index.'
+            'description': 'The highest representation for any index.',
             'format': '{:.,2f}',
         }
         table_config = {
--- a/multiqc/modules/interop/interop.py
+++ b/multiqc/modules/interop/interop.py
@@ -170,9 +170,10 @@ class MultiqcModule(BaseMultiqcModule):
 
         headers = OrderedDict()
         headers['Yield'] = {
-            'title': 'Yield',
-            'description': 'The number of bases sequenced.',
-            'scale': 'PuOr'
+            'title': '{} Bp Yield'.format(config.base_count_prefix),
+            'description': 'The number of bases sequenced ({})'.format(config.base_count_desc),
+            'scale': 'PuOr',
+            'shared_key': 'base_count'
         }
         headers['Aligned'] = {
             'title': 'Aligned (%)',
@@ -192,7 +193,7 @@ class MultiqcModule(BaseMultiqcModule):
         }
         headers['Intensity C1'] = {
             'title': 'Intensity Cycle 1',
-            'description': ''
+            'description': 'The intensity statistic at cycle 1.',
         }
         headers['%>=Q30'] = {
             'title': '% >= Q30',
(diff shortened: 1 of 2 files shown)


matthdsm:
For good measure, here's a preview of the output.
[multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/1362762/multiqc_report.html.zip)

Cheers
M

multiqc/modules/interop/interop.py
```diff
@@ -195,137 +206,198 @@ def run_metrics_summary_table(self, data):
...
         }
         headers['% Read Identified (PF)'] = {
             'title': '% Reads Identified (PF)',
-            'description': 'The total fraction of passing filter reads assigned to an index.'
+            'description': 'The total fraction of passing filter reads assigned to an index.',
+            'suffix': '%',
         }
         headers['CV'] = {
             'title': 'CV',
             'description': 'The coefficient of variation for the number of counts across all indexes.'
```
ewels: You're missing a comma on these lines, which is triggering the syntax error failures on Travis. Just add a comma at the end of the `'description'` and remove it after `'format'`. eg:

```python
headers['CV'] = {
    'title': 'CV',
    'description': 'The coefficient of variation for the number of counts across all indexes.',
    'format': '{:.,2f}'
}
```

ewels:
@matthdsm - let me know when Travis is passing again (don't worry about AppVeyor for now) and I'll take a look at this for you.

Phil

matthdsm:
Hi Phil,

Thanks! Will the Travis build be triggered automatically? I don't seem to have sufficient rights to trigger a new one.

Cheers
M

ewels:
Yes - Travis will automatically fire again if you push a new commit.

multiqc/modules/interop/interop.py
```diff
@@ -195,137 +206,198 @@ def run_metrics_summary_table(self, data):
...
+            'description': 'The number of clusters (in millions) passing filtering.',
+            'suffix': 'M',
         }
         headers['Cycles Error'] = {
             'title': 'Cycles Error',
             'description': 'The number of cycles that have been error-rated using PhiX, starting at cycle 1.'
         }
+        headers['Yield'] = {
+            'title': 'Yield',
+            'description': 'The number of bases sequenced which passed filter.',
```
ewels: Please use the following config for consistency with other modules:

```python
headers['Yield'] = {
    'title': '{} Bp Yield'.format(config.base_count_prefix),
    'description': 'The number of bases sequenced which passed filter ({})'.format(config.base_count_desc),
    'scale': 'PuOr',
    'shared_key': 'base_count'
}
```

The shared_key means that the scale bars will be shared with other read counts. This also automatically applies the lambda and allows the user to configure the count modifier.

(Note that this is base count instead of read count as below).

multiqc/modules/interop/interop.py
```diff
@@ -195,137 +206,198 @@ def run_metrics_summary_table(self, data):
...
+            'min': 0,
+            'max': 100,
+            'suffix': '%',
+            'scale': 'OrRd'
         }
         headers['Reads'] = {
             'title': 'Reads',
-            'description': 'The number of clusters (in millions).'
+            'description': 'The number of clusters (in millions).',
+            'suffix': 'M',
```
ewels: Please see the other comment below for how to handle table columns with read counts.

multiqc/modules/interop/interop.py
```diff
@@ -195,137 +206,198 @@ def run_metrics_summary_table(self, data):
...
-        }
-        headers['%>=Q30'] = {
-            'title': '%>=Q30',
-            'description': 'The percentage of bases with a quality score of 30 or higher, respectively.'
-        }
-        headers['Yield'] = {
-            'title': 'Yield',
-            'description': 'The number of bases sequenced which passed filter.'
+            'description': 'The number of clusters (in millions) passing filtering.',
+            'suffix': 'M',
```
ewels: Please see the other comment below for how to handle table columns with read counts.

multiqc/modules/interop/interop.py
```diff
@@ -195,137 +206,198 @@ def run_metrics_summary_table(self, data):
...
         }
         headers['Cycles Error'] = {
             'title': 'Cycles Error',
             'description': 'The number of cycles that have been error-rated using PhiX, starting at cycle 1.'
```
ewels: This is an `int` count right? Would be nice to strip decimal places: `'format': '{:.,0f}',`

multiqc/modules/interop/interop.py
```diff
@@ -195,137 +206,198 @@ def run_metrics_summary_table(self, data):
...
+            'min': 0,
+            'max': 100,
+            'suffix': '%',
+            'scale': 'OrRd',
+            'hidden': True
         }
         headers['Intensity C1'] = {
             'title': 'Intensity Cycle 1',
             'description': 'The corresponding intensity statistic at cycle 20 as a percentage of that value at the first cycle.',
             'help': 'equation: 100%x(Intensity at cycle 20)/(Intensity at cycle 1).'
```
ewels: I don't think that `help` does anything?

multiqc/modules/interop/interop.py
```diff
@@ -195,137 +206,198 @@ def run_metrics_summary_table(self, data):
...
         headers['Cluster PF'] = {
             'title': 'Cluster PF (%)',
-            'description': 'The percentage of clusters passing filtering, +/- 1 standard deviation.'
+            'description': 'The percentage of clusters passing filtering, +/- 1 standard deviation.',
+            'suffix': '%',
+        }
+        headers['Phased'] = {
+            'title': 'Phased (%)',
+            'description': 'The value used by RTA for the percentage of molecules in a cluster for which sequencing falls behind (phasing) or jumps ahead (prephasing) the current cycle within a read.',
+            'format': '{:.,2f}',
```
ewels: I think that this is the default and can be removed.

multiqc/modules/interop/interop.py
ewels: Also, I think that the description may be wrong - it looks to me like it should be _"The average of the A channel intensity measured at the first cycle averaged over filtered clusters."_

ewels: I guess this text is taken from _% Intensity Cycle 20_ which isn't here?

ewels: Also the other column with the same title ([here](https://github.com/matthdsm/MultiQC/blob/9435ac88ec3662931a5474f8ce33dce04d226c7d/multiqc/modules/interop/interop.py#L193-L196)) has an empty description...

multiqc/modules/interop/interop.py
```diff
@@ -195,137 +206,198 @@ def run_metrics_summary_table(self, data):
...
             'title': '% Reads Identified (PF)',
-            'description': 'The total fraction of passing filter reads assigned to an index.'
+            'description': 'The total fraction of passing filter reads assigned to an index.',
+            'suffix': '%',
         }
         headers['CV'] = {
             'title': 'CV',
-            'description': 'The coefficient of variation for the number of counts across all indexes.'
+            'description': 'The coefficient of variation for the number of counts across all indexes.',
+            'format': '{:.,2f}',
```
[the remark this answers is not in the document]
matthdsm: This one needs to stay. If I remove the formatting, the values gets rounded to 1 decimal and I'd like to see more (all)

multiqc/modules/interop/interop.py
matthdsm: Right you are!

matthdsm:
Hi Phil,

I pushed another commit. I hope these are the changes you're after.

Cheers
M

ewels:
Looks brilliant, thanks!

ewels:
Ok, and merged to master. Thanks again @matthdsm - will be out in the v1.3 release very shortly! 🎉 

Phil

Issue #470, opened 2017-06-01
Formatting when read_count_multiplier=1


vladsaveliev:
With `config.read_count_multiplier: 1`, reports still show 2 decimal digits (`1 234 567.00`). It would be nicer to show integer values just as `1 234 567`. Also, need to make sure that empty `config.read_count_desc` will not make the descriptions look ugly, for instance `'Total raw sequences ({})'` ends up with empty parenthesis.

A quick solution would be to look through all modules that report read numbers, and add a piece of code like 
```python
read_format = '{:,.2f} ' + config.read_count_prefix
if config.read_count_multiplier == 1:
    read_format = '{:,.0f}'
```

But it doesn't look very nice because it requires to copy this bit over and over, and does not guarantee that a developer of a new module will remember to add this. 

What do you think about this, is there any nice solution? Maybe a flag like `is_reads_field` that will make multiqc implicitly fix 'format' and `description` fields according to `read_count_multiplier`?

ewels:
Yes, when I was about halfway through my copy+paste marathon for `config.read_count_multiplier` I had the same thought - that it would be better to just add a "special case" for read counts and have the core table code set the rest of the config values automatically.

As you say, perhaps we can add a new config option (maybe `read_counts: True`?) and then set the other values accordingly.. Something like:

* `title`: prepend `config.read_count_prefix`
* `description`: add `({})` suffix
* `min`: set to `0`
* `shared_key`: set to `read_count`
* `modify`: set to `lambda x: x * config.read_count_multiplier`
* `format`: as suggested above - `{:,.2f}` or `{:,.0f}`

This also future-proofs it a little if we want to add more custom stuff in the future.

Phil

ewels:
ps. Maybe no need for a new field (`is_reads_field`) - could just see if `shared_key` == 'reads'`?

Would probably need some extra checks to avoid duplicating `title` and `description` strings for outdated modules though..

Phil

ewels:
This is now done and merged - thanks!

PR #607, opened 2017-10-11
Automatically modify read or base counts according shared_key set

vladsaveliev:
Will make the read and base counts in custom content affected by modifiers. 

The problem is that we can't assign lambdas in YAML, thus can't set proper `modify` value for custom content. Making use of `shared_key` to automatically set the modification would be nice, though the downside is that it makes it implicit and also breaks compatibility, which is probably not desired. So rather opening to start discussion.

### DIFF
--- a/multiqc/modules/qualimap/QM_BamQC.py
+++ b/multiqc/modules/qualimap/QM_BamQC.py
@@ -505,7 +505,6 @@ def general_stats_headers (self):
         'title': 'Ins. size',
         'description': 'Median insert size',
         'min': 0,
-        'suffix': 'bp',
         'scale': 'PuOr',
         'format': '{:,.0f}'
     }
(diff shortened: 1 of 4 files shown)


vladsaveliev:
Just remembered that I already created a related issue https://github.com/ewels/MultiQC/issues/470! So can continue moving from here.

ewels:
I think this looks good and I'm happy to merge as a first step if you don't intend to add anything else imminently.

What do you think about setting `min` and `format` automatically in the same blocks if they're not already set?

vladsaveliev:
Absolutely. Happy to add anything you think is reasonable. I guess everything from your list?
* `title`: prepend `config.read_count_prefix`
* `description`: add `({})` suffix
* `min`: set to `0`
* `modify`: set to `lambda x: x * config.read_count_multiplier`
* `format`: `'{:,.2f} ' + config.read_count_prefix if config.read_count_multiplier != 1 else read_format = '{:,.0f}'`

ewels:
Yeah - though `title` `description` are a little more controversial as we're adding to something that's already set. I was thinking that for the others we'd only set if not already set (so easy to override). But as long as we go through all of the core modules updating them so that the new code works it should be fine (a bit of clever regex copy and paste perhaps?).

Phil

ewels:
ps. I reran the failed travis job and it worked this time 👍

vladsaveliev:
Actually cannot agree more with the concerns on `title`, since in my reports I add `M` as a `suffix` rather than to the header. Often it appears to be more clear, but primarily it allows to save horizontal space since the column header is usually wider than the values.

vladsaveliev:
Added `min` and `format`, moved that code to a more proper location (before the header key defaults are set), and modified modules `qualimap` and `bcl2fastq` accordingly for testing and to start cleaning up modules. Can be merged as a first step.

multiqc/modules/qualimap/QM_BamQC.py
```diff
@@ -505,7 +505,6 @@ def general_stats_headers (self):
         'title': 'Ins. size',
         'description': 'Median insert size',
         'min': 0,
-        'suffix': 'bp',
```
ewels: Sorry - tiny thing: removing this is fine, but please add back to the description in that case so that the info is still there

ewels:
Brilliant stuff, thanks!

PR #608, opened 2017-10-11
Allow to config modules running multiple times, and add Bcftools options

vladsaveliev:
Add `bcftools` options `write_general_stats` and `write_separate_table` to control where the table stats go (useful when `bcftools` running multiple times)

So for bcbio MultiQC report, I wanted to run `bcftools` twice once for germline and once for somatic variants (https://github.com/chapmanb/bcbio-nextgen/issues/2081). The new MultiQC feature that allowed to run modules twice is really great, however since the `bcftools` module reports at least 5 useful metrics, making it twice makes the general stats table expand beyond the screen width. So I wanted to hide the "germline" stats by default, and rather move them to a separate table. I added a parameter for `bcftools` that controls where the stats would go (the first commit), which works nicely, however `module_order` doesn't allow user to set any custom config parameters specifically for a run. This PR tries to address that. If any custom value outside of default `name/anchor/target/href/info/comment/extra` is set in `module_order`, it will affect `config.<module_anchor>` dict, e.g.:
```
module_order:
- bcftools:
       name: 'Bcftools (somatic)',
       info: 'Bcftools stats for somatic varaiant calls only.'
       path_filters: ['*_bcftools_stats.txt']
       write_general_stats: true
       write_separate_table: true
- bcftools:
       name: 'Bcftools (germline)'
       info: 'Bcftools stats for germline variant calls only.'
       path_filters: ['*_bcftools_stats_germline.txt']
       write_general_stats: false
       write_separate_table: true
```

Let me know if you would implement it in other way and I'll fix it. For instance, one way would be to contain values in an explicit parameter `custom_config` provided to the `module_order` item, e.g.:
```
module_order:
- bcftools:
       name: 'Bcftools (germline)'
       info: 'Bcftools stats for germline variant calls only.'
       path_filters: ['*_bcftools_stats_germline.txt']
       custom_config:
            write_general_stats: false
            write_separate_table: true
```
More verbose, but less error-prone. Sticking to the first version however.

### DIFF
--- a/multiqc/modules/bcftools/bcftools.py
+++ b/multiqc/modules/bcftools/bcftools.py
@@ -25,7 +25,7 @@ class MultiqcModule(BaseMultiqcModule, StatsReportMixin):
         # Initialise the parent object
         super(MultiqcModule, self).__init__(
             name='Bcftools',
-            anchor='Bcftools', target='Bcftools',
+            anchor='bcftools', target='Bcftools',
             href='https://samtools.github.io/bcftools/',
             info=(" contains utilities for variant calling and manipulating VCFs and BCFs."))
 
--- a/multiqc/modules/bcftools/stats.py
+++ b/multiqc/modules/bcftools/stats.py
@@ -5,7 +5,7 @@
 import logging
 from collections import OrderedDict
 from multiqc import config
-from multiqc.plots import bargraph, linegraph
+from multiqc.plots import bargraph, linegraph, table
 
 # Initialise the logger
 log = logging.getLogger(__name__)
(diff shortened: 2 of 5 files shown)


vladsaveliev:
Hi Phil, just poking you about this PR :)

ewels:
Hi @vladsaveliev,

Sorry to be so slow in getting to this for you.. It looks great! The second variant with a nested dict is my preferred method though - I can imagine people getting really confused if there's the possibility of keys overlapping, or even just being in the same tier and not doing similar things.. Is there a reason that you stuck with version 1?

Phil

multiqc/modules/bcftools/bcftools.py
```diff
@@ -25,7 +25,7 @@ def __init__(self):
         # Initialise the parent object
         super(MultiqcModule, self).__init__(
             name='Bcftools',
-            anchor='Bcftools', target='Bcftools',
+            anchor='bcftools', target='Bcftools',
```
ewels: 👍

multiqc/modules/bcftools/stats.py
```diff
@@ -109,8 +109,15 @@ def parse_bcftools_stats(self):
             # Write parsed report data to a file
             self.write_data_file(self.bcftools_stats, 'multiqc_bcftools_stats')
 
-            # General Stats Table
-            self.bcftools_stats_genstats_table()
+            # Stats Table
+            stats_headers = self.bcftools_stats_genstats_headers()
```
ewels: Very nice method 👍

vladsaveliev:
The reason to stick with the option 1 was that when I tried to figure how to pass an option to bcftools section, that was my first guess - I tried to put that into the module_order item and it didn't work. Only yhen I looked up into the code and figured that all those `target` and `info` are BaseModule constructor parameters, not configurable options :) So my guess was the option 1 is natural and users might want start using it without looking into the docs. 

But I agree with your point too - even if it simplifies usage, the possibility of confusion and errors is a too high price for that. So just pushed another commit moving other options into `custom_config`, also added a line into the docs.

ewels:
Ok fantastic, thanks for the change 👍 Great new feature!

PR #610, opened 2017-10-12
HiCExplorer

joachimwolff:
This PR adds a MultiQC module for the quality log files of the [HiCExplorer](http://hicexplorer.readthedocs.io).

deepinscreenshot_20171012183437

Thanks for considering!

Ping: @fidelram @vivekbhr @bgruening


joachimwolff:
I am not sure if the fail for Python 3.5 and 3.6 on AppVeyor is my fault. Does someone know this?

ewels:
Hi @joachimwolff,

Thanks for this - it looks great! I'm away at the moment and pretty busy next week, but will try to review asap.

The test failure is just AppVeyor being rubbish, don't worry about that. It's been failing sporadically for weeks now, I think I'll probably remove those tests.

One thing I can see from the screenshot on my phone - would be great if the category descriptions could be moved into `helptext` instead of `description`.

Thanks!

Phil

Phil

multiqc/modules/hicexplorer/hicexplorer.py
```diff
@@ -0,0 +1,232 @@
...
+            'title': '% Used pairs',
+            'max': 100,
+            'min': 0,
+            'modify': lambda x: x * 100
+        }
+        headers['Unmapped'] = {
+            'title': '% Unmapped',
+            'max': 100,
+            'min': 0,
+            'modify': lambda x: x * 100
```
ewels: I like to give colour schemes that reflect the numbers in tables, where possible. For example, high numbers of `Unmapped`  is bad, so `OrRd` may be good to highlight big numbers with a red colour. See the palette names in the docs here: http://multiqc.info/docs/#step-3-adding-to-the-general-statistics-table

Note that for consistency with other modules, I'd actually prefer to have `% Mapped` instead of `Unmapped` here unless you have a strong preference. Should be easy to switch using the lambda.

multiqc/modules/hicexplorer/hicexplorer.py
```diff
@@ -0,0 +1,232 @@
...
+                'Unmapped': self.mod_data[file]['One mate unmapped'][0] / total_pairs,
+                'Min rest. site distance': self.mod_data[file]['Min rest. site distance'][0],
+                'Max rest. site distance': self.mod_data[file]['Max rest. site distance'][0],
+            }
+            data[self.mod_data[file]['File'][0]] = data_
+        headers = OrderedDict()
+
+        headers['Pairs considered'] = {
+            'title': 'Number of pairs',
+            'modify': lambda x: int(x)
```
ewels: Do you need the `int()` here? Nice to add `'format': '{:,.0f}'` to cut off decimal places if you're using integers too.

Better still is to use the MultiQC config behaviour to handle read counts here though, so that users can customise how to display the count. See the docs here: http://multiqc.info/docs/#number-base-multiplier

Also, please use `'shared_key': 'read_count'` to make the scale bars consistent with counts from other modules (eg. FastQC etc).

ewels: NB: Depending on timing, the number formatting stuff may be handled automatically with the shared key - see currently open PR https://github.com/ewels/MultiQC/pull/607

multiqc/modules/hicexplorer/hicexplorer.py
```diff
@@ -0,0 +1,232 @@
...
+        keys['One mate unmapped'] = {'color': self.colors[1], 'name': 'One mate unmapped'}
+        keys['One mate not unique'] = {'color': self.colors[2], 'name': 'One mate not unique'}
+        keys['One mate low quality'] = {'color': self.colors[3], 'name': 'One mate low quality'}
+        keys['Dangling end'] = {'color': self.colors[4], 'name': 'Dangling end'}
+
+        keys['Self ligation'] = {'color': self.colors[5], 'name': 'Self ligation'}
+        keys['One mate not close to rest site'] = {'color': self.colors[6], 'name': 'One mate not close to rest site'}
+        keys['Same fragment'] = {'color': self.colors[7], 'name': 'Same fragment'}
+        keys['Self circle'] = {'color': self.colors[8], 'name': 'Self circle'}
+        keys['Duplicated pairs'] = {'color': self.colors[9], 'name': 'Duplicated pairs'}
```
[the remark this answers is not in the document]
ewels: Also I think you don't need to set `name` if it's the same as the key. This defaults to the key anyway.

multiqc/modules/hicexplorer/hicexplorer.py
ewels: Yeah, just need `'shared_key': 'read_count'` now (`modify` and the others will be set automatically then).

multiqc/modules/hicexplorer/hicexplorer.py
```diff
@@ -0,0 +1,232 @@
...
+            total_pairs = self.mod_data[file]['Pairs considered'][0]
+            data_ = {
+                'Pairs considered': self.mod_data[file]['Pairs considered'][0],
+
+                'Pairs used': self.mod_data[file]['Pairs used'][0] / total_pairs,
+                'Unmapped': self.mod_data[file]['One mate unmapped'][0] / total_pairs,
+                'Min rest. site distance': self.mod_data[file]['Min rest. site distance'][0],
+                'Max rest. site distance': self.mod_data[file]['Max rest. site distance'][0],
+            }
+            data[self.mod_data[file]['File'][0]] = data_
```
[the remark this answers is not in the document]
joachimwolff: I am not sure if I understand this issue. `self.mod_data[file]['File'][0]` returns a sample name which is content of the log file. The log file name has nothing to do with it. 

A log file, let's call it `foo.log`, looks like this:

```

File	SRR443883.cool		
Pairs considered	130534234		
Min rest. site distance	300		
Max rest. site distance	800		

Pairs used	44899181	(34.40)	(54.12)
One mate unmapped	47576234	(36.45)	(57.35)
One mate not unique	231996	(0.18)	(0.28)
One mate low quality	6898437	(5.28)	(8.32)
dangling end	0	(0.00)	(0.00)
self ligation (removed)	0	(0.00)	(0.00)
One mate not close to rest site	0	(0.00)	(0.00)
same fragment (800 bp)	26096500	(19.99)	(31.46)
self circle	1752780	(1.34)	(2.11)
duplicated pairs	4831886	(3.70)	(5.82)
Of pairs used:
inter chromosomal	9381021	(20.89)
short range < 20kb	3965623	(8.83)
long range	31552537	(70.27)
inward pairs	8503350	(18.94)
outward pairs	9560389	(21.29)
left pairs	8741433	(19.47)
right pairs	8712988	(19.41)


```

In this case the sample name returned by `self.mod_data[file]['File'][0]` would be `SRR443883.cool`. In the linked description it is assumed that the file name should be the sample name, but this is here not the case. 

Thanks for clarifying!

joachimwolff:
@ewels Thanks for your review. I tried to cover your concerns, only the one with `self.clean_s_name` is left, see my comment.

multiqc/modules/hicexplorer/hicexplorer.py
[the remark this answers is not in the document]
ewels: Sorry, I wasn't clear. Yes it's great that it returns the input filename, but what happens if someone runs MultiQC with the sample sample that's been run 10 times in slightly different ways? The input filename will be identical and get overwritten. This is where the `-d` flag comes for MultiQC, which prepends the directory name for where the log file was found.

There are other things that this function does too. For example, it maybe be nice to add `.h5` and `.cool` to the [file name cleaning config](https://github.com/ewels/MultiQC/blob/master/multiqc/utils/config_defaults.yaml#L94-L98) to automatically remove these file extensions. This is user-configurable.

So it's fine to continue using this value as the sample name, but it should be passed through that function before being used. In most cases it shouldn't have any effect.

multiqc/modules/hicexplorer/hicexplorer.py
```diff
@@ -0,0 +1,184 @@
...
+                                            href="https://hicexplorer.readthedocs.io",
+                                            info=" addresses the common tasks of Hi-C analysis from processing to visualization.")
+
+        self.mod_data = dict()
+        for file in self.find_log_files('hicexplorer'):
+            self.mod_data[file['s_name']] = self.parse_logs(file['f'])
+            self.add_data_source(file)
+        if len(self.mod_data) == 0:
+            raise UserWarning
+        self.colors = ["#1f77b4",
```
ewels: Is there a specific reason to use custom colours for this module?

multiqc/modules/hicexplorer/hicexplorer.py
```diff
@@ -0,0 +1,184 @@
...
+        keys_list_contact_distance = ['Inter chromosomal', 'Short range', 'Long range']
+        keys_list_read_orientation = ['Inward pairs', 'Outward pairs', 'Left pairs', 'Right pairs']
+
+        # prepare the detail report section
+        self.add_section(
+            name='Pair categorization',
+            anchor='hicexplorer_pairs_categorized',
+            plot=self.hicexplorer_create_plot(keys_list_plot_pairs, 'HiCExplorer: Categorization of reads'),
+            helptext='''
+            This figure contains the number of reads that were finally used to build the
```
ewels: This is great, but now maybe a bit too far in the other direction! It's nice to have a short one-line description above the plot always showing (just not long paragraphs). So this first sentence would be perfect as a description, then the rests as `helptext`.

multiqc/modules/hicexplorer/hicexplorer.py
```diff
@@ -0,0 +1,184 @@
...
+            Self circles are defined as pairs within 25kb with \'outward\' read orientation
+
+            #### Self ligation
+
+            These are read pairs with a restriction site in between that are within 800 bp.'''
+        )
+
+        self.add_section(
+            name='Contact distance',
+            anchor='hicexplorer_contact_distance',
```
ewels: It would be nice to have some help text here saying what constitutes "Long range" etc.

multiqc/modules/hicexplorer/hicexplorer.py
```diff
@@ -0,0 +1,184 @@
...
+            plot=self.hicexplorer_create_plot(keys_list_contact_distance, 'HiCExplorer: Categorization of reads')
```
ewels: Is there a reason to have the plot title different to the section title? (I guess this is a mistake?)

multiqc/modules/hicexplorer/hicexplorer.py
```diff
@@ -0,0 +1,184 @@
...
+                'Pairs used': self.mod_data[file]['Pairs used'][0] / total_pairs,
+                'Mapped': self.mod_data[file]['One mate unmapped'][0] / total_pairs,
+                'Min rest. site distance': self.mod_data[file]['Min rest. site distance'][0],
+                'Max rest. site distance': self.mod_data[file]['Max rest. site distance'][0],
+            }
+            data[self.mod_data[file]['File'][0]] = data_
+        headers = OrderedDict()
+
+        headers['Pairs considered'] = {
+            'title': 'M Number of pairs',
```
ewels: This is a bit messy currently, but users can configure what multiplier to use for read counts (millions, thousands etc). So this should be as follows to reflect this:

```python
'title': '{} Pairs'.format(config.read_count_prefix),
```

See http://multiqc.info/docs/#number-base-multiplier for more info.

multiqc/modules/hicexplorer/hicexplorer.py
```diff
@@ -0,0 +1,184 @@
...
+        headers['Mapped'] = {
+            'title': '% Mapped',
+            'max': 100,
+            'min': 0,
+            'modify': lambda x: (1 - x) * 100,
+            'scale': 'RdYlGn'
+        }
+
+        headers['Min rest. site distance'] = {
+            'title': 'Min rest. site distance',
```
ewels: It would be nice to make this title shorter (to give a narrower column) but with a longer tooltip description. Maybe something like this?

```python
'title': 'Min RE dist',
'descripton': 'Minimum restriction site distance (bp)'
```

multiqc/modules/hicexplorer/hicexplorer.py
```diff
@@ -0,0 +1,184 @@
...
+        headers['Pairs considered'] = {
+            'title': 'M Number of pairs',
+            'shared_key': 'read_count'
+        }
+
+        headers['Pairs used'] = {
+            'title': '% Used pairs',
+            'max': 100,
+            'min': 0,
+            'modify': lambda x: x * 100
```
ewels: It's nice to add `'suffix': '%'` for these columns.

multiqc/modules/hicexplorer/hicexplorer.py
joachimwolff: I tried different colors for the plot and the current scheme was the one I liked most.

joachimwolff:
Hi Phil,

I changed the module for the HiCExplorer a bit because we have a slightly different format of the QC files in our new 1.8 release.

Thanks for your review!

Best,

Joachim

ewels:
Great, thanks Joachim! I'm hoping to get to this really soon, sorry it's taking me so long.

How easy is it to make the module work with both versions of the output? That's my usual approach when output formats change, where realistic. Failing that it would be good to at least catch the error triggered by the old format and give a warning log message with an explanation.

Phil

joachimwolff:
@fidelram Can you check if my small patch for HiCExplorer <= 1.7 is valid? Thanks a lot!

ewels:
Apologies, a minor merge conflict has crept in from me merging another PR just now. If you can fix this and we get both versions of the log files in the test data (and merged) then the automated tests should check the patch for us :)

ewels:
Many thanks for this module @joachimwolff - it's great! I've just gone through and instead of making lots of minor requests on the PR I've just made the changes myself. I hope this is ok - it's because I'd like to get this included in the release which I'm about to do now.

I'll make comments on the commit above (d28d6efa54a7232a6be2d3b26c37a5486cc4e8bd) explaining why I've made the changes - you're welcome to submit a new PR reverting any that you disagree with.

Phil

PR #623, opened 2017-11-07
Get around needless enum34 requirement in python3.4+ in the next release

dpryan79:
In the most recent bioconda recipe, we ran into an issue due to `enum34` being listed as a requirement. This isn't actually needed as of python-3.4 and conda-forge has started making it python-2.7 only, meaning that one can't `conda install enum34` under python-3.6. While one could simply continue to build enum34 under versions of python not needing it, it's preferable to simply not require it unless needed. This PR essentially does what the patch for `setup.py` is currently doing in bioconda and removes `enum34` as a dependency in versions of python that already come with the `enum` built-in module.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ MultiQC was written by Phil Ewels (http://phil.ewels.co.uk) at SciLifeLab Sweden
 """
 
 from setuptools import setup, find_packages
+import sys
 
 version = '1.4dev'
 dl_version = 'master' if 'dev' in version else 'v{}'.format(version)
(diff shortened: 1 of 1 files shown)


ewels:
Ok fantastic - many thanks for looking into this @dpryan79! I had missed your patches on bioconda, I didn't realise that there were problems so thanks for fixing that 👍 

The new `enum34` dependency is only used in a very small part of one of the new module additions by @winni2k [here](https://github.com/ewels/MultiQC/blob/74cb2d9caca1d9c9768c05dba7e95ecd69de7510/multiqc/modules/gatk/base_recalibrator.py#L14-L19) - I actually queried it with a concern [here](https://github.com/ewels/MultiQC/pull/555#discussion_r140164491) exactly because of worries about potential problems like this, but we chatted in person and decided that it was fine. But anyway, another option is that we just refactor that code to drop the dependency. This may be a cleaner solution than adding in new py-version specific fixes?

Phil

ewels:
@winni2k - what do you think? Would it be much of a change to refactor that code to work without the import?

winni2k:
@ewels no need to refactor the code using `enum`, as `enum` is part of the standard library from python 3.4 onward.  This PR does exactly the right thing: Only install `enum34` (which is just a drop-in replacement for `enum`) if python version is less than 3.4.

So, looks good to me.

mbargull:
Those changes work, but it'd be even simpler to leverage `setuptools`'s built-in capabilities, i.e., just replace `'enum34'` by `'enum34;python_version<"3.4"'`, see http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies

PR #634, opened 2017-11-30
Update macs2.py

avilella:
fragments vs tags for older MACS2 compatibility.


ewels:
Great - tests are passing now 🎉 

Thanks for this - if you have some truncated example files to add to the test data repo as well then that would be great (feel free to drag and drop into a comment here and I can upload them).

Phil

PR #649, opened 2017-12-22
Adding parsing of general error rate from qualimap

Cashalow:
Hello !

First thanks a lot for this very useful tool, I have discorded it recently and it's really awesome !

For my development I wanted to be able to study the general error rate of the reads, parameter that is calculated by qualimap but that was not included in your parsing. It is now included in the general table because I think it is a rather useful metrics but maybe it should be hidden by default ? Let me know what you think.

I have tried to add it and keep the code as close as yours, so please tell me if this is not sufficient or if quality is subpar.

Cheers !

### DIFF
--- a/multiqc/modules/qualimap/QM_BamQC.py
+++ b/multiqc/modules/qualimap/QM_BamQC.py
@@ -554,7 +554,7 @@ def general_stats_headers (self):
         'max': 100,
         'min': 0,
         'suffix': '%',
-        'scale': 'OrRd'
+        'scale': 'OrRd',
         'format': '{0:.2f}'
     }
 
--- a/multiqc/modules/qualimap/QM_BamQC.py
+++ b/multiqc/modules/qualimap/QM_BamQC.py
@@ -84,6 +84,7 @@ def parse_genome_results(self, f):
         'mean_insert_size': r"mean insert size = ([\d,\.]+)",
         'median_insert_size': r"median insert size = ([\d,\.]+)",
         'mean_mapping_quality': r"mean mapping quality = ([\d,\.]+)",
+        'general_error_rate': r"general error rate = ([\d,\.]+)",
     }
     d = dict()
     for k, r in regexes.items():
@@ -93,6 +94,7 @@ def parse_genome_results(self, f):
                 d[k] = float(r_search.group(1).replace(',',''))
             except ValueError:
                 d[k] = r_search.group(1)
+    print(d)
 
     # Check we have an input filename
     if 'bam_file' not in d:
@@ -546,6 +548,14 @@ def general_stats_headers (self):
         'scale': 'Blues',
         'shared_key': 'read_count',
         'hidden': True
+    }
+    self.general_stats_headers['general_error_rate'] = {
+        'title': 'Error rate'.format(config.read_count_prefix),
+        'description': 'General error rate of the reads',
+        'max': 100,
+        'min': 0,
+        'suffix': '%',
+        'scale': 'Blues'
     }
 
 def _calculate_bases_within_thresholds(bases_by_depth, total_size, depth_thresholds):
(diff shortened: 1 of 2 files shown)


multiqc/modules/qualimap/QM_BamQC.py
```diff
@@ -547,6 +548,15 @@ def general_stats_headers (self):
         'shared_key': 'read_count',
         'hidden': True
     }
+    self.general_stats_headers['general_error_rate'] = {
+        'title': 'Error rate'.format(config.read_count_prefix),
```
ewels: You can remove the `.format(config.read_count_prefix)` bit - that's for adding in the 'M' for million etc. eg:

```python
'title': 'Error rate',
```

multiqc/modules/qualimap/QM_BamQC.py
```diff
@@ -547,6 +548,15 @@ def general_stats_headers (self):
...
+        'description': 'General error rate of the reads',
```
ewels: Could you write a slightly longer description here please? eg. Specify that it's the alignment error rate, and perhaps how it's calculated or something?

Cashalow:
I think you are right for the hidden parameters, I will just modify this on the multiqc version I am using so that it appears automatically for me. For the general error rate description, I used the documentation of qualimap available here http://qualimap.bioinfo.cipf.es/doc_html/analysis.html

> Mismatches and indels
The section reports general alignment error rate (computed as a ratio of total collected edit distance to the number of mapped bases), total number of mismatches and total number of indels (computed from the CIGAR values). Additionally fraction of the homopolymer indels among total indels is provided. Note, the error rate and mismatches metrics are based on optional fields of a SAM record (NM for edit distance, MD for mismatches). The features are not reported if these fields are missing in the SAM file.

ewels:
Great stuff, thanks! Merged, then I've added a small edit to the description - made it a little shorter and changed it from `General error rate of the reads` (which I think is misleading - reads could be raw sequence reads) to `Alignment error rate` (which I think is clearer and more accurate). Shout if you disagree with this :)

Note that you don't need to have a modified version of MultiQC for yourself, you can override the hidden default by customising your MultiQC config - see the [docs](http://multiqc.info/docs/#hiding-columns). I'll add to the Qualimap section of the docs to mention this too.

Phil

ewels:
Ok, changes pushed. Note that the config that you need to show the error rate by default is the following:

```yaml
table_columns_visible:
    QualiMap:
        general_error_rate: True
```

Cashalow:
Thanks a lot ! 
I am using a modified version because some hard coded behaviour does not suit my need (for instance I'm working with microbial genomes and in the QUAST parsing, the L50 and L75 values are automatically divided by 1000 to show the number of contigs by thousands, and my assembly typically have less than 200 contigs. Also for QUAST I have removed the suffix 'bp' for N50 and Total Lenght because  the unit of a column is in Kbp or Mbp (for N50 and total lenght) :sweat_smile: 

I could try to make my fixes for QUAST better and submit another PR if you are interested.

ewels:
Ah great - yes a PR for QUAST would be brilliant! Should be possible to add some module-specific config parameters to control this behaviour instead of hardcoding it (as is done with Qualimap `qualimap_config` - could look for a config key called `quast_config`). These could then work much like the `read_count_multiplier` configs (see http://multiqc.info/docs/#number-base-multiplier). eg:

```python
qconfig = getattr(config, quast_config, {})
contig_length_multiplier: qconfig.get('contig_length_multiplier', 0.001)
contig_length_suffix: qconfig.get('contig_length_suffix', 'Kbp')
```
```python
headers['N75'] = {
    'title': 'N75 ({})'.format(contig_length_suffix),
    'description': 'N75 is the contig length such that using longer or equal length contigs produces 75% of the bases of the assembly ({})'.format(contig_length_suffix),
    'min': 0,
    'suffix': contig_length_suffix,
    'scale': 'RdYlGn',
    'modify': lambda x: x / contig_length_multiplier
}
```
Which would then enable an optional config such as:

```yaml
qconfig:
    contig_length_multiplier: 1
    contig_length_suffix: 'bp'
```

Phil

Cashalow:
Perfect I'll look into that and I'll get back to you !

Issue #650, opened 2018-01-03
peddy module - het_check plot


aledj2:
Hi Phil,

I'm looking to include a different output of peddy in the MultiQC report (the equivalent of the `het_check.png`) as this will show possible sample contamination.

Looking at the peddy module,I have written a new function which plots this (`x=median_depth`, `y=het_ratio`), based on the functions which create the other plots, however I think there's an issue with the way the data is read from the peddy files.

There are three csv files that peddy produces, and MultiQC reads. From what I can tell these are imported into the `self.peddy_data` dictionary, with the sample name as a key and then all the values from these files added to a sample specific dictionary.

The issue is that two of the peddy files have columns with the same name (but different values) so if trying to plot eg het_ratio from het_check, the value plotted is the het_ratio from the sex_check (sex_check is parsed after het_check).
 
I have got around this by emptying `self.peddy_data` and re-populating it just with the het_check file but I don't think this is ideal! 

If I get a chance I'll try and make something which prefixes the input file to the dictionary value name to make sure the correct values are plotted.

If I were to do this, and add in an extra function for the het_check plot do i need to create an anchor or something for the plot or add in an option to the config file?

I've attached what I've done so far, but this also includes removing the other plots and columns from the general statistics table.
[peddy.py.gz](https://github.com/ewels/MultiQC/files/1601291/peddy.py.gz)

Thanks!

ewels:
Hi @aledj2,

Many thanks for this and your pull request. Looking at your PR it seems that you've answered most of the questions that you had here (it looks great!), so I'll close this issue for now. If there's something I've missed, please raise it as a comment on the PR and we can continue the discussion there.

Cheers,

Phil

PR #651, opened 2018-01-04
Plot Peddy het check

aledj2:
New peddy plot for that shows `x=median_depth`, `y=het_ratio`. See #650

### DIFF
--- a/multiqc/modules/peddy/peddy.py
+++ b/multiqc/modules/peddy/peddy.py
@@ -218,10 +218,6 @@ class MultiqcModule(BaseMultiqcModule):
         # empty dictionary to add sample names, and dictionary of values
         data = {}
         
-        # set variable for y axis limits
-        ymin = 1
-        ymax = 0
-        
         # for each sample, and list in self.peddy_data
         for s_name, d in self.peddy_data.items():
             # check the sample contains the required columns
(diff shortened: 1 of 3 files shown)


multiqc/modules/peddy/peddy.py
```diff
@@ -205,3 +212,31 @@ def peddy_relatedness_plot(self):
...
+        
+        pconfig = {
+            'id': 'peddy_het_check_plot',
+            'title': 'Peddy Het Check Plot',
+            'xlab': 'median depth',
+            'ylab': 'proportion het calls',
+            }
+    
+        self.add_section (
+            name = 'Peddy Het Check Plot',
```
ewels: This section is already under Peddy in the nav, so no need to repeat this again - `Het Check` should be fine.

multiqc/modules/peddy/peddy.py
```diff
@@ -205,3 +212,31 @@ def peddy_relatedness_plot(self):
...
+                data[s_name] = {
+                    'x': d['median_depth_het_check'],
+                    'y': d['het_ratio_het_check']
+                }
+        
+        pconfig = {
+            'id': 'peddy_het_check_plot',
+            'title': 'Peddy Het Check Plot',
+            'xlab': 'median depth',
+            'ylab': 'proportion het calls',
```
ewels: Is it worth forcing a y axis scale from 0-1 here? Or would that mask differences that we're looking for?

multiqc/modules/peddy/peddy.py
```diff
@@ -205,3 +212,31 @@ def peddy_relatedness_plot(self):
...
+            if 'median_depth_het_check' in d and 'het_ratio_het_check' in d:
+                # add sample to dictionary with value as a dictionary of points to plot
+                data[s_name] = {
+                    'x': d['median_depth_het_check'],
+                    'y': d['het_ratio_het_check']
+                }
+        
+        pconfig = {
+            'id': 'peddy_het_check_plot',
+            'title': 'Peddy Het Check Plot',
```
ewels: Preferred syntax for plot titles is now to have a colon: `Peddy: Het Check`

multiqc/modules/peddy/peddy.py
ewels: Also, please add at least a `description` and preferably also a longer `helptext` for this section.

multiqc/modules/peddy/peddy.py
aledj2: Yea, I think 0-1 Y axis would make the differences harder to spot. I have set the y limits to +/- 0.1 from min/max values, but now I think about it I prefer it with the original padding! Will undo this change

ewels:
Looks great, thanks!

ewels:
Hi @aledj2,

A bit late, now that I've merged this, but I'm not completely clear of the use of plotting this against `median_depth`? Would this not be clearer as a simple bar plot of `proportion_het_calls`? I feel like that would be quite a bit easier to read..

Phil

PR #653, opened 2018-01-08
Quast modifiers

Cashalow:
Here's one try to adding modifiers to quast parsing

### DIFF
--- a/multiqc/modules/quast/quast.py
+++ b/multiqc/modules/quast/quast.py
@@ -108,7 +108,7 @@ class MultiqcModule(BaseMultiqcModule):
     def quast_general_stats_table(self):
         """ Take the parsed stats from the QUAST report and add some to the
         General Statistics table at the top of the report """
-        qconfig = getattr(config, quast_config, {})
+        qconfig = getattr(config, "quast_config", {})
 
         contig_length_multiplier= qconfig.get('contig_length_multiplier', 0.001)
         contig_length_suffix= qconfig.get('contig_length_suffix', 'Kbp')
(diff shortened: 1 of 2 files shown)


multiqc/modules/quast/quast.py
```diff
@@ -108,79 +122,83 @@ def parse_quast_log(self, f):
...
-            'title': 'N50 (Kbp)',
-            'description': 'N50 is the contig length such that using longer or equal length contigs produces 50% of the bases of the assembly (kilo base pairs)',
+            'title': 'N50 ({})'.format(self.contig_length_suffix),
+            'description': 'N50 is the contig length such that using longer or equal length contigs produces half (50%) of the bases of the assembly.',
             'min': 0,
-            'suffix': 'bp',
+            'suffix': self.contig_length_suffix,
             'scale': 'RdYlGn',
-            'modify': lambda x: x / 1000
+            'modify': lambda x: x / self.contig_length_multiplier
```
ewels: These need to be switched to multiply instead of divide. Otherwise the default of dividing by 0.000001 will make the counts a million times bigger ;)

Cashalow:
Yes indeed ! Sorry about that.

ewels:
No problem! I'm not really sure why I decided to do multiplying by a small value instead of dividing by a large value, it must have made sense at the time 😛 

Looks good anyway, thanks!

ewels:
Ok, I added some documentation here: http://multiqc.info/docs/#quast

Shout if anything is wrong!

PR #658, opened 2018-01-10
use file suffix when adding column to general stats

aledj2:
Added suffix to the self.peddy_data but didn't use this suffix when trying to add a column into general stats 'error' --> error_sex_check


ewels:
Good spot, thanks!

Issue #661, opened 2018-01-11
bcl2fastq2 ZeroDivisionError


aledj2:
Hi Phil,

Just ran the latest version of multiqc and have the following error from the bcl2fastq module.

Have attached the file which produced the error

[Stats.json.gz](https://github.com/ewels/MultiQC/files/1623091/Stats.json.gz)

```python
Module bcl2fastq raised an exception: Traceback (most recent call last):
  File "/home/dnanexus/Miniconda/lib/python2.7/site-packages/multiqc-1.4-py2.7.egg/EGG-INFO/scripts/multiqc", line 427, in multiqc
    output = mod()
  File "/home/dnanexus/Miniconda/lib/python2.7/site-packages/multiqc-1.4-py2.7.egg/multiqc/modules/bcl2fastq/bcl2fastq.py", line 21, in __init__
    self.parse_file_as_json(myfile)
  File "/home/dnanexus/Miniconda/lib/python2.7/site-packages/multiqc-1.4-py2.7.egg/multiqc/modules/bcl2fastq/bcl2fastq.py", line 171, in parse_file_as_json
    run_data[lane]["percent_Q30"] = (float(run_data[lane]["yieldQ30"]) / float(run_data[lane]["total_yield"])) * 100.0
ZeroDivisionError: float division by zero
```

aledj2:
Is a pretty horrible hack but have put some of the division lines in a try except statement, setting value to NA if it fails - not sure if that's appropriate

I'm not sure if it's even a real issue or if I've got a dodgy stats file.


[bcl2fastq.py.gz](https://github.com/ewels/MultiQC/files/1623145/bcl2fastq.py.gz)

ewels:
Sorry @aledj2, you just missed the v1.4 release - I pushed it out in a bit of a rush as pip installations were suddenly failing due to an upstream dependency change.

Sounds like a pretty common edge case scenario that comes up quite a bit in MultiQC. Here your total_yield must be 0. Doesn't really matter if you have a weird file or not - if you managed to trigger it then other people can presumably do the same, doesn't hurt to handle it nicely. Catching the error and setting to NA sounds like a sensible way to handle it.

Could you open a PR for the changes? They look fine. For shorter code could perhaps use an inline conditional instead:

```python
[...]

Issue #663, opened 2018-01-12
Conflicting conditional formatting gives double-labelled values


aledj2:
Hi,

The sex check column in peddy returns true if a sex error is seen and false if no sex error is seen (sex is as expected). Therefore because errors are True these are a in a nice green box where as  those which pass the test are red (==False) are in a red box.

I tried changing `config.yaml` using:
```yaml
table_cond_formatting_rules:
    mqc-generalstats-error_sex_check:
        fail:
            - s_eq: 'true'
        pass:
            - s_eq: 'false' 
```

However this resulted in both green and red boxes around each value

To get around this I've changed the peddy module so it swaps true and false to get the colours correct, and removed the code from the `config.yaml` but not sure if this is the best approach?

ewels:
> To get around this I've changed the peddy module so it swaps true and false to get the colours correct, and removed the code from the `config.yaml` but not sure if this is the best approach?

I'm not sure that there's a right or wrong answer here, but I think this is probably the best solution. Personally I tend to think that True = a good thing. Having a sex error is not A Good Thing, so it makes sense to me to switch this to _Correct Sex_ instead of _Sex error_.

> However this resulted in both green and red boxes around each value

This is a bug and shouldn't happen - the column specific config should have taken preference. I'll look into it.

PR #664, opened 2018-01-12
For ewels

aledj2:


aledj2:
fixes #661 and partially addresses that raised in #663

multiqc/modules/peddy/peddy.py
```diff
@@ -121,11 +122,16 @@ def parse_peddy_csv(self, f, pattern):
                 parsed_data[s_name] = dict()
                 for i, v in enumerate(s):
                     if i not in s_name_idx:
+                        if headers[i] == "error" and pattern == "sex_check":
+                            if v == "True":
```
ewels: Can compress this a bit I think:
```python
v = "True" if v == "False" else "False"
```

multiqc/modules/peddy/peddy.py
```diff
@@ -74,10 +74,11 @@ def __init__(self):
 
         # Relatedness plot
         self.peddy_relatedness_plot()
-
+
```
ewels: Would be nice if you could clean up these whitespace changes in the diff - they all get highlighted in a bright orange in my editor so it saves me removing them later 😉

aledj2:
verifybamid.py has a comma removed at the end of a dicitonary. I think that's correct - I don't get any errors but let me know if it's wrong

ewels:
Perfect, thanks! A comma on the final element of a Python iterable (dictionary, list etc) is not required, but it doesn't break if it's there either.

ewels:
Note that you may get some git nastyness when you pull changes after this squash merge. You'll probably need to rebase your git history to fix this. Worth doing straight away before you start making any future code changes :)

Phil

ewels:
ps. To avoid me doing squash merges in the future, you can use interactive rebasing to reduce the number of commits - there were 39 commits here for quite a minor change which is why I squashed them. See for example https://github.com/mockito/mockito/wiki/Using-git-to-prepare-your-PR-to-have-a-clean-history

PR #667, opened 2018-01-18
Slight adaptations of axis labellings for Slamdunk module.

t-neumann:
Hi Phil,

just some adaptations of axis labellings.

Cheers


ewels:
Ah ffs.. _More_ dependency problems??

```
error: networkx 2.1rc1 is installed but networkx==2.0 is required by {'spectra'}
```

PR looks great, thanks @t-neumann! Apologies, there are packaging woes happening at the moment, which is why the PR tests (and MultiQC installations currently) are failing. Let me know if you need this merged quickly, otherwise I'll come back to it as soon as the installation crap is fixed.

t-neumann:
Nope no urgency. Just needed it for the manuscript and figured I'll create the PR rightaway to not lose it on the way.

PR #687, opened 2018-02-19
[Bug fix] HiCExplorer plugin

joachimwolff:
Bug fix: if two matrices had the same name, the keys in the data dictionary were overwritten.


ewels:
Thanks, looks good! Tests are failing due to a recent bug in MatPlotLib (https://github.com/matplotlib/matplotlib/issues/10476), should be fixed soon hopefully.

Phil

PR #690, opened 2018-02-23
Add biscuit module

zwdzwd:
This adds support for QC output from BISCUIT. This includes: QC for mapping uniformity, retention rates, duplication uniformity, CpG coverage, methylation bias, mapping summary, etc. Initial testing should pass with the test data on MultiQC_TestData (also a pull request there). Please review. Thank you.


zwdzwd:
Hi @ewels, Thanks for the feedbacks. They are very helpful. I will go work on those and get back to you.

ewels:
Hi @zwdzwd,

Have you had a chance to make any progress on this PR recently?

Phil

zwdzwd:
@ewels 
Sorry for the late response. I got swamped by other things. I am still working on this. I will try get back to you by next week.

zwdzwd:
Hi @ewels ,
My apology again for delayed action. Could you please check this version? I managed to cut the number of sections to 11 by combining figures to tabs and combining tables. I removed a couple of sections of less importance. I wonder if this may be an OK number for initial implementation? I also double checked to make sure titles and axis labels are there and sample names are clean. Let me know if I am missing something or there is a better way of doing things. 

Thank you for making such a handy and beautiful tool!

Wanding

ewels:
Great stuff, thanks! No problem with the delay, I’ve been very busy recently too. I will spend a day working on MultiQC later this week and have another look then.

Thanks!

Phil

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+        self.mdata = {
+            'mapq': {},
+            'markdup': {},
+            'coverage': {},
+            'retention': {},
+        }
+
+        # Find and parse bismark alignment reports
+        for k in self.mdata:
+            for f in self.find_log_files('biscuit/%s' % k):
```
ewels: Minor thing, but I'd prefer format strings if possible - this is the code style that's used everywhere else in MultiQC. eg:

```python
'biscuit/{}'.format(k)
```

This doesn't make any functional difference, but helps with consistency :)

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+                d = getattr(self, 'parse_logs_%s' % k)(f['f'], f['fn'])
+                self.mdata[k][f['fn']] = d
+
+        num_parsed = 0
+        for k in self.mdata:
+            num_parsed += len(self.mdata[k])
+        if num_parsed == 0:
```
ewels: A slightly more condensed version could be:

```python
if sum([len(self.mdata[k]) for k in self.mdata]) == 0:
```

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+            raise UserWarning
+
+        # Basic Stats Table
+        self.biscuit_stats_table()
+
+        # Write out to the report
+        for k in self.mdata:
+            if len(self.mdata[k]) > 0:
+                self.write_data_file(self.mdata[k], 'multiqc_biscuit_%s' % k)
+                log.info("Found %d biscuit %s reports" % (len(self.mdata[k]), k))
```
ewels: The log message is prefixed with the module name (biscuit), so no need to repeat it here.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+                if k != 'markdup':
+                    getattr(self, 'chart_biscuit_%s' % k)()
```
ewels: What does this do?

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+
+    def biscuit_stats_table(self):
+
+        pd = {}
+        for sid, dd in self.mdata['mapq'].items():
+            if not sid.endswith('_mapq_table.txt'):
+                continue
+            sid = sid.replace('_mapq_table.txt','')
```
ewels: Please run through the `self.clean_s_name()` function as well. This is needed for the `-b` function and custom sample name cleanups.

See http://multiqc.info/docs/#custom-sample-names

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+                self.write_data_file(self.mdata[k], 'multiqc_biscuit_%s' % k)
+                log.info("Found %d biscuit %s reports" % (len(self.mdata[k]), k))
+                if k != 'markdup':
+                    getattr(self, 'chart_biscuit_%s' % k)()
+
+    def biscuit_stats_table(self):
+
+        pd = {}
+        for sid, dd in self.mdata['mapq'].items():
+            if not sid.endswith('_mapq_table.txt'):
```
ewels: I'd prefer to remove this. Sometimes MultiQC users change the filenames of tool outputs and then customise their MultiQC config to update the search patterns. This will break things for them as the filename could not have that string in for genuine reasons.

If you want to validate that the correct thing has been parsed, I'd recommend doing this in the parsing step instead (preferably by checking file contents / parsed data).

multiqc/modules/biscuit/biscuit.py
ewels: Ah, I just got to the search pattern config. Yes, so it's better to have a single search pattern config for every _type_ of file rather than relying on the filename for this reason. Then if each file type is parsed into a separate data structure, you shouldn't need to have this check.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+
+        pheader = OrderedDict()
+        pheader['%aligned'] = {'title':'% Aligned', 'max':100, 'min':0, 'suffix':'%','scale':'Greens'}
+        if hasPE:
+            pheader['%dupratePE'] = {'title':'% DupsPE', 'max':100, 'min':0, 'suffix':'%','scale':'Reds'}
+        if hasSE:
+            pheader['%duprateSE'] = {'title':'% DupsSE', 'max':100, 'min':0, 'suffix':'%','scale':'Reds'}
+        self.general_stats_addcols(pd, pheader)
+
+        # if len(mdata)>0 and all([dd['Duplication Rate'] is not None for dd in mdata.values()]):
```
ewels: Please remove old commented out code.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+        # self.general_stats_addcols(mdata, {'% Dup': {'suffix':'%', 'min':0, 'max':100}})
+
+
+    def parse_logs_mapq(self, f, fn):
+
+        data = {}
+        if fn.endswith('_mapq_table.txt'):
```
ewels: Yup, please split each of these out into their own function with their own search pattern.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+                elif int(mapq) >= 40:
+                    pd[sid]['OAligned'] += int(cnt)
+                else:
+                    pd[sid]['SAligned'] += int(cnt)
+
+        self.add_section(
+            name = 'Mapping Summary',
+            anchor = 'biscuit-mapping',
+            plot = bargraph.plot(pd, OrderedDict([
+                ('OAligned', {'name':'Optimally Aligned Reads', 'color': '#8bbc21'}),
```
ewels: Personally, I prefer charts to use colours from the existing MultiQC colour palette (the defaults here should actually be pretty good I think? Blue / Dark Blue / Orange off the top of my head). But it's your code, so if you really prefer these colours then it's fine to stick with them 😉

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+                if mapq == 'unmapped':
+                    pd[sid]['UAligned'] += int(cnt)
+                elif int(mapq) >= 40:
+                    pd[sid]['OAligned'] += int(cnt)
+                else:
+                    pd[sid]['SAligned'] += int(cnt)
+
+        self.add_section(
+            name = 'Mapping Summary',
+            anchor = 'biscuit-mapping',
```
ewels: Please add at least a short `description` and preferably a longer `helptext` to describe what to expect and what to look out for.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+            description = "<p>This plot shows the distribution of strand of mapping and strand of bisulfite conversion.</p>",
+            plot = bargraph.plot(pd, OrderedDict([
+                ('++', {'name':'Waston-Aligned, Waston-Bisulfite Conversion', 'color': '#F53855'}),
+                ('+-', {'name':'Waston-Aligned, Crick-Bisulfite Conversion', 'color': '#E37B40'}),
+                ('-+', {'name':'Crick-Aligned, Waston-Bisulfite Conversion', 'color': '#46B29D'}),
+                ('--', {'name':'Crick-Aligned, Crick-Bisulfite Conversion', 'color': '#324D5C'})
+            ]), {'id':'biscuit_strands',
+                 'title':'BISCUIT: Mapping Strand Distribution',
+                 'ylab':'Number of Reads',
+                 'cpswitch_c_active': True,
```
ewels: `True` is the default - maybe you mean `False` so that the bars display percentages by default? This would make sense to me, as we're looking for differences in proportions here, rather than counts.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+        )
+
+        # mapping strand distribution
+        pd = dict([(k.replace('_strand_table.txt', ''),v) for k,v in self.mdata['mapq'].items() if k.endswith('_strand_table.txt')])
+        self.add_section(
+            name='Mapping Strand Distribution',
+            anchor='biscuit-strands',
+            description = "<p>This plot shows the distribution of strand of mapping and strand of bisulfite conversion.</p>",
+            plot = bargraph.plot(pd, OrderedDict([
+                ('++', {'name':'Waston-Aligned, Waston-Bisulfite Conversion', 'color': '#F53855'}),
```
ewels: Same comment with colours as above, but fine to leave if you prefer.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+                 'cpswitch_counts_label': '# Reads'
+            })
+        )
+
+        # mapping strand distribution
+        pd = dict([(k.replace('_strand_table.txt', ''),v) for k,v in self.mdata['mapq'].items() if k.endswith('_strand_table.txt')])
+        self.add_section(
+            name='Mapping Strand Distribution',
+            anchor='biscuit-strands',
+            description = "<p>This plot shows the distribution of strand of mapping and strand of bisulfite conversion.</p>",
```
ewels: The description is parsed as markdown, so no need for HTML tags such as `<p>`.

It would be great if you could write a `helptext` talking about different possibilities here and what to expect. eg. different types of bisulfite library prep.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+            pd_mapping[sid] = dict(zip(range(61), mapqcnts))
+
+        pd_isize = dict([(k.replace('_isize_score_table.txt', ''),v['I']) for k,v in self.mdata['mapq'].items() if k.endswith('_isize_score_table.txt')])
+
+        self.add_section(
+            name = 'Mapping Quality and Insert Size',
+            anchor = 'biscuit-mapq-isize',
+            description = "<p>This plot shows the distribution of Primary Mapping Quality.</p>",
+            plot = linegraph.plot([pd_mapping, pd_isize],
+                {'id':'biscuit_mapping', 'title': 'Mapping Information', 'data_labels': [
```
ewels: Please prepend the title with the module name.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+                    {'name':'Mapping Quality', 'ylab': '% Primary Mapped Reads','xlab': 'Mapping Quality'},
+                    {'name':'Insert Size', 'ylab': '% Mapped Reads', 'xlab': 'Insert Size'}]}))
```
ewels: This graph is empty for me when I test. Please can you add code to check that there is data to plot so that the tab isn't added if there is no data..

multiqc/modules/biscuit/biscuit.py
ewels: Suggested additional config options:
```python
'ymin': 0,
'yCeiling': 100,
'yMinRange': 10,
'yLabelFormat': '{value}%',
'tt_label': '<strong>Q{point.x}:</strong> {point.y:.2f}% of reads'
```

multiqc/modules/biscuit/biscuit.py
ewels: Also, I know I asked you to combine plots and reduce number of sections, but I don't really see how insert size and mapping quality are related. So I think these _should_ be separate sections and plots. Sorry!

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+            dict([(k.replace('_covdist_cpg_table.txt',''),v) for k, v in self.mdata['coverage'].items() if k.endswith('_covdist_cpg_table.txt')]),
+            dict([(k.replace('_covdist_cpg_q40_table.txt',''),v) for k, v in self.mdata['coverage'].items() if k.endswith('_covdist_cpg_q40_table.txt')]),
+            dict([(k.replace('_covdist_cpg_q40_botgc_table.txt',''),v) for k, v in self.mdata['coverage'].items() if k.endswith('_covdist_cpg_q40_botgc_table.txt')]),
+            dict([(k.replace('_covdist_cpg_q40_topgc_table.txt',''),v) for k, v in self.mdata['coverage'].items() if k.endswith('_covdist_cpg_q40_topgc_table.txt')]),
+        ]
+        if len(mdata)>0:
+            self.add_section(
+                name = 'Cumulative Base Coverage',
+                anchor = 'biscuit-coverage-base',
+                description = "<p>This plot shows the cummulative base coverage. High and low GC content region are the top and bottom 10% 100bp window in GC content.</p>",
```
ewels: No HTML tags needed. Please also describe what the Q means. Help text would also be, well, helpful!

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+                plot = linegraph.plot(mdata, {'id':'biscuit_coverage_base',
+                    'title': 'Cumulative Base Coverage',
```
ewels: Please add an X axis, `xLabelFormat` with the `X` suffix to denote coverage,  prepend the title with the module name.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+                    'data_labels': [
+                        {'name': 'All', 'ylab':'Million Bases'},
```
ewels: Please adjust the ylab values to reflect which plot is shown. This is helpful when single plots are exported in isolation.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+                        {'name': 'CpG Q40', 'ylab':'Million CpGs'}, 
+                        {'name': 'CpG Q40 low GC', 'ylab':'Million CpGs'}, 
+                        {'name': 'CpG Q40 high GC', 'ylab':'Million CpGs'}, 
+                    ]})
+            )
+            
+            for sid, dd in mdata[0].items():
+                if sid not in basecov:
+                    basecov[sid] = {}
+                basecov[sid]['all'] = dd[1]/dd[0]*100
```
ewels: Note that you want to be careful with ints and floats here. `100` is an `int`, so if the others are parsed as ints then the result will also be an integer. I usually wrap stuff in `float()` and use `100.0` to force floating number results.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+                if sid not in basecov:
+                    basecov[sid] = {}
+                basecov[sid]['cpg_uniq'] = dd[1]/dd[0]*100
+
+        # base coverage >=1x table
+        if len(basecov)>0:
+            self.add_section(
+                name = 'Coverage by At Least One Read',
+                anchor = 'biscuit-coverage-base-table',
+                description = '<p>The fraction covered by at least one read.</p>',
```
ewels: The fraction of what? No HTML tags needed.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,688 @@
...
+        pheader['ct'] = {'title':'r.CpT', 'format':'{:,.2g}', 'min':0, 'max':100, 'description':'CpT', 'suffix':'%'}
+        pheader['bca'] = {'title':'b.CpA', 'format':'{:,.2g}', 'min':0, 'max':100, 'description':'CpA', 'suffix':'%'}
+        pheader['bcc'] = {'title':'b.CpC', 'format':'{:,.2g}', 'min':0, 'max':100, 'description':'CpC', 'suffix':'%'}
+        pheader['bcg'] = {'title':'b.CpG', 'format':'{:,.2g}', 'min':0, 'max':100, 'description':'CpG', 'suffix':'%'}
+        pheader['bct'] = {'title':'b.CpT', 'format':'{:,.2g}', 'min':0, 'max':100, 'description':'CpT', 'suffix':'%'}
+
+        self.add_section(
+            name = 'Cytosine Retention',
+            anchor = 'biscuit-retention',
+            description = "<p>This plot shows cytosine retention rate. 'r.' stands for read-averaging and 'b.' stands for 'base-averaging.</p>",
```
ewels: You can use markdown backticks for `r` which will look nicer that 'r' I think. No HTML tags needed.

ewels: It would be great to have help text explaining this too. I've worked with bisulfite data quite a bit and I'm not sure what cytosine retention means.

multiqc/utils/search_patterns.yaml
```diff
@@ -82,6 +82,32 @@ biobloomtools:
     num_lines: 2
 bismark/align:
     fn: '*_[SP]E_report.txt'
+biscuit/mapq:
```
ewels: File search patterns should be unique for each _type_ of data. The way that the code is written currently makes it impossible for people to customise these search patterns if they modify their data files, as the module code expects certain file paths. So instead, each file type should have its own search pattern:

```yaml
biscuit/mapq_mapq_table: fn: '*_mapq_table.txt'
biscuit/mapq_strand_table: fn: '*_strand_table.txt'
biscuit/mapq_isize_score_table: fn: '*_isize_score_table.txt'
```

multiqc/utils/search_patterns.yaml
```diff
@@ -82,6 +82,32 @@ biobloomtools:
...
+    - fn: '*_mapq_table.txt'
+    - fn: '*_strand_table.txt'
+    - fn: '*_isize_score_table.txt'
+biscuit/markdup:
+    fn: '*_markdup_report.txt'
```
ewels: This is quite a generic filename which could conceivably also be used by other tools. It would be good to also add a `content` string to confirm the identity of the file. Otherwise, even if the module code ignores it, the file will not be used by any other modules.

multiqc/utils/search_patterns.yaml
```diff
@@ -82,6 +82,32 @@ biobloomtools:
...
+    - fn: '*_covdist_cpg_q40_botgc_table.txt'
+    - fn: '*_covdist_cpg_q40_topgc_table.txt'
+    - fn: '*_covdist_cpg_table.txt'
+    - fn: '*_covdist_q40_botgc_table.txt'
+    - fn: '*_covdist_q40_table.txt'
+    - fn: '*_covdist_q40_topgc_table.txt'
+    - fn: '*_covdist_table.txt'
+    - fn: '*_cpg_cv_table.txt'
+    - fn: '*_cpg_dist_table.txt'
+    - fn: '*_dup_report.txt'
```
ewels: Same as above with the common filename pattern.

ewels:
Two more things: I only added one comment about using `self.clean_s_name()`, but that should be used every time a sample name is generated (this is quite a few places in the code). Also, parsed data structures should be filtered using `self.ignore_samples()` so that users can choose to remove certain samples.

.travis.yml
```diff
@@ -10,7 +10,7 @@ matrix:
 # commands to install dependencies
 install:
   - python setup.py install
-  - git clone https://github.com/ewels/MultiQC_TestData.git test_data
+  - git clone https://github.com/zwdzwd/MultiQC_TestData.git test_data
```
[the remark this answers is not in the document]
zwdzwd: This is reverted.

multiqc/modules/biscuit/biscuit.py
zwdzwd: I changed the code to this style.

multiqc/modules/biscuit/biscuit.py
zwdzwd: Thanks for showing me. This version is used.

multiqc/modules/biscuit/biscuit.py
zwdzwd: The repetition is removed.

multiqc/modules/biscuit/biscuit.py
zwdzwd: I added clean_s_name
```
s_name = self.clean_s_name(f['fn'], f['root']) # allow user to specify directory
self.mdata[k][s_name] = d
```

I did my best to separate search pattern for each file type and removed existing file checking best as I can.

multiqc/modules/biscuit/biscuit.py
zwdzwd: These are removed as mentioned above.

multiqc/modules/biscuit/biscuit.py
zwdzwd: The old commented code has been removed.

multiqc/modules/biscuit/biscuit.py
zwdzwd: Done.:)

multiqc/modules/biscuit/biscuit.py
zwdzwd: The description and helptext have been added.

multiqc/modules/biscuit/biscuit.py
zwdzwd: Yes, I changed it to False. Thanks

multiqc/modules/biscuit/biscuit.py
zwdzwd: I leave the color of this one as it appears to highlight the contrast better. I think the color code I was using comes from the Documentation IIRC.

multiqc/modules/biscuit/biscuit.py
zwdzwd: "<p>" are removed. helptext is added.

multiqc/modules/biscuit/biscuit.py
zwdzwd: These are added accordingly.

multiqc/modules/biscuit/biscuit.py
zwdzwd: I think the graph is empty because of the obsolete test data. I separated insert size and mapping quality section. Thank you.

multiqc/modules/biscuit/biscuit.py
zwdzwd: `p` is removed. I added help text.

multiqc/modules/biscuit/biscuit.py
zwdzwd: X-axis and xLabelFormat are added as suggested.

multiqc/modules/biscuit/biscuit.py
zwdzwd: Now ylab values are made unique to reflect the current plot shown. Thanks for the good idea.

multiqc/modules/biscuit/biscuit.py
zwdzwd: Thanks for the warning. I added `float()` and `100.0` to force floating number calculation.

multiqc/modules/biscuit/biscuit.py
zwdzwd: `<p>` removed. The description is updated to clarify it's the fraction of either genome or genomic CpGs (depends on tab).

multiqc/modules/biscuit/biscuit.py
zwdzwd: Thanks for the suggestion. I switched to `r` and removed `<p>`. I added help text to explain on cytosine retention/conversion.

multiqc/utils/search_patterns.yaml
zwdzwd: Yes, now each file type has its own search pattern.

multiqc/utils/search_patterns.yaml
zwdzwd: I updated the log file format to include a (quite) distinguishable header (BISCUITqc). I updated both search_pattern.yml and the script to detect this header. I hope this reduces the cross-talk between modules. Thanks for pointing me to `content`.

multiqc/utils/search_patterns.yaml
zwdzwd: Same solution as in response above. Thanks.

zwdzwd:
Thank you @ewels for making such an extensive review. I have addressed most of the comments which actually led to quite some refactoring of both the multiqc script and biscuit QC script itself. Yes, my PI did like the multiqc very very much and intended to use it as the primary reporting tool for Biscuit. The biggest change in the current code is to have each file type identified by a single search pattern. This indeed removed some previous check inside the script code to distinguish file type. I also added a few descriptions and help text as suggested. I will probably add more as the review goes on. I also added quite some annotation inside the script to make it more readable than the previous version.

Thanks again for your instruction, suggestions and feedback. Please certainly let me know of anything that I may still miss.

Wanding

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,830 @@
...
+
+""" MultiQC module to parse output from BISCUIT """
+
+from __future__ import print_function
+from collections import OrderedDict
+import logging
+import re
+
+from multiqc import config
+from multiqc.plots import beeswarm, linegraph, bargraph, table
```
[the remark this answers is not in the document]
zwdzwd: Also removed. Thanks.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,830 @@
...
+            # retention
+            'retention_dist': {},
+            'retention_dist_byread': {},
+            'retention_cpg_readpos': {},
+            'retention_cph_readpos': {},
+            'retention_rate_byread': {},
+            'retention_rate_bybase': {},
+        }
+
+        # Find and parse bismark alignment reports
```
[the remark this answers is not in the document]
zwdzwd: Sorry. My oversight. Now removed. :)

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,830 @@
...
+        for k in self.mdata:
+            for f in self.find_log_files('biscuit/{}'.format(k)):
+                d = getattr(self, 'parse_logs_%s' % k)(f['f'], f['fn'])
+                s_name = self.clean_s_name(f['fn'], f['root']) # allow user to specify directory
```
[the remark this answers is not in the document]
zwdzwd: This make sense, I still kept the `s_name = self.clean_s_name` part for a later comment. This is the place sample names were cleaned first round. They get more cleaning (in ad hoc ways) in each functions.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,830 @@
...
+        pd = {}
+        for sid, dd in self.mdata['align_mapq'].items():
+            sid = sid.replace('_mapq_table','')
+            allreads = sum([int(_) for _ in dd.values()])
+            pd[sid] = {'%aligned': float(allreads-int(dd['unmapped']))/allreads*100}
+        
+        hasPE = False
+        hasSE = False
+        for sid, dd in self.mdata['markdup_report'].items():
+            sid = sid.replace('_markdup_report','')
```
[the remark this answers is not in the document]
zwdzwd: That makes sense. As mentioned above, I now have "clean_s_name" at the time of parsing. The plotting code does contain some more cleaning but they tend to be targeting uncommon suffixes.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,830 @@
...
+                pd[sid]['%dupratePE'] = dd['dupRatePE']
+                hasPE = True
+            if 'dupRateSE' in dd and dd['dupRateSE'] is not None:
+                pd[sid]['%duprateSE'] = dd['dupRateSE']
+                hasSE = True
+
+        pheader = OrderedDict()
+        pheader['%aligned'] = {'title':'% Aligned', 'max':100, 'min':0, 'suffix':'%','scale':'Greens'}
+        if hasPE:
+            pheader['%dupratePE'] = {'title':'% DupsPE', 'max':100, 'min':0, 'suffix':'%','scale':'Reds'}
```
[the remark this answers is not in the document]
zwdzwd: Good information. It is deleted now.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,830 @@
...
+            description = 'This shows the fraction of optimally aligned reads, which is defined by mapQ >= 40.',
+            helptext = 'A good library should have high fraction of reads optimally aligned. Suboptimally aligned reads include both nonunique alignments and imperfect alignments.',
+            plot = bargraph.plot(pd, OrderedDict([
+                ('OAligned', {'name':'Optimally Aligned Reads'}),
+                ('SAligned', {'name':'Suboptimally Aligned Reads'}),
+                ('UAligned', {'name':'Unaligned Reads'})
+            ]), {'id':'biscuit_mapping_summary',
+                 'title':'BISCUIT: Mapping Summary',
+                 'ylab':'Number of Reads',
+                 'cpswitch_c_active': False,
```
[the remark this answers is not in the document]
zwdzwd: Sure, I deleted this for consistency.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,830 @@
...
+                 'cpswitch_counts_label': '# Reads'
+            })
+        )
+
+        # Mapping Quality together in one plot
+        total = {}
+        for sid, dd in self.mdata['align_mapq'].items():
+            sid = sid.replace('_mapq_table','')
```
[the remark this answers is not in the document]
zwdzwd: That makes sense. I regrouped the name cleaning operations at parsing.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,830 @@
...
+            dict([(sid, dd['CT']) for sid, dd in mdata.items()]),
+        ]
+        self.add_section(
+            name = 'Number of Retention Distribution',
+            anchor = 'biscuit-retention-read',
+            description = "This plot shows the distribution of the number of retained cytosine in each read, up to 10.",
+            plot = linegraph.plot(pd, {
+                'id': 'biscuit_retention_read_cpa', 
+                'xlab': 'Number of Retention within Read',
+                'title': 'Retention Distribution',
```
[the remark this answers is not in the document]
zwdzwd: Added. Thanks.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,830 @@
...
+            dict([(k.replace('_CpGRetentionByReadPos',''),v['2']) for k, v in self.mdata['retention_cpg_readpos'].items()]),
+        ]
+        self.add_section(
+            name = 'Retention vs. Base Position in Read',
+            anchor = 'biscuit-retention-cytosine',
+            description = "This plot (aka. mbias plot) shows the distribution of cytosine retention rate in read.",
+            plot = linegraph.plot(mdata, {
+                'id': 'biscuit_retention_cytosine',
+                'xlab': 'Position in Read', 'ymin':0, 'ymax':100, 'yMinRange':0, 'yFloor':0,
+                'title': 'Retention vs. Base Position in Read',
```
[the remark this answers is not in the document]
zwdzwd: Added. Thanks.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,830 @@
...
+        pheader['cv_all_botgc'] = {'title':'CV.low.gc','description':'Bottom Decile in GC Content'}
+        pheader['mu_cpg'] = {'title':'CG.Mu.gnm','description':'All CpGs'}
+        pheader['mu_cpg_topgc'] = {'title':'CG.Mu.high.gc','description':'Top Decile in GC Content'}
+        pheader['mu_cpg_botgc'] = {'title':'CG.Mu.low.gc','description':'Bottom Decile in GC Content'}
+        pheader['cv_cpg'] = {'title':'CG.CV.gnm','description':'All CpGs'}
+        pheader['cv_cpg_topgc'] = {'title':'CG.CV.high.gc','description':'Top Decile in GC Content'}
+        pheader['cv_cpg_botgc'] = {'title':'CG.CV.low.gc','description':'Bottom Decile in GC Content'}
+
+        self.add_section(
+            name = 'Sequencing Depth',
```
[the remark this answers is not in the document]
zwdzwd: Thanks for the comment and showing me this option. In fact the `mu` column is the depth. `cv` column shows the variation. Sequencing depth variation characterize the **sequencing uniformity** and is one of the most important QC metrics in my point of view, particularly for single cell/ low input libraries. Many of the single-cell libraries we have are strongly enriched in coming from specific territories of the genome, like, the region low or high in GC content or peri-centromeric satellites. This has been captured quite well using the `cv` column. I hope that justifies keeping this section?

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,830 @@
...
+
+        self.add_section(
+            name = 'BISCUIT Insert Size Distribution',
+            anchor = 'biscuit-isize',
+            description = "This plot shows the distribution of insert size.",
+            plot = linegraph.plot(pd_isize, 
+                {'id':'biscuit_isize', 'title': 'Insert Size Distribution',
+                 'ymin': 0, 'yLabelFormat': '{value}%', 
+                 'tt_label': '<strong>Q{point.x}:</strong> {point.y:.2f}% of reads', 
+                 'ylab': '% Mapped Reads', 'xlab': 'Insert Size'}))
```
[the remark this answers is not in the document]
zwdzwd: This is added. Great suggestion!

multiqc/modules/biscuit/biscuit.py
```diff
@@ -0,0 +1,830 @@
...
+        pheader['bca'] = {'title':'b.CpA', 'format':'{:,.2g}', 'min':0, 'max':100, 'description':'CpA', 'suffix':'%'}
+        pheader['bcc'] = {'title':'b.CpC', 'format':'{:,.2g}', 'min':0, 'max':100, 'description':'CpC', 'suffix':'%'}
+        pheader['bcg'] = {'title':'b.CpG', 'format':'{:,.2g}', 'min':0, 'max':100, 'description':'CpG', 'suffix':'%'}
+        pheader['bct'] = {'title':'b.CpT', 'format':'{:,.2g}', 'min':0, 'max':100, 'description':'CpT', 'suffix':'%'}
+
+        self.add_section(
+            name = 'Cytosine Retention',
+            anchor = 'biscuit-retention',
+            description = "This plot shows cytosine retention rate. `r.` stands for read-averaging and `b.` stands for 'base-averaging.",
+            helptext = "**Cytosine retention rate** is 1.0 - **cytosine conversion rate**. Assuming full (complete but not over) bisulfite conversion, **cytosine retention rate** is the average cytosine modification (including 5mC, 5hmC etc) rate.",
```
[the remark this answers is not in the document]
zwdzwd: The format is changed accordingly. Great suggestion!

zwdzwd:
Thanks for the very careful code review @ewels!  They are very helpful. I have responded to all the comments and addressed most suggested changes. Let me know of further issues.

PR #691, opened 2018-02-26
Added support for supernova 2.0 reports

remiolsen:
@ewels This one should be pretty straight-forward, I think. See https://github.com/ewels/MultiQC_TestData/pull/78

Also fixed a couple of annoying bugs.

### DIFF
--- a/multiqc/modules/supernova/supernova.py
+++ b/multiqc/modules/supernova/supernova.py
@@ -29,11 +29,18 @@ class MultiqcModule(BaseMultiqcModule):
                 'suffix': 'Mb',
                 'scale': 'YlGn'
         }
+        self.headers['% missing 10Kb'] = {
+                'rid': 'pct_missing_10Kb',
+                'description': '% of base assembly missing from scaffolds >= 10 kb',
+                'suffix': '%',
+                'scale': 'YlGn',
+        }
         self.headers['# Long scaffs'] = {
                 'rid': 'num_long_scaffs',
                 'description': 'number of scaffolds >= 10 kb',
                 'scale': 'YlGn',
                 'format': '{:,.0f}',
+                'hidden': True
         }
         self.headers['Scaff N50'] = {
                 'description': 'N50 scaffold size (in kilobases)',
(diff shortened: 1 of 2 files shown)


multiqc/modules/supernova/supernova.py
```diff
@@ -266,8 +311,12 @@ def __init__(self):
 
         # Conditional sections
         if len(molecules) > 0:
-            # Remove the long tail
-            max_x = self.trim_tail(molecules, 100000)
+            # Remove the long tail, or fail if this is a legacy empty json file
+            try:
+                max_x = self.trim_tail(molecules, 100000)
+            except IndexError:
+                log.debug('The file histogram file is empty. Skipping molecule lenght section')
```
ewels: lenght-> length

multiqc/modules/supernova/supernova.py
```diff
@@ -289,9 +338,12 @@ def __init__(self):
...
             )
         if len(kmers) > 0:
-            # Remove the long tail
-            max_x = self.trim_tail(kmers, 50)
-
+            # Remove the long tail, or fail if this is a legacy empty json file
+            try:
+                max_x = self.trim_tail(kmers, 50)
+            except IndexError:
+                log.debug('The file histogram file is empty. Skipping kmers section')
```
ewels: file histogram file? 😕

multiqc/modules/supernova/supernova.py
```diff
@@ -113,11 +126,44 @@ def __init__(self):
                 'scale': 'OrRd',
                 'hidden': True
         }
+        self.headers['BC usage'] = {
+                'description': 'fraction of barcodes used; between 0 and 1',
+                'scale': 'OrRd',
+                'hidden': True
```
ewels: How come so many hidden columns? Is it worth moving these into a separate table instead perhaps?

remiolsen: Hmm. I try to make a judgement about what values are the most important to display so as not to clutter this table too much.

There are some loosely defined categories you can put these values in. For example 10X barcode metrics, assembly metrics, values that are computed from the kmer composition.

ewels: > Hmm. I try to make a judgement about what values are the most important to display so as not to clutter this table too much.

Yup for sure, I'm not suggesting making them all visible in the same table. I'm just thinking about the fact that most people won't realise that these are here at all. So given that there are quite a few, putting them in their own table may make them more accessible.

Does however mean that the scatter plot option won't work against stuff from other modules though.

I leave it up to you! Let me know if you'd like to merge.

remiolsen: We still have this [line](https://github.com/ewels/MultiQC/pull/691/files#diff-d6febcd24ea91cf804b5b6d7ff887727L233) in the report. But I'm sure people will miss it :wink:

I can think any tables to split out from the main one that would make sense. So go ahead and merge.

Thanks.

PR #701, opened 2018-03-06
[rseqc - read distribution] plot the number of intergenic reads

moxgreen:
In the current plot of rseqc read distribution, the intergenic reads that do not fall inside the predefined region arount TSS are not represented at all.
This pull request add a new feature in the plot that account for this reads, with the label "other_intergenic".
image

### DIFF
--- a/multiqc/modules/rseqc/read_distribution.py
+++ b/multiqc/modules/rseqc/read_distribution.py
@@ -50,6 +50,9 @@ def parse_reports(self):
                 d['{}_tag_count'.format(k)] = int(r_search.group(2))
                 d['{}_tags_kb'.format(k)] = float(r_search.group(2))
 
+        d['other_intergenic_tag_count'] = d['total_tags']-d['total_assigned_tags']
+        d['other_intergenic_total_bases'] = 0
+
         # Calculate some percentages for parsed file
         if 'total_tags' in d:
             t = float(d['total_tags'])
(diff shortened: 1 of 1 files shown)


multiqc/modules/rseqc/read_distribution.py
```diff
@@ -50,6 +50,9 @@ def parse_reports(self):
                 d['{}_tag_count'.format(k)] = int(r_search.group(2))
                 d['{}_tags_kb'.format(k)] = float(r_search.group(2))
 
+        d['other_intergenic_tag_count'] = d['total_tags']-d['total_assigned_tags']
+        d['other_intergenic_total_bases'] = 0
```
ewels: What's this line used for?

ewels:
Looks good anyway - will merge and then remove that line if that's ok 👍 

Thanks!

multiqc/modules/rseqc/read_distribution.py
moxgreen: Sorry for the late reply. Do you mean the row
``d['other_intergenic_total_bases'] = 0``
?
Probably I just inserted it trying to be consistent with other record that have *_tag_count, *_total_bases and *_tags_kb.

ewels: Yup - ok I see. The variable wasn't been accessed or written anywhere that I could see (from memory I think the others are set and then incremented at a later stage?). Removing it didn't seem to have any effect, and I figure it's better to have cleaner and clearer code 👍

PR #703, opened 2018-03-08
added trimmed reads per sample [WIP]

matthdsm:
Hi,

I'm trying to add a small tweak to the bcl2fastq module. I'd like to add the percentage of trimmed bases/sample to the general stats table or something similar.

I've already added most of the requisite code (I suppose), but I'd like to know if it's up standard for MutliQC. When running the script I get a KeyError for `trimmed_bases` on line 199, but I'm not really sure where to add the key.

Could you have a look?

Thanks
M

### DIFF
--- a/multiqc/modules/bcl2fastq/bcl2fastq.py
+++ b/multiqc/modules/bcl2fastq/bcl2fastq.py
@@ -137,7 +137,8 @@ class MultiqcModule(BaseMultiqcModule):
                     "perfectIndex": 0,
                     "filename": os.path.join(myfile['root'],myfile["fn"]),
                     "yieldQ30": 0,
-                    "qscore_sum": 0
+                    "qscore_sum": 0,
+                    "trimmed_bases": 0
                 }
                 run_data[lane]["total"] += demuxResult["NumberReads"]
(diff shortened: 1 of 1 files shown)


multiqc/modules/bcl2fastq/bcl2fastq.py
```diff
@@ -152,6 +153,7 @@ def parse_file_as_json(self, myfile):
                     run_data[lane]["qscore_sum"] += readMetric["QualityScoreSum"]
                     run_data[lane]["samples"][sample]["yieldQ30"] += readMetric["YieldQ30"]
                     run_data[lane]["samples"][sample]["qscore_sum"] += readMetric["QualityScoreSum"]
+                    run_data[lane]["samples"][sample]["trimmed_bases"] += readMetric["TrimmedBases"]
             undeterminedYieldQ30 = 0
             undeterminedQscoreSum = 0
             if "Undetermined" in conversionResult:
```
ewels: You missed adding the `TrimmedBases` stats from the `Undetermined`. So when the later code gets to this "sample" it crashes with a `KeyError`.

PR #706, opened 2018-03-12
Samtools stats: error rate description

vladsaveliev:
10x LongRanger pipeline doesn't report the NM tag, so Error rate is calculated to 0. By description, it look like it only uses CIGAR, so not clear why it might equal 0. Clarifying that.


ewels:
Looks good, thanks!

PR #705, opened 2018-03-09
Clip&Merge Module attempt

apeltzer:
My attempt at creating a module for MultiQC to parse the Clip&Merge data

PR #708, opened 2018-03-12
Dedup adjustments

apeltzer:
- Added a little helptext to DeDup 
- Added functionality to parse file ID from logfile (and if this is not there, just use basename as before)

### DIFF
--- a/multiqc/modules/dedup/dedup.py
+++ b/multiqc/modules/dedup/dedup.py
@@ -50,6 +50,7 @@ class MultiqcModule(BaseMultiqcModule):
 
     def parse_dedup_log(self, f):
         regexes = {
+            'input_file': r"Input file:\s+(\S+)",
             'total_reads': r"Total reads:\s+(\d+)",
             'reverse_removed': r"Reverse removed:\s+(\d+)",
             'forward_removed': r"Forward removed:\s+(\d+)",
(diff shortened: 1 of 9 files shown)


ewels:
Looks like this PR scooped up a bunch of commits from #705 - I'll take a look at that one first.

multiqc/modules/dedup/dedup.py
```diff
@@ -104,6 +108,7 @@ def dedup_alignment_plot (self):
         config = {
             'id': 'dedup_rates',
             'title': 'DeDup: Deduplicated Reads',
+            'description': 'Read categories that were either not removed (= unique reads) or removed (= duplicates).'
```
ewels: Same thing - good description but in the wrong place. I'll move it in a second 👍

multiqc/modules/dedup/dedup.py
```diff
@@ -61,16 +62,19 @@ def parse_dedup_log(self, f):
...
         try:
             parsed_data['not_removed'] = parsed_data['total_reads'] - parsed_data['reverse_removed'] - parsed_data['forward_removed'] - parsed_data['merged_removed']
         except KeyError:
             log.debug('Could not calculate "not_removed"')
 
         if len(parsed_data) > 0:
-            # TODO: When tool prints input BAM filename, use that instead
             s_name = self.clean_s_name(os.path.basename(f['root']), f['root'])
+            if 'input_file' in parsed_data:
+                s_name = self.clean_s_name(parsed_data['input_file'], f['root'])
```
ewels: If you could add an example of this type of file to MultiQC_TestData that would be great!

multiqc/modules/dedup/dedup.py
ewels: Ignore me sorry - this is in a table, so it _is_ correct. Ok, pretty confusing! 😕

ewels: No wait! It's a plot! So the first comment was correct!

Ok, deleting comments and getting a coffee sorry..

PR #710, opened 2018-03-14
Bcl2fastq tweak

ewels:
Follow on from #703


ewels:
Hi @matthdsm,

See the few commits that I added above - this prevents the crash and tweaks the output a little. Notably, the data I tested had 0% of bases trimmed for all samples, which is kind of boring. So now the table column is hidden by default if this is the case. If you can check that it's shown by default with your files that have some trimmed bases, that would be great 👍 

Thanks for this!

Phil

matthdsm:
Hi Phil,

Thanks for the help! I just checked with some of my own files, and the percentages are shown. Works like a charm! Thanks!

M

ewels:
Great! Thanks for letting me know!

PR #711, opened 2018-03-21
Fix path_filters for config.top_modules and config.module_order

cpavanrun:
This PR fixes a bug when specifying multiple globs for `path_filters` for `config.top_modules` or `config.module_order`.

For instance, using the following configuration:
```yaml
top_modules:
    - moduleName:
        name: 'Module (filtered)'
        info: 'This section shows the module with different files'
        path_filters:
            - '*_special.txt'
            - '*_others.txt'
```

Would only select filename searches that match `*.special.txt` **and** `*_others.txt`. Wheras you would want it to match `*.special.txt` **or** `*_others.txt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## MultiQC v1.6dev
 
-_..nothing yet.._
+#### Bug Fixes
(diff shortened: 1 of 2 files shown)


ewels:
Thanks!

PR #714, opened 2018-03-23
New feature: path_filters_exclude setting

cpavanrun:
This PR adds the `path_filters_exclude` and makes some slight improvements to path filtering (_e.g._ debug statements and filter _generators_ instead of _lists_).

The `path_filters_exclude` allows you to exclude certain files when running modules multiple times.

A use case would be as follows:

Take the input files which would cause duplicate sample names when ran together:
- S1.HsMetrics_trimmed.txt
- S1.HsMetrics.txt
...
- S10.HsMetrics_trimmed.txt
- S10.HsMetrics.txt

One could run the module twice and preventing duplicate sample names clashes by adding the following to your MultiQC configuration:
```yaml
top_modules:
    - moduleName:
        name: 'Module (trimmed)'
        path_filters:
            - '*_trimmed.txt'
    - moduleName:
        name: 'Module (not trimmed)'
        path_filters_exclude:
            - '*_trimmed.txt'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## MultiQC v1.6dev
 
+#### New MultiQC Features:
+* Add `path_filters_exclude` to exclude certain files when running modules multiple times. You could previously only include certain files.
+
 #### Bug Fixes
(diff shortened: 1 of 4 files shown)


ewels:
This looks great, thanks! 👍

PR #716, opened 2018-03-28
Remove enum34 dependency

boulund:
A quick fix that removes enum/enum34 dependency in gatk module. Replaced it with a namedtuple that works kind of the same way as the enum in the original code. Only a slight modification to the part of the code that called the `human_readable`  method on the original enum types. 

I ran with the original version and my version and diffed the two output report outputs and found no differences (other than the line with run date and time, of course). Not sure if that is enough to solidly verify that it doesn't break anything, but I hope it's an indication that probably works.


ewels:
Super cool, thanks @boulund!

PR #718, opened 2018-04-04
Add BBDuk 'stats' table summary to report

boulund:
A few quick edits to include the BBDuk kmer-based adapter/contaminant filtering stats in a table. Includes two hidden columns with the observed counts that BBDuk uses internally to compute the proportion that is the only column this module shows for this filetype by default. Also added the extension '.stats' for stripping from filenames, as that is the most common file naming convention for this type of statistics output from BBDuk.

### DIFF
--- a/multiqc/modules/bbmap/bbmap.py
+++ b/multiqc/modules/bbmap/bbmap.py
@@ -53,13 +53,14 @@ class MultiqcModule(BaseMultiqcModule):
                 log.debug("section %s has %d entries", file_type,
                           len(self.mod_data[file_type]))
 
-                self.add_section(
-                    name = file_types[file_type]['title'],
-                    anchor =  'bbmap-' + file_type,
-                    description = file_types[file_type]['descr'],
-                    helptext = file_types[file_type]['help_text'],
-                    plot = self.plot(file_type)
-                )
+                if file_types[file_type]['plot_func']:
+                    self.add_section(
+                        name = file_types[file_type]['title'],
+                        anchor =  'bbmap-' + file_type,
+                        description = file_types[file_type]['descr'],
+                        helptext = file_types[file_type]['help_text'],
+                        plot = self.plot(file_type)
+                    )
 
             if any(self.mod_data[file_type][sample]['kv']
                    for sample in self.mod_data[file_type]):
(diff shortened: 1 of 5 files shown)


multiqc/utils/config_defaults.yaml
```diff
@@ -160,6 +160,7 @@ fn_clean_exts:
     - '_peaks.xls'
     - '.relatedness'
     - '.cnt'
+    - '.stats'
```
ewels: This is a little generic perhaps, and I'm worried that it could appear in the middle of genuine filenames and cause incorrect truncation. Could it be moved to `fn_clean_trim` instead so that it's only removed if it's at the end of the filename?

multiqc/modules/bbmap/bbmap.py
```diff
@@ -96,7 +96,17 @@ def parse_logs(self, file_type, root, s_name, fn, f, **kw):
...
+                    if len(line) == 3 and file_type == "stats":
+                        # This is a special case for the 'stats' file type:
+                        # The first line _might_ have three columns if processing paired-end reads, 
+                        # but we don't care about the first line.
+                        # The third line is always three columns, which is what we really want.
+                        if line[0] == "File": 
+                            continue
+                        kv["% filtered"] = float(line[2].strip("%"))
+                        kv[line[0]] = line[1] 
+                        continue
```
ewels: Do we need this `continue`?

ewels:
Thanks @boulund - is this working with the [bbmap test data](https://github.com/ewels/MultiQC_TestData/tree/master/data/modules/bbmap/sample1)?

When I run MultiQC with your code on this folder, I can't find the new output in the report (`BBDuk filtering statistics` etc).

boulund:
Oh. Sorry. I forgot to put this new file type in the bbmap test data. I'll take care of that asap!

boulund:
Just created a PR with a `sample1.stats.txt` for the bbmap test data repo.

boulund:
I also added a line about the new functionality in the docs, forgot about that as well.

multiqc/utils/config_defaults.yaml
ewels: Yup, should happen with both as long as `.txt` is first in the list.

ewels:
Thanks for this @boulund! Sorry it took me so long to get to the PR. Merged, plus I've added a few bits of my own (21fc012ad1341b1a5d0d95a330c3bc73fbf3b809) to improve the display of the columns and make them fit in with other MultiQC modules.

PR #727, opened 2018-04-11
Set several modules too-global search patterns to share

cpavanrun:
Went over the search_patterns.yaml . 

Those modules which entries I thought too global had `shared: true` added to them in order to prevent them from incorectly claiming a file. Mostly these are entries with `fn` without any `content`.

This should reduce the number of times a module *eats* a log when it's search pattern had a false-positive hit. I don't think this will result in too much of a performance hit.

Cheers

### DIFF
--- a/multiqc/utils/search_patterns.yaml
+++ b/multiqc/utils/search_patterns.yaml
@@ -4,13 +4,15 @@
 
 adapterRemoval:
     fn: '*.settings'
-    shared: true
+    contents: 'AdapterRemoval'
+    num_lines: 1
 afterqc:
     fn: '*.json'
     contents: 'allow_mismatch_in_poly'
 bamtools/stats:
     contents: 'Stats for BAM file(s):'
     shared: true
+    num_lines: 2
 bbmap/stats:
     contents: '#Name	Reads	ReadsPct'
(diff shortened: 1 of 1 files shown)


multiqc/utils/search_patterns.yaml
```diff
@@ -4,9 +4,11 @@
 
 adapterRemoval:
     fn: '*.settings'
+    shared: true
 afterqc:
     fn: '*.json'
     contents: 'allow_mismatch_in_poly'
+    shared: true
```
ewels: This shouldn't be needed, as it also needs the string `allow_mismatch_in_poly` in the file which should be specific enough (it's not claiming all JSON files).

cpavanrun:
Sorry, had this laying around for a bit tooooo long =/.

I agree that making the search-pattern entries more specific via file contents would be the better solution. However, checking all the test-data files for optimal search depth and keywords is a bit of work... I might pick this up in the near future. If I don't manage: I think the proposed fix will suffice for now.

ewels:
Great stuff, thank you for this! Relatively minor changes, but could actually solve a lot of "bugs" for a lot of people! 👍

PR #728, opened 2018-04-11
Picard: ValidateSamFile

cpavanrun:
Hej @ewels,

This PR adds a submodule to the Picard module that parses ValidateSamFile output.

Cheers,
Chris

**Example and Test Files**
I also uploaded an [example report](https://surfdrive.surf.nl/files/index.php/s/qWMOI14MZkShljx/download) wich shows the test files. 
A test-data PR exists that introduces the test files: https://github.com/ewels/MultiQC_TestData/pull/82

**Performance Profiling**
Performance profiling suggests that most computation is required for building the table after data was generated rather than parsing or processing the actual files:
* 6/397 ms was needed for parsing the (5) test files. 
* 391/397 ms seconds was needed for the table `plot`.


multiqc/modules/picard/ValidateSamFile.py
```diff
@@ -0,0 +1,369 @@
...
+        'max': 1,
+        'colour': '252,146,114',
+        'format': '{:.0f}',
+        'hidden': True,
+    }
+
+    return headers
+
+
+def _generate_overview_note(pass_count, only_warning_count, error_count, total_count):
```
ewels: Nice approach with the custom HTML! It might be nicer to have a stacked progress bar to show the proportions... What do you think? eg. https://getbootstrap.com/docs/3.3/components/#progress-stacked

multiqc/modules/picard/ValidateSamFile.py
```diff
@@ -0,0 +1,369 @@
...
+                </li>
+                '''.format(pass_count, total_count, sample='samples' if total_count > 1 else 'sample')
+            )
+
+    note_html.append('</ul>')
+
+    return "\n".join(note_html)
+
+
+def _generate_detailed_table(data):
```
ewels: This table is way too wide, 60 columns with the test data. This isn't really usable. Could you pick around 5-10 columns max to show here?

Perhaps better than this table would be to have a bar chart showing the numbers of errors and warnings per sample.. The reason for the report is to highlight patterns and outlier samples, users can always go back to the original files to dig into the details of every category.

multiqc/modules/picard/ValidateSamFile.py
```diff
@@ -0,0 +1,369 @@
...
+
+
+def _get_general_stats_headers():
+    """
+        Returns a header dict for the general stats collumns
+    """
+    headers = OrderedDict()
+
+    headers['file_validation_status'] = {
+        'title': "File-Validation Status",
```
ewels: It's better to keep column titles as short as possible. Perhaps just `Validation`? Then Description as `ValidateSamFile (File Validation)`.

multiqc/modules/picard/ValidateSamFile.py
```diff
@@ -0,0 +1,369 @@
...
+        'description': 'pass/warn/fail',
+        'hidden': False,
```
ewels: This is the default value, so not needed. Though it could be nice to have `True` if all values `pass`, hiding the column by default if there's nothing interesting to show.

multiqc/modules/picard/ValidateSamFile.py
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    headers['WARNING_count'] = {
+        'title': "# Validation Warnings",
```
ewels: Can drop `Validation` to make this shorter.

multiqc/modules/picard/ValidateSamFile.py
```diff
@@ -0,0 +1,369 @@
...
+        'description': 'number of warnings',
+        'scale': 'Set3',
+        'max': 1,
+        'colour': '255,237,160',
+        'format': '{:.0f}',
+        'hidden': True,
```
ewels: Again, can hide by default if all passing and show if any are failing.

multiqc/modules/picard/ValidateSamFile.py
```diff
@@ -0,0 +1,369 @@
...
+        'description': 'pass/warn/fail',
+        'hidden': False,
+    }
+
+    headers['WARNING_count'] = {
+        'title': "# Validation Warnings",
+        'description': 'number of warnings',
+        'scale': 'Set3',
+        'max': 1,
+        'colour': '255,237,160',
```
ewels: Instead of specifying a fixed colour, better to use a warning-style [scale](http://multiqc.info/docs/#table-colour-scales-1) such as `YlOrBr`. This overwrites the above `Scale`

multiqc/modules/picard/ValidateSamFile.py
```diff
@@ -0,0 +1,369 @@
...
+        'format': '{:.0f}',
+        'hidden': True,
+    }
+
+    headers['ERROR_count'] = {
+        'title': "# Validation Errors",
+        'description': 'number of errors',
+        'scale': 'RdYlGn-rev',
+        'max': 1,
+        'colour': '252,146,114',
```
ewels: Setting a fixed colour overwrites the `scale`.

multiqc/modules/picard/ValidateSamFile.py
```diff
@@ -0,0 +1,369 @@
...
+        'max': 1,
+        'colour': '255,237,160',
+        'format': '{:.0f}',
+        'hidden': True,
+    }
+
+    headers['ERROR_count'] = {
+        'title': "# Validation Errors",
+        'description': 'number of errors',
+        'scale': 'RdYlGn-rev',
```
ewels: A good colour scale would be `Reds`

multiqc/modules/picard/ValidateSamFile.py
```diff
@@ -0,0 +1,369 @@
...
+        'title': "File-Validation Status",
+        'description': 'pass/warn/fail',
+        'hidden': False,
+    }
+
+    headers['WARNING_count'] = {
+        'title': "# Validation Warnings",
+        'description': 'number of warnings',
+        'scale': 'Set3',
+        'max': 1,
```
ewels: Why max 1? Is it not better to see a representation of how many errors there are?

multiqc/modules/picard/ValidateSamFile.py
```diff
@@ -0,0 +1,369 @@
...
+    headers['file_validation_status'] = {
+        'title': "File-Validation Status",
+        'description': 'pass/warn/fail',
+        'hidden': False,
+    }
+
+    headers['WARNING_count'] = {
+        'title': "# Validation Warnings",
+        'description': 'number of warnings',
+        'scale': 'Set3',
```
ewels: `Set3` is categorical rather than continuous, so not really appropriate for this data.

multiqc/modules/picard/ValidateSamFile.py
```diff
@@ -0,0 +1,369 @@
...
+        'max': 1,
+        'colour': '255,237,160',
+        'format': '{:.0f}',
+        'hidden': True,
+    }
+
+    headers['ERROR_count'] = {
```
ewels: You could add `shared_key` to both of these (eg. `'shared_key': "ValidateSamEntries"`). Then the background bar scales will be common between the two columns.

multiqc/modules/picard/ValidateSamFile.py
```diff
@@ -0,0 +1,369 @@
...
+def _parse_reports_by_type(self):
+    """ Returns a data dictionary
+
+        Goes through logs and parses them based on 'No errors found', VERBOSE or SUMMARY type.
+    """
+
+    data = dict()
+
+    for f in self.find_log_files('picard/sam_file_validation', filehandles=True):
+        sample = f['s_name']
```
ewels: Do these files usually have `.SUMMARY` and `.VERBOSE` appended? These can be added to the `fn_clean_trim` config list if so to remove automatically.

cpavanrun: No, these files have no default filename what so ever.

The suffixes have been added:
*  In capitals because of a '.summary' search pattern of a different module.
* As an easy reference to keep assit debugging summary and verbose parsing.

docs/modules/picard.md
```diff
@@ -45,3 +46,11 @@ picard_config:
         - 10
         - 50
 ```
+
+#### ValidateSamFile Search Pattern
+Generally, Picard adds identifiable content to the output of function calls. This is not the case for ValidateSamFile. In order to identify logs the MultiQC Picard submodule `ValidateSamFile` will search for filesnames that contain 'validatesamfile' or 'ValidateSamFile'. One can customise the used search pattern by overwriting the `picard/sam_file_validation` pattern in your MultiQC config:
```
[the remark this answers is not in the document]
cpavanrun: Done: 180c7519

multiqc/modules/picard/ValidateSamFile.py
```diff
@@ -0,0 +1,369 @@
...
+
+        if sample in data:
+            log.debug("Duplicate sample name found! Overwriting: {}".format(sample))
+
+        filehandle = f['f']
+        first_line = filehandle.readline().rstrip()
+        filehandle.seek(0)  # Rewind reading of the file
+
+        if 'No errors found' in first_line:
+                sample_data = _parse_no_error_report(f)
```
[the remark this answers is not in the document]
cpavanrun: Heh, nice spot. It's cool that Python doesn't even seem to care: 180c7519

multiqc/modules/picard/ValidateSamFile.py
```diff
@@ -0,0 +1,369 @@
...
+        else:
+                sample_data = _parse_summary_report(f)
+
+        data[sample] = sample_data
+
+    return data
+
+
+def _parse_no_error_report(file):
+    log.debug("Parsing {} as an error-free ValidateSamFile report.".format(file['fn']))
```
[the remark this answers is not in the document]
cpavanrun: Done: b7a3198e

multiqc/modules/picard/ValidateSamFile.py
cpavanrun: The original thought was that even a single error should be clearly highlighted. Independent of how many errors another sample had. But I guess it's nice to see how bad things are in perspective =P

ewels: You could set up the new conditional formatting feature to really highlight > 0 if you want, in addition to scale bars in the background..

multiqc/modules/picard/ValidateSamFile.py
cpavanrun: I think this is indeed nicer. It comes with the problem of wanting to label them. I changed it to use a progress bar: ad978a04

cpavanrun:
Dear @ewels ,

Thank you for the detailed review!

I have added nearly all of your suggested changes, some internally suggested changes and in the process learned a bunch. 

It has led to the following [example report](https://surfdrive.surf.nl/files/index.php/s/b5sVlXPkwo1MuYC/download).

Cheers,
Chris

ewels:
Hi @cpavanrun,

Sorry for the delay in getting back to you. This is looking great! Thank you for the changes 👍 

Phil

ewels:
Hi @cpavanrun,

I ran a test locally after merging and the progress bar wasn't working. I fixed it and whilst I was there I refactored a little and added some hover text to give a little more information to make it clear what the bars are showing.

Thanks again!

Phil

PR #730, opened 2018-04-16
Modified the clean_s_name function so that if the cleaned sample name become empty return the original file name

moxgreen:
I have a problem with the parsing of featurecounts summary. My bam  #files are named like that
``STAR/fastq/__ACTUAL_SAMPLE_NAME__.STAR/Aligned.sortedByCoord.out.ribo.ex.bam``
I found that ``clean_s_name `` in ``base_module.py`` convert this path to an empty sting, thus reporting only one record in the featurecounts plot.

I have introduced a workaround  that may be of general interest: just return the original file name if clean_s_name will convert this name in an empty string.

### DIFF
--- a/multiqc/modules/base_module.py
+++ b/multiqc/modules/base_module.py
@@ -174,6 +174,7 @@ class BaseMultiqcModule(object):
         :config.prepend_dirs: boolean, whether to prepend dir name to s_name
         :return: The cleaned sample name, ready to be used
         """
+        bak_name=s_name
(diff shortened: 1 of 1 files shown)


ewels:
Thanks @moxgreen! The workaround looks good.

Quick question before I merge - with this, will your samples not all now be called `Aligned.sortedByCoord.out.ribo.ex.bam` and still be merged into one?

PR #734, opened 2018-04-18
Add background samples to the peddy ancestry PCA plot

roryk:
This makes a plot more like the one in the peddy documentation, with the test samples overlaid on the samples for each ancestry (see https://www.dropbox.com/s/k0viyfjzjpc9i4q/Screenshot%202018-04-18%2014.54.18.png?dl=0).

### DIFF
--- a/multiqc/modules/peddy/peddy.py
+++ b/multiqc/modules/peddy/peddy.py
@@ -5,7 +5,7 @@
 from __future__ import print_function
 from collections import OrderedDict
 import logging
-import simplejson as sj
+import json
 
 from multiqc import config
 from multiqc.plots import scatter
@@ -61,7 +61,7 @@ class MultiqcModule(BaseMultiqcModule):
         # parse background PCA JSON file, this is identitical for all peddy runs,
         # so just parse the first one we find
         for f in self.find_log_files("peddy/background_pca"):
-            background = sj.loads(f['f'])
+            background = json.loads(f['f'])
             PC1 = [x["PC1"] for x in background]
             PC2 = [x["PC2"] for x in background]
             ancestry = [x["ancestry"] for x in background]
@@ -196,7 +196,19 @@ class MultiqcModule(BaseMultiqcModule):
         }
         default_color = '#000000'
         default_background_color = 'rgb(211,211,211,0.05)'
-        data = dict()
+        data = OrderedDict()
+
+        # plot the background data first, so it doesn't hide the actual data points
+        d = self.peddy_data.pop("background_pca", {})
+        if d:
+            background = [{'x': pc1,
+                        'y': pc2,
+                        'color': default_background_color,
+                        'name': ancestry,
+                        'marker_size': 1}
+                        for pc1, pc2, ancestry in zip(d['PC1'], d['PC2'], d['ancestry'])]
+            data["background"] = background
+
         for s_name, d in self.peddy_data.items():
             if 'PC1_het_check' in d and 'PC2_het_check' in d:
                 data[s_name] = {
@@ -204,14 +216,6 @@ class MultiqcModule(BaseMultiqcModule):
                     'y': d['PC2_het_check'],
                     'color': ancestry_colors.get(d['ancestry-prediction'], default_color)
                 }
-            if s_name == "background_pca":
-                background = [{'x': pc1,
-                               'y': pc2,
-                               'color': default_background_color,
-                               'name': ancestry,
-                               'marker_size': 1}
-                              for pc1, pc2, ancestry in zip(d['PC1'], d['PC2'], d['ancestry'])]
-                data["background"] = background
 
         pconfig = {
             'id': 'peddy_pca_plot',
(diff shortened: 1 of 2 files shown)


ewels:
Ah nice, looks good! I think we talked about doing this when we first wrote the module - how big is the resulting HTML file? Is it at all sluggish? Could you attach it so I can take a look please? (no need to just dropbox - you can just drag and drop files onto the comment box)

ewels:
screenshot 2018-04-18 14 54 18

roryk:
HTML report is still small, 1.3 MB. https://www.dropbox.com/s/utcdrd8x11slihh/multiqc_report.html?dl=0 (HTML isn't supported in the box).

ewels:
Good point, you need to zip it 👍

I also just did a test run and got a similar (minimal) difference - a pleasant surprise! Ok, then this is great - I have no problems in that case.

multiqc/modules/peddy/peddy.py
```diff
@@ -54,6 +58,16 @@ def __init__(self):
                             self.peddy_data[s_name].update(parsed_data[s_name])
                         except KeyError:
                             self.peddy_data[s_name] = parsed_data[s_name]
+        # parse background PCA JSON file
+        for pattern in ['background_pca']:
```
ewels: Why this loop?

multiqc/modules/peddy/peddy.py
```diff
@@ -54,6 +58,16 @@ def __init__(self):
...
+            sp_key = 'peddy/{}'.format(pattern)
+            for f in self.find_log_files(sp_key):
```
ewels: Am I right in thinking that these files should always be identical? Can we parse just the first one found if so?

roryk:
Sweet, thanks for the fast review. I don't understand what the change is for the dependencies, what is it? I click see review but it doesn't tell me much.

multiqc/modules/peddy/peddy.py
```diff
@@ -148,30 +162,64 @@ def peddy_general_stats_table(self):
...
+        default_background_color = 'rgb(211,211,211,0.05)'
         data = dict()
         for s_name, d in self.peddy_data.items():
             if 'PC1_het_check' in d and 'PC2_het_check' in d:
                 data[s_name] = {
                     'x': d['PC1_het_check'],
                     'y': d['PC2_het_check'],
+                    'color': ancestry_colors.get(d['ancestry-prediction'], default_color)
                 }
+            if s_name == "background_pca":
```
ewels: Instead of doing this inside the loop, could you move it out and do it first? That way the sample points should always be on top, so that the hover tooltips work (in my test they were partially hidden):
image

image

ewels: (it'll need to be an `OrderedDict`, but I think it should work...)

ewels:
>  I click see review but it doesn't tell me much.

Oops, me either. I must have missed the button when writing the specific comment sorry.

multiqc/modules/peddy/peddy.py
```diff
@@ -5,6 +5,7 @@
 from __future__ import print_function
 from collections import OrderedDict
 import logging
+import simplejson as sj
```
ewels: Please can you use the regular JSON import instead of adding a new dependency here. As done in other MultiQC scripts, eg:

https://github.com/ewels/MultiQC/blob/e50a2b8a9161365dea861372d03e3c3958d0ed26/multiqc/utils/util_functions.py#L7

multiqc/modules/peddy/peddy.py
roryk: Oh, that's a good idea, I will do that as a stopgap. I was hoping for the future we could like turn off the mouseover for those background points, make them the right ancestry colors and put borders around the real points if, point-specific features get implemented. ;)

multiqc/modules/peddy/peddy.py
roryk: Will do-- was thinking there were other json files to parse when I was writing it, but turns out not.

multiqc/modules/peddy/peddy.py
ewels: \* googles simplejson \*

> `json` is `simplejson`, added to the stdlib. But since `json` was added in 2.6, `simplejson` has the advantage of working on more Python versions (2.4+).
> https://stackoverflow.com/a/712799/713980

roryk: Totally, will swap over to it.

roryk:
Sweet, I swapped it to just add the background points once, and to try to plot them first; it doesn't seem to help out with the background points bleeding through the actual data points though.

roryk:
Also simplejson -> json

ewels:
Looks good, thanks! The background points are still the last item in the list though, right? I think that they probably need to be the first... Haven't tried tinkering yet though so this is a bit of a guess.

roryk:
Thanks, should be the first I think-- I load them up first, add them to the ordered dict and then add the rest of the points after that.

ewels:
Ah you're right sorry. Did you try at the end instead to see what happens? Happy to merge now anyway, thanks! 👍

roryk:
Thanks! I tried adding them last as well but it didn't help at all.

PR #739, opened 2018-04-23
Move to GNU style option naming

ingkebil:
Just noticed the options are mixed style between `_` and `-`. What about moving everything to GNU style options? 

See: https://www.gnu.org/prep/standards/standards.html#Option-Table

What do you think?

### DIFF
--- a/scripts/multiqc
+++ b/scripts/multiqc
@@ -84,7 +84,7 @@ logger = config.logger
                     multiple = True,
                     help = "Use only modules which tagged with this keyword, eg. RNA"
 )
-@click.option( '--view_tags',
+@click.option( '--view-tags',
                     is_flag = True,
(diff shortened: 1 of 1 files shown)


ingkebil:
It seems one can have multiple long-opts and it would be possible to leave the `--view_tags` and `--cl_config` as is while also appending a name which has consistent naming.

I'm not exactly sure why this PR is breaking the appveyor check. I would need some help fixing that one

ewels:
Appveyor is being dumb, don't worry about that - it's failing on everything.

Sounds great about having both long forms! Would you mind adding that to the PR?

ingkebil:
Here you go!

ewels:
Awesome! I will do the same in reverse to the others too ;)

PR #743, opened 2018-04-23
sample name is now ordered fixes #712

### LINKED ISSUE #712: Fastq_Screen row order
I recently upgraded from multiqc .9 to 1.4.  The FastQ Screen plot is no longer in an expected order. What controls the order and what, if anything, can I do to fix it? Example attached. I would like it ordered by user (the first directory name). 
fastq_screen-1-2

Version .9 had them in the same order as the "General Statistics" table which still orders by directory name.

hassanfa:
Spotted and fixed this together with Henrikki Almusa. Fixes #712


ewels:
Great! Thank you both.

PR #745, opened 2018-04-23
Make 'view_tags' paramter work without mandatory data dir

### LINKED ISSUE #741: Make 'view_tags' as eager option in click
**Description of bug:**
To use `--view_tags` we also give the required data directory though it is not actually needed. Easy way to handle this via `click` as explained [here](http://click.pocoo.org/5/options/#callbacks-and-eager-options)

* **Command used to run MultiQC**
    * `multiqc .`
* **MultiQC Version:**
    * MultiQC v1.5

senthil10:
This is resolve #741 so it just works with `multiqc --view_tags`

### DIFF
--- a/scripts/multiqc
+++ b/scripts/multiqc
@@ -36,6 +36,9 @@ logger = config.logger
 
 # Callback function to view available tags and modules
 def view_all_tags(ctx, param, value):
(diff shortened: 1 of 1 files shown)


senthil10:
Thank you @ewels 🙂 👍

Issue #740, opened 2018-04-23
Module request: Longranger


remiolsen:
Related to #413 . This one is more specific for Longranger (version >= 2.2) and writing a module for the WGS sub-pipeline. I will also take a look at the exome ("targeted") pipeline, but I am missing example data. 

* **Name of tool:**
    * Longranger wgs (targeted?)
* **Tool description:**
    * "A set of analysis pipelines that perform sample demultiplexing, barcode processing, alignment, quality control, variant calling, phasing, and structural variant calling."
* **Tool homepage:**
    * https://www.10xgenomics.com/
* **Log filename pattern:**
    * `summary.csv`
* **Most interesting data for General Stats table:**
    * molecule_length,number_reads,mean_depth,phaseblock_n50

remiolsen:
Example data. I am starting work on this @ewels 

[P5357_1006_grch38_gatk.zip](https://github.com/ewels/MultiQC/files/1938008/P5357_1006_grch38_gatk.zip)

ewels:
Added test data to the MultiQC_TestData repository 👍

PR #750, opened 2018-05-03
Longranger module

remiolsen:
Module that supports the output from `longranger wgs` and `longranger targeted`

See #740 
*and* 
https://github.com/ewels/MultiQC_TestData/pull/86

@ewels

### DIFF
--- a/multiqc/modules/longranger/longranger.py
+++ b/multiqc/modules/longranger/longranger.py
@@ -1,0 +1,120 @@
+#!/usr/bin/env python
+
+""" MultiQC module to parse output from Longranger """
+
+from __future__ import print_function
+from collections import OrderedDict
+import logging
+import re
+import os
+import csv
+
+from multiqc import config
+from multiqc.modules.base_module import BaseMultiqcModule
+from multiqc.plots import table
+
+# Initialise the logger
+log = logging.getLogger(__name__)
+
+class MultiqcModule(BaseMultiqcModule):
+    """ Longranger module """
+
+    def __init__(self):
+
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__(name='Longranger', anchor='longranger',
+        href="https://www.10xgenomics.com/",
+        info="A set of analysis pipelines that perform sample demultiplexing, "
+        "barcode processing, alignment, quality control, variant calling, phasing, "
+        "and structural variant calling.")
+
+        self.longranger_data = dict()
+        self.paths_dict = dict()
+        for f in self.find_log_files('longranger/invocation'):
+            sid = self.parse_invocation(f['f'])
+            self.paths_dict[os.path.basename(f['root'])] = sid
+        
+        running_name = 1
+        for f in self.find_log_files('longranger/summary'):
+            data = self.parse_summary(f['f'])
+            updir, _ = os.path.split(f['root'])
+            base_updir = os.path.basename(updir)
+            sid = "Longranger_{}".format(running_name)
+            if base_updir in self.paths_dict.keys():
+                sid = self.paths_dict[base_updir]
+            else:
+                running_name += 1
+            
+            self.longranger_data[sid] = data 
+
+        # Filter to strip out ignored sample names
+        self.longranger_data = self.ignore_samples(self.longranger_data)
+
+        if len(self.longranger_data) == 0:
+            raise UserWarning
+
+        self.headers = OrderedDict()
+        self.headers['molecule_length_mean'] = {
+                'description': 'placeholder'
+        }
+        self.headers['longranger_version'] = {
+                'description': 'placeholder'
+        }
+        self.headers['mean_depth'] = {
+                'description': 'placeholder'
+        }
+        self.headers['mapped_reads'] = {
+                'description': 'placeholder'
+        }
+        self.headers['corrected_loaded_mass_ng'] = {
+                'description': 'placeholder'
+        }
+        self.headers['snps_phased'] = {
+                'description': 'placeholder'
+        }
+        self.headers['n50_phase_block'] = {
+                'description': 'placeholder'
+        }
+
+
+
+        ### Write the report
+        self.write_data_file(self.longranger_data, 'multiqc_longranger')
+        config_table = {
+            'id': 'longranger_table',
+            'namespace': 'longranger'
+        }
+        self.add_section (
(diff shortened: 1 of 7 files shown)


multiqc/modules/longranger/longranger.py
```diff
@@ -0,0 +1,330 @@
...
+
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__(name='Longranger', anchor='longranger',
+        href="https://www.10xgenomics.com/",
+        info="A set of analysis pipelines that perform sample demultiplexing, "
+        "barcode processing, alignment, quality control, variant calling, phasing, "
+        "and structural variant calling.")
+
+        self.headers = OrderedDict()
+        self.headers['longranger_version'] = {
```
ewels: Could you instead add this to the `info` string, if all values are the same? Then perhaps only add this column if they differ.

multiqc/modules/longranger/longranger.py
```diff
@@ -0,0 +1,330 @@
...
+                'scale': False,
+                'title': 'Instrument ID',
+                'description': 'The list of instrument IDs used to generate the input reads.',
+                'hidden': True
+        }
+        self.headers['gems_detected'] = {
+                'scale': 'RdYlGn',
+                'title': '# gems',
+                'modify': lambda x: float(x) / 1000000.0,
+                'suffix': 'M',
```
ewels: I think that the standard way of doing millions etc. is to add the `M` in the title (suffix is more for `%` etc). So for this one: `'title': 'M Gems'`. Also please add `(millions)` to the end of the description text.

Same goes for the other columns.

multiqc/modules/longranger/longranger.py
```diff
@@ -0,0 +1,330 @@
...
+        self.headers['mean_dna_per_gem'] = {
+                'scale': 'RdYlGn',
+                'title': 'DNA per gem',
+                'modify': lambda x: float(x) / 1000000.0,
+                'suffix': 'M',
+                'description': 'The average number of base pairs of genomic DNA loaded into each GEM. This metric is based on the observed extents of read-pairs on each molecule.',
+                'hidden': True
+        }
+        self.headers['bc_on_whitelist'] = {
+                'scale': 'RdYlGn',
```
ewels: Please avoid using the same colour scheme for everything. It's best if no two adjacent columns share the same scheme, it makes the table much easier to read and interpret.

multiqc/modules/longranger/longranger.py
```diff
@@ -0,0 +1,330 @@
...
+        }
+        self.headers['r2_q30_bases_fract'] = {
+                'scale': 'PuBu',
+                'title': 'R2 Q30',
+                'description': 'Fraction of bases in R2 with base quality >= 30.',
+                'hidden': True,
+                'suffix': '%',
+                'modify': lambda x: float(x) * 100.0
+        }
+        self.headers['si_q20_bases_fract'] = {
```
ewels: This is a mega huge table, though you've hidden most columns. Normally I'd say that you should just remove these columns, or split into several tables / plots etc. But I have a feeling that you're going to say that it's useful to have them there so that you can do scatter plots..?

But still, may be worth removing some of these :)

multiqc/utils/search_patterns.yaml
```diff
@@ -212,6 +212,14 @@ kallisto:
 leehom:
     contents: 'Adapter dimers/chimeras'
     shared: true
+longranger/summary:
+    fn: '*summary.csv'
+    contents: 'longranger_version,instrument_ids,gems_detected,mean_dna_per_gem,bc_on_whitelist,bc_mean_qscore,n50_linked_reads_per_molecule'
+    num_lines: 2
+longranger/invocation:
+    fn: '_invocation'
+    contents: 'call PHASER_SVCALLER_CS('
```
ewels: I think that this is probably a specific enough filename to not need the `contents` string..

multiqc/modules/longranger/longranger.py
remiolsen: I also use `PuBu` and `YlGn` ;)
Then I would just alternate the columns. `YlGn`, `PuBu`, `YlGn`, ...

ewels: But you have [so many to choose from](http://multiqc.info/docs/#table-colour-scales-1)! 😉

multiqc/utils/search_patterns.yaml
remiolsen: Perhaps. But the other pipelines from 10X and their `martian` worflow engine output these `_invocation` files.

multiqc/modules/longranger/longranger.py
remiolsen: Yeah. These R1, R2, si, etc, they're a bit much.

multiqc/utils/search_patterns.yaml
ewels: ok, good reason to leave as it is then 👍

multiqc/modules/longranger/longranger.py
remiolsen: I'm not sure I understand. Can you give an example of a module that does this?

ewels:
```python
longranger_versions = set([d['longranger_version'] for d in self.longranger_data])
if len(longranger_versions) == 1:
    infotext += "All samples were processed with Longranger version {}".format(longranger_versions[0])
else:
    self.headers['longranger_version'] = {
    #...
```

multiqc/modules/longranger/longranger.py
remiolsen: colour bomb

ewels: I had to [look that one up](https://www.youtube.com/watch?v=T9MAmWnOznI) 😉

ewels:
@remiolsen - I just pushed some changes in 7a502123332e76dbedbad0957a5629a06bc67272:

* I reorganised the table so that the columns were shown in a more logical order (roughly chronological from start of experiment to final SV calling, right to left).
* I added two plots - one for phasing and one for alignment. I made these table columns hidden by default.
* A bunch of other more minor tweaking to stuff.

Can you take a look and check that it all still makes sense? Thanks!

remiolsen:
@ewels I like it. This is much better, thank you!

PR #755, opened 2018-05-04
prepend_dirs config value getting clobbered

tsnowlan:
If setting `prepend_dirs` in config files, it will get clobbered here since `dirs` is only not False when `-d` is specified.

Also checking for an unset attr, just in case that causes an issue somewhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Many of these updates are thanks to the efforts of people who attended the [NASP
 #### Bug Fixes
 * Fix path_filters for top_modules/module_order configuration only selecting if *all* globs match. It now filters searches that match *any* glob.
 * Empty sample names from cleaning are now no longer allowed
+* Stop prepend_dirs set in the config from getting clobbered by an unpassed CLI option ([@tsnowlan](https://github.com/tsnowlan))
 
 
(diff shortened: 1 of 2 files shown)


tsnowlan:
Probably best to go with the explicit `if dirs is True` then. `if dirs is not None` will still eval to True if `dirs = False`, clobbering the value again.

I wasn't sure if `config.prepend_dirs` would be unset or not, since I didn't see it being set in `utils/config.py`. I see now it's getting loaded in from `config_defaults.yaml`, so can leave out that extra check.

ewels:
You're completely right sorry, I'd forgotten that it was a boolean flag. In that case there's no risk of people using weird strings so `if dirs` is fine 👍

ewels:
Ah, you already fixed it 😆 Looks good, thanks!

ewels:
Sorry, final thing if you still have your editor open - would be great if you could add a line to the Changelog under bugfixes, then add yourself to the list of contributors if you're not already there.. (I need to add this to the PR template text..)

tsnowlan:
Updated changelog. Does that look about right? Wasn't quite sure where else to set contributor status, since I think the list on the repo gets generated automatically.

ewels:
Thanks! There's a list of names on the readme too. But maybe I should just replace that with a link to the dynamic page of contributors as you say, that would be easier and less likely to miss people.

PR #759, opened 2018-05-14
Make general stat work for custom content

senthil10:
If a `custom_content` need to be added in the _general stat_ table, now it would be possible with having a file in appropriate format

### DIFF
--- a/multiqc/modules/custom_content/custom_content.py
+++ b/multiqc/modules/custom_content/custom_content.py
@@ -202,10 +202,12 @@ def custom_module_classes():
                         hs[k] = v
                 gsheaders = hs
 
-            # Add namespace if not specified
-            for k in gsheaders:
-                if 'namespace' not in gsheaders[k]:
-                    gsheaders[k]['namespace'] = c_id
+            # Add namespace and description if not specified
+            for h in gsheaders:
+                if 'namespace' not in gsheaders[h]:
+                    gsheaders[h]['namespace'] = mod['config'].get('namespace', k)
+                if 'description' not in gsheaders[k]:
+                    gsheaders[h]['description'] = mod['config'].get('description', "General stat for custom module {}".format(k))
 
             bm.general_stats_addcols(mod['data'], gsheaders)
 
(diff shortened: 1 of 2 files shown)


multiqc/modules/custom_content/custom_content.py
```diff
@@ -202,10 +202,12 @@ def dict_constructor(loader, node):
...
                 gsheaders = hs
 
-            # Add namespace if not specified
-            for k in gsheaders:
-                if 'namespace' not in gsheaders[k]:
-                    gsheaders[k]['namespace'] = c_id
+            # Add namespace and description if not specified
+            for h in gsheaders:
+                if 'namespace' not in gsheaders[h]:
+                    gsheaders[h]['namespace'] = mod['config'].get('namespace', k)
```
ewels: If you're going to switch `k` for `h`, make sure you do it everywhere 😉

multiqc/modules/custom_content/custom_content.py
```diff
@@ -202,10 +202,12 @@ def dict_constructor(loader, node):
...
+                if 'description' not in gsheaders[k]:
+                    gsheaders[h]['description'] = mod['config'].get('description', "General stat for custom module {}".format(k))
```
ewels: I think it's better to leave this blank. The hover text always begins the with the `namespace`, which we have already collected above. So if undefined, it should already default to something pretty sensible. I don't think it's useful for people reading the report to know that the data was parsed with custom content.

Correct me if I'm wrong and the default hover text looks bad without this.

multiqc/modules/custom_content/custom_content.py
senthil10: hehe.. I was expecting this 😅

multiqc/modules/custom_content/custom_content.py
senthil10: No, I agree too, I was not so certain myself if this would be completely helpful. Just thought could get your opinion 😉

The default _description_ would be the same as column name. Which I guess is suffice 🤷‍♂️ , so I will remove this

multiqc/modules/custom_content/custom_content.py
senthil10: ah ha.. no, the reason why I switched for `h` is to be able to use `k` from the main `for` loop _scope_. Because earlier you tried to use `c_id`, I assume you wanted to use the "custom module" name rather than the "column name" _(which I too prefer)_. 

`c_id` would not work as intended as its `for` loop end before and it will just contain the module name from last iteration.

ewels: ahhh, ok cool - good spot! 👍

ewels:
Ok great - final thing, please could you add a note to the `CHANGELOG.md` and I'll merge 👍

PR #760, opened 2018-05-15
Kallisto fullpath samplename fix rescues -d or -dd option

tiennes:
Given that the sample_name from kallisto stdout contains a full path and -d is used:

base_module.clean_s_name at line:
``s_name = os.path.basename(s_name.split(ext['pattern'], 1)[0])``
will discard any prior directories for naming

example: 
```bash 
BEFORE: "compare-refs | kallisto_all_rf_test | /compare-refs/kallisto_all_rf_test/myname.fastq.gz"
AFTER: "myname.fastq.gz"
```
This is mitgated by using os.path.basename on the s_name.match

### DIFF
--- a/multiqc/modules/kallisto/kallisto.py
+++ b/multiqc/modules/kallisto/kallisto.py
@@ -4,6 +4,7 @@
 
 from __future__ import print_function
 from collections import OrderedDict
+import os
 import logging
 import re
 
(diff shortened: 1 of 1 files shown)


tiennes:
exactly,
thx for the quick merge

PR #769, opened 2018-05-28
Fix linegraph and scatter plots data conversion to work with data_labels.

### LINKED ISSUE #766: Picard WGS Coverage Counts Histogram
The Picard WGS Coverage "Counts Histogram" doesn't seem to work. This is using the test data from his git. 

image

image

cpavanrun:
Fixes #766 by making sure that any labels set in data_label overwrite those set in pconfig when converting the data holding dictionaries to a format that HighCharts can handle.

This is now the case for both `linegraph` and `scatter`. The rest of the plots didn't really seem to use any data-conversions based on `pconfig` content.

Cheers,
Chris

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ Some of these updates are thanks to the efforts of people who attended the [NASP
 * Stop prepend_dirs set in the config from getting clobbered by an unpassed CLI option ([@tsnowlan](https://github.com/tsnowlan))
 * Modules running multiple times now have multiple sets of columns in the General Statistics table again, instead of overwriting one another.
 * Prevent tables from clobbering sorted row orders.
+* Fix linegraph and scatter plots data conversion (sporadically the incorrect `ymax` was used to drop data points) ([@cpavanrun](https://github.com/cpavanrun))
 
 
 ## [MultiQC v1.5](https://github.com/ewels/MultiQC/releases/tag/v1.5) - 2018-03-15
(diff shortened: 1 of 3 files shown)


ewels:
Cool! So this is something that is documented already here, right? http://multiqc.info/docs/#switching-datasets

And this PR is because it was documented but not implemented? This makes me fractionally suspicious that it is implemented somewhere but broken or something.

Anyway, code looks good 👍 Will wait for the tests to pass then happy to merge, thanks!

A `git --rebase` might be nice 😉

cpavanrun:
This PR is to fix an issue. =P

In short: the pconfig `ymax` was used to drop data points. You would expect to be able to overwrite this in one of the data_labels. You can't. So if you have a large missmatch between Y-axis sizes (e.g. percentage vs actual counts) things become werid.

Mhhh a `git rebase` feels a bit dangerous if it's already in the cloud... I'll try though. Hold on!

ewels:
Yeah, but my point is that the docs have this function as an example already:

```python
config = {
    'data_labels': [
        {'name': 'Reads', 'ylab': 'Number of Reads'},
        {'name': 'Bases', 'ylab': 'Number of Base Pairs', 'ymax':100}
    ]
}
```

So it seems weird that I would add that as an example if it didn't already work! 😅 (But entirely possible)..

Yes - `git rebase` will have to be followed by `git push --force` (which is the scary bit). Don't worry if you're not sure - I can do a squash merge. Then just discard your branch and pull the upstream fork to master.

ewels:
Hmm.. Not quite sure what happened with your rebasing there, but all of your previous commits are still visible 😉 I'll just do a squash merge 👍

ewels:
Thanks very much!

cpavanrun:
Awesome! Forgot the `--force` on the push so it automerged. Tried to redo the rebase but you were quicker than I was =D.

PR #775, opened 2018-06-11
Add Hap.py module

tsnowlan:
Illumina's hap.py is useful for benchmarking control output against gold standards from GIAB or Platinum Genomes. This module generates a table from the csv and makes it possible to track quality over time in MegaQC.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -47,6 +47,7 @@ MultiQC Modules:
     featureCounts: modules/featureCounts.md
     GATK: modules/gatk.md
     goleft_indexcov: modules/goleft_indexcov.md
+    Hap.py: modules/happy.md
     HiCExplorer: modules/hicexplorer.md
     HOMER: modules/homer.md
     HTSeq: modules/htseq.md
--- a/docs/modules/happy.md
+++ b/docs/modules/happy.md
@@ -1,0 +1,7 @@
+---
+Name: Hap.py
+URL: https://github.com/Illumina/hap.py
+Description: >
+    Hap.py is a set of programs based on htslib to benchmark variant calls
+    against gold standard truth datasets. Som.py output not currently supported.
+---
--- a/multiqc/modules/happy/__init__.py
+++ b/multiqc/modules/happy/__init__.py
@@ -1,0 +1,4 @@
+from __future__ import absolute_import
+
+from .happy import MultiqcModule
+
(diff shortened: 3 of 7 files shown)


tsnowlan:
Need anything else for this one?

tsnowlan:
Nah, no big rush. I can use the fork branch until you get to it.

ewels:
A super fast glance over the code looks fine, however I'm generally against just making huge tables. If it's possible to cut down on the number of columns in the table (ideally ~ 5-8) and use some of the numbers to generate plots instead, that would be great. You can also split the table up into a few to make them easier to read, or just remove / hide columns which are not likely to be super interesting.

Phil

tsnowlan:
Yeah, the bulk of the columns are hidden, but I could probably hide some more.  It'll generally contain only one subject's data per batch, since it's just for testing how the control did, so having plots didn't seem terribly useful.

Honestly, this is mostly just a vehicle for getting the data into MegaQC for trend analysis. Might be good to eventually have some way of submitting data directly to MegaQC without having to go through MultiQC. A bit outside the scope of "MegaQC is for tracking multiple MultiQC runs" though.

ewels:
Yes, I have wondered previously about having an option for MultiQC to just save data to files / MegaQC objects and not produce anything in the report. It's kind of messy though.

ewels:
Ok, finally got to pulling your code and testing it. Works fine and report looks ok. A few minor requests before merging if that's ok:

- How come some columns are general stats and some are in their own table?
   - Given that there are so many columns in general stats, I think this would be better as it's own table in the module section.
   - One or two key stats that are useful for comparing against other modules could be suitable in general stats. Otherwise, just skip that table and have the second stand-alone only.
- More description text, especially drawing attention to the point that there are lots of hidden columns available
- Some help text, describing what the columns show and what to be expected if possible. Ok if it's a summary or only the columns shown by default.

Again, I prefer having plots where possible but I understand if there's nothing useful to plot here. Thanks again for this!

Phil

tsnowlan:
Sure, those all make sense. I'll get the changes made sometime today or tomorrow.

tsnowlan:
Or maybe in 2 weeks. Whoops.

How do these look?

ewels:
I noticed that the module was logging on every MultiQC run, even when no files were found, so I ended up doing a bit of refactoring. Maybe double check that the results are all the same - hope this is ok!

Phil

tsnowlan:
Ah, good call. Looks fine to me. Thanks!

PR #778, opened 2018-06-15
Add mean and sd of WGScoverage to general stats

cpavanrun:
By default these stats are hidden. In practice we tend to end up diving into the actual metrics file to find these two stats.

Also mention the changes in changelog

Cheers,
Chris


ewels:
Great stuff, thanks!

Issue #669, opened 2018-01-23
Module request: fastp


AMChalkie:
Not a bug exactly, but I am using fastp instead of afterqc and notice it is not supported, even though it is supposed to be and implementation of the same thing (not sure if/how the output file has changed in the reimplementation).

So: multiqc doesn't recognise fastp files (that are created to be the same as afterqc). 

Also a useful addition would be: 
- listing an invalid module choice in alphabetical order.

Thanks for the tool, it is great!

ewels:
Hi @AMChalkie,

You can see the example data from AfterQC that were used for the MultiQC module here: https://github.com/ewels/MultiQC_TestData/tree/master/data/modules/afterqc

MultiQC finds the output files from AfterQC using the following search rules: https://github.com/ewels/MultiQC/blob/master/multiqc/utils/search_patterns.yaml#L7-L9

If the output is identical and still matches these search patterns, it _should_ work. If you can identify the differences then we should be able to modify the module to work with both. If you're able to upload some example outputs from fastp to the MultiQC_TestData above in a comparable folder structure, that would be a great start 👍 

Regarding the alphabetical order thing, you mean this error message?

```
Error: Invalid value for "-m" / "--module": invalid choice: fastp. (choose from rna_seqc, theta2,
 peddy, bismark, preseq, fastqc, sargasso, sortmerna, cutadapt, deeptools, goleft_indexcov,
 bamtools, qorts, kallisto, busco, samtools, clusterflow, picard, disambiguate, rseqc, quast, interop,
 bbmap, gatk, hicup, bowtie2, qualimap, conpair, vcftools, featureCounts, star, adapterRemoval,
 flexbar, bcftools, supernova, rsem, salmon, fastq_screen, methylQA, htseq, hicexplorer, snpeff,
 verifybamid, bowtie1, trimmomatic, jellyfish, homer, bcl2fastq, prokka, slamdunk, leehom,
 biobloomtools, hisat2, afterqc, macs2, skewer, tophat, custom_content, samblaster)
```

Yes, that's probably a good idea! It was probably easy enough to read when I originally wrote that log message with about 5 modules, but times have changed since then 😉 

Phil

ewels:
Ok, module names now alphabetically sorted 👍 

[...]

PR #782, opened 2018-06-21
Add fastp module

florianduclot:
In reference to #669 , this PR adds a new module for reports created by [Fastp](https://github.com/OpenGene/fastp).

In its current state, this module mostly parses the JSON report from fastp to populate the General statistics table, and generate a barplot for filtering metrics. Fastp report a lot of other information, but as most of these are redundant with other popular tools (such as FastQC--to some degrees), these may not be needed. Maybe this should be the subject of future update to this module?

With regards to the code, `fastp.py` is mostly based on the _afterqc_ module, to which _fastp_ is a successor (same author). As this is my first attempt at a module, and python to this extent, the code and approach may not be optimal. I'm especially wondering whether errors are handled the best way possible. Thanks for any pointers/suggestions.

This module has been tested with the example JSON report provided by _fastp_ ([here](http://opengene.org/fastp/fastp.html)), and verified to work on Python 2 and 3 (both manually and Travis).

Thanks for considering adding this module and for any suggestion for improvement.


ewels:
Thanks for this @florianduclot! I'll get to reviewing it soon. Next couple of weeks are pretty packed but will be as quick as I can...

ewels:
Many thanks again for this @florianduclot! It's a great start and looks good. I think I'd like to do a bit of refactoring and adding additional plots, so I've changed your PR to point to a dev branch for this module - hope this is ok! I'll merge there and then do some more work on it myself. I'll then open up a second PR and ping you for feedback.

Cheers,

Phil

florianduclot:
Sounds good! Thanks for your time working on this!

As I briefly mentioned above, most plots could be considered redundant with other tools such as FastQC and that's why I didn't include them as I was worried it could make the multiqc report too busy by default for some. 
I do like the possibility to plot both read1 and read2 independently for most metrics, though, so I was entertaining this idea.
Looking forward to hearing more.
Cheers,

ewels:
I kind of think that if the MultiQC report plot has these metrics and plots, then people don't need to run FastQC and the other similar tools? People can always easily disable modules if they want anyway, so I'm not too concerned by that.

PR #787, opened 2018-06-28
Peddy: add suffixes to multi-file headers

chapmanb:
In doing more Peddy testing I discovered some missing columns from the main table due to naming mismatches. This fixes those and gets all the improvements in development for peddy working.

Peddy input files have duplicated columns, so the combined inputs get
file specific suffixes. Two of the main stats table columns were missing
because the header did not match the suffixed column name. This fixes to
include these.


ewels:
Great stuff!

PR #791, opened 2018-07-03
Module: fastp

ewels:
New module for fastp.

Started in #782 by @florianduclot and then added to by @ewels with a bunch of new plots and other tidying.

 - [x] `README.md` is updated

@florianduclot - if you could have a play, do some testing that would be brilliant!


ewels:
@florianduclot - I should also say, if you have any additional test data kicking around that would be very useful. Especially in cases where certain steps are not run (trimming, filtering etc). I've tried to keep the code so that it easily tolerates missing data, but it's a little tricky to be certain of it without seeing the different possible outputs from the tool.

ewels:
I'm going to merge this now as I'm happy with the code. But would be great to continue the discussion here :)

Issue #794, opened 2018-07-11
Skewer: don't use fixed ymax 100 for percentage plot


lucacozzuto:
Plots with percentage should NOT have a fixed of Y limit 100%. It is better to have a variable value depending on the highest value (i.e. like having the function Y-limits off).

A test case could be the Skewer plot.

ewels:
See discussion on Gitter: https://gitter.im/ewels/MultiQC?at=5b45cfab0a17e01fccd719eb

Specifically the Skewer `Read length after trimming` plot:

![screen-shot-2018-07-11-at-11 37 20](https://user-images.githubusercontent.com/465550/42564533-4c60075e-8501-11e8-85d0-a637e325fa9a.png)

![screen-shot-2018-07-11-at-11 37 27](https://user-images.githubusercontent.com/465550/42564528-497d8a5c-8501-11e8-8780-06983091b7fe.png)

Other plots may also need updating in a similar way. Fix is to not specify a fixed `ymax` but instead use `yCeiling` and `yMinRange`

vpresnyak:
I would love to not have a fixed yMax for the FastQC Sequence Duplication Levels plot as well - I'm getting plots that look like this:

![image](https://user-images.githubusercontent.com/36459283/42847669-a8f11c86-89ea-11e8-9e19-2ad21194d7fd.png)

The corresponding plot in FastQC itself looks much more interpretable with a higher y limit.

![image](https://user-images.githubusercontent.com/36459283/42847684-c21ecdb6-89ea-11e8-8074-5650bbb18e7e.png)

vpresnyak:
Working to implement these changes. The simplest solution seems to be simply remove the ymax limit and allow it to adjust to the ymax of the data. Is that an acceptable solution?

ewels:
Hi @vpresnyak,

> I would love to not have a fixed yMax for the FastQC Sequence Duplication Levels plot as well 

This is super weird! I've not seen this before. What version of FastQC are you using? I thought that plot always gave a maximum of 100% - not sure how you can have 2500% duplicate reads?

> Is there a specific reason to use yCeiling and yMinRange here?

Setting `yCeiling` to `100` prevents HighCharts from autoscaling the axis beyond that. So you don't get a percentage axis that goes to 120%.

[...]

PR #803, opened 2018-07-19
Fixes to axis limits

vpresnyak:
## This PR is _not_ a new module
---
There are 2 key changes here.

1. ymax limits have been removed for skewer read length plot and fastqc sequence duplication plot. Fixing a hard limit on the y axis was leading to undesired behavior in some cases. See #794 for examples.
2. Floor and Ceiling axis limits have been reworked. Floor limits now set ymin and Ceiling limits set ymax (floor below and ceiling above).

Both of these changes now seem to behave as expected in my hands, hopefully they will for others as well!


ewels:
Hi @vpresnyak,

This looks good! Any reason you closed it? Reopening now so that I don't forget it.

Phil

multiqc/plots/linegraph.py
```diff
@@ -341,13 +341,13 @@ def matplotlib_linegraph (plotdata, pconfig=None):
...
             ymin = pconfig['ymin']
-        elif 'yCeiling' in pconfig:
-            ymin = min(pconfig['yCeiling'], default_ylimits[0])
+        elif 'yFloor' in pconfig:
+            ymin = max(pconfig['yFloor'], default_ylimits[0])
         ymax = default_ylimits[1]
         if 'ymax' in pconfig:
             ymax = pconfig['ymax']
-        elif 'yFloor' in pconfig:
-            ymax = max(pconfig['yCeiling'], default_ylimits[1])
```
ewels: Mismatch between `yFloor` and `yCeiling` keys?? 😱 

This bug has been hiding there for a loooong time 😆 Great spot!

vpresnyak:
I had closed it because I figured out the sequence duplication scale was a FastQC version issue - I was using FastQC v0.10, which set the value at 1 to 100% and scaled from there. Versions of FastQC after 0.11 set the total to be 100%, so everything is fractional from there.

The Floor/Ceiling fix could still be useful.

ewels:
Yup, and the skewer fix should be good right? See the issue about FastQC - I think we can make this conditional on the max value being over 100%.

vpresnyak:
For the skewer, I just removed the ylim and let it scale with the data. It seemed appropriate in my testing, but I'm not sure if there are corner cases where that wouldn't be appropriate.

vpresnyak:
Just a note about the floor and ceiling functions: beyond just fixing the floor/ceiling mismatch, I also flipped which variable they set, which is a major alteration how these variable work.

In the existing code, `yCeiling` was setting a minimum value and `yFloor` was setting a maximum. It wasn't clear to me if that was intentional - normally it would be the opposite, with ceiling would imply adjusting the top of a graph and floor would imply adjusting the bottom. I've flipped them to match my expectations, but that may not match yours! Please give them a second look.

vpresnyak:
It's more than just the mismatch: the old code looked like this, where `yCeiling` was setting `ymin` to the smaller of the passed value and the default, and `yFloor` was setting `ymax` (and producing a mismatch with `yCeiling`) to the larger of the passed value or the default. 
```
        elif 'yCeiling' in pconfig:
            ymin = min(pconfig['yCeiling'], default_ylimits[0])
...
        elif 'yFloor' in pconfig:
            ymax = max(pconfig['yCeiling'], default_ylimits[1])
```
The new code fixes the mismatch, but also flips the relationship as I described above. `yCeiling` now sets `ymax` (to the _smaller_ of the passed value and the default, hence imposing a limit on the top of the plot) and `yFloor` sets `ymin` (to the _larger_ of the passed value and the default, imposing a limit to the bottom of the plot). I just wanted you to be aware of the change in behavior.

```
        elif 'yFloor' in pconfig:
            ymin = max(pconfig['yFloor'], default_ylimits[0])
...

        elif 'yCeiling' in pconfig:
            ymax = min(pconfig['yCeiling'], default_ylimits[1])
```

ewels:
Ok, gotcha 👌 I'm not sure that we should change this behaviour though. As I remember, the reason that I wrote the `min` test was to ensure that no data is hidden. If it should be a hard limit hiding data then we can just use `ymax` instead. In fact, thinking about it, this should mean that no fix is required for the FastQC plot as it should show all data anyway in the older versions for this reason.

Does that make sense? Am I understanding the situation correctly?

ewels:
Ok, I've gone and had a proper look at the code now and everything makes more sense. In short - yes you're totally correct and your fix is the way it I must have surely intended it to work originally (apologies - reviewing code on my phone whilst sat on a bus apparently isn't always ideal!).

Part of the reason that I was confused was that I thought this must be affecting all plots, and I didn't understand how that behaviour of the config could be broken. Looking now I see that it's for the flat matplotlib versions of the charts only, and when I run on the MultiQC_TestData repo with every single module, these conditional lines were not executed even once. After applying your changes manually, it is now applied in a handful of plots when using the flat plotting mode, and those all look correct.

Anyway - sorry for my incorrect concern and thank you for the fix! Great spot!

ewels:
Many thanks again for this! Note that I made a minor tweak in c4b93c52b289b2c930a8147a5d9d5cf9075f97cc so that the FastQC duplicate `ymax` is only removed if the data is larger than 100%.

PR #808, opened 2018-07-30
Modification of bcl2fastq module

ddesvillechabrol:
Hello !

I slightly modified the bcl2fastq module to have a BCL quality control for the platform NGS of Institut Curie:

- Top 20 of undetermined barcodes by lane are showed in a barplot.
- Informations for R1/R2 in "General Statistics" are splitted.
- Use sampleID instead of SampleName because in Chromium experiment several sample have the same SampleName.

I saw that I modified trimmed percent for trimmed bases for R1/R2.
Maybe you preferred R1_trimmed_percent and R2_trimmed_percent.


ewels:
Great stuff, thanks @ddesvillechabrol! I'll review this as soon as I get a chance, though it could be a few weeks sorry. In the mean time, it would be great if you could update the changelog, describing what you've done.

Phil

ewels:
Hi @ddesvillechabrol,

I'm really sorry for leaving this PR open for so long. I couldn't quite believe it when I sat down just now! I generally really like all of your modifications, with a couple of very minor exceptions:

* I prefer the original Lane Statistics table colour scales which were different from one another
* For the test data, the R1 and R2 trimmed columns are all zero. It would be good to check if this is the case and auto-hide those columns if so
* I like having the sample IDs and sample names for 10x data, but perhaps we can do this only if the two are different from one another, to avoid unnecessarily polluting the sample names.
* The new underdetermined barcodes plot is great! It would be nice if there could be a percentage view as well though, to see proportions across lanes with different outputs.

Given the very long review time and the fact that I'd like to push a release today, I will merge this into a branch and make these changes myself. I hope this is ok! Let me know if you have any thoughts on the above.

Thanks again!

Phil

ewels:
Oops, I merged to master. Ah well, I'll make the changes on master then I guess! 😆

ddesvillechabrol:
Ahaha !  Classic mistake with github PR 😄 
Everything is good for me ^^
Thanks Phil !

Dimitri

ewels:
Great! Happy Christmas!

PR #809, opened 2018-08-02
Add new module for miRTrace

chuan-wang:

 - [X] `README.md` is updated


ewels:
Hmm, tests are failing: https://travis-ci.org/ewels/MultiQC/jobs/411186222#L2581-L2594

multiqc/modules/mirtrace/mirtrace.py
```diff
@@ -0,0 +1,381 @@
...
+            self.parse_complexity(f)
+
+        # Filter to strip out ignored sample names
+        self.summary_data         =   self.ignore_samples(self.summary_data)
+        self.length_data          =   self.ignore_samples(self.length_data)
+        self.contamination_data   =   self.ignore_samples(self.contamination_data)
+        self.complexity_data      =   self.ignore_samples(self.complexity_data)
+
+        # Warning when no files are found
+        if len(self.summary_data) == 0:
```
ewels: You should only have one of these checks and raise if there are none of any of the reports. We want to have the module run if we find anything, right? For example:

```python
if max(len(self.summary_data), len(self.length_data), len(self.contamination_data), len(self.complexity_data)) == 0:
    raise UserWarning
```

You can then use these checks later when adding the report sections, eg:

```python
if len(self.length_data) > 0:
    self.add_section (
        name = 'Read Length Distribution',
        anchor = 'mirtrace_length',
        plot = self.mirtrace_length_plot()
    )
```

multiqc/modules/mirtrace/mirtrace.py
```diff
@@ -0,0 +1,381 @@
...
+                if s_name in self.summary_data:
+                    s_name = s_name + '_' + record['filename']
+                self.add_data_source(f, s_name)
+                self.summary_data[s_name] = parsed_data
+
+
+    # Parse a miRTrace mirtrace-stats-length.tsv file
+    def parse_length(self, f):
+        try:
+            with open(f['fn'], 'rb') as file:
```
ewels: If you use `self.find_log_files('mirtrace/length', filehandles=True):` then MultiQC will give you already opened file handles. Then you can simplify this code a little by removing the `with` statement:

```python
try:
    reader = csv.reader(f['f'], delimiter="\t")

File: multiqc/modules/mirtrace/mirtrace.py [hunk 3]
```diff
@@ -0,0 +1,381 @@
...
+                reader = csv.reader(file, delimiter="\t")
+                header = reader.next()
+                body = {}
+                for row in reader:
+                    body[row[0]]=row[1:len(row)]
+        except ValueError as e:
```
ewels: I don't really see why you need to use `try`/`except` to catch this `ValueError` if then all you do is re-`raise` it. I think you can get rid of these two lines and the code will work in exactly the same way.

multiqc/modules/mirtrace/mirtrace.py
```diff
@@ -0,0 +1,381 @@
...
+                parsed_data = {}
+                idx = header[1:len(header)].index(record)
+                for length in body:
+                    parsed_data[length] = int(body[length][idx])
+                if s_name in self.length_data:
+                    s_name = s_name + '_' + ''.join(random.choice(string.ascii_uppercase + string.ascii_lowercase + string.digits) for _ in range(6))
+                self.add_data_source(f, s_name)
+                self.length_data[s_name] = parsed_data
+        else:
+            log.debug('No valid data for read length distribution')
```
ewels: It's good to specify which file this log message relates to (mention `f['fn']`)

multiqc/modules/mirtrace/mirtrace.py
```diff
@@ -0,0 +1,381 @@
...
+
+        if header[0] == 'LENGTH':
+            for record in header[1:len(header)]:
+                s_name = self.clean_s_name(record, f['root'])
+                parsed_data = {}
+                idx = header[1:len(header)].index(record)
+                for length in body:
+                    parsed_data[length] = int(body[length][idx])
+                if s_name in self.length_data:
+                    s_name = s_name + '_' + ''.join(random.choice(string.ascii_uppercase + string.ascii_lowercase + string.digits) for _ in range(6))
```
ewels: Why do you need to do this? When would the `s_name` be the same? Usually in these cases we overwrite the previous data and flag a debug log message:
https://github.com/ewels/MultiQC/blob/dd262ff4bbc5926b2ace4135f4a7ed36c021085f/multiqc/modules/salmon/salmon.py#L45-L46

multiqc/modules/mirtrace/mirtrace.py
```diff
@@ -0,0 +1,381 @@
...
+        # Specify the order of the different possible categories
+        keys = OrderedDict()
+        keys['adapter_removed_length_ok']   =   { 'color': '#006837', 'name': 'Reads ≥ 18 nt after adapter removal' }
+        keys['adapter_not_detected']        =   { 'color': '#66bd63', 'name': 'Reads without adapter' }
+        keys['length_shorter_than_18']      =   { 'color': '#fdae61', 'name': 'Reads < 18 nt after adapter removal' }
+        keys['low_complexity']              =   { 'color': '#d73027', 'name': 'Reads with low complexity' }
+        keys['low_phred']                   =   { 'color': '#a50026', 'name': 'Reads with low PHRED score' }
+
+        # Construct a data structure for the plot - duplicate the samples for read 1 and read 2
+        data = {}
```
ewels: There's no need to create this data dict here as the structure and keys are the same as `self.summary_data`. MultiQC will ignore any data keys not specified in `keys`, so you can just call the function with the raw data directly here:

```python
return bargraph.plot(self.summary_data, keys, config)
```

multiqc/modules/mirtrace/mirtrace.py
```diff
@@ -0,0 +1,381 @@
...
+
+        # Config for the plot
+        config = {
+            'id': 'mirtrace_rna_categories_plot',
+            'title': 'miRTrace: RNA Categories',
+            'ylab': '# Reads',
+            'cpswitch_counts_label': 'Number of Reads'
+        }
+
+        return bargraph.plot(data, keys, config)
```
ewels: As above, you can just use `self.summary_data` directly instead of creating `data`

multiqc/modules/mirtrace/mirtrace.py
```diff
@@ -0,0 +1,381 @@
...
+
+
+    # miRTrace Contamination Check
+    def mirtrace_contamination_check(self):
+        """ Generate the miRTrace Contamination Check"""
+
+        # A library of 24 colors. Should be enough for this plot
+        color_lib = ['rgb(166,206,227)', 'rgb(31,120,180)', 'rgb(178,223,138)', 'rgb(51,160,44)', 'rgb(251,154,153)', 'rgb(227,26,28)', 'rgb(253,191,111)', 'rgb(255,127,0)', 'rgb(202,178,214)', 'rgb(106,61,154)', 'rgb(255,255,153)', 'rgb(177,89,40)', 'rgb(141,211,199)', 'rgb(255,255,179)', 'rgb(190,186,218)', 'rgb(251,128,114)', 'rgb(128,177,211)', 'rgb(253,180,98)', 'rgb(179,222,105)', 'rgb(252,205,229)', 'rgb(217,217,217)', 'rgb(188,128,189)', 'rgb(204,235,197)', 'rgb(255,237,111)']
```
ewels: Any reason not to use the default colours here?

multiqc/modules/mirtrace/mirtrace.py
```diff
@@ -0,0 +1,381 @@
...
+
+        idx = 0
+
+        # Specify the order of the different possible categories
+        keys = OrderedDict()
+        for clade in self.contamination_data[self.contamination_data.keys()[0]]:
+            keys[clade] = { 'color': color_lib[idx], 'name': clade }
+            idx = idx + 1
```
ewels: Shorthand for this is:

```python
idx += 1
```

You can also loop over a list and get the keys directly with `enumerate`:

```python
for idx, clade in enumerate(self.contamination_data[self.contamination_data.keys()[0]]):
    keys[clade] = { 'color': color_lib[idx], 'name': clade }
```

Though, if you do want to keep the custom colours, you should reset `idx` once it goes past 24 so that the colours recycle (otherwise this will break with > 24 categories if it ever happens)

multiqc/modules/mirtrace/mirtrace.py
```diff
@@ -0,0 +1,381 @@
...
+        # Config for the plot
+        config = {
+            'cpswitch_c_active': False,
+            'id': 'mirtrace_contamination_check_plot',
+            'title': 'miRTrace: Contamination Check',
+            'ylab': '# miRNA detected',
+            'cpswitch_counts_label': 'Number of detected miRNA'
+        }
+
+        return bargraph.plot(data, keys, config)
```
ewels: Can just use `self.contamination_data` directly instead of making `data` again I think.

multiqc/utils/config_defaults.yaml
```diff
@@ -295,7 +295,9 @@ module_order:
     - verifybamid:
         module_tag:
             - DNA
-
+    - mirtrace:
+        module_tag:
+            - smRNA
```
ewels: Maybe also `miRNA`

multiqc/modules/mirtrace/mirtrace.py
chuan-wang: Hi, I was thinking about the situation that multiple input fastq files were anlayzed for one sample. It will not happen based on our routine, but maybe someone else will run into such situation.

multiqc/modules/mirtrace/mirtrace.py
ewels: Ah ok, makes sense. The default bar plotting colours are from HighCharts actually, not colorbrewer. But the same thing applies - only 10 defaults before cycling: https://api.highcharts.com/highcharts/colors

Fine to keep it as there's a reason 👍

chuan-wang:
Hi Phil,

I have fixed the codes based on your comments. It seems that the
csv.reader() function does not work with file handlers (I tested f, f['f']
and f['fn'] but none worked). Therefore I switched to splitlines() insteads.

Best,
Chuan

PR #810, opened 2018-08-03
samtools: add "% proper pairs" to General Statistics table

bwlang:
title says it all...


ewels:
Thanks for this, looks good! One request - this currently shows up even if the data is single end, which is a little confusing. Can we hide the column by default if all samples are exactly `0.0%`? eg:

```python
'hidden': True if max([x['reads_properly_paired_percent'] for x in self.samtools_stats.values()]) == 0) else False
```

PR #820, opened 2018-08-28
Test

SherryH1229:

 - [ ] `README.md` is updated


SherryH1229:
Sorry for mis-pulling. I just want to pull to my forked branch.

ewels:
No problem, and thanks for the explanation :)

Note that if you have an internal-only module, then you can write it as a [plugin](http://multiqc.info/docs/#multiqc-plugins) (see [example](

ewels:
No problem, and thanks for the explanation :)

Note that if you have an internal-only module, then you can write it as a [plugin](http://multiqc.info/docs/#multiqc-plugins) (see [example](https://github.com/MultiQC/example-plugin)). Then the code and installation is separate from the main MultiQC code base and you don’t need to worry about pulling updates.

If this is for a tool that is open then it’s better to keep it as it is and add it back for everyone at some point in the future though.. :)

ewels:
(Oops, clumsy fingers on my mobile sorry)

PR #832, opened 2018-09-13
Stacks module

remiolsen:
* Eeeh. It complains locally about `LINT: Module 'phantompeakqualtools' found in docs/README.md, but not installed modules`
 - [x] `README.md` is updated
* One is 7 columns excluding the sample name. I couldn't trim it down further without losing sleep over it.


multiqc/modules/stacks/stacks.py
```diff
@@ -0,0 +1,287 @@
...
+
+        self.gsheaders = OrderedDict()
+        self.gsheaders['n_loci'] = {
+                'title': '# loci',
+                'description': 'Number of loci built',
+                'format': '{:,.i}',
+                'scale': 'RdYlGn'
+        }
+        self.gsheaders['n_used_fw_reads'] = {
+                'title': 'K reads used',
```
ewels: Instead of this, I wonder if it's better to use the built-in functionality to handle number multipliers: https://multiqc.info/docs/#number-base-multiplier

Did you have a reason not to use this?

multiqc/modules/stacks/stacks.py
```diff
@@ -0,0 +1,287 @@
...
+        }
+        self.add_section (
+            name = 'Sample statistics',
+            anchor = 'stacks-gstacks',
+            description = 'The sample specific statistics for Stacks',
+            helptext = '''**Note!** The sample names have the following scheme `<run folder name> | <input fastq file prefix>`.
+                        This data is obtained from the gstacks program run after builing sample and catalog loci merge
+                        paired-ends and call variants.
+                        These numbers are obtained from the `gstacks.log.distribs` file''',
+            plot = table.plot(self.cov_data, self.gsheaders, config_table)
```
ewels: Any reason not to use General Statistics here?

multiqc/modules/stacks/stacks.py
```diff
@@ -0,0 +1,287 @@
...
+                for i in range(1, len(content)):
+                    cdict[list(cdict.keys())[i-1]] = content[i]
+                if s_name is not None:
+                    out_dict[s_name+" | "+content[0]] = cdict
+                else:
+                    out_dict[content[0]] = cdict
+
+        return out_dict
+
+    def parse_sumstats(self, file_contents, s_name=None, f=None):
```
ewels: Better not to let `s_name` default to `None` here as there's no method for replacing that. It'd be a confusing error if this does end up being `None`, so better to let it crash I think.

multiqc/modules/stacks/stacks.py
remiolsen: I had the vague idea that these samples are ideally in a 1:1 ratio to the raw number of samples. This isn't always the case for Stacks. A common case is to run Stacks several times with varying parameters (where the column values of these samples can vary). Another less common case is that Stacks supports in-read barcodes, so that it can do additional demultiplexing.

multiqc/modules/stacks/stacks.py
ewels: No, it's deliberately report-wide instead of module specific. You could make a new module-specific config option to make this adjustable, but if you think that `K reads` will basically always give numbers in a sensible range then I guess it's fine to hardcode it...

ewels:
Remaining comments are very minor and can be addresses in later PRs if needed. Looks good, thanks! 👍

remiolsen:
Awesome! Thank you.

PR #851, opened 2018-10-30
QualiMap RNAseq: fix parsing integer 5'/3' bias

vladsaveliev:
Regex to parse 5'/3' bias presumes it's a value with a decimal point. However when the result is exact integer, it will be written by QualiMap without decimal points, failing MultiQC to parse it:

```
>>>>>>> Transcript coverage profile

    5' bias = 0.74
    3' bias = 0.71
    5'-3' bias = 1
```

Fixing to support such edge cases as well.


ewels:
Looks good, thanks!

PR #854, opened 2018-11-04
Adds minionqc module

ManavalanG:
 - [x] `README.md` is updated
 - [x] Everything that can be represented with a plot instead of a table is a plot (yes, if I understand this properly)


ewels:
Hi @ManavalanG,

This module is a great start, but needs a little bit of work still I think:

* The plots are good, but several stats are repeated in a few places.
* The YAML file contains some metrics spread over a range, it would be nice to plot this range instead of having bar plots of a single value

I'd like to get a MultiQC release out today, so I will quickly work on this myself and then merge if that's ok.

On the [MionQC homepage](https://github.com/roblanf/minion_qc#output-details-for-minion) there are loads of really nice plots which would be nice to add in too - do you know if there is any text / data for these plots saved? Or is it graphic output files only?

Phil

PR #874, opened 2018-12-11
Picard Module: Add a chart to RnaSeqMetrics for correct strand mapping.

bichkd:
Add a chart to display CORRECT_STRAND_READS and INCORRECT_STRAND_READS.

Add a warning if Picard CollectRnaSeqMetrics was run without a RIBOSOMAL_INTERVALS files. Tests whether the RIBOSOMAL_BASES field is blank, which seemed more direct than parsing the picard command line from the header.


ewels:
Hi @bichkd,

Many thanks for this addition - it's great! I made two minor additions before merging, I hope that's ok:

* The code now lists the sample names that are missing rRNA data (or how many there are if there are lots). This is because I thought I had a bug in the test data but was actually blind and just didn't notice that one of the samples was missing the data. So I figured that it would be helpful for others too! 😅 
* In my test data, the strandedness counts are all zero. So I made that plot only show for samples where the correct plus incorrect counts are > 0

Cheers,

Phil

PR #875, opened 2018-12-19
Add DamageProfiler

apeltzer:

 - [x] `README.md` is updated


apeltzer:
```

[ERROR  ]          report : LINT: >damageprofiler/damageprofiler.py< HTML ID was not clean ('damageprofiler-3'-misincorporation-plot' -> 'damageprofiler-3_-misincorporation-plot') ## plot = self.threeprime_plot()
[ERROR  ]          report : LINT: >damageprofiler/damageprofiler.py< HTML ID was not clean ('damageprofiler-5'-misincorporation-plot' -> 'damageprofiler-5_-misincorporation-plot') ## plot = self.fiveprime_plot()
```

apeltzer:
So plot should have red colors for the 3' first sample and blue color for the 5' first sample, which I couldn't get to work, unfortunately. All other samples can then proceed with whatever color is chosen by MultiQC.

apeltzer:
Tried doing this with this `color : <dict>` here, but that didn't work as expected... 

https://multiqc.info/docs/#line-graphs

apeltzer:
ToDos:

- [x] Make parsing YAML output work fine
- [x] Add lgdistribution data to plots 
- [x] Add 3' and 5' DNA damage plots 
- [ ] Add first two bases of % DNA damage to general table
- [ ] Compute average read length out of DamageProfiler histogram 
- [ ] Compute median read length out of DamageProfiler histogram

apeltzer:
@ewels Could you have a look at this? I can add more stuff if required, but wanted to get a brief overview on what to extend/change ... Python skills are still limited but I tried to document what I did + added as much details as possible what I want to do :-)

apeltzer:
Also a question: How can I compute a median/average read length of the `lgdistribution.yaml` data I have here? Can I use some libraries for this in MultiQC or is this not possible?

ewels:
> Fix HTML tag twice occuring linting error

It's not that it's occurring twice, it's that MultiQC had to adjust the ID that you gave it to be HTML safe. This is because the ID contains an apostrophe. Remove the `'` from the IDs and the lint error should go.

For the computation of the averages, it's fine to get MultiQC to work this out. This is super fast so not a problem.

multiqc/modules/damageprofiler/damageprofiler.py
```diff
@@ -0,0 +1,231 @@
...
+        """ Parse the JSON output from DamageProfiler and save the summary statistics """
+        try:
+            parsed_json = json.load(f['f'])
+        except Exception as e:
+            print(e)
+            log.warn("Could not parse DamageProfiler JSON: '{}'".format(f['fn']))
+            return None
+        
+        #Get sample name from JSON first
+        s_name = parsed_json['metadata']['sample_name']
```
ewels: Please clean sample names :) https://multiqc.info/docs/#custom-sample-names

multiqc/modules/damageprofiler/damageprofiler.py
```diff
@@ -0,0 +1,238 @@
...
+
+        # Filter to strip out ignored sample names
+        self.threepGtoAfreq_data         =   self.ignore_samples(self.threepGtoAfreq_data)
+        self.fivepCtoTfreq_data          =   self.ignore_samples(self.fivepCtoTfreq_data)
+        self.lgdist_fw_data   =   self.ignore_samples(self.lgdist_fw_data)
+        self.lgdist_rv_data      =   self.ignore_samples(self.lgdist_rv_data)
+
+
+        # Write parsed data to a file
+        #self.write_data_file(self.threepGtoAfreq_data, 'multiqc_damageprofiler_3pGtoAfreq')
```
ewels: I usually don't write plot data to files. There is already an option to do this from core MultiQC.

multiqc/modules/damageprofiler/damageprofiler.py
```diff
@@ -0,0 +1,238 @@
...
+    
+    #### Tables from here on 
+
+    def dmgprof_misinc_stats(self, dict_to_plot, readend, substitution):
+        """ Take the parsed stats from the DamageProfiler and add it to the
+        basic stats table at the top of the report """
+
+        headers = OrderedDict()
+        headers['1'] = {
+            'title': readend + ' ' + substitution + ' 1st base ',
```
ewels: Instead of concatenating strings like this, it's better to use `format()`:
```suggestion
            'title': '{} {} 1st base'.format(readend, substitution),
```

multiqc/modules/damageprofiler/damageprofiler.py
```diff
@@ -0,0 +1,238 @@
...
+            'description': readend + ' ' + substitution + ' 1st base ',
```
ewels: When you're ready, would be nice to write something a little more meaningful here 😉

multiqc/modules/damageprofiler/damageprofiler.py
```diff
@@ -0,0 +1,238 @@
...
+            'min': 0,
+            'max': 100,
+            'suffix': '%',
+            'scale': 'GrRd',
+            'format': '{:,.2f}',
```
ewels: I think this is the default and can be omitted..

multiqc/modules/damageprofiler/damageprofiler.py
```diff
@@ -0,0 +1,238 @@
...
+            'description': readend + ' ' + substitution + ' 2nd base ',
+            'min': 0,
+            'max': 100,
+            'suffix': '%',
+            'scale': 'GrRd',
+            'format': '{:,.2f}',
+            'modify': lambda x: x * 100.0
+        }
+
+        # Create new small subset dictionary for entries (we need just the first two data (k,v) pairs from each report)
```
ewels: I have no idea what you're doing here 😆

multiqc/modules/damageprofiler/damageprofiler.py
```diff
@@ -0,0 +1,238 @@
...
+            try:
+                data[s_name] = {int(d): int (dict_to_use[s_name][d]) for d in dict_to_use[s_name]}
+            except KeyError:
+                pass
+        if len(data) == 0:
+            log.debug('No valid data for forward read lgdist input!')
+            return None
+        
+        config = {
+            'id': 'length-distribution-' + orientation,
```
ewels: Would be nice to set `'xDecimals': False,`

multiqc/modules/damageprofiler/damageprofiler.py
```diff
@@ -0,0 +1,238 @@
...
+            'smooth_points': 50,
```
ewels: This seems kind of overkill? I usually smooth down to 500 - 1000 if there can be many thousands of points. If this is a read length plot for illumina data then no need to smooth I think.

multiqc/utils/search_patterns.yaml
```diff
@@ -204,6 +204,8 @@ cutadapt:
     contents: 'This is cutadapt'
     # contents: 'cutadapt version' # Use this instead if using very old versions of cutadapt (eg. v1.2)
     shared: true
+damageprofiler:
+    fn: 'dmgprof.json'
```
ewels: Sometimes nice to add an `*` at the start of a file pattern like this in case people add a filename prefix.

multiqc/modules/damageprofiler/damageprofiler.py
apeltzer: Done :-)

multiqc/modules/damageprofiler/damageprofiler.py
apeltzer: Removed this :-)

multiqc/modules/damageprofiler/damageprofiler.py
apeltzer: Done :-)

multiqc/modules/damageprofiler/damageprofiler.py
apeltzer: :-)

multiqc/modules/damageprofiler/damageprofiler.py
apeltzer: Wrote a little doc for that :-)

apeltzer:
I fixed all remaining linting errors, tested it and found no further issues :tada:

Also the summary metrics are now in place :-)

ewels:
This seems a little suspicious:
image

I don't think that this is an error in the MultiQC code, as both `std` and `median` have values of _exactly_ `37.474989985505474` in the test data: https://github.com/ewels/MultiQC_TestData/blob/master/data/modules/damageprofiler/dmgprof.json

I guess this is an upstream bug?

Phil

ewels:
Awesome stuff - thanks! Note that I made a few tweaks in  e834fb1 - please pull and have a check over them to see if you agree with what I did.

It would be great to get some description and help texts for each plot if possible. See the [FastQC output](https://multiqc.info/examples/rna-seq/multiqc_report.html#fastqc) for an example of what these look like in a report. These are the `description` and `helptext` keys when adding a section, see [the docs](https://multiqc.info/docs/#step-5-create-report-sections). If you could pop some of these in and do a new PR, that would be fab ✨ 

Phil

PR #880, opened 2018-12-21
Help & Description & real percentages

apeltzer:
Hi Phil!

this adds:

- descriptions for all sections
- helptext for all sections

Also, I multiply the % with 100 for the plots, to really get percentages in the misincorporation plots

### DIFF
--- a/multiqc/modules/damageprofiler/damageprofiler.py
+++ b/multiqc/modules/damageprofiler/damageprofiler.py
@@ -244,7 +244,7 @@ class MultiqcModule(BaseMultiqcModule):
         for key in self.threepGtoAfreq_data:
             pos = list(range(1,len(self.threepGtoAfreq_data.get(key))))
             #Multiply values by 100 to get %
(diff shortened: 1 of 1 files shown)


multiqc/modules/damageprofiler/damageprofiler.py
```diff
@@ -227,7 +243,9 @@ def threeprime_plot(self):
         # Create tuples out of entries
         for key in self.threepGtoAfreq_data:
             pos = list(range(1,len(self.threepGtoAfreq_data.get(key))))
-            tuples = list(zip(pos,self.threepGtoAfreq_data.get(key)))
+            #Multiply values by 100 to get %
+            tmp = [i * 100 for i in self.threepGtoAfreq_data.get(key)]
```
ewels: You can easily get caught out with variable types here if you're not careful. I basically always multiply by `100.0` so that you get floats with proper accuracy.

apeltzer: Added it :-)

Issue #881, opened 2018-12-23
samtools flagstat beeswarm: use read_count_multiplier


jfy133:
(apologies if this is in the wrong issue section)

**Issue**
When working with ancient DNA, we often have very low endogenous DNA and we get very few on-target reads to our organism (hundreds/thousands max).

The [nf-core/eager](https://github.com/nf-core/eager) pipeline utilises samtools for some of the statistics generation, however the samtools MultiQC module reports all numbers in millions of reads (e.g. 0.0M).

Thus, when I have e.g. 3,000 mapped reads  (example stats file attached), the plot cannot display the number and could misdirect a user to believe there is no mapping reads. The same also goes to the 'general statistics' table.

**Request**
Provide an option (or auto-detect) to show 'real numbers' rather than in millions of reads for the samtools module and/or general statistics table.

**Additional Info**
samtools flagstat: [ABM006.A0101_S0_L002_R1_001.sorted.stats.txt](https://github.com/ewels/MultiQC/files/2706060/ABM006.A0101_S0_L002_R1_001.sorted.stats.txt)

multiqc report: [multiqc_report.zip](https://github.com/ewels/MultiQC/files/2706061/multiqc_report.zip)

ewels:
Good news - this feature already exists! 🎉 

https://multiqc.info/docs/#number-base-multiplier

jfy133:
Does this also apply to the samtools plots though? 

When I tested those parameters with MultiQC 1.7 (with the above file) it only changed the general statistics table, and not the samtools plots).

ewels:
There has to be support written into the module code, so it's possible that it's lacking that.

..yes, this needs modifying: https://github.com/ewels/MultiQC/blob/c89816e5e47a7c2d7514c0f7be470f18ffd2f14a/multiqc/modules/samtools/flagstat.py#L54

jfy133:
Great thanks!

PR #883, opened 2019-01-07
Add read_count_multiplier functionality to flagstat plot

apeltzer:
This should address https://github.com/ewels/MultiQC/issues/881 

### DIFF
--- a/multiqc/modules/samtools/flagstat.py
+++ b/multiqc/modules/samtools/flagstat.py
@@ -52,7 +52,7 @@ class FlagstatReportMixin():
             reads = {
                 'min': 0,
                 'modify': lambda x: float(x) * config.read_count_multiplier,
-                'suffix': 'M reads',
+                 'suffix': '{} reads'.format(config.read_count_prefix),
                 'decimalPlaces': 2,
                 'shared_key': 'read_count'
             }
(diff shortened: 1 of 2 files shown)


multiqc/modules/samtools/flagstat.py
```diff
@@ -51,7 +51,7 @@ def parse_samtools_flagstats(self):
             keys = OrderedDict()
             reads = {
                 'min': 0,
-                'modify': lambda x: float(x) / 1000000.0,
+                'modify': lambda x: float(x) * config.read_count_multiplier,
                 'suffix': 'M reads',
```
ewels: Please can you use `config. read_count_prefix` here

ewels: And please add `config.read_count_desc` for the ones that have descriptions too..

ewels: `M reads` is short for `Million reads`. But it may not always be a million. `config.read_count_prefix` has either `M` or `K` or a blank string or whatever, which should correspond properly.

`read_count_desc` has the long form of this. See https://multiqc.info/docs/#number-base-multiplier

ewels:
```suggestion
                'suffix': '{} reads'.format(config.read_count_prefix),
```

The GitHub suggestion thing isn't working properly for me, but this is what I mean..

ewels: Ah, it was just the preview that wasn't working..

PR #898, opened 2019-02-05
Custom modules: fix assigning default namespace

vladsaveliev:
...and rename too generic variable names for readability.

A somewhat relate question. If I only populate general stats for a custom content, MultiQC will raise `UserWarning` and print `No samples found: custom_content`, and also skip generating the report at all if custom content is the only input (e.g. when testing). Wondering if that's intentional or just hard to work around? Not a big deal, just wondering.

Vlad

### DIFF
--- a/multiqc/modules/custom_content/custom_content.py
+++ b/multiqc/modules/custom_content/custom_content.py
@@ -191,7 +191,7 @@ def custom_module_classes():
 
     # Go through each data type
     parsed_modules = list()
-    for k, mod in cust_mods.items():
+    for module_id, mod in cust_mods.items():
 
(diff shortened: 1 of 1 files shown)


ewels:
Not intentional, but I did know about it. It's basically only possible with custom content I think, and then you can just use a table instead of general stats (if you're not trying to integrate with other tools).

I _should_ fix it (or at least make an issue about if there isn't already one), but it's a fairly low priority currently.

vladsaveliev:
Thank for merging!

Yeah, I figured you would never need that apart from testing of your custom content. Thanks for the clarification!

Issue #893, opened 2019-01-28
Hicexplorer module cannot parse log files


jakevc:
**Description of bug:**

The hicexplorer module broke when trying to parse log files from hicBuildMatrix.

**MultiQC Error log:**
```
(multiqc) -bash-4.2$ multiqc .
[INFO   ]         multiqc : This is MultiQC v1.7
[INFO   ]         multiqc : Template    : default
[INFO   ]         multiqc : Searching '.'
Searching 64 files..  [####################################]  100%
[ERROR  ]         multiqc : Oops! The 'hicexplorer' MultiQC module broke...
  Please copy the following traceback and report it at https://github.com/ewels/MultiQC/issues
  If possible, please include a log file that triggers the error - the last file found was:
    ./10kb/SRX764945_rs_QC/QC_table.txt
============================================================
Module hicexplorer raised an exception: Traceback (most recent call last):
  File "/home/users/vancampe/install/miniconda3/envs/multiqc/bin/multiqc", line 440, in multiqc
    output = mod()
  File "/home/users/vancampe/install/miniconda3/envs/multiqc/lib/python3.6/site-packages/multiqc/modules/hicexplorer/hicexplorer.py", line 42, in __init__
    self.mod_data[data_]['Pairs mappable, unique and high quality'] = self.mod_data[data_]['Pairs considered']
KeyError: 'Pairs considered'
============================================================
[WARNING]         multiqc : No analysis results found. Cleaning up..
[INFO   ]         multiqc : MultiQC complete
```

**File that triggers the error:**

```
$ cat rest_site/SRX764944_rs_QC/QC_table.txt
File	Pairs considered	Min rest. site distance	Max library insert size	Pairs mappable, unique and high quality	Pairs used	One mate unmapped	One mate not unique	One mate low quality	dangling end	self ligation (removed)	One mate not close to rest site	same fragment	self circle	duplicated pairs	inter chromosomal	short range < 20kb	long range	inward pairs	outward pairs	left pairs	right pairs
/home/groups/hoolock/u1/jvc/candidate-tad-boundaries/data/hicBuildMatrix/rest_site/SRX764944_rs.h5	117634613	300	1000	55359180	50535214	37858076	4687153	19730204	0	0	45121	4300781	0	478064	13164174	4983274	32387766	9280444	9354719	9374219	9361658
```

**MultiQC run details (please complete the following):**

Operating System: RHEL CentOS 7

```
conda list
[...]

PR #902, opened 2019-02-20
Updated hicexplorer search patterns, compatible with previous versions

jakevc:

 - [X] (optional but recommended): https://github.com/ewels/MultiQC_TestData contains test data for this change

 - [ ] `README.md` is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 * **VCFTools**
     * Fixed a bug where `tstv_by_qual.py` produced invalid json from infinity-values.
 * **Hicexplorer** 
-    * Fixed bug in search patterns, and added new test data for hicexplorer=2.2
+    * Fixed bug where module tries to parse QC_table.txt, a new log file in hicexplorer v2.2.
 
 #### New MultiQC Features:
 _..no updates yet.._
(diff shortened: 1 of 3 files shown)


jakevc:
Fixes https://github.com/ewels/MultiQC/issues/893

ewels:
Hi @jakevc,

Can you explain how this fixes the error in #893? I don't see how changing the search pattern would make any difference..?

Phil

jakevc:
The log files produced by hicBuildMatrix of hicexplorer=v2.2, produces a new log file QC_table.txt with a format not designed to be parsed by this module. It looks like the log file that was designed to be parsed by this module is QC.log(v2.2,v1.8,v1.7), or QC.txt(v2.0).

ewels:
Ok gotcha, thanks for the explanation 👍 

This filename fix is good, but it would be nice to also put something into the Python code to specifically ignore these new log files. People can (and often do) create their own search configs so it's good to be able to reject these from within the parsing code as a fallback.

Are you ok to add that to this PR as well? It could just be searching the file for a string that is unique to the file that we want to ignore (eg. `mappable, unique and high quality pairs` or something).

Phil

jakevc:
Okay great, I'll work on a more robust and integrated solution.
Thanks,
JVC

ewels:
No need to revert the search pattern change - that was good! Just an extra backup in the code would be a good bonus feature 🚀

jakevc:
This new change made it so we don't need the search pattern. Should it still be included?

ewels:
Ok great 👍

PR #911, opened 2019-03-03
Support opening MultiQC on websites with CSP `script-src 'self'` with some sha256 exceptions

lukaw3d:
So far MultiQC reports could not be opened on websites that use [Content-Security-Policy](https://content-security-policy.com/) header to limit XSS attacks (it would require `script-src 'unsafe-inline'`).

With this pullrequest, data passed into html is separated from javascript. This allows using CSP with hash exception for each script (because hash doesn't change with data anymore)
`script-src 'self' 'sha256-<script>' 'sha256-<script>' ...;`

It isn't a perfect solution, as server admins using CSP will have to update exceptions with MultiQC releases, but at least they can avoid `'unsafe-inline'`.

## List of CSP hash codes for this MultiQC version

```
Content-Security-Policy:
    script-src 
        'self'
        'sha256-X9boLn1Jzgn6VfdR/mbleypCzQB7wVaQ45tYrsWmfBQ=' // load mqc_compressed_plotdata and mqc_config
        'sha256-+F66a8s8ponoNnBnP06kgZ4LvQ6IBedFeTJHpTbA0O0=' // load_fastqc_passfails();
        'sha256-hBQTjaW1/EOhnTxMz3aTwQHziBXZnlSP6On3OlSNP0A=' // load_fastqc_seq_content();
        'sha256-6Iyez8LRQbIQe12+IHNW66k9OJvzlLF1yPb2gK11T5o=' // fastq screen inline script

        'sha256-hVVnYaiADRTO2PzUGmuLJr8BLUSjGIZsDYGmIJLv2b8=' // jquery-3.1.1.min.js
        'sha256-iWO8z1sbOFlW9jqHm6PDldXb2wCcqo3gBNPZxg/TBBs=' // jquery-ui.min.js
        'sha256-U5ZEeKfGNOja007MMD3YBI0A3OSZOQbeG6z2f2Y0hu8=' // bootstrap.min.js
        'sha256-QwUrND/LJhYUBcYRAAxtMNWjZhZI+Yr8N7fpvHSFd4E=' // highcharts.js
        'sha256-X5AUZ+9MCHrVWj/x7N3Rkc/02lN9YB7fEx9RODM7A+k=' // highcharts.heatmap.js
        'sha256-+9tTm/2PzXj/LVTpB+PDvRvk7tG4QGM5YZIKs4DE45o=' // highcharts.exporting.js
        'sha256-Vndz/rd9+SJEYoVVpV/BqYrULpr5CA+pwbE6rbynp/0=' // highcharts.offline-exporting.js
        'sha256-qRzmKGpTQ9kQfjgKtYTcIxkIAwipJOcF3c6psri9FAc=' // highcharts.export-csv.js
        'sha256-AhtCowp6HzQCEXZP0Bdbk1Jg9PEYtx8/jZkfHJgXY2c=' // jquery.tablesorter.min.js
        'sha256-mYqjlBuTYmeoEFTjuPCrwns2stAp2HOJyXR5X2xjP70=' // clipboard.min.js
        'sha256-FPJJt8nA+xL4RU6/gsriA8p8xAeLGatoyTjldvQKGdE=' // FileSaver.min.js
        'sha256-nRoO8HoupfqozUr7YKBRgHXmdx40Hl/04OSBzv7e7L8=' // lz-string.min.js
        'sha256-v69SxdPeknzXy0hkSHL/SrchLvO8QDQS4N7V50xE0Ks=' // jquery.toast.min.js
        'sha256-7gUXukWB8g8TYIxnHrzoQrukADhWL1n8vcdPVzDvf9o=' // multiqc.js
        'sha256-bXJ/DLuHU3+9ybEepMTgDM965n8Cp2jD7JEJZtsWxGc=' // multiqc_tables.js
        'sha256-+pOrbPU8yLRc3gIYFSyzUfvh3zDWjv4P+WgZr3tm8yg=' // multiqc_plotting.js
        'sha256-8kPZc83fQr/qzo5ISaYAcLl4M5meLuuTzHFVRGb4GhU=' // multiqc_mpl.js
        'sha256-1pWjCVoi5AoQvT52zb72qGCpIpfJzlc9trKcbcarXUY=' // multiqc_toolbox.js
        'sha256-AzuP74MO1xsRljs7mAfk56aLW7JT0TNECViHCvfj7c8=' // multiqc_fastqc.js
    ;
```

## This PR is _not_ a new module

### DIFF
--- a/multiqc/modules/fastqc/assets/js/multiqc_fastqc.js
+++ b/multiqc/modules/fastqc/assets/js/multiqc_fastqc.js
@@ -17,11 +17,19 @@ max_bp = 0;
 current_single_plot = undefined;
 
 fastqc_passfails = {}; // { <module>: { <section>: { <sample>: { data } } }
+fastqc_seq_content = {}; // { <module>: { <sample>: data } }
 
 function load_fastqc_passfails() {
     $('.fastqc_passfails').each(function (i, elem) {
         var key_value = JSON.parse(elem.innerHTML);
         fastqc_passfails[key_value[0]] = key_value[1];
+    });
+}
+
+function load_fastqc_seq_content() {
+    $('.fastqc_seq_content').each(function (i, elem) {
+        var key_value = JSON.parse(elem.innerHTML);
+        fastqc_seq_content[key_value[0]] = key_value[1];
     });
 }
 
@@ -33,7 +41,7 @@ function fastqc_seq_content_heatmap() {
     sample_statuses = [];
     var p_data = {};
     var hidden_samples = 0;
-    $.each(fastqc_seq_content_data, function(s_name, data){
+    $.each(fastqc_seq_content['fastqc'], function(s_name, data){
         // rename sample names
         var t_status = fastqc_passfails['fastqc']['per_base_sequence_content'][s_name];
         $.each(window.mqc_rename_f_texts, function(idx, f_text){
@@ -171,6 +179,8 @@ function fastqc_seq_content_heatmap() {
 
 // Set up listeners etc on page load
 $(function () {
+    load_fastqc_seq_content();
+    fastqc_seq_content_heatmap();
 
     // Go through each section in case FastQC is there multiple times
     $('.mqc-module-section').each(function(i){
@@ -315,7 +325,7 @@ $(function () {
     });
     $('#fastqc_per_base_sequence_content_plot').on('mqc_plotexport_data', function(e, cfg){
         if(cfg['ft'] == 'json'){
-            json_str = JSON.stringify(fastqc_seq_content_data, null, 2);
+            json_str = JSON.stringify(fastqc_seq_content['fastqc'], null, 2);
             var blob = new Blob([json_str], {type: "text/plain;charset=utf-8"});
             saveAs(blob, cfg['fname']);
         } else {
@@ -346,11 +356,11 @@ $(function () {
 
         // Update the key with the raw data for this position
         var hover_bp = Math.max(1, Math.floor((x/c_width)*max_bp));
-        var thispoint = fastqc_seq_content_data[s_name][hover_bp];
+        var thispoint = fastqc_seq_content['fastqc'][s_name][hover_bp];
         if(!thispoint){
             var nearestkey = 0;
             var guessdata = null;
-            $.each(fastqc_seq_content_data[s_name], function(bp, v){
+            $.each(fastqc_seq_content['fastqc'][s_name], function(bp, v){
                 bp = parseInt(bp);
                 if(bp < hover_bp && bp > nearestkey){
                     nearestkey = bp;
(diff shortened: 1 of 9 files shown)


multiqc/modules/fastqc/assets/js/multiqc_fastqc.js
```diff
@@ -24,12 +24,20 @@ function load_fastqc_passfails() {
         var key_value = JSON.parse(elem.innerHTML);
         fastqc_passfails[key_value[0]] = key_value[1];
     });
+
+    // Backward compatible global variables
```
lukaw3d: Last commit can be removed if there's no need to keep backward compatibility of global javascript variables

mstajdohar:
👍@lukaw3d
@ewels This feature is quite important for us. What's holding you back?

ewels:
Sorry, just became a dad to two new twins - hopefully a pretty good excuse, though I appreciate it doesn't help you much! I will get to this PR and everything else when I can, but it could be a while I'm afraid. I'll do my best though..

lukaw3d:
Ah that's fun, good luck with them :D

multiqc/modules/fastqc/assets/js/multiqc_fastqc.js
ewels: These files are basically always bundled together into a single HTML file, so I don't think that there is any need for backwards compatibility, no 👌

ewels:
@lukaw3d - out of curiosity, is this only a problem when using the `default_dev` template? That is, when the HTML file references external javascript files? The `default` template includes all of these javascript files inline in to the HTML file so should be fine... Right?

lukaw3d:
Hello :)

It is a problem for both templates, because it came from inline scripts with variable content (inline scripts that included data in script text).

<br>

( inline scripts are blocked by CSP, unless whitelisted by their content's sha256
```html
before:
<script type="text/javascript">mqc_compressed_plotdata = '{{ report.plot_compressed_json }}';</script>
= infinite sha256s


after:
<script type="text/plain" id="mqc_compressed_plotdata">{{ report.plot_compressed_json }}</script>
<script type="text/javascript">mqc_compressed_plotdata = document.getElementById('mqc_compressed_plotdata').innerHTML;</script>
= sha256-Lzz6mIjrROikTpWwcmmXGEOCVptLy+sUXzO+IgFNZmI=
```

ewels:
Ok nice, thanks for the explanation 👍 If you could remove the backwards-compatible code then I'll be happy to merge.

One note on the CSP hash codes in your first comment: I wonder if MultiQC can automatically calculate these hashes and print them somewhere useful.. eg. In the comment block in the HTML `head`..? What do you think?

Phil

lukaw3d:
I have an idea. How hard is it for a unittest to create a report.html with all modules included?

lukaw3d:
I made https://github.com/lukaw3d/MultiQC/commit/a8eb8099d85006dca111cd9ba3941dae82a58dc3
- would add [CSP_HEADER.txt](https://github.com/lukaw3d/MultiQC/blob/csp-headers-test/CSP_HEADER.txt)
- travis would run `cd test && multiqc ../test_data/data/` to create report with all(?) modules
- and [test/browser-test.js](https://github.com/lukaw3d/MultiQC/commit/a8eb8099d85006dca111cd9ba3941dae82a58dc3#diff-d0dfbc5a966f6cba67a192f9c1f00924R88) that prints new sha256 when CSP_HEADER.txt is outdated
  (it opens /test/multiqc_report.html in headless browser with CSP set through markup and logs errors)

ewels:
Wow! That is some impressive stuff 🚀 

The hashes are just sha256 of static files though, right? Can we not just do that in Python within MultiQC? I just wonder if this would be simpler and not require adding in new test complexities.

Also I guess that with Travis we can detect when changes are required, but if we generate in Python then we can just always provide an up to date file at run time without needing any manual updates..

lukaw3d:
Yes, maybe :D
I have counterpoints:
- I was a bit worried about possibly inaccurate parsing by python compared to a browser
- Or if any scripts are added by javascript libraries in runtime
- It would be nice to keep hashes for old versions in the list too (for servers hosting multiple versions) 
  - either keep old reports for python to go through, or have an additive list of hashes
- With the new test framework, it is easy to add other runtime tests, e.g. build a report with all modules duplicated and check if all visualizations show up after a delay
`duplicatesReport.find('.hc-plot').length === 2 * normalReport.find('.hc-plot').length`

multiqc/templates/default/head.html
```diff
@@ -16,8 +16,26 @@
 <title>{{ config.title + ': ' if config.title != None }}MultiQC Report</title>
 
 <!-- JSON plot data -->
+<script type="text/plain" id="mqc_compressed_plotdata">{{ report.plot_compressed_json }}</script>
+
+<script type="application/json" id="mqc_config">{{
+{
+    "num_datasets_plot_limit": config.num_datasets_plot_limit,
+    "sample_names_rename": config.sample_names_rename,
+    "decimalPoint_format": config.decimalPoint_format,
```
ewels: I think that this can print the text `None` if it's not set, which breaks stuff. This is why it was inside a conditional statement in the old code.

ewels:
The first three are pretty minor I think 😉 But you're completely right about adding HTML tests for other stuff. I've been thinking about doing this for years, but never got around to figuring out how such tests would work.

Ok cool - let's wrap up this PR and then start a new one with the npm test suite..

Phil

multiqc/templates/default/head.html
lukaw3d: This combination works well because ` | tojson` converts None into null

Problematic code:
```html
<script type="application/json" id="mqc_config">
{
    "num_datasets_plot_limit": {{config.num_datasets_plot_limit}},
    "sample_names_rename": {{config.sample_names_rename | tojson}},
    "decimalPoint_format": "{{config.decimalPoint_format}}",
    "thousandsSep_format": "{{config.thousandsSep_format}}",
}
</script>
```

ewels: Ok great, thanks!

Issue #886, opened 2019-01-13
Export static plots from fastqc results in incorrect colour assignment to background


abeu1990:
**Description of bug:**
When using the export function in multiqc to get png files from fastqc output the colour assignment for pass/fail is incorrect. 0-10 is red, 10-15 is green, 15-25 is yellow and 25-40 is green. This does not seem to be a problem if you generate the html without static images.  

See attached file. 

![mqc_fastqc_per_base_sequence_quality_plot_1](https://user-images.githubusercontent.com/46658905/51091826-38891100-17e4-11e9-9c75-50d5ebc8d407.png)

ewels:
Fixed in #916 by @epakarin

PR #916, opened 2019-03-04
#886 Fixing matplotlib plot bands

epakarin:
The default alignment for matplotlib plot bars is to be centered, but the relevant (x/y)PlotBands array is originally constructed (e.g. in the fastqc module) with the assumption that the bar should begin at the specified point. This fixes the broken plot bars when MultiQC auto-switches from highcharts to matplotlib.

### DIFF
--- a/multiqc/plots/linegraph.py
+++ b/multiqc/plots/linegraph.py
@@ -401,11 +401,11 @@ def matplotlib_linegraph (plotdata, pconfig=None):
         if 'yPlotBands' in pconfig:
             xlim = axes.get_xlim()
             for pb in pconfig['yPlotBands']:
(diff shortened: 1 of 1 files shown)


ewels:
Hi @epakarin,

Sorry it's taken so long for me to get to this. The change looks fine, but I don't quite understand what's broken - I don't remember seeing any plot bands that look wrong? Any chance you could post a before and after image?

Cheers,

Phil

epakarin:
Hey,

I've posted this fix in response to issue #886 which contains a screenshot, but below are before-and-after images from when we ran into this issue ourselves. The issue is caused by missing explicit alignment options. Plot bands end up getting centered on the Y-values they should be edge-aligned with. As a result, a quality score of e.g. 13 ends up in the green, whereas it should be deep in the red zone.

pre-patch
MultiQC_plot_fix_before

post-patch
MultiQC_plot_fix_after

ewels:
Perfect, thank you for the explanation!

PR #922, opened 2019-03-08
Enable syntax highlighting for Snakemake example

johanneskoester:


johanneskoester:
Just saw you already activated yaml highlighting. Thanks, in fact it seems to look even better for such simple examples.

ewels:
Yes, it definitely helps - thanks!

PR #927, opened 2019-03-19
Add TOC to pipeline examples

johanneskoester:
By that, we make available examples visible from the start.


ewels:
The side navigation already has a table of contents showing the available examples:


..and if you view the docs on a mobile device, you'll see that there's actually an in-line ToC there already: 


..it's just hidden by default on larger devices because there's room to show the side navigation.

I agree that it wouldn't hurt to highlight that there are multiple examples though. Could you do this as a sentence instead of a ToC though, for consistency? Or just a list without the heading.

Cheers,

Phil

johanneskoester:
Didn't see that, thanks!

Issue #924, opened 2019-03-12
Module request: add module for Mosdepth


fgvieira:
**Tool Details**
 - Name of tool: Mosdepth
 - Tool homepage: https://github.com/brentp/mosdepth
 - Tool description: fast BAM/CRAM depth calculation for WGS, exome, or targeted sequencing
 - Data suitable for MultiQC plot(s): cumulative coverage density plot, coverage density plot (not calculated by `mosdepth`, but can be inferred; see code below)
 - Most interesting data for General Stats table: fraction of sites above X coverage; it is not given by `mosdepth` but can be calculated from `*.thresholds.bed.gz`

**Tool output:**
 - Log filename pattern: `*.mosdepth.region.dist.txt` if available. If previous not available, then `*.mosdepth.global.dist.txt`. And also `*.thresholds.bed.gz`
 - File contents

[$prefix.mosdepth.region.dist.txt](https://github.com/ewels/MultiQC/files/2957217/mosdepth.mosdepth.region.dist.txt)

[$prefix.mosdepth.global.dist.txt](https://github.com/ewels/MultiQC/files/2957222/mosdepth.mosdepth.global.dist.txt)

[$prefix.thresholds.bed.txt](https://github.com/ewels/MultiQC/files/2957224/mosdepth.thresholds.bed.txt)

**NOTES**

* To increase precision, mosdepth should be run as:
```
MOSDEPTH_PRECISION=5 mosdepth <args>
```

* It seems that someone already tried to include `mosdepth` into MultiQC on the repo [bcbio](https://github.com/MultiQC/MultiQC_bcbio/blob/master/multiqc_bcbio/bcbio.py)

* Here is some R code to illustrate the plotting of both distributions
```r
#!/usr/bin/env Rscript

library(ggplot2)

input_file <- "$prefix.mosdepth.global.dist.txt"

df <- read.table(input_file, sep = "\t", header = FALSE, col.names = c("chr", "coverage", "percent"))
df <- subset(df, chr == "total")
df$diff <- c(df$percent[1], diff(df$percent))

## Cumulative sum plot
ggplot(data = subset(df, percent > 0.02)) + geom_line(aes(x = coverage, y = percent))

## Histogram-like plot
ggplot(data = subset(df, percent > 0.02)) + geom_line(aes(x = coverage, y = diff))
```

ewels:
Thanks @fgvieira!

[...]

PR #913, opened 2019-03-04
Add snpeff effects

sachalau:
- [x] This comment contains a description of changes (with reason)
 - [x] `CHANGELOG.md` has been updated
 - [ ] (optional but recommended): https://github.com/ewels/MultiQC_TestData contains test data for this change

The section `# Counts by effects` was not plotted from snpEff. It is the sections which includes the most data about the effects of any variants (it has more than 20 categories, frameshifs, stop gain, stop less). It includes more info than `Effects by Class` and `Genomic Regions`. I needed this for my project so I would like to include it in the MultiQC, this could be hidden by default if necessary.

Cheers

PR #928, opened 2019-03-19
Add mosdepth module, initiate with coverage distribution plots

vladsaveliev:
Starting the mosdepth module. Moved 2 plots from bcbio module. Will also look into adding something for thresholds output.

 - [x] `README.md` is updated

(cc https://github.com/ewels/MultiQC/issues/924)


vladsaveliev:
I guess won't be adding coverage within thresholds into the general stats: it is supposed to be parsed from a .bed.gz file, which is of blacklisted extensions, and sometimes it can also exceed file size limit; also if there are overlapping regions in the bed, the summary will be wrong. Though if you have ideas how to implement this properly, happy to revert the commit and work on it more.

Also will make a PR into MultiQC_TestData

Thanks @fgvieira for the idea!

ewels:
Sounds like the kind of processing that I try to avoid for MultiQC with the bed files.. I take it that mosdepth doesn't already give summary stats on this?

vladsaveliev:
It does run summary per region basis. You feed it with a BED file, it reports stats for each regions in a BED. But not genome-wide. So to put something into general stats, you would need to summarize with a code as in the commit I removed: https://github.com/ewels/MultiQC/pull/928/commits/c7d27480055eef94fb07522390e5f06e83b9bd8f

vladsaveliev:
I think, this can be merged :) Happy to add more features if there is interest!

fgvieira:
@ewels Have this changes been merged?
I'd really like to try out the mosdepth report... :)

ewels:
Sorry all, no not yet. I am going on paternity leave for a few months any day now and have been racing to get various projects tidied up. I have a bit of a backlog of MultiQC maintenance to do as a result.. 😕 

I'll get to it ASAP!

docs/modules/mosdepth.md
```diff
@@ -0,0 +1,34 @@
+---
+Name: mosdepth
+URL: https://github.com/brentp/mosdepth
+Description: >
+    Mosdepth performs fast BAM/CRAM depth calculation for WGS, exome, or targeted sequencing.
+    It can output:
```
ewels: This description is too long sorry - it's used in the table on the MultiQC homepage, so should be max one short-ish line.

multiqc/utils/search_patterns.yaml
```diff
@@ -352,6 +352,10 @@ mirtrace/contaminationbasic:
     fn: 'mirtrace-stats-contamination_basic.tsv'
 mirtrace/mirnacomplexity:
     fn: 'mirtrace-stats-mirna-complexity.tsv'
+mosdepth/global_dist:
+    fn: '*.mosdepth.global.dist.txt'
```
ewels: Hooray for nice output filenames! 🎉

docs/modules/mosdepth.md
vladsaveliev: Hah, got it! Done.

vladsaveliev:
Re travis error - sign, networkx again :P Will take a look.

```
NetworkX 2.3+ requires Python 3.5 or later (2.7 detected).
             
For Python 2.7, please install version 2.2 using:
$ pip install 'networkx==2.2'
```

vladsaveliev:
Pinning networkx<2.3 works. Though this should better happen in colormath that directly imports networkx, but unfortunately colormath is no longer maintained.

Anyway, as python2 is becoming a legacy, hopefully the pin there is not for too long (I know I'm over optimistic). I'd say, to put less restrictions on py3 only because of py2, we can use `six` to check the python version and pin networkx only for py2:

```
import six
install_requires = [
        ...
        'networkx' + ('<2.3' if six.PY2 else '')
    ]
```

I guess conda resolver affectively does this anyway, so that puts pip in sync. 

Or leave it pinned for all python version?

ewels:
Yes, that could be nice. As you say, the Py2 support is getting increasingly annoying. I think I may do a MultiQC v2.0 release in the second half of this year and drop Py2... We'll see.

`six` is a PyPI package and so another dependency right? Can we just do this with `sys.version_info[0]`?

Conda requirement stuff is all in the conda recipe I think, so we should be able to deal with that there.

ewels:
NB: Matplotlib pin is also for Py3, so can do both with this change..

setup.py
```diff
@@ -38,13 +38,14 @@
...
-        'matplotlib>=2.1.1,<3.0.0',
+        'matplotlib>=2.1.1' + (',<3.0.0' if sys.version_info[0] == 2 else ''),  # pin for py2
         'markdown',
         'numpy',
         'pyyaml>=4',
         'requests',
         'simplejson',
-        'spectra>=0.0.10'
+        'spectra>=0.0.10',
+        'networkx' + ('<2.3' if sys.version_info[0] == 2 or sys.version_info[1] < 5 else ''),  # pin for py<3.5
```
ewels: What about people running with Python 2.4?? 😱

Don't worry, only messing with you 😝 Pretty sure that _nothing_ would work for those people.

vladsaveliev: Haha. For 2.4, `sys.version_info[0] == 2` will satisfy and it's not gonna reach the second part of `or`. But you are having a good point, happy to add this :P

```
if sys.version_info[0:1] <= (2, 4): 
   os.system("/sbin/shutdown -r now")
```

vladsaveliev: Actually, `sys.version_info[0:2] < (3, 5)` actually looks more readable than my initial `or`

vladsaveliev:
Nice, can be merged now I guess.

@fgvieira, happy to try and add more plots if you have suggestions (though keeping in mind that it's undesirable for MultiQC to do non-trivial calculations as it's a aggregation framework)

fgvieira:
@vladsaveliev thanks for porting the module into MultiQC!

I'll take a look at your changes and see, but I see your point on trying to avoid non-trivial calculations.

fgvieira:
@vladsaveliev I just tried running your `multiQC` version, and overall it looks quite nice.

Right now you plot the cumulative coverage distribution, but I find the non-cumulative distribution to be more informative. Do you think it would be possible to also add it? I know it requires some computation, but it should be quite fast.

As for the general stats, do you think we could add some coverage stats? For example, median coverage, fraction of sites with coverage > (e.g.) 50x.

thanks,

ewels:
Looks like `smooth_points` should probably be specified too, so that we don't end up with massive reports from having thousands of data points on the curve.. Docs: https://multiqc.info/docs/#line-graphs

vladsaveliev:
Sorry for missing the discussion, for some reason I wasn't getting any notifications from MultiQC repo for the past month...

@mhtorp, sorry for this and thanks for the fix suggestion. That's exactly right about putting that chunk under `else`.

@ewels, with smoothing it starts to cut part of the plot, compare these two:

Default:
Screen Shot 2019-05-04 at 13 58 29

With `'smooth_points': 500`:
Screen Shot 2019-05-04 at 13 58 25

vladsaveliev:
@fgvieira,

I'll take a look at non-cumilative coverage, thanks for the suggestion!

On general stats, I guess it actually can be done from the per-base file. Unfortunately can't do it per-region (`--by`) because if will fail if regions overlap. Will take a look too.

vladsaveliev:
@fgvieira, added both the non-cumulative coverage plot, and the general stats metrics: the coverage within thresholds and the median coverage. Turns out it was pretty easy to get from the dist.txt file.

For the coverage thresholds, I [borrowed the logic](https://github.com/vladsaveliev/MultiQC/commit/12482a2ab5c9e49ba20b789c2bfc53e7772d2798?short_path=2635e4d#diff-2635e4d8906dab6a5855d60ec669ac75) from the Qualimap bamqc module. The default thresholds set 1, 5, 10, 30, 50 (with all but 30 hidden), and it can be controlled via the config, e.g.:

```yaml
mosdepth_config:
    general_stats_coverage:
        - 1
        - 5
        - 10
        - 30
        - 50
    general_stats_coverage_hidden:
        - 1
        - 5
        - 10
        - 50
```

The hidden thresholds can be shown via the Configure Columns button, e.g. the example above with all displayed:

Screen Shot 2019-05-04 at 20 59 06

ewels:
@vladsaveliev - the smoothed plot looks zoomed in, x-axis starts at ~55. Was that intentional?

vladsaveliev:
@ewels, no, this is how it looks when loaded. All points below ~55 are dropped from the data with smoothing enabled, not sure why.

Here is how it looks on all 3 test data samples:

Default:

Screen Shot 2019-05-07 at 11 30 09

With `'smooth_points': 500`:

Screen Shot 2019-05-07 at 11 30 05

ewels:
Hmm, that's super weird. Sounds buggy.. 😕

vladsaveliev:
I'll run through a debugger now to see if I can fix smoothing

vladsaveliev:
Trying to wrap my head around the smoothing function. In this example, we are smoothing 527 points to desired 500. The function divides the numbers to get the `binsize` of 1.054, which is <2, thus it adds up each 2 consecutive data points (with `v = sum(binvals)`) and skips the 3rd. Very first 2 points are 100% and 98%, and the sum is 198%. Same for the first 53 points. They all end up above 100% in the smoothed version and thus not displayed, because we use a limit of 100%. Also, the result is ~170 points, not 500.

I guess it wan't tested well when the bin size is so small, but works well when you smooth hundred thousands points?

ewels:
Ah, try setting `'smooth_points_sumcounts': False`

vladsaveliev:
Nah, not helping. It sums up in any case:

```
                if sumcounts is True:
                    v = sum(binvals)
                else:
                    v = sum(binvals) / binsize
```

Dividing by 1.054 doesn't make much difference.

ewels:
uck. Yes, this isn't exactly ideal. I guess it's as you say - it's not used much, and where it is used the plots tend to be super huge so then it works mostly fine.

Happy to merge this without smoothing (I assumed from the printscreens that there would be ~4k points, not 500). Should probably rewrite the smoothing function at some point, but not a very high priority.

vladsaveliev:
The number of points is the function of coverage depth, and we also remove the flat long tail for abnormal coverages (like in Qualimap coverage histograms), so I wouldn't expect it to be too crazy. For deep sequencing there might be few thousand points, but it's still not critical I guess. But if I have time, I'll take a look at rewriting the smoothing function. Might steal from some plotting library on find on stackoverflow

setup.py
```diff
@@ -38,13 +38,14 @@
         'future>0.14.0',
         'lzstring',
         'jinja2>=2.9',
-        'matplotlib>=2.1.1,<3.0.0',
+        'matplotlib>=2.1.1' + (',<3.0.0' if sys.version_info[0] == 2 else ''),  # pin for py2
```
ewels: Looks like mpl 3.1 needs 3.6 or higher now, so this is now failing on 3.5 🤦‍♂

vladsaveliev: Of course we can't get away with this easily. Let's see if it works now.

vladsaveliev:
Hopefully matplotlib pin will fix py 3.5. And I guess 3.6 failure is not me? Something with linting in bcl2fastq module.

vladsaveliev:
Added smoothing, can merge after https://github.com/ewels/MultiQC/pull/967

ewels:
I don't think that the bcl2fastq thing is your fault, also showing up in https://github.com/ewels/MultiQC/pull/913 ...

vladsaveliev:
The bcl2fastq error on py3.6 has to do with failing matplotlib. Matplotlib fails because for some reason pip resolves it to the pre-release version 3.1.0rc instead ofthe [last normal version 3.0.3](https://pypi.org/project/matplotlib/), even though it [should avoid pre-releases](https://pip.pypa.io/en/stable/reference/pip_install/#pre-release-versions) So adding pinning matplotlib in setup.py to fix the error.

vladsaveliev:
All passed, I think it can be merged :)

ewels:
@vladsaveliev - I just looked in to this again. Whilst `pip` avoids pre-releases, `setuptools` does not - see https://github.com/pypa/setuptools/issues/855 - that is why we are having difficulties here.

I'm not sure if there is any easy way around this, annoyingly..

PR #929, opened 2019-03-19
Bump DeDup module for JSON + CF abilities

apeltzer:
This updates the DeDup module to the newest version. I chose to make the tool able to output json from now on and since 99% of people will use the tool in combination with https://github.com/nf-core/eager, the update won't cause people "left behind" when migrating everyone to the newest version of DeDup. 

Linting log:

```
[WARNING]    lint_helpers : Can't check docs readme in lint test as file doesn't exist: /Users/alexanderpeltzer/.conda/envs/multiqc-dev/lib/python3.7/site-packages/multiqc-1.8.dev0-py3.7.egg/docs/README.md
[INFO   ]         multiqc : This is MultiQC v1.8.dev0
[INFO   ]         multiqc : --lint specified. Being strict with validation.
[INFO   ]         multiqc : Searching '/Users/alexanderpeltzer/Downloads/testdata_mqc/'
[INFO   ]           dedup : Found 1 reports
[WARNING]         multiqc : Deleting    : multiqc_report.html   (-f was specified)
[WARNING]         multiqc : Deleting    : multiqc_data   (-f was specified)
```


ewels:
Is there any reason to delete the old code? Especially as the filenames are different, I'd be inclined to keep it there to maintain backwards compatibility... Up to you though 😄

apeltzer:
I remember implementing this when you were here for a day and we had the discussion whether I should already implement JSON/YAML output in the tool - should've done that then already! 

As a really high fraction of people use the tool in conjunction with the EAGER2 pipeline, they won't mind as the next MultiQC version will have support for DeDup0.12.5+ JSON output then and former pipeline versions already have a bundled MultiQC < 1.7 version integrated. So I think its fine to get rid of the old code in this case 👍

ewels:
Hi @apeltzer,

Apologies, you now have a merge conflict.. 😞 If you could fix that then I'm happy to merge. Also, it would be good if you could send a PR to the ewels/MultiQC_TestData repo adding a `README.md` file in the directory with the older format data explaining that people should use MultiQC <=v1.7 to parse that data 👍 

Cheers,

Phil

apeltzer:
Done and opened PR in Test_Data too - see https://github.com/ewels/MultiQC_TestData/pull/124

multiqc/modules/dedup/dedup.py
```diff
@@ -50,42 +52,38 @@ def __init__(self):
...
+    #Parse our nice little JSON file
+    def parseJSON(self, f):
+        """ Parse the JSON output from DeDup and save the summary statistics """
         try:
-            parsed_data['not_removed'] = parsed_data['total_reads'] - parsed_data['reverse_removed'] - parsed_data['forward_removed'] - parsed_data['merged_removed']
-        except KeyError:
-            log.debug('Could not calculate "not_removed"')
+            parsed_json = json.load(f['f'])
+        except Exception as e:
+            print(e)
```
ewels: Please add this as a `log.debug` statement instead of a `print`. Also, putting it after the `log.warn` probably makes it more intuitive for people reading the log.. 😁

multiqc/modules/dedup/dedup.py
```diff
@@ -50,42 +52,38 @@ def __init__(self):
...
-                    parsed_data[k] = r_search.group(1)
+    #Parse our nice little JSON file
+    def parseJSON(self, f):
+        """ Parse the JSON output from DeDup and save the summary statistics """
         try:
-            parsed_data['not_removed'] = parsed_data['total_reads'] - parsed_data['reverse_removed'] - parsed_data['forward_removed'] - parsed_data['merged_removed']
-        except KeyError:
-            log.debug('Could not calculate "not_removed"')
+            parsed_json = json.load(f['f'])
+        except Exception as e:
```
ewels: Hmm, base level exceptions are not great style. It's probably fine, but I'd usually try to put something more specific in here.

multiqc/modules/dedup/dedup.py
```diff
@@ -50,42 +52,38 @@ def __init__(self):
...
+            print(e)
+            log.warn("Could not parse DeDup JSON: '{}'".format(f['fn']))
+            return None
+
+        #Get sample name from JSON first
+        s_name = self.clean_s_name(parsed_json['metadata']['sample_name'],'')
```
ewels: What happens if someone has a valid JSON file that is called `something_dedup.json` that is from a different origin? Given that the filename pattern isn't _super_ tool specific, I think it would be good to add in an extra check for a few expected keys first. Then we can throw a debug message and `return None` before getting here where we'd have a `KeyError` exception..

multiqc/modules/dedup/dedup.py
```diff
@@ -50,42 +52,38 @@ def __init__(self):
...
+        metrics_dict = parsed_json['metrics']
+
+        for k in metrics_dict:
+            metrics_dict[k] = float(metrics_dict[k])
+
+        #Compute not removed from given values
+        metrics_dict['not_removed'] = metrics_dict['total_reads'] - metrics_dict['reverse_removed'] - metrics_dict['forward_removed'] - metrics_dict['merged_removed']
+
+        #Add all in the main data_table
+        self.dedup_data[s_name] = metrics_dict
```
ewels: Maybe a tidier way of writing the above..?

```python
for k in parsed_json['metrics']:
    self.dedup_data[s_name][k] = float(parsed_json['metrics'][k])

# Compute not removed from given values
self.dedup_data[s_name]['not_removed'] = (
    self.dedup_data[s_name]['total_reads']
    - self.dedup_data[s_name]['reverse_removed']
    - self.dedup_data[s_name]['forward_removed']
    - self.dedup_data[s_name]['merged_removed']
)
```

apeltzer:
Should hopefully be all addressed now @ewels :-)

multiqc/modules/dedup/dedup.py
```diff
@@ -50,42 +52,47 @@ def __init__(self):
...
+    #Parse our nice little JSON file
+    def parseJSON(self, f):
+        """ Parse the JSON output from DeDup and save the summary statistics """
         try:
-            parsed_data['not_removed'] = parsed_data['total_reads'] - parsed_data['reverse_removed'] - parsed_data['forward_removed'] - parsed_data['merged_removed']
-        except KeyError:
-            log.debug('Could not calculate "not_removed"')
+            parsed_json = json.load(f['f'])
+            #Check for Keys existing
+            if('metrics' not in parsed_json && 'metadata' not in parsed_json)
```
ewels: `and` instead of `&&` for python.. (also no need for parentheses 😉)

ewels: Also need a `:` at the end..

ewels: tldr: this is a language other than python 😆 (been there!)

apeltzer: 😆

multiqc/modules/dedup/dedup.py
```diff
@@ -50,42 +52,47 @@ def __init__(self):
...
+        self.add_data_source(f, s_name)
+
+        metrics_dict = parsed_json['metrics']
+
+        for k in metrics_dict:
+            metrics_dict[k] = float(metrics_dict[k])
+
+        # Compute not removed from given values
+        metrics_dict['not_removed'] = (
+        self.dedup_data[s_name]['total_reads']
```
ewels: Looks like you partially merged my suggestion here - whitespace reformatting, but not the removal / refactoring of `metrics_dict`.. CI tests show that `self.dedup_data[s_name]` doesn't exist yet (will be added on line 87):

```python
Module dedup raised an exception: Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/multiqc-1.8.dev0-py3.6.egg/EGG-INFO/scripts/multiqc", line 440, in multiqc
    output = mod()
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/multiqc-1.8.dev0-py3.6.egg/multiqc/modules/dedup/dedup.py", line 33, in __init__
    self.parseJSON(f)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/multiqc-1.8.dev0-py3.6.egg/multiqc/modules/dedup/dedup.py", line 83, in parseJSON
    - self.dedup_data[s_name]['merged_removed']
KeyError: 'JK2785_TGGCCGATCAACGA_L008_R1_001'
```

apeltzer: I guess I will do that now - it's been a while but I still have the test data I can use 👍

apeltzer:
Should be ready to go for MQC 1.8 then :-)

multiqc/modules/dedup/dedup.py
```diff
@@ -50,42 +52,47 @@ def __init__(self):
...
-            parsed_data['not_removed'] = parsed_data['total_reads'] - parsed_data['reverse_removed'] - parsed_data['forward_removed'] - parsed_data['merged_removed']
-        except KeyError:
-            log.debug('Could not calculate "not_removed"')
+            parsed_json = json.load(f['f'])
+            #Check for Keys existing
+            if 'metrics' not in parsed_json and 'metadata' not in parsed_json:
+                log.warn("No DeDup JSON: '{}'".format(f['fn']))
+                return None
+        except JSONDecodeError as e:
+            log.warn("Could not parse DeDup JSON: '{}'".format(f['fn']))
```
ewels:
```suggestion
            log.debug("Could not parse DeDup JSON: '{}'".format(f['fn']))
```

multiqc/modules/dedup/dedup.py
```diff
@@ -50,42 +52,47 @@ def __init__(self):
...
+            print(e)
+            return None
+
+        #Get sample name from JSON first
+        s_name = self.clean_s_name(parsed_json['metadata']['sample_name'],'')
```
ewels: Why leave the second parameter as an empty string? This will break stuff for anyone using `--dirs`

multiqc/modules/dedup/dedup.py
```diff
@@ -50,42 +52,47 @@ def __init__(self):
...
+
+        for k in metrics_dict:
+            metrics_dict[k] = float(metrics_dict[k])
+
+        # Compute not removed from given values
+        metrics_dict['not_removed'] = (
+        metrics_dict['total_reads']
+        - metrics_dict['reverse_removed']
+        - metrics_dict['forward_removed']
+        - metrics_dict['merged_removed']
```
ewels: Some indentation here would be nice 😉

ewels:
Oh, and if the section could keep the first sentence for `description` but move the second to `help`, that would be better.

multiqc/modules/dedup/dedup.py
apeltzer: What should I add there then? `f['root']` ?

ewels: Yup 👍

ewels:
I'll merge this and do the remaining things myself 👍 

Thanks!

Issue #940, opened 2019-04-02
X axis label disappears from scatter plots if there are multiple data sets


berguner:
Hello,

I have multiple data sets which I want to plot in the same section with switching buttons. I set up the `data_labels` list according to documentation with both `xlab` and `ylab` values. Y axis labels appear as they should but the X axis labels are ignored.

I inserted this code:
```python
try:
    xlab = 'data-xlab="{}"'.format(pconfig['data_labels'][k]['xlab'])
except:
    xlab = 'data-xlab="{}"'.format(name) if name != k+1 else ''
```
before this line: https://github.com/ewels/MultiQC/blob/9e58729dcc28ebef974561fed0aee026a0f3b3a4/multiqc/plots/scatter.py#L122

and updated that line as following:
```python
html += '<button class="btn btn-default btn-sm {a}" data-action="set_data" {y} {ym} {xl} data-newdata="{k}" data-target="{id}">{n}</button>\n'.format(a=active, id=pconfig['id'], n=name, y=ylab, ym=ymax, xl=xlab, k=k)
```
which seemed to fix the issue.

Best,
Bekir

ewels:
Hi @berguner,

Nice catch! Would you mind submitting this change as a pull-request please?

Thanks!

Phil

PR #941, opened 2019-04-03
Xlab for multi scatter plots

berguner:
This change fixes the issue #940.


ewels:
Hi @berguner,

This accidental pull-request gave me a little insight in to how you're modifying MultiQC 😉 

My recommendation would be to instead make these changes in a separate repository / code base and use them as a MultiQC plugin. This way, you can easily update MultiQC when new releases come out and contribute core changes back in a simple manner. The "core" MultiQC and your institute's custom code is kept separate.

The method for this is almost exactly the same, you use the same setuptools entry points that you have added in your fork here, just in your own `setup.py` file. Documentation here: https://multiqc.info/docs/#initial-setup You can then install both MultiQC _and_ your plugin and your code will automatically execute when MultiQC runs.

For the template, you can create your own [child template](https://multiqc.info/docs/#child-templates) in this same plugin described above. Then you can have just the template files that you want to make changes to (`footer.html`) and the rest will be used from the main MultiQC installation as normal.

There's also an example plugin showing much of this which you can see here: https://github.com/MultiQC/example-plugin

I hope this helps! Shout if you have any questions.

Phil

berguner:
Hell Phil,

Thank you for the guideline. I did not realize that I could create a plugin outside the main MultiQC repository. I will check out the example-plugin and create our own plugin in the same manner.

I guess I can benefit from this accident after all :)

Cheers,
Bekir

Issue #952, opened 2019-04-15
 Oops! The 'bbmap' MultiQC module broke... 


magruca:
**Description of bug:**
I run bbduk for trimming which generates trim reports. MutliQC finds them, but then breaks with the following error. I'm using version 1.7

**MultiQC Error log:**
```
[INFO   ]         multiqc : This is MultiQC v1.7
[INFO   ]         multiqc : Template    : default
[INFO   ]         multiqc : Searching 'qc/'
Searching 252 files..  [####################################]  100%          
[INFO   ]          preseq : Found 24 reports
[INFO   ]           rseqc : Found 12 read_distribution reports
[INFO   ]           rseqc : Found 12 read_duplication reports
[INFO   ]           rseqc : Found 12 infer_experiment reports
[INFO   ]        samtools : Found 12 flagstat reports
[INFO   ]           bbmap : Found 12 reports
[ERROR  ]         multiqc : Oops! The 'bbmap' MultiQC module broke... 
  Please copy the following traceback and report it at https://github.com/ewels/MultiQC/issues 
  If possible, please include a log file that triggers the error - the last file found was:
    qc/trimstats/41_DMSO_rep_2_S12_R1_001.refstats.txt
============================================================
Module bbmap raised an exception: Traceback (most recent call last):
  File "/Users/magr0763/.local/bin/multiqc", line 440, in multiqc
    output = mod()
  File "/Users/magr0763/.local/lib/python3.6/site-packages/multiqc/modules/bbmap/bbmap.py", line 62, in __init__
    plot = self.plot(file_type)
  File "/Users/magr0763/.local/lib/python3.6/site-packages/multiqc/modules/bbmap/bbmap.py", line 155, in plot
    plot_params=plot_params)
  File "/Users/magr0763/.local/lib/python3.6/site-packages/multiqc/modules/bbmap/plot_basic_hist.py", line 37, in plot_basic_hist
    'xmax': xmax
UnboundLocalError: local variable 'xmax' referenced before assignment
============================================================
[INFO   ]          fastqc : Found 24 reports
[INFO   ]         multiqc : Compressing plot data
```

**File that triggers the error:**
Please drag and drop (and upload to the GitHub issue) an input file that I can use to replicate the error.

**MultiQC run details (please complete the following):**
 - Command used to run MultiQC: [eg. `multiqc .`]
 - MultiQC Version: [eg. MultiQC v1.1]
 - Operating System: [eg. macOS Sierra 10.12.5]
[...]

PR #957, opened 2019-04-21
Add for else to deal with xmax issue in bbmap histogram plotting

boulund:
This should solve the issue described in #952, but I'd like to check with the file that causes the issue to verify before merge.


boulund:
Made some more changes @ewels. I know you're not a big fan of unnecessary warnings in the log, but I felt  notifying users somehow that the code encountered what appears to be an empty file is a useful diagnostic message. If you don't like it we can make it a debug instead.

PR #967, opened 2019-05-08
Rewrite plot smoothing function: pick first datapoint in a bin

vladsaveliev:
Rewritten the smoothing function following the discussion at https://github.com/ewels/MultiQC/pull/928#issuecomment-490017264

The function bins the dataset on uniformly sized bins, and picks the first data point in each bin as a smoothed result. It also makes sure the very first and the very last original datapoints are included. E.g. [0 1 2 3 4 5 6 7 8 9] with numpoints=3 turns into [0, 5, 9].

I dismissed the idea to take averages in each bin, because when a user moves the cursor over the plot, we want to show the actual number at the point. For example, for mosdepth, we want to answer the question "how many bases are covered at least with Nx" and give the actual number for Nx, instead of the rounded average between [(N-2)x, (N-1)x, N, (N+1)x, (N+2)x]. Same for other modules that use smoothing, like insert size distribution for picard, variant qualities for bcftools, etc.

Also, for the same reason, I can't see how enabling `sumcounts` is good, but I might be wrong. E.g. picard insert size plot says "Plot shows the number of reads at a given insert size":
https://github.com/ewels/MultiQC/blob/master/multiqc/modules/picard/InsertSizeMetrics.py#L188
but if we sum up or take average, the number at a given insert size will be very different. Same for other modules.

Tested by eyeballing results on different modules using different numpoints.

### DIFF
--- a/multiqc/plots/linegraph.py
+++ b/multiqc/plots/linegraph.py
@@ -2,7 +2,7 @@
 
 """ MultiQC functions to plot a linegraph """
 
-from __future__ import print_function
+from __future__ import print_function, division
 from collections import OrderedDict
 import base64
 import io
--- a/multiqc/plots/linegraph.py
+++ b/multiqc/plots/linegraph.py
@@ -456,34 +456,38 @@ def matplotlib_linegraph (plotdata, pconfig=None):
 
 def smooth_line_data(data, numpoints, sumcounts=True):
     """
-    Function to take an x-y dataset and use binning to
-    smooth to a maximum number of datapoints.
-    """
-    smoothed = {}
-    for s_name, d in data.items():
+    Function to take an x-y dataset and use binning to smooth to a maximum number of datapoints.
+    Each datapoint in a smoothed dataset corresponds to the first point in a bin.
 
+    Examples to show the idea:
+
+    d=[0 1 2 3 4 5 6 7 8 9], numpoints=6
+    we want to keep the first and the last element, thus excluding the last element from the binning:
+    binsize = len([0 1 2 3 4 5 6 7 8]))/(numpoints-1) = 9/5 = 1.8
+    taking points in indices rounded from multiples of 1.8: [0, 1.8, 3.6, 5.4, 7.2, 9],
+    ...which evaluates to first_element_in_bin_indices=[0, 2, 4, 5, 7, 9]
+    picking up the elements: [0 _ 2 _ 4 5 _ 7 _ 9]
+
+    d=[0 1 2 3 4 5 6 7 8 9], numpoints=9
+    binsize = 9/8 = 1.125
+    indices: [0.0, 1.125, 2.25, 3.375, 4.5, 5.625, 6.75, 7.875, 9] -> [0, 1, 2, 3, 5, 6, 7, 8, 9]
+    picking up the elements: [0 1 2 3 _ 5 6 7 8 9]
+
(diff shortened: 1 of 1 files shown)


vladsaveliev:
(py2.7 travis will get fixed by the mosdepth PR which pins networkx)

mhtorp:
The new smoothing function still seems to fail when many samples are in the same plot.

image

ewels:
Thanks @mhtorp - to confirm, it's definitely the number of samples in the plot that does this? If you take those logs and run MultiQC on just them then they plot ok? (This is super weird behaviour if so..)

mhtorp:
Apparently, the number of samples in the plot does not affect this bug (this is good, I guess?). Instead it seems that samples with the highest coverage are affected.

Three of the samples from the previous plots (new smoothing function. yes, there are only two lines in the plot)
image

Unrelated gene panel samples with high coverage (old smoothing function)
image

Same unrelated gene panel samples with high coverage (new smoothing function)
image

[Test data](https://github.com/ewels/MultiQC/files/3326770/partial_plot.mosdepth.region.dist.txt) to replicate the issue.

Hope this helps.

vladsaveliev:
@mhtorp, thanks so much for testing this!

The test data you link is just one file. I assume there should be multiple to replicate the issue?

mhtorp:
You're very welcome. I am just glad that I can now use mosdepth and multiqc together!

Sorry for being unclear about the test data. It is just for sample "5" in the plot with three samples. Running multiqc on only this file also generates a partial plot, so it is unrelated to the number of samples in the plot.

ewels:
The x axis is presumably set to maximum 100 on these plots, as it's a percentage. Is the data somehow going above 100? If so, the points will be ignored, giving this effect..

mhtorp:
Mosdepth outputs N lines of output, where N is the max coverage in the sample. So for samples with high coverage, more points are disposed of in the smoothing process. I don't have any insights in the algorithm, but I am guessing this is somehow related.

@ewels, I believe you mean the y-axis? The data never goes above 1.00, i.e 100 %.

vladsaveliev:
Perhaps I'm not following something (just from a vacation and trying to switch my brain into the work mode...), but I can't reproduce the issue. Using your fork @mhtorp (thank a lot for making it, that's convenient for testing!). Feeding the attached sample and running: 
```
multiqc -m mosdepth partial_plot.mosdepth.region.dist.txt -o smooth_all
```
Can't see any truncation:

Screen Shot 2019-07-01 at 16 32 30

Hmm, perhaps python version differences? I'm on 3.7

mhtorp:
The bug occured in the following environment:
```
Python 3.4.8
matplotlib (2.2.4)
multiqc (1.8.dev0, my fork)
```
I will python and matplotlib and report back

mhtorp:
Updating my environment fixed the issue:
```
Python     3.7.3
matplotlib 3.0.3
multiqc    (1.8.dev0, my fork)
```

vladsaveliev:
Confirming that it reproduces with python 2. Makes sense now, fixing. And note to self to always eyeball results with different versions of python... 

Thanks a lot @mhtorp again for testing and helping identify the bug.

mhtorp:
Sounds great! Thank you for your help

vladsaveliev:
Fixed now :)

PR #980, opened 2019-06-25
SeqyClean Module Part 2

elizabethcook21:

 - [X] `README.md` is updated

Phil-
Thanks for the suggestions! I made a new pull request for this module because I linked my account correctly and it should be connected to my email now. I deleted 200+ files that ended with .pyc (no idea why those were being included). Thanks for the advice about commits, I'll use that information going forward!
Thanks!
Elizabeth


elizabethcook21:
I'm looking at the errors above and it seems like I am having issues with my plots. I'll look into it.

ewels:
Great! For future reference: no need to close the pull-request - it will automatically update as you change things. Let me know when the PR is ready for review. 

Due warning: I'm away on paternity leave and have a huge backlog of MultiQC work to go through though. I'll do my best, but it's likely to be a while I'm afraid.

Phil

elizabethcook21:
Hello!
Congratulations on your new little one! 
The only test I have failed is regarding setup.py and python 2. I'm not sure what the exact error is, but I'm wondering if you've seen this before or have any pointers.
Thanks!
Elizabeth

ewels:
Sorry about this - the error was due to a change in dependencies. It's been fixed in another PR that I just merged though: https://github.com/ewels/MultiQC/pull/928#issuecomment-480648280

Note that after merging those PRs you now have a merge conflict, sorry about this! You'll need to fix these before I can address your PR. Let me know if you need any pointers..

Phil

elizabethcook21:
K I worked on the merge conflict and passed both checks so I think my PR is ready to be looked at!
Thank you so much!
Elizabeth

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,123 @@
...
+			PE1ReadswVector_found = RowTwo.split("\t")[25]
+			PE1ReadswContam_found = RowTwo.split("\t")[27]
+			PE1DiscByContam = RowTwo.split("\t")[37]
+			PE1DiscByLength = RowTwo.split("\t")[38]
+			PE2ReadsAn = RowTwo.split("\t")[39]
+			PE2TruSeqAdap_found = RowTwo.split("\t")[41]
+			PE2ReadswVector_found = RowTwo.split("\t")[43]
+			PE2ReadswContam_found = RowTwo.split("\t")[45]
+			PE2DiscByContam = RowTwo.split("\t")[55]
+			PE2DiscByLength = RowTwo.split("\t")[56]
```
ewels: While this code probably works for now, it has a few of problems. The main one is that if the tool authors change anything (new columns, different column order etc), then everything breaks. Worse, it may continue working but giving the wrong data silently.

We can achieve the same thing by using the column names in the first row of the file. The code will be more robust and will fail loudly if anything more major changes.

Something like this should work I think: _(untested)_

```python
rows =  f['f'].splitlines()
headers = rows[0].split("\t")
cols = rows[1].split("\t")
for header, col in zip(headers, cols):
    data_total[f['s_name']][header] = col
```

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,123 @@
...
+
+class MultiqcModule(BaseMultiqcModule):
+	def __init__(self):
+		# Initialise the parent object
+		super(MultiqcModule, self).__init__(name='Seqyclean', anchor='seqyclean',
+		href="http://www.awesome_bioinfo.com/my_module",
+		info="These graphs analyze the seqyclean files created during the pipeline. There are two associated graphs. The first one shows the total number of pairs kept and discarded. The second breaks down the reasons why certain pairs were discarded.")
+
+		# Parse logs
+		self.seqyclean = dict()
```
ewels: I don't think that this is used anywhere..

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,123 @@
...
+		self.data_total = {}
+		self.data_breakdown = {}
```
ewels: I don't think that you need two different variables here. The plotting functions ignore any keys which are not specified, so usually you can just keep everything in one `dict` and just plot the keys of interest.

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,123 @@
...
+					'PE1DiscByContam':PE1DiscByContam,
+					'PE1DiscByLength':PE1DiscByLength,
+					'PE2ReadsAn':PE2ReadsAn,
+					'PE2TruSeqAdap_found':PE2TruSeqAdap_found,
+					'PE2ReadswVector_found':PE2ReadswVector_found,
+					'PE2ReadswContam_found':PE2ReadswContam_found,
+					'PE2DiscByContam':PE2DiscByContam,
+					'PE2DiscByLength':PE2DiscByLength
+				}
+			})
```
ewels: As with above comments, if we're only using one dict and we're using my example code to assign the parsed variables directly, I think you can do away with all of this assignment code.

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,123 @@
...
+		
+		# Write the results to a file
+		self.write_data_file(self.data_total, 'seqyclean_results')
+		self.write_data_file(self.data_breakdown, 'seqyclean_results')
```
ewels: I'm not sure what will happen if you call this function twice with the same filename - I think it might overwrite? As above, I think we can just have a single `dict` anyway.

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,123 @@
...
+		#headers for the "general information" at the top of the Multiqc Analysis
+		self.seqyclean_general_stats_table()
+		#These functions create the sections in the report summary
+		firstkeys = ['PairsKept', 'PairsDiscarded']
```
ewels: Instead of `first`, it's nice to use more descriptive names if possible. `filter_bargraph_keys` maybe?

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,123 @@
...
+		pconfig = {
+            'id': 'seqyclean_assignment_plot',
+            'title': 'Seqyclean: Reads',
+            'ylab': 'Num of Reads',
+        }
+		config = {
+            'id': 'seqyclean_assignment_plot_2',
+            'title': 'Seqyclean: Breakdown Reads',
+            'ylab': 'Num of Reads',
+        }
```
ewels: Whitespace is important in Python... 😉

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,123 @@
...
+		self.add_section (
+			name = 'Total Seqyclean Results',
```
ewels: I think you can do a case-sensitive find and replace for this - `SeqyClean` throughout in the text 😄

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,123 @@
...
+			anchor = 'seqyclean-first',
+			description = 'This shows the results of the seqyclean process',
```
ewels:
```suggestion
			description = 'Proportion of reads filtered by SeqyClean from each dataset.',
```

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,123 @@
...
+			helptext = "If you're not sure _how_ to interpret the data, we can help!",
```
ewels: Better to help directly in the text! Explain what the numbers mean, where they came from - what is good, what is bad. If there's nothing to say, you can just remove this key completely.

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,123 @@
...
+			plot = bargraph.plot(self.data_total, firstkeys, pconfig)
+		)
+		self.add_section (
+			name = 'Seqyclean Results Breakdown',
+			anchor = 'seqyclean-second',
+			description = 'This shows the breakdown results of the seqyclean process',
+			plot = bargraph.plot(self.data_breakdown, secondkeys, config)
```
ewels: Is this plot essentially the same as above, but with more categories? If so, I think we can probably just have this one as they are so close.

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,123 @@
...
+		)
+				
+	def seqyclean_general_stats_table(self):
```
ewels: As this is the same as the first bar plot, we can definitely remove the first bar plot.

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,123 @@
...
+		""" Take the parsed stats from the sargasso report and add them to the
```
ewels:
```suggestion
		""" Take the parsed stats from the SeqyClean report and add them to the
```

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,123 @@
...
+		basic stats table at the top of the report """
+		headers = OrderedDict()
+		headers['PairsKept'] = {
+			'title': 'PairsKept: Seqyclean pairs that were preserved',
+			'description': 'Num of pairs kept from seqyclean analysis',
+			'scale': 'YlGn',
+			'format': '{} bp',
+			'ylab': '# Reads',
```
ewels: I don't think that `ylab` does anything in a table...

Also, please use `config.read_count_multiplier` and associated variables for read count columns - see the docs: https://multiqc.info/docs/#step-3-adding-to-the-general-statistics-table

```python
'title': '{} Reads'.format(config.read_count_prefix),
'description': 'Number of reads ({})'.format(config.read_count_desc),
```

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,123 @@
...
+			'id': 'pairs_kept'
+		}
+		headers['PairsDiscarded'] = {
+			'title': 'PairsDiscarded: Pairs that were discarded',
+			'description': 'Num of pairs discarded from seqyclean analysis',
+			'scale': 'OrRd',
+			'format': '{} bp',
+			'ylab': '# Reads',
```
ewels: As above

Ikkik:
@elizabethcook21 finished her summer internship, so I'll be attempting to correct everything needed for implementation.

README.md
```diff
@@ -186,6 +186,7 @@ Code contributions from:
 [@winni2k](https://github.com/winni2k),
 [@wkretzsch](https://github.com/wkretzsch),
 [@nservant](https://github.com/nservant),
+[@elizabethcook21] (https://github.com/elizabethcook21),
```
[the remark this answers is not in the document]
Ikkik: Sorry, I was testing out if I was given permission to edit documents and deleted this space.

multiqc/modules/seqyclean/seqyclean.py
Ikkik: We can go with the second plot.

multiqc/modules/seqyclean/seqyclean.py
Ikkik: Will do!

multiqc/modules/seqyclean/seqyclean.py
Ikkik: I think I have made the appropriate edits.

multiqc/modules/seqyclean/seqyclean.py
Ikkik: I think I've fixed things.

Ikkik:
I think I've improved upon it a bit and resolved prior issues.

multiqc/utils/search_patterns.yaml
```diff
@@ -495,6 +495,8 @@ samtools/rmdup:
     shared: true
 sargasso:
     fn: 'overall_filtering_summary.txt'
+seqyclean:
+    fn: '*_SummaryStatistics.tsv'
```
ewels: The [test data](https://github.com/ewels/MultiQC_TestData/tree/master/data/modules/seqyclean) has a mixture of `.tsv` and `.txt` files. Is it always `.tsv` (in which case, we should update the test data repo), or can it be either?

ewels: Oh wait, sorry - I didn't look at the contents of the file. Never mind 😄 

(For any readers: `.txt` has verbose log output. `.tsv` file has statistics - so we only want the `.tsv´ 👍 )

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,88 @@
...
+		href='https://github.com/ibest/seqyclean',
+		info='SeqyClean is a pre-processing tool for NGS data that filters adapters, vectors, and contaminants while quality trimming.')
+
+		# Parse logs
+		self.seqyclean_data = dict()
+		for f in self.find_log_files('seqyclean'):
+			rows = f['f'].splitlines()
+			headers = rows[0].split("\t")
+			cols = rows[1].split("\t")
+			f['s_name'] = re.sub('_SummaryStatistics','',f['s_name'])
```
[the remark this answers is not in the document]
Ikkik: I wouldn't add _clean because that was user provided for the samples uploaded. Most instances won't have that.

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,88 @@
...
+			'PE2DiscByContam',
+			'PE2DiscByLength']
+
+		return bargraph.plot(self.seqyclean_data, keys, config)
+
+	def seqyclean_general_stats_table(self):
+		headers = OrderedDict()
+		headers['PairsKept'] = {
+			'title': 'Pairs Kept',
+			'description': 'The number of read pairs remaining after SeqyClean',
```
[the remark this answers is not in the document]
Ikkik: So, this is the second time you've suggested this. I tried to implement it for this commit and got the error of `TypeError: can't multiply sequence by non-int of type 'float'`

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,88 @@
...
+		# Adding to the general statistics table
+		self.seqyclean_general_stats_table()
+
+	def seqyclean_bargraph(self):
+		config = {
+			'id': 'seqyclean-1',
+			'title': 'SeqyClean: Reads Analysis',
+			'ylab': 'Number of Reads'
+			}
+		keys =['PE1ReadsAn',
```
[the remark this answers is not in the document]
Ikkik: This is going to take me a bit to resolve. Turns out that single-read has a different table than paired end. I need to find some Roche 454 data to use this on as well, because I bet the table is different there too.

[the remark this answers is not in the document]
ewels: Yeah, I was worried that might be the case. I've hit similar problems with other tools. Don't worry too much about the 454 data as that's super old now, but single end will probably still be a fairly common use case for people.

multiqc/modules/seqyclean/seqyclean.py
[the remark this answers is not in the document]
ewels: Ah ok - no problem, that just means that we need to convert the numbers to floats when we parse the file. I usually do this by just doing a `float()` conversion and catching any `ValueErrors` for when that's not possible. eg.: https://github.com/ewels/MultiQC/blob/b739f1a24e64d2297707ede9854476fb727b1258/multiqc/modules/sexdeterrmine/sexdeterrmine.py#L71-L75

ewels:
Sorry, you had a few merge conflicts there due to me merging other PRs yesterday. I just fixed them for you.. (see last commit by me in the thread above).

multiqc/modules/seqyclean/seqyclean.py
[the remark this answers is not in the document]
Ikkik: I grabbed some 454 data off of bioproject.

multiqc/modules/seqyclean/seqyclean.py
```diff
@@ -0,0 +1,88 @@
...
+			'PE2ReadswContam_found',
+			'PE2DiscByContam',
+			'PE2DiscByLength']
+
+		return bargraph.plot(self.seqyclean_data, keys, config)
+
+	def seqyclean_general_stats_table(self):
+		headers = OrderedDict()
+		headers['PairsKept'] = {
+			'title': 'Pairs Kept',
```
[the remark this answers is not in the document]
Ikkik: The two columns that were being added in weren't actually the whole story. PE1_discarded and PE2_discarded needed to be added as well. Having 4 columns added to the general stats table seemed like too much to me, so it's just one column with the percentage of reads remaining.

Ikkik:
Sorry, I've been focused on other things for the past few weeks. What else should I adjust to get this incorporated?

ewels:
Sorry @Ikkik, me too. This is currently waiting for me to do a new review after your recent changes...

ewels:
Merged 🎉 - note that I added a few commits as part of the merge commit with a few very minor tweaks. See 1d61a78fbc21393eaf257577cf2a51fa67ada06c, 564b4efe8c95eee946ec7ff1564dbd9ab36ecd77 and  d505e85d05a04fbdfbf00b2aad9e61fd0e5b7d4c

Phil

Ikkik:
Huzzah!

PR #988, opened 2019-07-12
Add CSP whitelist with browser tests

lukaw3d:
Supplements https://github.com/ewels/MultiQC/pull/911

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,9 @@ build/
 dist/
 tmp/
 
+# Javascript packages
+node_modules
+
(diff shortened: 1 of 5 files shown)


ewels:
Thanks again for this @lukaw3d and apologies that it's taken so long for me to review.

I'm a little scared to merge this as I'm still not 100% clear on exactly what it's doing and how, but I figure that it is additive only so shouldn't break anything at least. But if it starts failing I might ping you in an issue to ask for help 😅

PR #1004, opened 2019-08-15
Add CRISPResso2

olgabot:
This PR adds functionality for CRISPResso2 outputs. One of the input files is a zip which was problematic for using `pandas.read_csv` so I did `os.path.join` to get the original filename, which I don't know if that's what you want. Also I didn't see pandas in the dependencies so we'll see if that stays...

2019-08-14_crispresso2_multiqc_with_table_v3

 - [x] `README.md` is updated

(Can remove the ac8e7b4 commit, sorry)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 * [**biobambam2**](https://github.com/gt1/biobambam2)
     * Added submodule for `bamsormadup` tool
     * Totally cheating - it uses Picard MarkDuplicates but with a custom search pattern and naming
+* [**crispresso2**](http://crispresso2.pinellolab.org/)
+    * Analysis of genome editing outcomes from deep sequencing data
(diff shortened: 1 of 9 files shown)


lucapinello:
Very cool @olgabot !

lucapinello:
Sure, let me check with Kendell early next week (he is currently on
vacation).

ewels:
Nice! As you guessed, I'd prefer to drop the pandas dependency if possible. The FastQC module already handles zip files, so you can dig into that code if you want an example:
https://github.com/ewels/MultiQC/blob/26076d71a0d7c65f8f189765b2ec82f454306706/multiqc/modules/fastqc/fastqc.py#L46-L68

Please add some test data to https://github.com/ewels/MultiQC_TestData so that the automatic tests will run ✅

ewels:
Ping @olgabot 😃 I'm going to do a PR marathon soon, aiming to get a release out. Any updates?

PR #1009, opened 2019-08-30
Fgbio group reads by umi

Smeds:
Module used to process fgbio GroupReadsByUmi data

 - [x] `README.md` is updated


ewels:
x-ref https://github.com/ewels/MultiQC_TestData/pull/125

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## MultiQC v1.8dev
 
 #### New Modules:
+* ['**fgbio**](http://fulcrumgenomics.github.io/fgbio/)
```
ewels:
```suggestion
* [**fgbio**](http://fulcrumgenomics.github.io/fgbio/)
```

multiqc/modules/fgbio/groupreadsbyumi.py
```diff
@@ -0,0 +1,80 @@
...
+
+""" MultiQC module to parse output from fgbio GroupReadsByUmi
+"""
+
+############################################################
+######  LOOKING FOR AN EXAMPLE OF HOW MULTIQC WORKS?  ######
+############################################################
+#### Stop! This is one of the most complicated modules. ####
+#### Have a look at Kallisto for a simpler example.     ####
+############################################################
```
ewels: I think you can remove this comment.. It's not that complicated 😉

multiqc/modules/fgbio/groupreadsbyumi.py
```diff
@@ -0,0 +1,80 @@
...
+
+from __future__ import print_function
+from collections import OrderedDict
+import io
+import json
+import logging
+import os
+import re
+import zipfile
```
ewels: Most of these imports aren't used - please delete any that aren't needed.

multiqc/modules/fgbio/groupreadsbyumi.py
```diff
@@ -0,0 +1,80 @@
...
+    def __init__(self):
+
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__(name='Fgbio', anchor='fgbio',
+        href="http://fulcrumgenomics.github.io/fgbio/tools/latest/GroupReadsByUmi.html",
+        info="fgbio is a command line toolkit for working with genomic and particularly next generation sequencing data.")
+
+        self.umi_data = dict()
+
+        # Find and parse unzipped FastQC reports
```
ewels:
```suggestion

```

multiqc/modules/fgbio/groupreadsbyumi.py
```diff
@@ -0,0 +1,80 @@
...
+        for f in self.find_log_files('fgbio/groupreadsbyumi'):
+            self.parse_hist_report(f)
+
```
ewels: If no logs are found, we need to raise a `UserWarning` here to stop the module code going any further.

https://multiqc.info/docs/#no-files-found

ewels: Please also filter the samples and print to the sources file:

https://multiqc.info/docs/#filtering-by-parsed-sample-names
https://multiqc.info/docs/#printing-to-the-sources-file

multiqc/modules/fgbio/groupreadsbyumi.py
```diff
@@ -0,0 +1,80 @@
...
+            'title': 'Familizy size count',
+            'ylab': '# UMI',
+            'xlab': 'Reads supporting UMI',
+            "xmax": 15,
+            'xDecimals': False
+        }
+
+        return linegraph.plot(self.umi_data, config)
+
+    def get_s_name(self, f):
```
ewels: Instead of this function, please just add `.histo` to the `extra_fn_clean_trim` dict in the config defaults.

See https://multiqc.info/docs/#sample-name-cleaning

multiqc/utils/search_patterns.yaml
```diff
@@ -257,6 +257,8 @@ fastqc/theoretical_gc:
 featurecounts:
     fn: '*.summary'
     shared: true
+fgbio/groupreadsbyumi:
+    fn: '*.histo.tsv'
```
ewels: Instead of searching by filename, I think we can look for the presence of the string `fraction_gt_or_eq_family_size` in the first line of the file. This is a pretty specific column title, so should be unique to this tool.

This makes it easier for people to run with the tool output, and also makes it less likely for the module to trip up with other histogram data.

```suggestion
    contents: 'fraction_gt_or_eq_family_size'
    num_lines: 1
```

ewels: Please also remove the other help text about needing the filename pattern if you agree with the above 👍

CHANGELOG.md
ewels: Please also add to the main `README.md` table

multiqc/modules/fgbio/groupreadsbyumi.py
Smeds: dependencies have been removed.

multiqc/modules/fgbio/groupreadsbyumi.py
Smeds: Warning is now raised when no logs are found. Sample filtering should now work. printing-to-the-sources-files should also be working

multiqc/modules/fgbio/groupreadsbyumi.py
Smeds: we now use the content of the file to locate files that should be processed I will stop trying to remove .histo. since user may not be including histo in their file names.

ewels:
Nice - thanks for working on this! I might make a few tweaks before I merge if that's ok, but probably easier if I just do them instead of doing a lot of back and forth. Let me know when it's safe for me to do so without clobbering anything that you're actively coding right now..

Smeds:
@ewels  comments should now have been resolved.

ewels:
Thanks for this! I worked on the code a little myself before merging, see 1876e150f5cd133f0d29f7d239e2aae3023f8195 and 333cd9ddea7f256bc5c23461d55f6fcb53f4c2c4 - it was fairly minor stuff though. Let me know if you find any problems with the changes I made.

Phil

PR #1020, opened 2019-10-01
bcl2fastq: parse undetermined barcodes when lane indexes do not start at 1

cschu:
 - [n/a] (optional but recommended): https://github.com/ewels/MultiQC_TestData contains test data for this change

This PR addresses (non-standard) situations in which bcl2fastq was run on lanes not starting with lane index 1. Before the fix, the "Undetermined barcodes per lane" plot would be missing, instead  an error message would be shown that the plot could not be created. The logfile would show a warning that there was no data available to generate the bargraph. The cause of the issue is the assumption that lanes present in the bcl2fastq reports start at 1. The specific code section used this assumption to index into the list contained in the UndeterminedBarcodes section and pull out the data at a specific [lane] index. In the affected situations (e.g. only lanes 3, 4 are present in the report instead of 1, 2, 3, 4), the code would try to access the list at [index - 1 (e.g. 3 - 1= 2 => out of bounds)], raising an IndexError, which would then be dealt with by setting the extracted data to None. The fix uses the LaneIndex value of each item in the UndeterminedBarcodes section, which is more robust in any case.

<tldr> UndeterminedBarcodes data were previously accessed based on an unreliable assumption. This PR fixes this. </tldr>


multiqc/modules/bcl2fastq/bcl2fastq.py
```diff
@@ -166,14 +166,13 @@ def parse_file_as_json(self, myfile):
...
-                    (item['Barcodes'] for item in content['UnknownBarcodes'] if item['Lane'] == 8),
-                    None
-                )
-            run_data[lane]['unknown_barcodes'] = unknown_barcode
+            for lane_data in content.get("UnknownBarcodes", list()):
+                if lane_data["Lane"] == l:
+                    unknown_barcode = lane_data["Barcodes"]
+                    break
+            else:
+                unknown_barcode = dict()
```
ewels: Instead of `break` / `else` syntax, I'd prefer to just initialise this empty `dict()` before the `for` loop and then it can be overwritten. The `break` / `else` syntax isn't super common and I'm worried that future edits might lead to a bug that could be quite tricky to track down.

cschu: Makes sense. Done.

PR #1029, opened 2019-10-16
Sexdetermine

apeltzer:
Work in progress, hopefully, done today 🤞 

 - [ ] `README.md` is updated


ewels:
You can't add the key `'NR Aut'` to `dict_to_plot[s_name]` if `s_name` isn't already initialised in `dict_to_plot`. You need to create an empty dict first:

```python
dict_to_plot[s_name] = {}
```

PR #1042, opened 2019-10-28
Proper __main__.py file

ewels:
Hi @KerstenBreuer,

Here's my final PR after playing around with your changes a bit. Your version was good except I found a couple of minor problems:

* This code never ran, because the namespace had changed:
https://github.com/ewels/MultiQC/blob/482c007445b32f62e17f3030a24c69a6ddccbd45/scripts/multiqc#L799-L807
* The main log messages were prefixed with `__main__` instead of `multiqc`
* The `click` command line help was showing `__main__.py`:
     ```
     Usage: __main__.py [OPTIONS] <analysis directory>
     ```

My solution is pretty horrible, with the `multiqc` name being thrown around everywhere 😰 but I _think_ everything now works... At least, both `python -m multiqc` and `multiqc` seem to work and give output identical to before I think.

If you could have a look over this and give it a quick review, that would be great 👍 

Thanks,

Phil


ewels:
Ugh, python 2x module imports failing 🤦‍♂

KerstenBreuer:
Hi,

Looks very good to me and works like a charm. Except of course with python2. But is this critical to you. Python2 is now officially depreciated and most packages are dropping support in newer versions.

Nevertheless, I will try to find the reason for the import error.

Thanks for reviewing my changes,
Kersten

ewels:
Hi @KerstenBreuer,

Sorry, I wrote that as I was running out of the door to go home. I know why the import error is happening - it's because of conflicting `multiqc` namespaces. It should be an easy fix I think - change `multiqc.py` to `somethingelse.py`. But then this breaks the logging output, which is what I was trying to solve in the first place.

I will be dropping Python 2x support soon, but I was hoping to wait until 2020 (like everyone else - https://pythonclock.org/).

I'll have a play with this again and see if I can bring myself to compromise on the name. It's just me being too much of a perfectionist 🤓  

Phil

ewels:
Gah, I hate this kind of bugfixing. I fixed the import problems for py2, but now we're hitting encoding errors (works on simple stuff, fails on [difficult stuff](https://github.com/ewels/MultiQC_TestData/tree/master/data/special_cases/%C3%B6dd_ch%C3%A5r%C3%A4ct%C3%A9rs)). Before the code restructuring it was fine.

I have no idea why moving the same code around into different files could have this effect. Python 3 seems fine still.

ewels:
And with your code in 066db2f tests were _also_ passing.. And I really don't see what I changed since then that would have this effect.. 🤔 (I tried deleting the default encoding strings and other things that I added since then, but no change). Any ideas?

ewels:
Ahhh I've got it! I found a nasty hack on the web but thought that it was a bit rough and I'd rather avoid it. I just realised that I put exactly that hack in the Python 2 code 4 years ago 😆 https://github.com/ewels/MultiQC/commit/455321d4e8ffa563a8d04a8a764b1984b72b8597

I must have dropped this when switching the scripts over, hence suddenly getting encoding errors...

ewels:
Ok brilliant. I’m super happy with this now - thank you @KerstenBreuer!

KerstenBreuer:
Cool :) I am happy I could help. @ewels thank you so much. Now, I can integrate MultiQC into CWLab.

KerstenBreuer:
Do you have a rough idea, when the next release will be?

ewels:
Soon hopefully, getting there now: https://github.com/ewels/MultiQC/milestone/14

PR #1056, opened 2019-11-14
Stacks: Fixed parsing of multiline populations sum stats

### LINKED ISSUE #906: Stacks population summary stats limited to the last population
**Description of bug:**
The summary stats of  the stacks module return an array listing only the last population of a data sets with more than one pop.

**MultiQC Error log:**
```
No error log for this behavior.
```

**File that triggers the error:**
In line 259 of the file : `multiqc/modules/stacks/stacks.py` there is only one entry that is created in the out_dict. The content is iteratively replaced by values  from a loop over the analysed populations 
```python
out_dict[s_name] = cdict
```

This should be 
```python
out_dict[content[0]] = cdict
```

`content[0]` is the population label

Best regards

remiolsen:
When specifying multiple populations for `--popmap` in Stacks, the resulting `populations.sumstats_summary.tsv` file was not parsed correctly.

closes #906

### DIFF
--- a/multiqc/modules/stacks/stacks.py
+++ b/multiqc/modules/stacks/stacks.py
@@ -250,8 +250,6 @@ class MultiqcModule(BaseMultiqcModule):
         for l in var_lines:
             if l.startswith("# Pop ID	Private	Num_Indv"):
                 headers = l.split("\t")
-            elif l.startswith("# All positions (variant and fixed)"):
-                break
             elif headers is not None:
                 cdict = OrderedDict()
                 content = l.split("\t")
--- a/multiqc/modules/stacks/stacks.py
+++ b/multiqc/modules/stacks/stacks.py
@@ -248,14 +248,16 @@ class MultiqcModule(BaseMultiqcModule):
         var_lines = fl[fl.index("# Variant positions")+1:fl.index("# All positions (variant and fixed)")]
         headers = None
         for l in var_lines:
-            if l.startswith("# Pop ID"):
+            if l.startswith("# Pop ID	Private	Num_Indv"):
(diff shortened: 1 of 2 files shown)


multiqc/modules/stacks/stacks.py
```diff
@@ -248,14 +248,16 @@ def parse_sumstats(self, file_contents, s_name):
         var_lines = fl[fl.index("# Variant positions")+1:fl.index("# All positions (variant and fixed)")]
         headers = None
         for l in var_lines:
-            if l.startswith("# Pop ID"):
+            if l.startswith("# Pop ID	Private	Num_Indv"):
                 headers = l.split("\t")
+            elif l.startswith("# All positions (variant and fixed)"):
```
ewels: Instead of this, can we just check the number of elements in `content`?

```suggestion
            elif headers and len(headers) != len(content):
```

remiolsen: I don't know how that `elif` statement got in there, it's redundant. See [L248](https://github.com/ewels/MultiQC/pull/1056/files#diff-0ddfc5f28b0365aabb99df67f1a415bdR248) 🤦‍♂ 

The parsing is only going to happen between those two lines to grab that table. I don't find the `# All positions` table terribly interesting, but that's just me.

remiolsen:
Wait for https://github.com/ewels/MultiQC_TestData/pull/133 I guess

multiqc/modules/stacks/stacks.py
ewels: Ah, nice approach 👍

PR #1013, opened 2019-09-20
Allow multiple libraries with Picard's MarkDuplicates

singerj:
Many thanks for this great tool. I find it very informative and use a quite a lot.

## If this PR is _not_ a new module
 - [x] This comment contains a description of changes (with reason)
 - [x] `CHANGELOG.md` has been updated
 - [x] (optional but recommended): https://github.com/ewels/MultiQC_TestData contains test data for this change

Issue #402, opened 2017-02-17
Mark Duplicates: "Error - was not able to plot data."


astewart-twist:
Any idea what this error means?

ewels:
Hi @astewart-twist,

Thanks for the bug report - this error was in the HTML report, right? I think it's the super general catch-all message that tries to stop the entire page from crashing if something goes wrong with a plot. So it could be a lot of things.

Would it be possible to send through the report where you see the error? Could you give a bit more information about how you generated the report, what it contained, whether other plots loaded, what system and browser you're using? Finally, if you're able to view the browser developer console and see any other error messages, that would be helpful.

Thanks,

Phil

astewart-twist:
Apologies @ewels , this was actually a picard tools bug.  Thanks though!

ewels:
Ok, that's always a relief to hear! What happened, out of curiosity? Could help me in the future if someone else has the same problem, plus in an ideal world MultiQC should complain and not try to plot anything if there's a problem with the data.

Phil

astewart-twist:
Sure, @ewels.  It turns out it had something to do with the way I was running `picard MarkDuplicates`.  There would be a row with an empty sample name and empty stats.  

I can paste an example output:

```
## htsjdk.samtools.metrics.StringHeader
# picard.sam.markduplicates.MarkDuplicates INPUT=[sample.bam] OUTPUT=sample.deduped.bam METRICS_FILE=sample.metrics VALIDATION_STRINGENCY=SILENT    MAX_SEQUENCES_FOR_DISK_READ_ENDS_MAP=50000 MAX_FILE_HANDLES_FOR_READ_ENDS_MAP=8000 SORTING_COLLECTION_SIZE_RATIO=0.25 PROGRAM_RECORD_ID=MarkDuplicates PROGRAM_GROUP_NAME=MarkDuplicates REMOVE_DUPLICATES=false ASSUME_SORTED=false DUPLICATE_SCORING_STRATEGY=SUM_OF_BASE_QUALITIES READ_NAME_REGEX=[a-zA-Z0-9]+:[0-9]:([0-9]+):([0-9]+):([0-9]+).* OPTICAL_DUPLICATE_PIXEL_DISTANCE=100 VERBOSITY=INFO QUIET=false COMPRESSION_LEVEL=5 MAX_RECORDS_IN_RAM=500000 CREATE_INDEX=false CREATE_MD5_FILE=false 
## htsjdk.samtools.metrics.StringHeader
[...]

PR #1058, opened 2019-11-15
Picard markdups mergelibs

ewels:
Continuation of https://github.com/ewels/MultiQC/pull/1013

Original description by @singerj:

> **Updated MarkDuplicates.py for multiple libraries**
> The current implementation only considers the firs library of the Picard's MarkDuplicates metrics file. This pull request combines the different libraries by simply adding them up. The only exception is `PERCENT_DUPLICATION`, which is recomputed at the end.
>
> Note that also `ESTIMATED_LIBRARY_SIZE` is added up even though it is really a library feature. However, currently only the first library is used.

---

With a few tweaks I managed to get the code working on all test-data files in the repository. However, I don't want to merge into `master` just yet as this is a fairly major change in behaviour and may affect a lot of people. My main concern is for files like [this test file](https://github.com/ewels/MultiQC_TestData/blob/master/data/modules/picard/MarkDuplicates/empty_sample.txt) which may have `unknown` / `undetermined` etc and we want to ignore these? But maybe this never happens or the current default is wrong?

I think that this feature is probably well suited to an opt-in approach via a new config option, but I thought it was worth a little discussion first.

Any thoughts? @singerj / @parlundin / others?


singerj:
Hi @ewels ,

Sorry for the late reply. 

I totally agree with your concerns regarding the unknown/undetermined reads. However, in general I am interested in the overall statistics of the read distribution of a particular file, without caring too much which library the reads belong to. Therefore, I would include the unknown/undetermined reads. Maybe one could raise a warning, letting the user know that there is a problem? Or quit MultiQC and tell the user to specify a certain config option before relaunching? 

Also, the [test file](https://github.com/ewels/MultiQC_TestData/blob/master/data/modules/picard/MarkDuplicates/empty_sample.txt) looks a little bit strange to me. Row 7 does not contain a library name and has a "?" at the end. Also, why is there row 8 if there were no reads at all? Do I miss something?

Many thanks for taking over and continuing working on this problem!

ewels:
Hi @singerj,

> Also, the test file looks a little bit strange to me. Row 7 does not contain a library name and has a "?" at the end. Also, why is there row 8 if there were no reads at all? Do I miss something?

That test file comes from this issue thread: https://github.com/ewels/MultiQC/issues/402#issuecomment-281819604 - I added it as files like that used to crash MultiQC and I wanted to keep testing for it.

I've gone over your code in quite a bit of detail and ended up working on it quite a lot myself - see the last two commits. The merging behaviour is now configurable, but on by default. Libraries can be kept separate with the following MultiQC config though:

```yaml
picard_config:
    markdups_merge_multiple_libraries: False
```

To make this work I had to refactor quite a lot of the module code. I also added in functionality to recalculate the `ESTIMATED_LIBRARY_SIZE` field, which also needed recalculation in addition to the `PERCENT_DUPLICATION` field. I took this from the Picard codebase and converted to Python, it seems to work well..

Please have a go with this new version and let me know if you find any bugs!

Phil

PR #1064, opened 2019-11-26
RNA-SeQCv2 Module

m-ish:
Hello! Here is a new module for the changed output of rnaseqc
 - [x] `README.md` is updated


ewels:
Nice! Thanks for this @m-ish. Before I properly review, couple of quick thoughts / suggestions from a 20 second code skim:

* As this isn't technically a new tool, I think it would be good to keep a single RNA-SeQC tool. They seem to still be calling it RNA-SeQC and not RNA-SeQC2, so I think we should do the same.
* It would be good to keep the existing functionality for old log files. I think we can probably keep all of that code untouched and just have new file search patterns?
* If you want to keep the code for the two versions separate for cleaner code / easier maintenance, you can look at how [samtools](https://github.com/ewels/MultiQC/tree/master/multiqc/modules/samtools) and [picard](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/picard/picard.py) and others organise their code for different sub-commands

Apologies if you're already doing some of this, it was a very quick skim.. 😅

m-ish:
I shall work on that now :) thanks!

m-ish:
Hello again @ewels ! Finished combining both RNASeQC versions 1 and 2 into a single file. 
Having problems with your checks however - to be honest I have little experience with Travis CI and Appveyor - how may I fix these problems?

Hope you can shed some light, and thanks a bunch!

Hamish

ewels:
I've actually just removed Travis and AppVeyor and replaced them with GitHub Actions. Not sure what happens now - I guess if you do another commit then they will be replaced?

Anyway, you can see the errors here: https://travis-ci.org/ewels/MultiQC/jobs/619451371#L2631

If you run locally with `--lint` you should also get them too. Looks like you are trying to set two page elements with the same ID.

Phil

m-ish:
Ah my bad! I shall run multiqc with the test data you supplied - I think that is why I am not seeing it on my end, even with --lint!

Cheers!

m-ish:
Hey mate! All done now - let me know if there are any issues!

Cheers,

Hamish

ewels:
Hi @m-ish,

Thanks again for working on this. I just had a go over the code and have now merged it, albeit after a fairly major refactor / rewrite. Although you had merged your code into the existing RNA-SeQC module, it was still pretty much separate meaning that if you ran MultiQC with a mixture of outputs you would get  two of each of the sections, with a somewhat random mixture of the different samples in each (the same with the table columns).

Most of the code was basically an identical copy+paste from the previous version of the module, so I was able to remove your new copies and it pretty much ran fine (with a couple of tweaks). Now all of the data from both versions of the tool sit together and there is a lot less code duplication (easier to maintain).

Please have a check over the new code and test it if you are able, let me know if you find any problems.

Thanks,

Phil

Issue #1076, opened 2019-12-05
Cannot order multiple-run samtools gen. stats. columns when changing name 


jfy133:
**Description of bug:**
I want to run the samtools module twice, picking up a `samtools flagstat` log file before and after a BAM filtering step.

In the MultiQC config file, I have specified under `top_modules:` two entries of the `file path_filter` accordingingly, each time specifying a unique name (i.e. Samtools Flagstat (pre-samtools filter) and Samtools Flagstat (post-samtools filter)).

When I then try and specify in the General Stats table where the two columns should be positioned using `table_columns_placement`, the ordering is ignored when using the 'custom' unique name to refer to each column accordingly.

If removing the 'custom name', the ordering works correctly (i.e. Samtools Flagstat).

This is not ideal as then it is not clear to a usable which column comes from which step.

The custom names _do_ work for other modules in the same config (e.g. FastQC). 

**MultiQC Error log:**

This contains all the log files and the config file.

[multiqc_samtools_issue.zip](https://github.com/ewels/MultiQC/files/3927098/multiqc_samtools_issue.zip)

**MultiQC run details (please complete the following):**
 - Command used to run MultiQC: `multiqc -f . -c multiqc_config.yaml`
 - MultiQC Version: 1.8
 - Operating System: Ubuntu 18.04
 - Python Version: 3.7.1
 - Method of MultiQC installation: conda

 **Additional context**

jfy133:
Maybe due to: https://github.com/ewels/MultiQC/blob/e4a6dfd8a008b85835139106b0d95f1008cce3d0/multiqc/modules/samtools/flagstat.py#L46

in

```
            flagstats_headers['mapped_passed'] = {
                'title': '{} Reads Mapped'.format(config.read_count_prefix),
                'description': 'Reads Mapped in the bam file ({})'.format(config.read_count_desc),
                'min': 0,
                'modify': lambda x: x * config.read_count_multiplier,
                'shared_key': 'read_count',
                'placement' : 100.0
            }
```

This entry isn't present in the FastQC module, for example (AFAICS)

PR #1086, opened 2019-12-20
Remove' hardcoded' samtools and damage profiler column name(s)

jfy133:
Addresses: https://github.com/ewels/MultiQC/issues/1076

The current version hardcodes the general stats column names using the `headers` variableof the `general_addcols()` function in `base_module.py`.

This PR removes this requirement, allowing filling "in required config variables if not supplied" by the function.

https://github.com/ewels/MultiQC/blob/7e0f2fa84285dcff848bcbb0b1584686c336271f/multiqc/modules/base_module.py#L277-L315

**EDIT** Throwing in an extra minor fix where I couldn't inititally customise DamageProfiler columns in the general stats table (via custom config) as the module 'name' was `DamageProfiler:` rather than `DamageProfiler` as in other tools. Also makes it prettier so columns are now `DamageProfiler: <my_column>` rather than `DamageProfiler:: <my_column>


ewels:
Thanks for this @jfy133! I guess this was mostly a cosmetic addition so that you can see which samtools subtool the numbers come from when you hover on a column. But it doesn't add much, and if removing this makes your code work then I'm happy with that 👍

Issue #1055, opened 2019-11-14
Illumina DRAGEN On-Premise


pellonen:
**Tool Details**
 - Name of tool: Dragen Bio-it on-premise
 - Tool homepage: https://emea.illumina.com/products/by-type/informatics-products/dragen-bio-it-platform.html#learn-more-onprem
 - Tool description: accelerated software/hardware package for alignment & variant calling
Dragen Enrichment, Dragen Germline, Dragen RNA pipe, Dragen Somatic pipe

 - Data suitable for MultiQC plot(s): GATK/SAMTOOLS equivalents
 - Most interesting data for General Stats table: SAMTOOL equivalents

Demo data available in BaseSpace, RNA example:
https://basespace.illumina.com/s/LucXlKME2uYC
Thanks, -pekka

ewels:
I can't use the crappy BaseSpace downloader tool because it won't ask MacOS Catalina for permission to access the disk properly, so it says that I have 0 bytes available.

If you can download the log files locally, could you please make a pull-request to https://github.com/ewels/MultiQC_TestData with them please? (in a folder `data/modules/dragen`).

Thanks!

ewels:
(not big BAM files etc, but all log files is fine)

pellonen:
Hi,
Sorry, haven’t had time for this. Is this still relevant? Met John Wilson today at FIMM, he was presenting Dragen here, great stuff!
-Pekka

ewels:
I can't write the module without example data, and I can't currently download the example data myself.. So this issue will be on hold until I can get a copy.. :)

Looking forward to hearing more about it in person!

vladsaveliev:
Hi @ewels, we are started evaluating Dragen as well, and I was looking up if there is is a MultiQC module for it, so ended u here :) Attaching an example output for tumor/normal alignment and variant calling Dragen pipeline. 

[T_SRR7890936_50pc.tar.gz](https://github.com/ewels/MultiQC/files/3895507/T_SRR7890936_50pc.tar.gz)

I'm gonna take look if I can write up a module for this pipeline. I don't have an example for RNA yet - that's gonna be next.

Vlad

ewels:
Ok fantastic! Thanks @vladsaveliev, looking forward to it..

vladsaveliev:
[...]

PR #1098, opened 2020-01-25
Illumina Dragen module

vladsaveliev:
Just creating a pull request again after getting feedback from Illumina. They are working on more QC stuff for Dragen, so will just open pull requests for any new stuff in the future.

---

Following the request at https://github.com/ewels/MultiQC/issues/1055, started a module that supports QC files from a tumor/normal variant calling pipeline (targeted or whole genome). Reporting on:
- mapping metrics (similar to `samtools stats`)
- coverage stats (similar to `mosdepth` and `qualimap bamqc`)
- variant calling QC
- run time metrics
- fragment size metrics 
- and ploidy estimation.

Perhaps doesn't fit in perfectly as a typical MultiQC module because Dragen is more like a Swiss army knife that does tons of jobs from demultiplexing to SV calling, instead of focusing more or less on one certain task, so it's hard to even classify it as "pre-processing", "aligner" or "post-alignment". However we will be using Dragen in UMCCR and this module is gonna be very useful for us, so perhaps it makes sense to have it in one way or another. Open to suggestions!

Screen Shot 2020-01-06 at 15 26 49

Full example report for a WGS tumor/normal variant calling pipeline:
[multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/4038848/multiqc_report.html.zip)

 - [x] There is example tool output for tools in the https://github.com/ewels/MultiQC_TestData repository - https://github.com/ewels/MultiQC_TestData/pull/134
 - [x] `README.md` is updated


ewels:
Hi @vladsaveliev,

Just getting to this now - apologies for the delay. This is a colossal PR so reviewing it is going to be a bit of a challenge, but I'll do my best.

Before I start digging into the code, a few general points for discussion:

1. At the top of the report I get a comment box about the genome size:
image
     * I don't love this to be honest. It doesn't follow the standard MultiQC pattern, in that no other modules do this. I'm also not sure that it's important enough information to warrant such a prominent position in the report? Finally, what happens if there are multiple genomes with different sizes?
     * Any reason that this info can't just go into a table like everything else?
2. As you say, this module covers a massive number of tools. How are these analyses run? The way that you've written the code is very clear and well separated, but if there is a way to split these into multiple modules in a way that makes sense with how Dragen is run, that could be nice.
    * For example, maybe have a module called _"Dragen Mapping"_, then one called _"Dragen Coverage"_,  _"Dragen Variant Calling"_ etc? It would make the report a little easier to navigate if nothing else.
    * I'm assuming that you've already thought about this. If this is the case then I'm happy with how it's currently written as the code is already well organised :+1:
3. The _General Statistics_ table has waaay too many columns. I know most are hidden, but having 148 columns is still kind of extreme. It makes the "Configure Columns" dialogue kind of unusable and bloats the report with a lot of data that most people won't use.
    * Can we remove a lot of these? My mantra is that MultiQC should only report the most interesting stuff and that people can always go back to the original reports to dig into every metric.
    * I would question the utility of some of the visible columns too. Do people really care about Raw + Aligned bases in addition to reads and percentages for example?
    * I imagine that quite a few people may run additional downstream variant analysis on these files, so we can expect to have several more columns added on in most reports.
    * If you want to maintain flexibility, you can use the [same approach](https://github.com/ewels/MultiQC/blob/0a2e583b4d2e0ed16174cc351ec81a3d46f442ed/multiqc/modules/picard/HsMetrics.py#L211-L254) that I used in a recent refactoring for Picard HsMetrics - have a default list of columns to include but allow this to be overwritten by a MultiQC config option.
4. General stats columns should typically be ordered in chronological order of analysis from right to left.
     * Currently it looks like the order is the opposite of this, with raw data on the left and VCF analysis off the screen to the right. This will be confusing when combined with other modules.
5. Order of sections is also a bit backwards
    * As with _General Statistics_, report sections should be ordered from the start of analysis at the bottom to end at the top (eg. mapping > coverage analysis > variant analysis)
6. I think the run-time barplot could be misleading and should probably be removed
     * Although it's cool, I'm not sure how useful this plot will be for most downstream users? It could also be misleading if there are other tools combined in the report
     * Generally I would leave this kind of plotting to the analysis framework outside of MultiQC, rather than within the MultiQC report itself.
     * Maybe this could be disabled by default and enabled with a MultiQC config option?
7. _Coverage Metrics_ section has a long description. Most of this should be moved to the help text.
8. The _Mapped and paired per RG_ barplot is... bright... 🤢 😆 Any reason for this colour choice?

Once again, thank you for this massive amount of work! Let me know what you think about the above then I'll start digging into the code a bit more.

Cheers,

Phil

ewels:
ps. Are those _tabs_ in `mqc_colour.py`?? Please replace them with spaces 😓

vladsaveliev:
Thanks for reviewing! Agree with all your comments. I guess an issue here is that DRAGEN is not a small specific tool that can be a component of a larger report, like it's intended for a MultiQC module. My goal was to make a self-sufficient report similar to what bcbio-nextgen generates, that covers all parts of the analysis. So I'm thinking I'm gonna extract the essentials for the module, and leave the rest for a separate codebase. And still I guess many metrics should be removed anyway in both, good point that users can alway go back to raw files.

Vlad

ewels:
Ok great - that sounds like a good plan. Keep MultiQC relatively lightweight but also build a full report. Let me know if I can help and of course feel free to use MultiQC for the full dragen module if you'd like (it could be a plugin if you want?). Whatever is easiest for you.

I'll mark this PR as a WIP for now in that case. Let me know when it's ready for review.

Phil

vladsaveliev:
Thanks again for the review! Think I addressed all your comments.

> At the top of the report I get a comment box about the genome size. I don't love this to be honest. It doesn't follow the standard MultiQC pattern, in that no other modules do this. I'm also not sure that it's important enough information to warrant such a prominent position in the report? Finally, what happens if there are multiple genomes with different sizes? Any reason that this info can't just go into a table like everything else?

If there are different genomes, the metric would have appeared in the main table as another standard column. The idea was to move away repeated metrics into the header to avoid rows of repeated values in the table. I primarily used it to show the size of the target for exome/amplicon sequencing, but in this WGS example it's not so useful. But agree that it doesn't follow standard MultiQC pattern, so I just dropped it.

> As you say, this module covers a massive number of tools. How are these analyses run? The way that you've written the code is very clear and well separated, but if there is a way to split these into multiple modules in a way that makes sense with how Dragen is run, that could be nice. For example, maybe have a module called "Dragen Mapping", then one called "Dragen Coverage", "Dragen Variant Calling" etc? It would make the report a little easier to navigate if nothing else.

Tried to implement it this way, but turns out it's not so elegant after all: need to keep separate docs sections for each "module", can't run with simple `-m dragen`, and so on. The argument for keeping it as a single module is that Dragen almost always run as a single command starting from raw fastq (or even basecalls), so in a way it can be treated as an aligner with additional features on top. Users will unlikely use variant calling or coverage analysis outside of read mapping. So I'll just treat it as an alignment tool and place it accordingly in the module_order list, in docs, etc.

> The General Statistics table has waaay too many columns. I know most are hidden, but having 148 columns is still kind of extreme.

Cut it down to 36 columns :) And only 9 columns displayed by default. A lot of optional metrics are still available in specific tables in corresponding sections (i.e. Variant calling, Coverage metrics, Mapping metrics), so no can drop them from the general stats without regrets. 

> General stats columns should typically be ordered in chronological order of analysis from right to left.
> Order of sections is also a bit backwards

Reordered.

> I think the run-time barplot could be misleading and should probably be removed

Dropped it.

> Coverage Metrics section has a long description. Most of this should be moved to the help text.

Moved to the help section.

> The Mapped and paired per RG barplot is... bright... 🤢 😆 Any reason for this colour choice?

Huh, no reason, just wanted it to be different from another bar plot (Mapped and duplicated) to make them visually distinguishable. Now made the colors less bright, hope it looks better now.

> ps. Are those tabs in mqc_colour.py?? Please replace them with spaces 😓

Oooops, sorry :)

Example of an updated report:

[multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/4535088/multiqc_report.html.zip)


Vlad

ewels:
> > ps. Are those tabs in mqc_colour.py?? Please replace them with spaces 😓
>
> Oooops, sorry :)

Not your fault! The tabs were already there, must have been me I guess...

ewels:
Thanks @vladsaveliev for the speedy response - looking great! I'll do one more set of general comments here from looking over the report and then start doing code suggestions.

> The idea was to move away repeated metrics into the header to avoid rows of repeated values in the table.

I think removing it is probably best, but what I've done in other modules is hide such columns if they all contain a single value and show it if there are multiple. Same logic..

> A lot of optional metrics are still available in specific tables in corresponding sections (i.e. Variant calling, Coverage metrics, Mapping metrics), so no can drop them from the general stats without regrets.

Tables are looking much better! Many of the hidden columns in General Stats seem to be the same as those in the tables below though, can we not just remove them from General Stats completely? Is there a reason to keep them there?

Ok, general points from eyeballing the new report:

1. Tables with read counts
    * This is a personal thing, but typically I've done _"M Input Reads"_ in the column header and then _not_ had a `M` text suffix for every table cell. I find this easier to read myself, but it's a minor thing.
2. Could we have a tab for normalised data in the coverage plots?
    * The _Coverage distribution_ and _Coverage per contig_ plots only have read counts, which means that one high-depth sample can make the plot kind of unreadable. It needs custom code to normalise, but you could add a tab for that data here if you think it's a good idea.
3. _Coverage metrics_
    * More of a question: When is `50x..100x` interesting when you already have `>50x` etc. columns?
    * Do we need all of the `>n X` percentages when the above plot is showing this data with greater detail already? I would be tempted to drop all of these columns and show all of the remaining, but that's just me (and I'm not using this kind of data a lot).
3. _Mapping metrics per RG_ > _Median insert size_: I guess these are ints? All have `.0` in the demo report. Can configure not to show decimals.
4. _Mapped and duplicated per RG_ and _Mapped and paired per RG_ plots
    * I wonder if we can combine these into a single plot, with two separate datasets? They are pretty similar data and the read group stuff is probably not super interesting most of the time, so I feel that it would make a natural "second" dataset. [[docs](https://multiqc.info/docs/#switching-datasets)].
5. _Fragment length hist_
    * Plot is fairly noisy and has quite a lot of points - 1383 in the demo report. Could maybe look cleaner by using `smooth_points` with the added bonus of reducing report size? (Would need some trial and error - could also not look great of course).

Getting to the smaller stuff now!

CHANGELOG.md
```diff
@@ -78,6 +78,8 @@ to break. If you haven't already, **you need to switch to Python 3 now**.
     * Relative coverage and error rate of X and Y chromosomes
 * [**SNPsplit**](https://github.com/FelixKrueger/SNPsplit)
     * Allele-specific alignment sorting
+* [**DRAGEN**](https://www.illumina.com/products/by-type/informatics-products/dragen-bio-it-platform.html)
```
ewels, line 81: Needs moving up to v1.9dev section, sorry.

multiqc/modules/dragen/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+        # <output prefix>.wgs_contig_mean_cov_tumor.csv    - same
+
+        self.add_mapping_metrics()
+        # <output prefix>.mapping_metrics.csv              - general stats table, a dedicated table, and a few barplots
+
+        self.add_fragment_length_hist()
+        # <output prefix>.fragment_length_hist.csv         - a histogram plot
+
+        # self.parse_time_metrics()
+        # <output prefix>.time_metrics.csv                 - ideally an overlapping barplot, but for now a beeswarm plot with dots for time each stage for each sample has finished
```
ewels, line 66: Can remove?

ewels:
Ah, another small thing - the MultiQC log currently looks like this when I test the module:

```
Searching 42 files..  [####################################]  100%
[INFO   ]      vc_metrics : Found variant calling metrics for 1 DRAGEN output prefixes
[INFO   ] ploidy_estimation_metrics : Found ploidy estimation metrics for 1 samples
[INFO   ]   coverage_hist : Found DRAGEN coverage histogram for 2 DRAGEN output prefixes
[INFO   ] coverage_metrics : Found DRAGEN coverage metrics for 2 DRAGEN output prefixes
[INFO   ] coverage_per_contig : Found DRAGEN per-contig coverage histogram for 2 DRAGEN output prefixes
[INFO   ] mapping_metrics : Found mapping metrics for 1 samples
[INFO   ] fragment_length : Found fragment length histograms for 1 DRAGEN output prefixes
```

The long names exceed the typical length, messing up the alignment and also the messages don't conform to the usual _"Found x reports"_ syntax (not strictly enforced, but the _"Found x"_ bit usually is - see a [typical CI run](https://github.com/ewels/MultiQC/runs/621763864?check_suite_focus=true#step:8:11).

There's also multiple log lines reporting the number of samples when they will typically be the same number across all sub-modules.

Instead, it could be nicer to return the number of samples found from each log file type and then just have a single log line with the max value for all of the DRAGEN module. I think this is what I've done in other similar modules, eg. Picard etc. This should also resolve the long name thing as _"dragen"_ is pretty short and also corresponds to the module name that is used.

Sound ok?

Phil

multiqc/modules/dragen/coverage_hist.py
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env python
+from __future__ import print_function
+
+import itertools
+import math
```
ewels, line 5: Is this used? I can't see it anywhere. Same for `itertools`.

```console
$ importchecker multiqc/modules/dragen/
multiqc/modules/dragen/coverage_hist.py:4: itertools
multiqc/modules/dragen/coverage_hist.py:5: math
multiqc/modules/dragen/coverage_hist.py:7: OrderedDict
multiqc/modules/dragen/coverage_hist.py:10: bargraph, table
multiqc/modules/dragen/coverage_metrics.py:6: math
multiqc/modules/dragen/coverage_metrics.py:8: OrderedDict
multiqc/modules/dragen/coverage_metrics.py:10: linegraph, bargraph
multiqc/modules/dragen/coverage_per_contig.py:4: itertools
multiqc/modules/dragen/coverage_per_contig.py:5: math
multiqc/modules/dragen/coverage_per_contig.py:9: bargraph
multiqc/modules/dragen/fragment_length.py:5: OrderedDict
multiqc/modules/dragen/mapping_metrics.py:7: OrderedDict
multiqc/modules/dragen/mapping_metrics.py:8: config
multiqc/modules/dragen/mapping_metrics.py:10: base_format, read_format
multiqc/modules/dragen/mapping_metrics.py:11: beeswarm
multiqc/modules/dragen/ploidy_estimation_metrics.py:7: linegraph
multiqc/modules/dragen/time_metrics.py:7: Metric
multiqc/modules/dragen/time_metrics.py:8: linegraph, table, beeswarm
multiqc/modules/dragen/vc_metrics.py:5: OrderedDict, defaultdict
multiqc/modules/dragen/vc_metrics.py:7: linegraph
```

([importchecker](https://pypi.org/project/importchecker/) is a nice little tool!)

multiqc/modules/dragen/coverage_metrics.py
```diff
@@ -0,0 +1,240 @@
...
+            - Duplicate reads and clipped bases are ignored.
+            - Only reads with MAPQ > min MAPQ and bases with BQ > min BQ are considered
+            Considering only bases usable for variant calling, i.e. excluding:
+                        1. clipped bases,
+                        2. bases in duplicate reads,
+                        3. reads with MAPQ < min MAPQ (default 20),
+                        4. bases with BQ < min BQ (default 10) 
+                        5. reads with MAPQ = 0 (multimappers); 
+                        6. overlapping mates are double-counted.
+                        """.replace('\n', '<br>'),
```
ewels, line 72: Should be able to use use regular markdown (maybe needs a little more whitespace?):

```suggestion
            helptext="""
            The following criteria are used when calculating coverage:
            
            * Duplicate reads and clipped bases are ignored.
            * Only reads with `MAPQ` > `min MAPQ` and bases with `BQ` > `min BQ` are considered
            
            Considering only bases usable for variant calling, _i.e._ excluding:
            
            1. Clipped bases
            2. Bases in duplicate reads
            3. Reads with `MAPQ` < `min MAPQ` (default `20`)
            4. Bases with `BQ` < `min BQ` (default `10`)
            5. Reads with `MAPQ` = `0` (multimappers)
            6. Overlapping mates are double-counted
            
            """,
```

multiqc/modules/dragen/coverage_metrics.py
```diff
@@ -0,0 +1,240 @@
...
+        if percentage is not None:
+            data[metric + ' pct'] = percentage
+
+    m = re.search(r'(.*).wgs_coverage_metrics_?(\S*)?.csv', f['fn'])
+    sample, phenotype = m.group(1), m.group(2)
+    f['s_name'] = sample
+    return {phenotype: data}
+
+
+def parse_wgs_hist(f):
```
ewels, line 190: Looks like you can ditch this?

multiqc/modules/dragen/coverage_metrics.py
```diff
@@ -0,0 +1,240 @@
...
+    PCT of bases in wgs with coverage [10x:15x), 0.80
+    PCT of bases in wgs with coverage [3x:10x), 0.78
+    PCT of bases in wgs with coverage [1x:3x), 0.49
+    PCT of bases in wgs with coverage [0x:1x), 3.42
+
+    Fully covered by _coverage_metrics_normal.csv and _coverage_metrics_tumor.csv, so ignoring.
+    """
+
+
+def parse_wgs_overall_mean_cov(f):
```
ewels, line 208: This too..

multiqc/modules/dragen/coverage_per_contig.py
```diff
@@ -0,0 +1,158 @@
...
+        main_contigs_by_sample = {sn: data[0] for sn, data in perchrom_data_by_sample.items()}
+        other_contigs_by_sample = {sn: data[1] for sn, data in perchrom_data_by_sample.items()}
+
+        self.add_section(
+            name='Coverage per contig',
+            anchor='dragen-coverage-per-contig',
+            description='Average coverage per contig or chromosome. '
+                        'Calculated as the number of bases (excluding duplicate marked reads, reads '
+                        'with MAPQ=0, and clipped bases), divided by the length of the contig or '
+                        '(if a target bed is used) the total length of the target region spanning that contig.',
```
ewels, line 53: I've started using multi-line strings a lot more for these, I find them quite a bit easier to read:

```suggestion
            description='''
            Average coverage per contig or chromosome.
            Calculated as the number of bases (excluding duplicate marked reads, reads
            with `MAPQ=0`, and clipped bases), divided by the length of the contig or
            _(if a target bed is used)_ the total length of the target region spanning that contig.
            ''',
```

multiqc/modules/dragen/coverage_per_contig.py
```diff
@@ -0,0 +1,158 @@
...
+            plot=linegraph.plot(main_contigs_by_sample, pconfig={
+                 'id': 'dragen_coverage_per_contig',
+                 'title': 'Average coverage per contig' + (' (main contigs)' if other_contigs_by_sample else ''),
+                 'ylab': 'Average coverage',
+                 'xlab': 'Region',
+                 'categories': True,
+                 'tt_label': '<b>{point.x}</b>: {point.y:.1f}x',
+            }) + (
+                '<br>' +
+                'Non-main contigs: '
```
ewels, line 63: I think it might be nicer to have these as separate report sections? I didn't realise that they weren't at first. I think that they're different enough to warrant it, and makes the code cleaner and report navigation easier.

multiqc/modules/dragen/coverage_per_contig.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+
+
+
+
+
+
+
+
```
ewels, line 158:
```suggestion
```
🤓

multiqc/modules/dragen/mapping_metrics.py
```diff
@@ -0,0 +1,447 @@
...
+
+        # Alternatively can consider a beeswarm plot, similar to samtools flagstats:
+        # self.add_section(
+        #     name='Mapping metrics per RG',
+        #     anchor='dragen-mapping-metrics',
+        #     description="A dot plot showing DRAGEN mapping metrics for each input read group. "
+        #                 "All read counts in " + str(config.read_count_desc) + ", " +
+        #                 "all bases counts in " + str(config.read_count_desc),
+        #     plot=beeswarm.plot(data_by_rg, beeswarm_keys, {'id': 'dragen-mapping-metrics-dp'})
+        # )
```
ewels, line 101: Can remove this I think. Or, maybe better, disable it by default and look for a module-specific config option to turn it on ([docs](https://multiqc.info/docs/#user-configuration)).

multiqc/modules/dragen/mapping_metrics.py
```diff
@@ -0,0 +1,447 @@
...
+
+
+
+
+
+
+
+
+
+
```
ewels, line 447:
```suggestion
```
🤓

multiqc/modules/dragen/time_metrics.py
```diff
@@ -0,0 +1,157 @@
...
+
+# Initialise the logger
+import logging
+log = logging.getLogger(__name__)
+
+
+NAMESPACE = 'DRAGEN runtime'
+
+
+class DragenTimeMetrics(BaseMultiqcModule):
```
ewels, line 18: If we are dropping this section, this file should be deleted..

multiqc/modules/dragen/utils.py
```diff
@@ -0,0 +1,125 @@
...
+            col['scale'] = 'Purples'
+        elif metric.unit == 'x' or metric.namespace == 'Coverage':
+            col['scale'] = 'Blues'
+            if metric.unit == '%':
+                col['scale'] = 'RdBu'
+                if metric.the_higher_the_worse:
+                    col['scale'] = 'YlOrRd'
+                    # col['scale'] += '-rev'
+        elif metric.namespace == 'Variants':
+            col['scale'] = 'Purples'
```
ewels, line 64: This is cool! 😎

multiqc/modules/dragen/utils.py
```diff
@@ -0,0 +1,125 @@
...
+            )
+            if metric.unit == 'reads':
+                pct_col['scale'] = 'RdYlGn'
+            elif metric.unit == 'bases':
+                pct_col['scale'] = 'RdGy'
+            elif metric.unit == 'x' or metric.namespace == 'Coverage':
+                pct_col['scale'] = 'RdBu'
+            if metric.the_higher_the_worse:
+                pct_col['scale'] = 'YlOrRd'
+                # pct_col['scale'] += '-rev'
```
ewels, line 82:
```suggestion
```

multiqc/modules/dragen/vc_metrics.py
```diff
@@ -0,0 +1,238 @@
...
+        self.general_stats_addcols(data_by_sample, gen_stats_headers, namespace=NAMESPACE)
+
+        self.add_section(
+            name='Variant calling',
+            anchor='dragen-vc-metrics',
+            description='Variant calling metrics. Metrics are reported for each sample in multi sample VCF '
+                        'and gVCF files. Based on the run case, metrics are reported either as standard '
+                        'VARIANT CALLER or JOINT CALLER. All metrics are reported for post-filter VCFs, '
+                        'except for the "Filtered" metrics which represent how many variants were filtered out '
+                        'from pre-filter VCF to generate the post-filter VCF.',
```
ewels, line 52:
```suggestion
            description='''
                 Variant calling metrics. Metrics are reported for each sample in multi sample VCF
                 and gVCF files. Based on the run case, metrics are reported either as standard
                 `VARIANT CALLER` or `JOINT CALLER`. All metrics are reported for post-filter VCFs,
                 except for the _"Filtered"_ metrics which represent how many variants were filtered out
                 from pre-filter VCF to generate the post-filter VCF.
                 ''',
```

ewels:
So I'm not quite sure why, but `--lint` isn't throwing errors on some stuff that I expected it to. For example, all plot titles should be in the format `Dragen: xxx` (so that if they're exported then they can stand alone).

That was the main thing really. It could be that I haven't implemented it for line graphs yet and that your bar graphs happened to fit the (fairly loose) format requirements by accident.

Phil

ewels:
Ah, it looks like the module doesn't raise a `UserWarning` if no samples are found, meaning that MultiQC generates a report even if you run it on an empty directory (again I thought I had a CI check for this, but apparently not).

This is another reason for doing a single summed count of samples for all sections. [[docs](https://multiqc.info/docs/#no-files-found)].

ewels:
(NB: Updated GitHub actions to fail if running on an empty directory produces a report, so should be easier to catch this in the future)

ewels:
(NB 2: Also added in linting for line graphs now, so should catch those problems automatically in the future too..)

ewels:
Hi @vladsaveliev,

Just to note that I tried to fix the merge conflicts for you but I couldn't push changes to the PR sorry (I guess because it's coming from an organisation, so you don't get the little "allow changes to this PR" checkbox).

Phil

vladsaveliev:
Alright, thanks again for thoroughly reviewing the PR and all the detailed comments! Addressing the first part now.

> Tables are looking much better! Many of the hidden columns in General Stats seem to be the same as those in the tables below though, can we not just remove them from General Stats completely? Is there a reason to keep them there?

I was thinking that users might wanna prioritize other metrics, and this way we allow them to configure hidden/shown columns differently for their own project. Though agree that unlikely anyone will be doing this. But I'll leave it like it is now. 

> This is a personal thing, but typically I've done /"M Input Reads"/ in the column header and then/not/had a M text suffix for every table cell. I find this easier to read myself, but it's a minor thing.

I think it's actually better to be consistent across MultiQC, so I'm gonna move it into the header like in other modules. Though I personally prefer the suffix in each cell, saves horizontal space. Minor thing.

> Could we have a tab for normalised data in the coverage plots? The /Coverage distribution/ and /Coverage per contig/ plots only have read counts, which means that one high-depth sample can make the plot kind of unreadable. It needs custom code to normalise, but you could add a tab for that data here if you think it's a good idea.

Good point. Analogous plots in mosdepth and qualimap use percentages, but their data is already pre-normalize. Worth doing it here, but I'll shelf for now.

> More of a question: When is 50x..100x interesting when you already have>50x etc. columns?

My idea was that it's a "usable" range of coverage, excluding regions with extreme coverage - too low and too high. But yeah, the range might be very different for different experiments. Better hide that metric.

> 4. /Mapping metrics per RG/>/Median insert size/: I guess these are ints? All have `.0` in the demo report. Can configure not to show decimals.

Done.

> 5. /Mapped and duplicated per RG/ and /Mapped and paired per RG/ plots. I wonder if we can combine these into a single plot, with two separate datasets? They are pretty similar data and the read group stuff is probably not super interesting most of the time, so I feel that it would make a natural "second" dataset. 

Fair. Done that too.

> 6. /Fragment length hist/. Plot is fairly noisy and has quite a lot of points - 1383 in the demo report. Could maybe look cleaner by using smooth_points with the added bonus of reducing report size? (Would need some trial and error - could also not look great of course).

Done that.

multiqc/modules/dragen/coverage_hist.py
vladsaveliev: I'm actually using PyCharm, it highlights unused imports right in the editor :) 

Removed unused ones now.

multiqc/modules/dragen/coverage_per_contig.py
vladsaveliev: Normally I avoid multiline strings because you need to deal with all the whitespace and empty lines from indentation. But here it makes total sense, it gets converted into html so all whitespace is stripped.

vladsaveliev:
> Ah, it looks like the module doesn't raise a UserWarning if no samples are found, meaning that MultiQC generates a report even if you run it on an empty directory (again I thought I had a CI check for this, but apparently not). This is another reason for doing a single summed count of samples for all sections. [docs].

Huh. That's why it's worth following the manual instead of just looking at existing modules for the reference :) Added the UserException now.

vladsaveliev:
> Just to note that I tried to fix the merge conflicts for you but I couldn't push changes to the PR sorry (I guess because it's coming from an organisation, so you don't get the little "allow changes to this PR" checkbox).

Huh, good to know. I'll pull-rebase now and it should resolve the conflicts for now.

multiqc/modules/dragen/coverage_per_contig.py
vladsaveliev: Not even sure now why I did it this way. Changed into 2 separate sections now

vladsaveliev:
Okay, looks good now! Will add normalized coverage in some other PR. Along with few more upcoming updates for the module (FastQC-like QC metrics, and structural variants QC).

Again, really appreciate such a detailed review. That's awesome that you are taking such care of MultiQC even though you've probably got piles of new projects on you.

multiqc/modules/dragen/coverage_hist.py
ewels: Hmm, I wonder if there's an Atom plugin to do this.. 🤔

multiqc/modules/dragen/coverage_per_contig.py
ewels: Yup - even better, I remove the whitespace at the start of the block for you - as long as it's even:

https://github.com/ewels/MultiQC/blob/c5d3c1d8a0b865366aa50d8ef94a18727150c5a6/multiqc/modules/base_module.py#L167

ewels:
Great stuff - many thanks for this epic effort @vladsaveliev, I'm sure it will be very appreciated by a lot of people! 🎉 

I made a couple of tiny tweaks in c66262a, I hope that's ok. Let me know if you'd like me to revert anything.

Also note that there is a new option for line graphs to have report buttons for log-transformation, in the same way that you can with bar graphs. That could be useful for future normalisation PRs - see https://github.com/ewels/MultiQC/pull/1140/files for the addition and also for an obs/exp addition to a similar plot type.

Thanks again!

Phil

vladsaveliev:
Yay, thanks for merging!

All the tiny changes make total sense to me.

Re the plot buttons - yeah, I was looking at that direction, mainly I was trying to understand how the percentage buttons for barplots implemented and if I can do that for a line plot, but figured I need more time to wrap my head around. Will get back to that.

Vlad

ewels:
From (distant) memory, HighCharts has built-in functionality for making a stacked bar plot show percentages which is why it’s easy to have a config option. I don’t think that the same exists for line graphs, which is why we need to compute the values. However, I guess we could still have a similar config option and then do the computation in the core plotting code if we wanted, to make it generic... :thinking:

PR #1111, opened 2020-02-19
Add khtools extract_coding parser

olgabot:
Add parser for khtools extract_coding which predicts the protein-coding sequence of RNA-seq reads. Just got started!

 - [ ] `README.md` is updated

Used in https://github.com/czbiohub/nf-predictorthologs/pull/1 and https://github.com/nf-core/kmermaid/pull/40

### DIFF
--- a/multiqc_config.yaml
+++ b/multiqc_config.yaml
@@ -1,0 +1,4 @@
+# STAR logs are renamed to "log.final.out" instead of "Log.final.out"
+sp:
+    star:
+        fn: '*log.final.out'


ewels:
Hi @olgabot!

I'm a little confused - all that seems to be in this PR is a custom MultiQC config that overwrites a default search pattern. Did something go missing?

Phil

ewels:
Ok! I’ll mark this PR as a WIP until there is some code to review then 😀 Let me know when it’s ready..

PR #1115, opened 2020-03-03
Samblaster should clean it's sample name

nh13:


ewels:
As I guess you already found out @nh13 - `f['s_name']` is already passed through `self.clean_s_name()` in the base module 😄

PR #1116, opened 2020-03-03
Adding support for various fgbio and Picard tools

nh13:
# If this PR is for a new module
 - [x] There is example tool output for tools in the https://github.com/ewels/MultiQC_TestData repository (see https://github.com/ewels/MultiQC_TestData/pull/139)


multiqc/modules/picard/QualityByCycleMetrics.py
```diff
@@ -0,0 +1,52 @@
...
+
+    # Plot the data and add section
+    pconfig = {
+        'id': 'picard_quality_by_cycle',
+        'title': 'Picard: Mean Base Quality by Cycle',
+        'ylab': 'Mean Base Quality',
+        'xlab': 'Cycle #',
+        'xDecimals': False,
+        'tt_label': '<b>cycle {point.x}</b>: {point.y:.2f} %',
+        'ymax': 100,
```
ewels: As above, from reading the Picard / GATK docs I think it sounds like this is not a percentage?

ewels:
@nh13 I made some changes in 8fd3a4c as I went through the code for the new Picard tools. The main thing I want to check with you is where I have left comments above about percentages / phred scores. Could you confirm that my suspicions are correct here please?

multiqc/modules/picard/QualityByCycleMetrics.py
```diff
@@ -0,0 +1,52 @@
...
+    self.write_data_file(all_data, 'multiqc_picard_quality_by_cycle')
+
+    # Plot the data and add section
+    pconfig = {
+        'id': 'picard_quality_by_cycle',
+        'title': 'Picard: Mean Base Quality by Cycle',
+        'ylab': 'Mean Base Quality',
+        'xlab': 'Cycle #',
+        'xDecimals': False,
+        'tt_label': '<b>cycle {point.x}</b>: {point.y:.2f} %',
```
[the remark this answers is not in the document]
nh13: good catch

multiqc/modules/picard/QualityByCycleMetrics.py
nh13: copy-paste error

nh13:
LGTM!

ewels:
Thanks!

PR #1119, opened 2020-03-05
Adding the gathered template

nh13:
This template does not group each sub-module by module like "default" (ex. all Picard sub-module sections under Picard).  Instead, this template groups all sections (for modules and sub-modules) into a "Detailed Statistics" section.  The end-user of this template doesn't care about _which_ tool generated the section, so this reduces the number of clicks.  It also presents all the sections under one top-level section in the navigation bar.


ewels:
Nice! It's very rare that I see anyone make new templates, so I get excited when someone does it 😅 

I don't completely understand how this is less clicking to be honest, but I understand that removing tool names can make it easier for non-bioinformaticians to navigate.

Note that this breaks a few things. I noticed two:

* Custom JavaScript in the FastQC _Per Base Sequence Content_ plot
* Any module sections that don't have titles don't end up in the side-navigation (hence my tweak to Cutadapt in this PR).

Neither of these are the fault of the template though, so I don't see this as a blocker. I may look into adding a lint test to check that all sections have titles.

Thanks! 👍

PR #1114, opened 2020-03-03
[bug fix] could not find WgsMetrics

nh13:
See above

PR #1120, opened 2020-03-09
Make the WgsMetrics module compatible with GATK4

berguner:
PICARD tools WgsMetrics results starts with the line:
`## METRICS CLASS        picard.analysis.CollectWgsMetrics$WgsMetrics`
whereas GATK4 WgsMetrics results starts with the line:
`## METRICS CLASS        picard.analysis.WgsMetrics`

The new code should be able to match the results from both PICAD tools and GATK4 after removing the `CollectWgsMetrics$` pattern from the search string.

### DIFF
--- a/multiqc/modules/picard/WgsMetrics.py
+++ b/multiqc/modules/picard/WgsMetrics.py
@@ -50,7 +50,7 @@ def parse_reports(self):
                     s_name = self.clean_s_name(s_name, f['root'])
 
             if s_name is not None:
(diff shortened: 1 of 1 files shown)


jasonwalker80:
@berguner I've also encountered this with a newer version of Picard (not just the GATK version). You may also want to take a look at #1114 which is a similar solution.

ewels:
Ah sorry @berguner - you were beaten to it in #1114 and I read that first, so it's just been merged. Hopefully that does the job for you as well. Let me know if not.

Thanks anyway!

Phil

Issue #1124, opened 2020-03-10
Hide/Show samples buttons


Maarten-vd-Sande:
**Is your feature request related to a problem? Please describe.**
There is already an option to add buttons to rename samples:
![image](https://user-images.githubusercontent.com/15323399/76333518-0bee5d80-62f2-11ea-8bf7-5c6823c40376.png)
 And there is already a show/hide samples option in the toolbox:
![image](https://user-images.githubusercontent.com/15323399/76333625-2f190d00-62f2-11ea-8b5c-cd1312f0a4f9.png)

It would be nice if there were buttons to hide or show certain samples.

**Describe the solution you'd like**
Clickable buttons would make the usage for non-programmers much much easier.

PR #1125, opened 2020-03-11
add show/hide buttons

Maarten-vd-Sande:
I mentioned this yesterday on Gitter and #1124.

ezgif com-video-to-gif(1)

I never worked with jinja before, and I last made a website 5 years ago, so please let me know if anything is not up to standards (no hurt feelings :smile: )

I basically just copied the same logic from the rename samples buttons but then adapted it for filtering.


Maarten-vd-Sande:
So I turned off fail-fast in the tests, but somehow the linux tests are failing?!

ewels:
> So I turned off fail-fast in the tests, but somehow the linux tests are failing?!

Apologies, updates to dependencies broke the Python 2 install and all CI tests started failing. I've removed official Py2k support and CI now, and things are back up and running 👍

multiqc/templates/default/assets/js/multiqc_toolbox.js
```diff
@@ -37,6 +37,32 @@ $(function () {
...
+    // Get requested pattern and whether to show or hide the pattern
+    var j = $(this).data('index');
+    if(j == 0){
+      apply_mqc_hidesamples();
+    } else {
+      var pattern = mqc_config['show_hide_patterns'][j];
+      var show_or_hide = mqc_config['show_or_hide'][j];
+
+      // Apply the changes
+      $('#mqc_hidesamples_filters').append('<li><input class="f_text" value="' + pattern + '" tabindex=" /><button type="button" class="close" aria-label="Close"><span aria-hidden="true">&times;</span></button></li>');
```
ewels, line 61: Also need to update the radio button for show or hide status.

multiqc/multiqc.py
```diff
@@ -113,6 +113,10 @@
                     type = click.Path(exists=True, readable=True),
                     help = "File containing alternative sample names"
 )
+@click.option('--show-patterns', 'show_patterns',
```
ewels, line 116: Not totally sure about this name. Maybe `--sample-filters`?

multiqc/utils/config.py
```diff
@@ -206,6 +206,27 @@ def load_sample_names(snames_file):
         logger.error("Error loading sample names file: {}".format(e))
     logger.debug("Found {} sample renaming patterns".format(len(sample_names_rename_buttons)))
 
+def load_show_hide(sh_file):
+    global show_hide_buttons, show_hide_patterns, show_or_hide
+    try:
+        with open(sh_file) as f:
+            logger.debug("Loading sample renaming config settings from: {}".format(sh_file))
+            show_hide_buttons.append("all")
```
ewels, line 214:
```suggestion
            show_hide_buttons.append("Show all")
```

multiqc/utils/config.py
```diff
@@ -206,6 +206,27 @@ def load_sample_names(snames_file):
...
+            show_hide_patterns.append("")
+            show_or_hide.append("show")
+            for l in f:
+                s = l.strip().split("\t")
+                if len(s) == 3:
+                    show_hide_buttons.append(s[0])
+                    show_hide_patterns.append(s[1])
+                    if not s[2] in ["show", "hide"]:
+                        raise IOError
```
ewels, line 223: Maybe an `assert` statement would be better here..

multiqc/utils/config_defaults.yaml
```diff
@@ -96,6 +96,9 @@ sample_names_ignore: []
 sample_names_ignore_re: []
 sample_names_rename_buttons: []
 sample_names_rename: []
+show_hide_buttons: []
+show_hide_patterns: []
+show_or_hide: []
```
ewels, line 101: Nice, so this already works with config files! Just needs documenting.

multiqc/utils/config.py
```diff
@@ -206,6 +206,27 @@ def load_sample_names(snames_file):
...
+                    show_or_hide.append(s[2])
+
+    except (IOError, AttributeError) as e:
+        logger.error("Error loading show patterns file: {}".format(e))
+    logger.debug("Found {} show/hide patterns".format(len(show_hide_buttons) - 1))
```
ewels, line 228: This probably deserves a `logger.info`. We should also check that there is more than 1 in case of a failed parse - if only one, remove it again.

multiqc/multiqc.py
```diff
@@ -113,6 +113,10 @@
...
+                    type = click.Path(exists=True, readable=True),
+                    help = "File containing show/hide patterns"
```
ewels, line 118:
```suggestion
                    help = "File containing show/hide patterns for the report"
```

docs/customisation.md
```diff
@@ -98,6 +98,25 @@ sample_names_rename:
...
+of rows containing the button name, the pattern, and whether the pattern should be hidden 
+or shown. For example your want to be able to filter on read pair groups, you can add a file 
+like this:
+
+```
+read_group_1  _R1 show
+read_group_2  _R2 show
+```
+
+Note that MultiQC automatically adds an `all` button which reverts back to showing all samples.
```
ewels, line 118:
```suggestion
Note that MultiQC automatically adds a `Show all` button which reverts back to showing all samples. 
```

ewels: It would be good to document that this can also be done using just a MultiQC config file too.

```yaml
show_hide_buttons:
  - read_group_1
  - read_group_2
show_hide_patterns:
  - _R1
  - _R2
show_or_hide:
  - show
  - show
```

docs/customisation.md
```diff
@@ -98,6 +98,25 @@ sample_names_rename:
...
+
+It's possible to supply a file with one or more patterns to filter samples on using the
+`--show-patterns` command line option. This file should be a tab-delimited file with any number 
+of rows containing the button name, the pattern, and whether the pattern should be hidden 
+or shown. For example your want to be able to filter on read pair groups, you can add a file 
+like this:
+
+```
+read_group_1  _R1 show
+read_group_2  _R2 show
```
ewels, line 115: My only feature-addition suggestion is that I can imagine people requesting that a button applies _multiple_ filter patterns. This could be accommodated fairly easily if we move the `show`/`hide` to the second column and then allow any number of columns after that for multiple patterns.

```suggestion
read_group_1  show _R1
read_group_2  show _R2 _read2
```

multiqc/utils/config.py
ewels: I think we should do this somewhere outside of the file parsing code. That way, if people set their patterns in a MultiQC config file instead of supplying a file the `Show all` button is still added..

ewels:
CI error was unexpected and from a recent module addition, sorry about that. Hopefully fixed it in the above commit.

I might go on and make my suggested changes if that's ok, so that I can push on with merging the PR.

Maarten-vd-Sande:
> I might go on and make my suggested changes if that's ok, so that I can push on with merging the PR.

Of course that's okay! I fully agree with all your comments. Let me know if you need some help :smile:.

multiqc/utils/config.py
ewels: Instead of this I just added to the if statement, so that it skips any lines that don't meet our expectations instead of stopping the parsing of the whole file

multiqc/utils/config.py
ewels: I changed my mind with this and left it as a debug. No need to check the parse with the change in how the prepended `Show all` button no works.

ewels:
Great stuff - please check my changes and let me know if there's anything I missed / you think can be improved 👍 

Thanks again for this, I think a lot of people will like this new option! 🎉 

Phil

Maarten-vd-Sande:
I think you missed the renaming of `show-patterns` -> `sample-filters` in the docs: https://github.com/ewels/MultiQC/blame/master/docs/customisation.md#L109

I like the option to have multiple patterns. Thanks a lot for this awesome tool!

ewels:
Ah good catch, thanks! 👍

bobamess:
Would be nice to include regex in patterns used with sample-filters. Reason being "_R1", "_R2" could be in the middle of a sample name.  This is even worse if you use "_1", "_2" to signify reads, which is common. So could "_R1$", "_R2$" be used to select only matches to the end of sample names?

Maarten-vd-Sande:
@bobamess you are absolutely right! Soon after implementing this I had the same problem! There is an open PR for this (and coincidentally) I accidentally deleted the fork today:

https://github.com/ewels/MultiQC/pull/1222

bobamess:
That’s good to know.  I will look forward to a successful PR.

Issue #447, opened 2017-04-12
Merge Atropos module into MultiQC


jdidion:
I have a working MultiQC module for [Atropos](http://github.com/jdidion/atropos) here: https://github.com/jdidion/atropos-multiqc. Would you be willing to add it to MultiQC? If so, can you pull it directly from my repo, or should I submit it as a pull request?

ewels:
Hi @jdidion,

This looks great! Really nice that you decided to support MultiQC in this way. You're a little ahead of the curve by writing your own module like this, sort of fitting into a model of MultiQC development that I thought about almost exactly [one year ago](https://github.com/ewels/MultiQC/issues/181).

Anyway, we have two options going forward:

1. Merge your code into the main MultiQC repo
    * _Pro:_ Everyone gets out-of-the-box support for Atropos when they install MultiQC
    * _Con:_ You have to mess around copying and pasting and lose the existing repo + history
    * _Con:_ You will have to put up with me nitpicking over your code in a PR 😉 
2. Keep your code separate
    * You're welcome to fork it into the [MultiQC organisation](https://github.com/MultiQC) to help keep things centralised. This is the way that I envisage things moving in the future, though not much has happened yet.
    * _Pro_: Basically no additional work
    * _Con_: As it stands, people have to run a second installation to use your module

As described in that issue linked above, I quite like the idea of having all modules as stand-alone MultiQC "plugins" in this way as it keeps the main codebase a lot smaller and less co-dependent on releases. However, it also introduces quite a few problems (requirement of an internet connection etc) and work, which is why it hasn't happened yet. It's likely to be quite a long time until this happens, so my recommendation at this point would be for option 1.

Sorry for the somewhat long response - hope that makes sense. I'd welcome and thoughts or suggestions!

Phil

jdidion:
@ewels - Option 1 does sound like the best approach at this point, although I do like the idea of an ecosystem where tool developers can create plugins within the MultiQC organization but not as a part of the core software, and the software is able to install any requested/missing modules on the fly.

Having someone nitpick my code sounds like a good idea. I mostly borrowed the code from the Cutadapt and FastQC modules, so it will probably end up being you nitpicking your own code :) My plan was always to get this working as a stand-alone module and then merge into MultiQC, so I’m fine with losing the repo history.

ewels:
Ok, sounds good. In that case then yes - please go ahead and for this repo, add your code, and submit a PR.

ewels:
[...]

PR #1132, opened 2020-03-24
fix missing module_key in report.search_file within BaseMultiqcModule

antgonza:
Not sure if this needs an entry in the CHANGELOG.md, please let me know either way.

Turns out that the error I reported in #447 is because `report.search_file` is missing a parameter: `module_key`; which I currently defined as `None` but happy to use whatever is best. 

Note that with this change, installing the atropos plugin as an external module works as expected and raises the expected warnings:
```bash
$ multiqc -m atropos_trim atropos_qc_json/atropos_logs/
[INFO   ]         multiqc : This is MultiQC v1.9.dev0 (661f82d)
[INFO   ]         multiqc : Searching   : /Users/antoniog/Downloads/multiqc/atropos_qc_json/atropos_logs
[INFO   ]         multiqc : Only using modules atropos_trim
[WARNING]     base_module : Depreciation Warning: Atropos - Please use new style for find_log_files()
```

Please let me know, how to proceed. BTW this also include another small pip install typo.

Thanks. 

cc: @ElDeveloper
 
----------

 - [ ] `README.md` is updated


ewels:
Many thanks for this @antgonza - good find! I couldn't work out why this wasn't causing more problems, but I realise now that it's the old method for searching files that I depreciated 3 years ago:

https://github.com/ewels/MultiQC/blob/04876cf7f71d2ae83843537c925edfebae7a8a31/multiqc/modules/base_module.py#L77

I guess none of the modules use this any more and I must have updated the function without realising that it would break this code. I should probably remove it as promised really to avoid this in the future.

Anyway, I'll merge now and probably say the same thing in another 3 years 😅 - thanks for the fix!

Phil

PR #1137, opened 2020-03-29
Fix #1136: heatmap and toolbox hide/highlight need x and y to be swapped

### LINKED ISSUE #1136: heatmap plotting function is incompatible with toolbox's hide/highlight
**Description of bug:**
built-in `heatmap` plotting function is incompatible with the MultiQC toolbox, specifically `highlight` and `hide` functionality - at times it wouldn't hide anything or wouldn't hide what it is supposed to hide, and highlight seem to "break" heatmap altogether sometimes - please check out screencast for clarification:
heatmap_bug

in this gif I try to hide `SRR` pattern (nothing gets hidden apart from broken labels), then undo it, then try to hide other pattern - something "bad" happens)) i undo it; then i try to highlight a couple of groups, first nothing happens, then it "breaks", then I try to sort - it just goes crazy at that point .

**MultiQC Error log:**
does not apply - it is just an interactive javascript issue

**File that triggers the error:**
This bug is relevant for any report using the built-in [`heatmap`](https://multiqc.info/docs/#heatmaps) function - I used "FastQC Status Check" for demonstration: `MultiQC_TestData/data/modules/fastqc`

**MultiQC run details (please complete the following):**
 - Command used to run MultiQC: `multiqc --module fastqc --outdir ./ ./MultiQC_TestData/data/modules/fastqc`
 - MultiQC Version: v1.9.dev0 (2e16d7e) [latest from `master`]
 - Operating System: Ubuntu 18.04.1
 - Python Version: Python 3.7.2
 - Method of MultiQC installation: pip

**Additional context:**
I think i was able to figure it out, so PR is coming ...

sergpolly:
issue descibed in #1136 seem to originate from the x/y swap - i.e. `xcats` and `ycats` of the heatmap (rows and columns respectively) are referring to wrong indexes in the sparse-matrix variable `data` that stores all of the heatmap's data: `x` refers to `data[n][1]` and `y` refer to `data[n][0]`, but it should be `x` <-->`data[n][0]` and `y` <--> `data[n][1]` instead.

`data` - seem to be a sprase matrix in COO format and it seem to follow the `[row, col, value]` layout
Screenshot from 2020-03-28 22-56-50

In this PR i'll carefully flip/swap `x/y` `0/1` when needed , and add some comments to hopefully ease community-maintenance.

Just one commit for now, fixing the `hide` aspect of the toolbox
[to be continued]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@
 * **bcl2fastq**
     * Samples with multiple library preps (i.e barcodes) will now be handled correctly ([#1094](https://github.com/ewels/MultiQC/issues/1094))
 
+#### Bug Fixes:
+* `heatmap` plotting function is now compatible with MultiQC toolbox `hide` and `highlight`
+
+
 ## [MultiQC v1.8](https://github.com/ewels/MultiQC/releases/tag/v1.8) - 2019-11-20
 
 #### New Modules:
--- a/multiqc/templates/default/assets/js/multiqc_plotting.js
+++ b/multiqc/templates/default/assets/js/multiqc_plotting.js
@@ -1171,6 +1171,8 @@ function plot_heatmap(target, ds){
   var data = JSON.parse(JSON.stringify(mqc_plots[target]['data']));
   var xcats = JSON.parse(JSON.stringify(mqc_plots[target]['xcats']));
   var ycats = JSON.parse(JSON.stringify(mqc_plots[target]['ycats']));
+  // "xcats" and "ycats" are labels of columns and rows respectively
+  // data[n] has form of [x,y,value], x/y are indices of columns/rows
 
   // Rename samples
   if(window.mqc_rename_f_texts.length > 0){
@@ -1228,6 +1230,7 @@ function plot_heatmap(target, ds){
           if(xcat_hl[i] == hl){
             new_xcats.push(xcats[i])
             for (j=0; j < data.length; j++) {
+              // data[j] element is [x,y,val], get "x"
               if(data[j][0] == i){ newdata[j][0] = xidx; }
             }
             xidx += 1;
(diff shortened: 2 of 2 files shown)


ewels:
This is fab, thanks @sergpolly - good detective work! I've labelled this PR as a work in progress for now. Let me know when the `console.log` statements are gone and it's good to merge.. 👍

sergpolly:
hey @ewels - I think this one is done:
 - I completed partial x/y <-> 0/1 swap fix (partial because it was correct in some places - e.g. "sort by highlight")
 - I removed debug-related code, minimized `diffs` but decided to keep some comments there ... 
 - updated `changelog`

here is the demo after fixing:
hide_highligh

all seemed great! ....

PS
working on this I realized another potential issue with applying toolbox to the heatmaps ...
right now `xcats` and `ycats` in the `heatmap` code are treated absolutely symmetrically - i.e. `rename/hide/highlight` are applied to both `xcats` and `ycats` regardless of the context. This should work as intended for a square sample-to-sample similarity/correlation matrix/heatmap. For other cases, this, of course, can lead to some unintended interference between sample names and the other category names (the non-sample one), although arguably it is rare scenario. The issue becomes absolutely *screaming*, when one chooses to `hide` in `Show only matching sa mples`-regime - see another demo:
hide_show_matching
`SRR106` matches a bunch of samples but none of the `xcats` - column labels - which makes `heatmap` yield "blank" - should I open another issue to keep this in the record ?

ewels:
> `SRR106` matches a bunch of samples but none of the `xcats` - column labels - which makes `heatmap` yield "blank" - should I open another issue to keep this in the record ?

Ah pants - yes that's annoying. A new issue would be fab 👍 I'm not really sure how we address that... 🤔  Maybe have different behaviour depending on whether the heatmap is square or not?

ewels:
This is great - tested and working really well in my hands. Thanks for this!

Issue #1118, opened 2020-03-05
Samtools Idxstats - Additional Normalisation Plot


StephenRicher:
Hi,

First off thanks very much for this fantastic tool! I was wondering if it would it be possible to represent the `samtools idxstats` output with an additional plot normalised by chromosome/contig length and by total read count?

As I understand the current normalised count is normalised by sequencing depth for comparison between samples.

However, I think it would be really cool to have an additional normalisation plot that also normalises by contig length in addition to total read count.  This would enable comparison between samples but also enable a quick assessment of any contigs that have higher coverage than expected.

Thanks and I look forward to hearing your thoughts,

Stephen

ewels:
Hi @StephenRicher,

I think this is a great idea! We already have a plot normalising by total read count, so I guess you mean normalising both of those measures at once? Essentially adding a third plot tab here:

![image](https://user-images.githubusercontent.com/465550/77843582-e214b200-719e-11ea-8020-faaf0c205014.png)

I agree that this would be more helpful as it's typically what people are looking for in this plot 👍 

Fancy having a stab at it? 😉 

Phil

StephenRicher:
Hi Phil,

Thanks very much for your email, sure I am happy to have a go! :)

Thanks,
Stephen

ewels:
Closed in #1140

PR #1140, opened 2020-03-29
Idxstats update

StephenRicher:
From my suggestion in [#1118](https://github.com/ewels/MultiQC/issues/1118) - I have added an additional normalisation plot to the `samtools idxstats` module. This computes the log10 transformed observed over expected read counts for each chromosome.

I only had to make some minor modifications to the code - the most significant being that the `parse_single_report()` function now also parses the chromosome size from the `idxstats` report. So now the value for each chromosome key is a list with index `0` being the number of mapped reads and the index `1` being the chromosome sizes.

For the plot itself I first compute the total genome size for each sample on [line 101](https://github.com/ewels/MultiQC/pull/1140/files#diff-030248f96efed9963a11e93f20ea9c7aR101). I then calculate the expected read count given the total genome size and library size for each chromosome. From these two values I calculate the observed over expected value.

I have chosen to `log10` transform my data (such that perfect evenly distributed coverage would result in a straight line at `y = 0`). I chose to do this because my test data (see attached [report](https://github.com/ewels/MultiQC/files/4399099/idxstats_update_multiqc.zip)) has a vastly over-represented number of sequences on the `Mt` chromosome and without a log transformation it is difficult to interpret the other chromosomes.

    - Updated but hope I have done so as is intended as not done this before!
    - I don't think any additional test data is required, beyond what is already present.


ewels:
Many thanks for this addition @StephenRicher - looks great!

The only thing I wanted to change was the hardcoded Log10 transformations. There's nothing wrong with it _per-se_, but I usually prefer to let the plotting library handle this as it gives nicer axes and leaves the values in tact.

The only problem with this was that when I tried to change the behaviour, I found that it was essentially impossible to do with line graphs! Looks like I wrote the functionality for bar graphs but that was all. So I ported this over to line graphs too and then removed the transformation from the python code.

Please have a play and let me know if you find any problems or have any feedback..

Cheers,

Phil

Issue #1142, opened 2020-03-31
MarkDuplicates/biobambam2 metrics interpretation issue


gt1:
**Description of bug:**
The metrics tables produced by Picard's MarkDuplicates and biobambam2's bamsormadup contain some columns referring to read pairs and some which refer to single reads. MultiQC currently does not distinguish between these and puts them together in a single bar plot.

**MultiQC Error log:**
There is no error in the output file. The result plot is wrong.

**File that triggers the error:**
[metrics.txt](https://github.com/ewels/MultiQC/files/4407747/metrics.txt)

**MultiQC run details (please complete the following):**
 - Command used to run MultiQC: multiqc
 - MultiQC Version: 1.9
 - Operating System: Ubuntu 19.10
 - Python Version: Python 3.7.5
 - Method of MultiQC installation: manual via git clone

 **Additional context**

* Referring to single reads:
    * `UNPAIRED_READS`
    * `UNMAPPED_READS`
    * `UNPAIRED_READ_DUPLICATES`
* Referring to read pairs:
    * `READ_PAIRS_EXAMINED`
    * `READ_PAIR_DUPLICATES`
    * `READ_PAIR_OPTICAL_DUPLICATES`

ewels:
Ah, the code comment that I wrote 4 years ago finally comes around 😅 

https://github.com/ewels/MultiQC/blob/9cf37d7d2302dcd349c97bea27599c757c1e8f54/multiqc/modules/picard/MarkDuplicates.py#L94-L101

So I guess the fix is to divide `UNPAIRED_READS`, `UNMAPPED_READS` and `UNPAIRED_READ_DUPLICATES` by two?

Is there a reliable way to know whether the log file is from a paired-end or single-end dataset?

Phil

gt1:
I would suggest to multiply the PAIR fields by two as the graph reports actual numbers as well as percentages and half a read sounds weird.

I don't think there is an absolutely reliable way to determine from the metrics file wether the data is paired-end or single-end. But you could always take a chance and assume that it is highly unlikely for a dataset to be paired end but the READ_PAIRS_EXAMINED field still being zero (i.e. sequencing was paired end but not a single read pair has both ends aligned).

Best,
German

ewels:
> Is there a reliable way to know whether the log file is from a paired-end or single-end dataset?

[...]

PR #1143, opened 2020-03-31
Fix MarkDuplicates/biobambam2 bar plots

gt1:
This pull request is an attempt at fixing issue #1142. Test data for this change can be found within the issue. Values for read pair related metrics columns are multiplied by two so become comparable with the values for single/unpaired reads.


ewels:
This is great - thanks for finally sorting it out @gt1!

I've done a few minor tweaks to your code but nothing serious. Mostly just some description and help text in the report describing how the numbers are calculated, plus an entry in the changelog.

PR #1109, opened 2020-02-17
Hicexplorer3

LeilyR:
Updated the format of the report to fits the changes which have been applied to the QC report of hicexplorer v3.3
Many thanks to contributing to MultiQC!

## If this PR is _not_ a new module
 - [x] This comment contains a description of changes (with reason)
 - [x] `CHANGELOG.md` has been updated
 - [x] (optional but recommended): https://github.com/ewels/MultiQC_TestData contains test data for this change

PR #1144, opened 2020-03-31
Hicexplorer3 - continuation of #1109

ewels:
Continue refactoring from #1109 that ended up a little larger than originally intended:

* Refactored parsing to flatted results dict (avoid using `[0]`)
* Added saved results files to `multiqc_data/multiqc_hicexplorer.{csv,txt,json,yaml}`
* Refactored how sample names are parsed and used
  * Only one dict key used now, and is correctly passed through `clean_s_name` so flags such as `-d` and `-s` no longer crash the module
* Standardised some variable names (eg. using `s_name` more like other modules)
* Reverted changes to `setup.py`
* Other minor cleanup stuff


ewels:
You can see specifically what I changed by looking at the diff of the final two commits in the Files tab. Direct link [here](https://github.com/ewels/MultiQC/pull/1144/files/80ba4d48d0188e370e512fffd92bd60bdbdb4537..7ac9c5edc0164b2b5806a6b8b1d819383b5f8dd0).

When I generate reports using `MultiQC_TestData/data/modules/hicexplorer/` before and after my changes, they appear identical with one exception:  `v1.7_QC_small_test_rf` gets `69.3%`/ `83.4%` for `% Used pairs` / `% Mapped` compared to `36.6%` / `91.2%`. Looking at the raw report I think that this latter value after the changes is probably the correct one though. I'm not entirely which change causes this switch (but I haven't really looked after I figured that it is now correct).

@LeilyR / @joachimwolff - I will merge this PR now and get on with fixing the FastQ Screen linting errors, but please have a look over the changes and comment / review if you spot anything wrong or have any suggestions. I'd be happy to make further changes if required.

Thanks!

Phil

PR #1152, opened 2020-04-03
Hicup bug fix

StephenRicher:
Previously, the sample name was cleaned with `.lstrip('HiCUP_output/')` which removes any of the individual characters from the left side of the same name. I discovered this bug when my sample names (which began with `H`) had the `H` removed. I think instead, `re.sub('^HiCUP_output/', '', s_name)` should be used to ensure the prefix is correctly removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,15 @@
     * Dropped Travis and AppVeyor, everything is now just on GitHub
     * Still testing on both Linux and Windows, with multiple versions of Python
     * CI tests should now run automatically for anyone who forks the MultiQC repository
-    
+
 #### Module updates:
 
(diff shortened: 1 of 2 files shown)


ewels:
Many thanks for spotting and fixing this @StephenRicher - the `.lstrip()` thing is a classic Python mistake! Apologies if it was me who introduced that.

I slightly adjusted your fix to just match and remove the string without having to use regexes. But it's basically the same.

Cheers,

Phil

ewels:
(see https://github.com/ewels/MultiQC/commit/f29a78e1198ab71894b865846d290f456696c99f#diff-bfc7071b6ba7c3532e3f4aa9ec31989e )

PR #1154, opened 2020-04-07
New mirtop module

FlorianThibord:
## This PR is for a new module
 - [x] `README.md` is updated


ewels:
Many thanks for this @FlorianThibord! I've tested the module and the code works really nicely, plots look great 👍 

I've just pushed a few changes to the code in the above commit, namely:

* Remove logging statements
* Clean sample names properly
* Handle read count formatting in tables
* Reorder and rename a few things
* A lot of whitespace munging to make code more readable

I hope that's ok! I might also refactor the code a little more if that's ok. Making the 3 sections in a loop looks efficient, but personally I find it much harder to read and maintain so I think I'd prefer to just have it as three separate sections (even if this does mean a little bit of code duplication). So more updates coming soon, then merge..

Phil

multiqc/modules/mirtop/mirtop.py
```diff
@@ -0,0 +1,242 @@
...
+            'title': '{} Reads'.format(config.read_count_prefix),
+            'description': 'Total read counts - both isomiRs and reference miRNA ({})'.format(config.read_count_desc),
+            'modify': lambda x: x * config.read_count_multiplier,
+            'shared_key': 'read_count',
+            'scale': 'BuGn'
+        }
+
+        self.general_stats_addcols(self.mirtop_data, headers)
+
+    def mirtop_barplot_section(self, aggregate_snps):
```
ewels, line 102: Is there a reason to have `aggregate_snps` as a function argument? It seems that it is always `True`?

ewels: I have removed this in my changes and just always apply this function now. Let me know if this is a mistake.

ewels:
Ok, refactored - please have a look over the code now @florianduclot and check that I haven't made any mistakes / that it looks ok to you. The report output seems to be identical to me, so I hopefully haven't messed it up too much.

I know it's a bit more verbose now and this is mainly a personal style thing, but I figure it will help with maintenance in the future.

Cheers,

Phil

multiqc/modules/mirtop/mirtop.py
FlorianThibord: Yes it's fine! I used it to test the output with/without the SNP aggregation, but it's easier to read with the aggregation.

FlorianThibord:
Thanks a lot for the refactoring @ewels !! I will check the changes asap. 
I will also keep in mind to prioritize readability over efficiency for future contributions...

ewels:
Yes sorry, I felt a bit guilty about the refactoring as it was very much a question of personal style. But I guess it’s not a bad thing to be a consistent as possible with how other MultiQC modules are written (but I’m sure that there are plenty of exceptions to that rule)

Phil

PR #1155, opened 2020-04-08
New module to include Somalier results

fgvieira:
Me, @brentp and @tstannius have made a MultiQC module for [somalier](https://github.com/brentp/somalier)

No test data at the moment, but hopefully @brentp and @tstannius can add some.

 - [x] `README.md` is updated


ewels:
Hi @fgvieira / @brentp / @tstannius,

Many thanks for this PR - it will be great to add Somalier into MultiQC! Could someone please provide some test data though? Without it I can't test the code or review your pull-request.

If possible, please make a PR to https://github.com/ewels/MultiQC_TestData, or just drop a zip file here and I can do it myself 👍 

Thanks!

Phil

brentp:
This is on me. Sorry to all for the delay. I'm working on it. Just checking on permissions on the data I'd like to use.

ewels:
No worries! Soon-ish is kind of good if possible as I'm pushing through the PR backlog now so that I can do a release, so if you miss that then it could be a little while before the module makes it into the next release after that. I probably have at least a week or more of work to do before I can get there though, so it's not super critical.

Note that the CI tests are currently failing, so they need fixing:

```python
Module somalier raised an exception: Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.8.2/x64/lib/python3.8/site-packages/multiqc/multiqc.py", line 556, in run
    output = mod()
  File "/opt/hostedtoolcache/Python/3.8.2/x64/lib/python3.8/site-packages/multiqc/modules/somalier/somalier.py", line 66, in __init__
    for f in self.find_log_files('somalier/ancestry', filehandles=True):
  File "/opt/hostedtoolcache/Python/3.8.2/x64/lib/python3.8/site-packages/multiqc/modules/base_module.py", line 93, in find_log_files
    for f in report.files[sp_key]:
KeyError: 'somalier/ancestry'
```

It looks like the search pattern config has `somalier/somalier-ancestry` but the code looks for `somalier/ancestry`..

multiqc/modules/somalier/somalier.py
```diff
@@ -0,0 +1,616 @@
...
+            if parsed_data is not None:
+                for s_name in parsed_data:
+                    s_name = self.clean_s_name(s_name, f['root'])
+                    if s_name in self.somalier_data.keys():
+                        log.debug("Duplicate sample name found! Overwriting: {}".format(s_name))
+                    self.add_data_source(f, s_name)
+                    self.somalier_data[s_name] = parsed_data[s_name]
+
+        # parse somalier ancestry files
+        for f in self.find_log_files('somalier/ancestry', filehandles=True):
```
brentp, line 66: the output of somalier is now `${output_prefix} + "somalier-ancestry.tsv"

brentp:
here is some example data that can be used
[somalier-example-output.tar.gz](https://github.com/ewels/MultiQC/files/4566427/somalier-example-output.tar.gz)

brentp:
as far as I know, I'm not the bottleneck for this. let me know if that's not the case.
@ewels it would be nice if there was a clear citation for specific tools in multiqc so that users benefiting from somalier or peddy output could (be reminded to) cite the tool(s) they are using in addition to multiQC.

multiqc/modules/somalier/somalier.py
ewels: Yup - this is just the key for the search patterns. You can see the full search pattern below:

https://github.com/ewels/MultiQC/pull/1155/files#diff-7f1a8e845586e3f80f703c3a85e4b869R385

```yaml
somalier/somalier-ancestry:
    fn: '*.somalier-ancestry.tsv'
```

(though note the mismatch in the keys)

ewels:
Fab - thanks @brentp! I've added that to the test data, see above referenced commit.

We need to get the tests passing and the code reviewed. It would be great if @fgvieira could resolve these problems. Otherwise when I'm done with the other PR reviews I'll have a look.

> @ewels it would be nice if there was a clear citation for specific tools in multiqc so that users benefiting from somalier or peddy output could (be reminded to) cite the tool(s) they are using in addition to multiQC.

Yes, this is a nice idea! Each tool already links to a web page, but a citation as well could be a nice addition if implemented nicely. I'll add an issue.

Phil

fgvieira:
I think I've fixed the issue but, not sure why, it is taking a long time to finish the checks.
@ewels is there a way to check why?

ewels:
Odd, I've not seen MultiQC tests time out like this before. I've just restarted them all - hopefully it was just a temporary hiccup.

Phil

ewels:
Ok great - now we're back to regular failures instead of timeouts :)

```
LINT: >somalier/somalier.py< Bargraph pconfig was missing key 'title'
LINT: >somalier/somalier.py<  Bargraph title did not match format 'Module: Plot Name' (found '')
```

You can replicate these locally by running MultiQC with the `--lint` option.

fgvieira:
Seems to be stalled again.. :/

@ewels, can you check?

ewels:
Restarted and passing now ✅

ewels:
Pulled in latest code from `master` and updated the search patterns to be a little more specific (`*samples.tsv` would likely match a lot of stuff not from Somalier).

Will continue with rest of the code review soon 👍 

Phil

ewels:
Hi @fgvieira,

I've just gone through the code and it looks fab, you've done a great job!

My only question / suggestion is the number of (mostly hidden) columns in the General Statistics table. I think it's a bit much, and I wonder if some of these might be useful and worth showing. How about moving them into a new somalier-specific table instead, leaving just a couple of the key ones in _General Statistics_?

Phil

fgvieira:
Hi @ewels,

I thought about that, but decided to include all the columns. On one hand because it did not seem to slow down things, but mostly because I did not have a good criteria to choose these columns; important columns for some users might not be important for others.

However, if you think it is too much, maybe we can remove some of them and move them to another table. Maybe @brentp has some ideas?

ewels:
I think I'd prefer to move them to a separate table and then show more of them unless you're very against that. I think that many MultiQC users are not super familiar with hidden columns so most will not use them and just not realise that they are there. Maybe more helpful to have the columns visible, but in a separate table so as not to swamp the rest of General Statistics..

Phil

ewels:
Fixed the upstream problem with the tests ✅

brentp:
I haven't been following closely, so hopefully this is what you mean... I would prioritize the fields (for sample data) that are in the links below the plots (see here for example: https://brentp.github.io/somalier/ex.html)
so: x_het, x_depth_mean, y_depth_mean, n_unknown, p_middling_ab, gt_depth_mean, ab_mean.

ewels:
Hi both,

I've had a play - moved most of the columns into their own table and fiddled around with some best guesses for hiding and showing stuff. You can see the report from the example data here: [multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/4611874/multiqc_report.html.zip)

I'll merge this PR now to keep things moving, but feel free if you would like the columns to be handled differently (either via a comment or better via a PR).

Cheers,

Phil

brentp:
looks good to me. FWIW, I used to have the heatmap in somalier output, but never found it that useful.
it's also useful to differentiate parent-child pairs from sib-sib pairs. somalier reports exactly 0.5 for parent-child pairs and 0.49 for sib-sib pairs so they can be colored differently.

I would drop the Genot Depth Std colum (which has an incorrect mouseover label).

I would also keep only "Sample Name" for the first 6 columns.

Also, is: "Heterozigosity" british spelling?

Sorry for late feedback... These are just small nits, looks great!

ewels:
> it's also useful to differentiate parent-child pairs from sib-sib pairs. somalier reports exactly 0.5 for parent-child pairs and 0.49 for sib-sib pairs so they can be colored differently.

Is this in the _Allele Balance_ value?

> I would drop the Genot Depth Std colum (which has an incorrect mouseover label).

Ah, this looks like a copy + paste error - good spot. Three columns had the same title. I've fixed this - I guess you were actually looking at _Allele balance_, which I've left there.

> I would also keep only "Sample Name" for the first 6 columns.

I don't understand this, sorry?

> Also, is: "Heterozigosity" british spelling?

No 😉  Fixed 👍

ewels:
Updated report: [multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/4612015/multiqc_report.html.zip)

brentp:
> > it's also useful to differentiate parent-child pairs from sib-sib pairs. somalier reports exactly 0.5 for parent-child pairs and 0.49 for sib-sib pairs so they can be colored differently.
> 
> Is this in the _Allele Balance_ value?

This is for the "Relatedness" plot. Currently there are only 3 colors. The example data doesn't have sibs (my bad), but parent-child and sib-sib pairs both have an expected relatedness (calculated from the pedigree file) of 0.5. But I report parent-child as exactly 0.5 and sib-sib as 0.49. See [here](http://home.chpc.utah.edu/~u6000771/somalier/ceph-new.html) for an example. So I'm letting you know that splitting into `less than 0.25, 0.25 - 0.5, greather than 0.5.` is not ideal. Also "Greather". :)

> 
> > I would drop the Genot Depth Std colum (which has an incorrect mouseover label).
> 
> Ah, this looks like a copy + paste error - good spot. Three columns had the same title. I've fixed this - I guess you were actually looking at _Allele balance_, which I've left there.
> 
> > I would also keep only "Sample Name" for the first 6 columns.
> 
> I don't understand this, sorry?
> 

The samples.tsv file is a proper .ped / .fam file so it contains "phenotype", "sex", "paternal_id", "maternal_id", etc, but those arent really useful in the per-sample QC (except maybe sex).

> > Also, is: "Heterozigosity" british spelling?
> 
> No  Fixed

ewels:
Ok, gotcha. Despite the help text, it looks like it's actually `>= 0.5` that's green, not `> 0.5` - see the code:

https://github.com/ewels/MultiQC/blob/82cc3ce2b44240fdb56a5d622e031c2fe53f9d33/multiqc/modules/somalier/somalier.py#L442-L448

It's also `0.125`, not `0.25`. So I think it's sufficient to just update the description text.. I'll also move the numbers into variables so that they can't fall out of sync again.

ewels:
Looking at your example:

image

What are these threshold ranges? Should we do the same colouring in MultiQC?

brentp:
there is no thresholding. I just give a color for each unique value.

ewels:
ok, so the hardcoded names are `unrelated = 0`, `sib-sib = 0.49` and `parent-child = 0.5`. Then you assign index colours to all other unique values?

brentp:
ah, I see what you mean. that's correct, yes.

ewels:
The example data has `-1.0` for a lot of samples - just checking that it should be printed as such as not be `Unrelated` also?

brentp:
yes, -1.0 is not even in same family. 0.0 is same family_id but not related (e.g. husband and wife). But I just put them in same group of "Unrelated".

ewels:
ok, so `unrelated` is `= 0` or `= -1`. That makes sense, I wondered if that would be the case.

ewels:
Ok, and I've hidden the columns that you said aren't usually useful (people can always show them again in the table config). [Uploading multiqc_report.html.zip…]()

Thanks for the help @brentp!

brentp:
looks great! thanks for polishing this.

fgvieira:
it looks very good!!
thanks both for all the final polishing.

fgvieira:
Hi @ewels,

just testing the new MultiQC module, and I might have found a small typo on the Somalier stats table, where one column says `Prop of XXX sites`.

Is this intended? @brentp, do you have a better suggestion for the label on that column? The description says:
`Proportion of sites with allele balance < 0.2 or > 0.8`

best,
thanks,

brentp:
oops. I didn't see that. Yes, the latter label you describe is accurate.

ewels:
Good spot @fgvieira - thanks! Changed short form to `Allele balance < 0.2, > 0.8` and left long-form hover text as it was.

fgvieira:
Hi both,

while playing a bit with this module, I cannot seem to find a way to show/hide columns from the multiqc config file independently. That is, right now some columns are duplicated between the general stats and the somalier specific stats. While initially I thought it would be a good idea, now I am not so sure, since I cannot find a way to control them separately. If I set:
```
table_columns_visible:
   somalier:
       ancestry:         False
       gt_depth_mean:    False
```

then the columns disappear from both tables. Is it possible to control the columns on just one table?
If not, then maybe it is better if the somalier stats are not in the main general stats table.

ewels:
Hah, that is an edge case that I have never considered before! Happy to remove those columns from the main General Stats table :+1:

fgvieira:
thanks, hopefully will be the last fixes! :)

fgvieira:
Hi @ewels and @brentp 

sorry to keep this thread alive, but I just realized that, apart from `relatedness`, there is no easy way to visualize the `pairs` stats in MultiQC.

Not sure how it would be the best way to plot this data, but I was thinking we could:
 - replace the relatedness scatter plot (now just with IBS0 and IBS2) for a more general scatter-plot where the user could choose what to plot from some drop-down boxes.
 - add a drop-down box on the heatmap where you could choose which stats to plot.

what do you think?

ewels:
Hi @fgvieira - I nearly lost this comment sorry, maybe future comments we can start having in new issues 😉 

I don't really  know what you mean by the `pairs` stats sorry, as I am not very familiar with Somalier outputs.

Dropdown boxes are difficult to do though, as that's not part of the standard plotting toolset of MultiQC. Scatter plots can have tabs for different datasets though.

Note that as a general principle I suggest that MultiQC modules should only show the key top stats from tools, not completely reproduce entire reports. I like the idea that anything odd can be spotted, then users can go back to the original tool outputs for more details.

Phil

fgvieira:
Hi @ewels 
`pairs` stats are all those that come from the pairwise comparison of samples (e.g. relatedness, IBS0, IBS2). For `relatedness` we opted to display it as a heatmap, but there are a lot more stats that could also be plotted.

Either way, I'll open an issue so that @brentp can also comment.
best,

PR #1159, opened 2020-04-17
iVAR Module for MultiQC 1.9

apeltzer:
 - [ ] `README.md` is updated


apeltzer:
Linting is fine:

```

--------------------------------
 MultiQC installation complete!
--------------------------------
For help in running MultiQC, please see the documentation available
at http://multiqc.info or run: multiqc --help

[WARNING]    lint_helpers : Can't check docs readme in lint test as file doesn't exist: /Users/alexanderpeltzer/.conda/envs/multiqc-dev/lib/python3.7/site-packages/multiqc-1.9.dev0-py3.7.egg/docs/README.md
[INFO   ]         multiqc : This is MultiQC v1.9.dev0
[INFO   ]         multiqc : --lint specified. Being strict with validation.
[INFO   ]         multiqc : Searching   : /Users/alexanderpeltzer/IDEA/MultiQC_TestData/data/modules/ivar
[INFO   ]         multiqc : Only using modules ivar
[INFO   ]            ivar : Found 2 reports
[WARNING]         multiqc : Deleting    : multiqc_report.html   (-f was specified)
[WARNING]         multiqc : Deleting    : multiqc_data   (-f was specified)
```

apeltzer:
Example Report (with *.txt ending - please do open in browser as HTML):

[multiqc_report_ivar.html.txt](https://github.com/ewels/MultiQC/files/4496843/multiqc_report_ivar.html.txt)

apeltzer:
Also made this already extendible - the current file pattern searches for ivar/trim instead of ivar in general - thus people could add other ivar components quite easily and extend this to be more useful 👍

drpatelh:
Looking great @apeltzer 😎 Should be spelt as `iVar` everywhere. Also, I want to try running it without the `-e` flag too to see if the message changes and the parsing still works. Doing that now. Ill post the output here.

drpatelh:
Also, do the numbers add up to 100% of the input reads in the report? 🤔 Just checked on an example output:

```
Trimmed primers from 93.37% (2524989) of reads.
1.53% (41280) of reads were quality trimmed below the minimum length of 30 bp and were not writen to file.
6.44% (174041) of reads started outside of primer regions. Since the -e flag was given, these reads were written to file.
```

drpatelh:
[SAMPLE1_PE.trim.ivar.log](https://github.com/ewels/MultiQC/files/4496894/SAMPLE1_PE.trim.ivar.log)
[SAMPLE3_SE.trim.ivar.log](https://github.com/ewels/MultiQC/files/4496895/SAMPLE3_SE.trim.ivar.log)

Ok. Uploaded a couple of reports (1 SE and 1 PE) from running `ivar trim` without the `-e flag`.

apeltzer:
> Also, do the numbers add up to 100% of the input reads in the report? 🤔 Just checked on an example output:
> 
> ```
> Trimmed primers from 93.37% (2524989) of reads.
> 1.53% (41280) of reads were quality trimmed below the minimum length of 30 bp and were not writen to file.
> 6.44% (174041) of reads started outside of primer regions. Since the -e flag was given, these reads were written to file.
> ```

I'm not using these - just taking the counts from the table, ignoring % if available (only V2+ does this apparently).

apeltzer:
- Adjustments from @drpatelh s comments (e.g. iVar everywhere)
- parsing of `-e` and wildtype `trim` output works now

Questions still existing:

- 100% adding up issue with the data  - completely unclear how we should deal with this
- V1 output doesn't have the "mapped/unmapped reads" information, only V2 has that - can't plot this that easily then :-( 



[multiqc_report_ivar.html.txt](https://github.com/ewels/MultiQC/files/4496959/multiqc_report_ivar.html.txt)

multiqc/modules/ivar/ivar.py
```diff
@@ -63,7 +63,7 @@ def __init__(self):
     def parse_ivar(self, f):
         parsed_data = dict()
         regexes = {
-            'mapped_reads': r'(?:Found\s)(\d+)(?:\smapped)',
+            'mapped_reads': r'(?:Found\s)(\d+)(?:\smapped)  ',
```
apeltzer, line 66:
```suggestion
            'mapped_reads': r'(?:Found\s)(\d+)(?:\smapped)',
```

apeltzer:
@gkarthik might be interested in seeing this :-)

gkarthik:
This is neat! However, regarding the percentages not summing up to 100%: There is an overlap between the three categories: primer trimmed, quality trimmed and reads without primers so they don't sum up to 100%. 

Another potential addition would be to show the trimmed counts per primer. However, this is only from `V2` so it may not support older versions.

drpatelh:
Thanks for the clarification @gkarthik :) So a stacked barplot of the overall counts may be a misleading representation if the numbers don't add up to 100%. Maybe a table will suffice unless @ewels has any other suggestions. Be nice to report the percentages to save having to calculate them on the fly :exploding_head: Maybe we just add support for `v1.2` onwards?

I guess the problem with the trimmed counts per primer would be summarising this information across numerous samples :thinking:

apeltzer:
Many thanks for the input @gkarthik ! 

I'd say drop the plot then if @ewels also agrees on it...

On the fly calculation is easy, but I can also have extra regexes to get the % into the report too - nothing big to add. 

>I guess the problem with the trimmed counts per primer would be summarising this information across numerous samples 🤔

Yup. I didn't come to a good idea on how to do this, so I just left it out - if there is a nice one, we can have it but I doubt its going to be feasible, also given that there might be hundreds of primers that we'd have in the table then ...?

gkarthik:
An unstacked bar chart would be nice to compare samples 😄 . 

I agree that number of primers would complicate things 🤔. Depending on the type of library prep, counts per primer would be a good estimate of amplicon specific stats and this would be nice to assess if there was amplicon drop out. Right now we have ad hoc python scripts in our pipelines to generate a heatmap but would be very useful to have in multiqc since we use it regularly 😄.

apeltzer:
> I agree that number of primers would complicate things 🤔. Depending on the type of library prep, counts per primer would be a good estimate of amplicon specific stats and this would be nice to assess if there was amplicon drop out. Right now we have ad hoc python scripts in our pipelines to generate a heatmap but would be very useful to have in multiqc since we use it regularly 😄.

There is support for heatmaps in MultiQC - i might have a look at this with the primers then, some reformatting is fine as long as its not computationally too expensive for MultiQC :-)

apeltzer:
Been playing a bit with it and I think I can add something like a heatmap that then shows how many reads were found for primer X, according to: 

https://github.com/ewels/MultiQC_TestData/blob/master/data/modules/ivar/v1.2/sample1.trim.ivar.log

1.) Will make that work only for V2 by additionally checking whether there is V2 output (easy)
2.) Parse the table of information we need and then create kind of a heatmap for it - that should be somehow feasible :-)

apeltzer:
Might need input from @ewels on this heatmap thing. My parser doesn't find the primer info although the regex101 page shows the regex to work just fine and find all of them.

apeltzer:
[multiqc_report_ivar_heatmap.html.txt](https://github.com/ewels/MultiQC/files/4505069/multiqc_report_ivar_heatmap.html.txt)
ivar-primer-count-heatmap

apeltzer:
Pleas checkout aforementioned Heatmap plot + Report with details. The primer information is now parsed from V2 reports too and automatically added to the report for samples where it exists. Using a `log.info` to display to the user in the multiqc log whether there were primers found or not, to keep things transparent 👍 

@drpatelh @gkarthik might be interested :-)

gkarthik:
This is definitely useful! Thank you for putting this together.

 Two quick comments,
1. Can the heatmap be wider? It seems to be occupying half the space now?
2. I notice in the regex that you can parsing based on `LEFT` and `RIGHT`. However, primer names can be arbitrary. Maybe an alternate approach would be to look for the header,

```
Primer Name	Read Count
```

and then parse all the following lines that have `\t` as a delimiter.

apeltzer:
> This is definitely useful! Thank you for putting this together.
> 
> Two quick comments,
> 
>     1. Can the heatmap be wider? It seems to be occupying half the space now?
> 
>     2. I notice in the regex that you can parsing based on `LEFT` and `RIGHT`. However, primer names can be arbitrary. Maybe an alternate approach would be to look for the header,

Regarding 1.) Adjusted accordingly :-) 
ivar-primer-count-heatmap 

And screenshot:

Bildschirmfoto 2020-04-20 um 20 38 44
[multiqc_report_ivar_heatmap.html.txt](https://github.com/ewels/MultiQC/files/4505458/multiqc_report_ivar_heatmap.html.txt)


Made 2.) more generic. Its now finding everything with `<String>\t<number>` which works fine with my 5 test samples I have here :-)

gkarthik:
This is great! Will definitely integrate this into our pipelines.

drpatelh:
This is looking really cool now @apeltzer @gkarthik 😎 So are we only supporting `v1.2` onwards? No point adding samples to the heatmap for the older versions right? Maybe a good idea to keep the same parsing strings in future releases @gkarthik 😉 and possibly to have a parameter to write this info to file as opposed to piping from `stdout`. Could also mean we can give it a "standardised" name that MultiQC can specifically look for as opposed to looking through all files.

How does the counts table look @apeltzer? Have you changed anything?

apeltzer:
> This is looking really cool now @apeltzer @gkarthik 😎 So are we only supporting `v1.2` onwards? No point adding samples to the heatmap for the older versions right? 

V1 data works but misses some information such as details on the primer counts, which are simply not there. I think its fair to keep both information in, if @gkarthik adds more stuff in future releases, some updates might be required but that can be streamlined with a release then :+1:

>Maybe a good idea to keep the same parsing strings in future releases @gkarthik 😉 and possibly to have a parameter to write this info to file as opposed to piping from `stdout`.
 
+1 from me, would be nicer / neater! Even better: Write out a JSON file, that can be parsed even easier :-) I added specific support for this as an extra log format to my pipelines - keep the "old" text report for users, JSON for MultiQC. 

> How does the counts table look @apeltzer? Have you changed anything?

Yes - newer data has more info now, e.g. the "Mapped Reads" that are unfortunately not there in V1 data. 

Bildschirmfoto 2020-04-20 um 21 37 07

apeltzer:
I guess I'll drop the barplot now too - it doesn't yield any useful information.

gkarthik:
+1 on both counts! Logging to file is on my TODO list.

apeltzer:
[multiqc_report_ivar_heatmap.html.txt](https://github.com/ewels/MultiQC/files/4505805/multiqc_report_ivar_heatmap.html.txt)

- Nicer heatmap
- V1 + V2 supported
- Proper columns

Please check the collumn names and propose better ones before @ewels gets to this PR - I'd love to really get this to V1.9 of MultiQC.

drpatelh:
Just had a look @apeltzer. I think the column names below are more appropriate. @gkarthik what do you think?

Mapped reads -> "Total mapped reads"
Total input reads -> "Primer trimmed reads"
Number of reads too short after trimming -> "Fail minimum length reads"
Reads outside primer region -> "Fail primer region reads"

apeltzer:
Will wait for feedback from him too and then update 👍👍👍

gkarthik:
My 2 cents,

Mapped reads -> "Total mapped reads" ✅ 
Total input reads -> "Primer trimmed reads" ✅ 
Number of reads too short after trimming -> "Fail minimum length reads" ✅ 
~~Reads outside primer region -> "Fail primer region reads"~~ ❌ 

Would retain `Reads outside primer region` or `Reads starting outside primer region` since depending on the library prep method, these reads technically didn't fail but rather are an artifact of intended fragmentation.

apeltzer:
All done and adjusted ✅

CHANGELOG.md
```diff
@@ -36,6 +36,8 @@ to break. If you haven't already, **you need to switch to Python 3 now**.
 
 * [**MultiVCFAnalyzer**](https://github.com/alexherbig/multivcfanalyzer)
     * combining multiple VCF files into one coherent report and format for downstream analysis.
+* **iVar**
+    * Added support for [iVar](https://github.com/andersen-lab/ivar)
```
ewels, line 40:
```suggestion
* **[iVar](https://github.com/andersen-lab/ivar)**
    * Added support for iVar: a computational package that contains functions broadly useful for viral amplicon-based sequencing.
```

docs/README.md
```diff
@@ -22,6 +22,7 @@ MultiQC Modules:
     FLASh: modules/flash.md
     Flexbar: modules/flexbar.md
     InterOp: modules/interop.md
+    iVar: modules/ivar.md
```
ewels, line 25: Needs to go into the table as well

docs/modules/ivar.md
```diff
@@ -0,0 +1,10 @@
+---
+Name: iVar
+URL: https://github.com/andersen-lab/ivar
+Description: >
+    Functions for viral amplicon-based sequencing.
+---
+
+iVar is a computational package that contains functions broadly useful for viral amplicon-based sequencing.
+
+This module parses the output from the `ivar trim` command and creates a table view. Both output from V1 and V2 of the tool are supported and parsed accordingly. The executable used can easily be installed from the BioConda channel using `conda install -c bioconda ivar=1.2` for example. The barplot shows a simple classification of reads that are passing criteria, are too short for further usage or outside of primer regions.
```
ewels, line 10:
```suggestion
This module parses the output from the `ivar trim` command and creates a table view. Both output from V1 and V2 of the tool are supported and parsed accordingly. The executable used can easily be installed from the BioConda channel using `conda install -c bioconda ivar=1.2` for example.
```

multiqc/modules/ivar/ivar.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+
+        # Find and load iVar trim results
+        self.ivar_data = dict()
+        self.parsed_primers = dict()
+        for f in self.find_log_files('ivar/trim', filehandles=True):
+            parsed_data = self.parse_ivar(f)
+            if parsed_data is not None and len(parsed_data) > 0:
+                self.ivar_data[f['s_name']] = parsed_data
+                self.add_data_source(f, f['s_name'])
```
ewels, line 37:
```suggestion
                self.add_data_source(f, section='trimming')
```
No need to add `s_name` if it's the default. As we have more than one data source type, we need to supply the `section` argument though.

multiqc/modules/ivar/ivar.py
```diff
@@ -0,0 +1,167 @@
...
+        for f in self.find_log_files('ivar/trim'):
+            parsed_primers = self.parse_ivar_primer_stats(f)
+            if parsed_primers is not None and len(parsed_primers) > 0:
+                self.parsed_primers[f['s_name']] = parsed_primers
+                self.add_data_source(f, f['s_name'])
```
ewels, line 42:
```suggestion
```

Are these the same files? Don't need to add the data source twice then.

ewels: Might be nice to only loop once if possible, but minor thing.

I think the only difference to the parsing code would be to switch `f['f'].splitlines()` for `f['f']` if this function ran in the above loop with the `filehandles=True`.

multiqc/modules/ivar/ivar.py
```diff
@@ -0,0 +1,167 @@
...
+            # Search regexes for stats
+            for k, r in regexes.items():
+                match = re.search(r, l)
+                if match:
+                    parsed_data[k] = int(match.group(1))
+        
+        return parsed_data
+
+    #Parse Primer stats appropriately
+    def parse_ivar_primer_stats(self, f):
```
ewels, line 85: Any reason that this has to be a separate parsing function? This is almost identical to the above, so might be cleaner to just loop through the file once.

multiqc/modules/ivar/ivar.py
```diff
@@ -0,0 +1,167 @@
...
+
+    # Add to general stats table
+
+    def ivar_general_stats_table(self):
+        """ Take the parsed stats from the iVAR report and add it to the
+        basic stats table"""
+
+        headers = OrderedDict()
+        headers['mapped_reads'] = {
+            'title': 'Total mapped reads',
```
ewels, line 106: Please see https://multiqc.info/docs/#number-base-multiplier

Also `shared_key`: `read_counts` https://multiqc.info/docs/#step-3-adding-to-the-general-statistics-table

multiqc/modules/ivar/ivar.py
```diff
@@ -0,0 +1,167 @@
...
+        headers['reads_too_short_after_trimming'] = {
+            'title': 'Fail minimum length reads',
+            'description': 'Number of reads too short (<30bp) after primer trimming',
+            'min': 0,
+            'scale': 'RdYlGn',
+            'format': '{:,.0f}'
+        }
+        headers['reads_outside_primer_region'] = {
+            'title': 'Reads outside primer region',
+            'description': 'Number of reads outside the primer region',
```
ewels, line 128: Any reason this doesn't have `'min': 0`?

multiqc/modules/ivar/ivar.py
```diff
@@ -0,0 +1,167 @@
...
+            'scale': 'RdYlGn-rev',
+            'format': '{:,.0f}'
+        }
+        self.general_stats_addcols(self.ivar_data, headers)
+    
+
+    def status_heatmap(self):
+        """ Heatmap showing information on each primer found for every sample """
+        data = self.parsed_primers
+        #print(data)
```
ewels, line 138:
```suggestion
```

multiqc/modules/ivar/ivar.py
```diff
@@ -0,0 +1,167 @@
...
+            'description': 'Number of reads outside the primer region',
+            'scale': 'RdYlGn-rev',
+            'format': '{:,.0f}'
+        }
+        self.general_stats_addcols(self.ivar_data, headers)
+    
+
+    def status_heatmap(self):
+        """ Heatmap showing information on each primer found for every sample """
+        data = self.parsed_primers
```
ewels, line 137:
```suggestion
```
I don't think you need to reassign this? You don't modify `data` afterwards..

multiqc/modules/ivar/ivar.py
```diff
@@ -0,0 +1,167 @@
...
+        #print(data)
+        #Top level dict contains sample IDs + OrderedDict(primer, counts)
+        
+        final_data = list()
+        final_xcats = list()
+        final_ycats = list()
+
+        for k,v in data.items():
+            sample = data[k]
```
ewels, line 146:
```suggestion
```
I don't think that you need this either

multiqc/modules/ivar/ivar.py
```diff
@@ -0,0 +1,167 @@
...
+            final_ycats.append(k)
+            tmp_prim_val = list()
+            tmp_xcat = list()
+            for prim,val in sample.items():
```
ewels, line 150:
```suggestion
            for prim,val in v.items():
```

multiqc/modules/ivar/ivar.py
```diff
@@ -0,0 +1,167 @@
...
+        """ Heatmap showing information on each primer found for every sample """
+        data = self.parsed_primers
+        #print(data)
+        #Top level dict contains sample IDs + OrderedDict(primer, counts)
+        
+        final_data = list()
+        final_xcats = list()
+        final_ycats = list()
+
+        for k,v in data.items():
```
ewels, line 145:
```suggestion
        for k,v in self.parsed_primers.items():
```

multiqc/modules/ivar/ivar.py
```diff
@@ -0,0 +1,167 @@
...
+    
+
+    def status_heatmap(self):
+        """ Heatmap showing information on each primer found for every sample """
+        data = self.parsed_primers
+        #print(data)
+        #Top level dict contains sample IDs + OrderedDict(primer, counts)
+        
+        final_data = list()
+        final_xcats = list()
```
ewels, line 142:
```suggestion
        final_xcats = set()
```
If you use a `set` instead of a `list` then the values must be unique (like a `factor` in R), so then you can just update it freely.

multiqc/modules/ivar/ivar.py
```diff
@@ -0,0 +1,167 @@
...
+
+        for k,v in data.items():
+            sample = data[k]
+            final_ycats.append(k)
+            tmp_prim_val = list()
+            tmp_xcat = list()
+            for prim,val in sample.items():
+                tmp_xcat.append(prim)
+                tmp_prim_val.append(val)
+            final_xcats = tmp_xcat #just need this once
```
ewels, line 153:
```suggestion
```

multiqc/modules/ivar/ivar.py
```diff
@@ -0,0 +1,167 @@
...
+        
+        final_data = list()
+        final_xcats = list()
+        final_ycats = list()
+
+        for k,v in data.items():
+            sample = data[k]
+            final_ycats.append(k)
+            tmp_prim_val = list()
+            tmp_xcat = list()
```
ewels, line 149:
```suggestion
```

multiqc/modules/ivar/ivar.py
```diff
@@ -0,0 +1,167 @@
...
+            for prim,val in sample.items():
+                tmp_xcat.append(prim)
```
ewels, line 151:
```suggestion
                final_xcats.add(prim)
```

multiqc/modules/ivar/ivar.py
```diff
@@ -0,0 +1,167 @@
...
+        if data is not None:
+            pconfig = {
+                'id': 'ivar-primer-count-heatmap',
+                'decimalPlaces': 0,
+                'square': False, 
+                'title': 'iVar: Number of primers found for each sample'
+            }
+
+            self.add_section (
+                name = 'iVar Primer Counts',
```
ewels, line 165: A description, ID and help text would be nice please 😄

multiqc/utils/search_patterns.yaml
```diff
@@ -321,6 +321,8 @@ interop/summary:
     contents: 'Level,Yield,Projected Yield,Aligned,Error Rate,Intensity C1,%>=Q30'
 interop/index-summary:
     contents: 'Total Reads,PF Reads,% Read Identified (PF),CV,Min,Max'
+ivar/trim:
+    contents: 'Number of references'
```
ewels, line 325:
```suggestion
    contents: 'Number of references'
    num_lines: 8
```

multiqc/modules/ivar/ivar.py
ewels: Ah, quite possibly. Maybe better to leave it as it is if it works 👍

ewels: Yeah, looks like sets are unordered at least in Python 2x. Lists are probably better then, sorry.. 🙏

apeltzer:
Should be fine now - the two parsing loops aren't working for me and I don't know why 🤔 Texts, help & co updated to match hopefully what you expected ;-)

multiqc/modules/ivar/ivar.py
ewels: I'll take a look at it when I come to merge 👍

ewels:
Thanks all! My thoughts on how to make a nice log file are well documented if you're curious 😉 http://tallphil.co.uk/writing-good-log-files/ - I don't think that much has changed since 2016..

Note that it is possible to have bar plots in MultiQC that are not stacked ([`'stacking': None`](https://multiqc.info/docs/#bar-graphs)), but it doesn't scale very well with lots of samples and the plots can be fairly hard to read. So I generally try to avoid them.

Other modules have had similar issues in the past and just write in the description that numbers may exceed total read counts. I think that's acceptable - the plot can still be useful, even if some reads are double-counted. Of course if you have the info available then you can always have a category called _"Multiple reasons"_ or whatever, which is probably the most optimal solution.

Phil

Issue #121, opened 2015-12-09
kraken


ewels:
Module request from SeqAnswers: kraken

kopardev:
Hello. Is there any update on this? I would also like multiqc to have a kraken module.

ewels:
Hi @kopardev,

Not yet, I need some example data before I can make a start on the module. If you have an example run or two that you don't mind me using it would be great if you could send over the data! (Or submit as a PR to [ewels/MultiQC_TestData](https://github.com/ewels/MultiQC_TestData)). No actual data files needed, just the logs as they are typically generated by the tool.

Cheers,

Phil

tomrconnor:
Hi Phil,

Are you still looking for some example Kraken data? We could probably send you some over. 

BW,

Tom.

ewels:
Hi @tomrconnor,

Yes please! Until I have some example data I can't do anything. At the start of MultiQC dev I would install and get test data and run tools to get the example logs, but it took way longer than actually writing the module code. So now I insist that anyone requesting a module provides these example logs before I make a start. This one has been waiting longer than most..!

Phil

vladsaveliev:
I was just about to upload an example of kraken data!

ewels:
@dfornika - thanks again for the test data!
@vladsaveliev - It's never a problem to have too much test data 😉 

Questions now that I've seen some test data:

1. Do the filenames always end in `.report` or is that just your convention @dfornika? (I think it's just your convention?)
2. Do you know what version of kraken these came from? (not critical to know but nice)
3. What does everyone want to see in the report?
    * `% Unclassified` in _General Statistics_?
    * Bar plot of third-level hierarchy counts? Which whitespace indentation level to plot could be configurable by the user.

Notes for myself:

* Website: http://ccb.jhu.edu/software/kraken/
* Output generated by `kraken-report` command
* Output format documentation: http://ccb.jhu.edu/software/kraken/MANUAL.html#sample-reports

dfornika:
1. `.report` is just my convention. I don't know of a standard file extension for kraken reports.
2. Those reports came from the `0.10.5-beta` version.
3. This is something we're still pondering. I think the most useful information for our purposes is the percentage (column 1) for the species-level phylogenetic rank. As you said, the user could select a different phylogenetic rank to report. But it might get tricky to deal with ranks that are reported with `-`.

ewels:
[...]

PR #1161, opened 2020-04-20
Kraken Module

apeltzer:
On making a module for KRAKEN2 based on some work by @dfornika. Will need to adjust some parsing stuff, but I guess its doable (and I've been on multiple MQC plugins recently so could use a new challenge 😆 ) 

 - [x] There is example tool output for tools in the https://github.com/ewels/MultiQC_TestData repository (yes, thanks to people from #121, e.g. @dfornika)
 - [ ] `README.md` is updated


apeltzer:
Thinking about adding this here: 

https://www.highcharts.com/docs/chart-and-series-types/parallel-coordinates-chart

ewels:
Yes, I've thought about / talked about adding a parallel coordinates plot for a while now. It's one of the newer HighCharts plot types so wasn't available when I first wanted it - some modules kind of fake it with a regular line plot.

I think it would be a nice addition! Adding a new core plot type is quite a lot of work though, be warned. It will probably require the HighCharts version to be updated (a bit of a task in itself) first, then all of the javascript has to be written before adding the core python plot handler. It also needs to then replicated in MatPlotLib for static images for huge sample numbers. This is why I haven't added it yet 😉

ewels:
Outputs from Kraken and Kraken2 look essentially the same in the test data repo ([v1](https://github.com/ewels/MultiQC_TestData/blob/master/data/modules/kraken/sample_01.report), [v2](https://github.com/ewels/MultiQC_TestData/blob/master/data/modules/kraken/v2.0.8_beta/sample2.kraken2.report)) and the module seems to work fine on both. So I'm renaming this from Kraken2 to just Kraken.

Issue #1101, opened 2020-02-05
biscuit module KeyError


efratushava:
**Description of bug:**
My command failes, and the error massage instructed me to report an issue...

**MultiQC Error log:**
```
[ERROR  ]         multiqc : Oops! The 'biscuit' MultiQC module broke...
  Please copy the following traceback and report it at https://github.com/ewels/MultiQC/issues
  If possible, please include a log file that triggers the error - the last file found was:
    ./biscuit_QC/File_biscuitQC/File_covdist_q40_botgc_table.txt
[File_covdist_q40_botgc_table.txt](https://github.com/ewels/MultiQC/files/4159241/File_covdist_q40_botgc_table.txt)

============================================================
Module biscuit raised an exception: Traceback (most recent call last):
  File "lib/python2.7/site-packages/multiqc/multiqc.py", line 546, in run
    output = mod()
  File "lib/python2.7/site-packages/multiqc/modules/biscuit/biscuit.py", line 113, in __init__
    getattr(self, 'chart_{}'.format(k))()
  File "lib/python2.7/site-packages/multiqc/modules/biscuit/biscuit.py", line 740, in chart_retention_cpg_readpos
    dict([(k,v['1']) for k, v in self.mdata['retention_cph_readpos'].items()]),
KeyError: '1'
============================================================

```

**File that triggers the error:**
[File_covdist_q40_botgc_table.txt](https://github.com/ewels/MultiQC/files/4159243/File_covdist_q40_botgc_table.txt)

**MultiQC run details (please complete the following):**
 - Command used to run MultiQC: `multiqc -f --config multiqc_config.yaml . -m fastqc -m biscuit`
 - MultiQC Version: multiqc, version 1.8
 - Operating System: Debian- Kernel Linux 4.4.146-aufs-1

 - Python Version: Python 2.7.15
 - Method of MultiQC installation: tried with pip or conda,( conda was through docker/singularity)

 **Additional context**

PR #1167, opened 2020-04-23
Update BISCUIT Module

jamorrison:
- Module was overhauled due to changes in the script that generates the input files that MultiQC searches for when running the BISCUIT module.
  - Fix was included to deal with bug reported in Issue [#1101](https://github.com/ewels/MultiQC/issues/1101).
  - MultiQC_TestData for the BISCUIT module was updated to account for changes in search patterns used by MultiQC when looking for BISCUIT related files.


ewels:
Many thanks for this update @jamorrison! I will get to the review as soon as I can.

After skimming over the code _very_ quickly, I assume that you have removed the functionality that handles logs from the previous version of BISCUIT. Do you think it is possible to maintain both or is that not practical?

Cheers,

Phil

jamorrison:
@ewels I did remove the functionality to handle logs from the previous version of BISCUIT. Practically, trying to maintain both would be quite a headache. Further, we have some breaking changes that are going into the upcoming release of BISCUIT, so it's important for people to update to the newer version. If it works for you, I can add a warning message into the BISCUIT module that states what the minimum version of BISCUIT that people should be using is.

jamorrison:
@ewels I saw how you merged the MultiQC_TestData for BISCUIT. I think I have a way that I can do a similar system with the actual MultiQC module for BISCUIT, where one version works with the old QC format and the other works with the new. Let me know if this solution sounds reasonable to you and I can work on implementing that.

ewels:
Hi @jamorrison,

Yes that would be nice if we could keep support for both old and new log styles. I’m planning to have a look at the code myself later today to think about this..

Phil

ewels:
Hi @jamorrison,

I've had a go through the code and am generally happy with it. I've made some tweaks - mostly in the code formatting (whitespace, multiline strings etc) and plot descriptions and help text. I hope that's ok!

#### Data files
One thing you may disagree with is that I removed the code that wrote data files to `multiqc_data`. There were a couple of reasons for this, the main one being that much of the data is purely plot data, and will be exported by core MultiQC if `--export` is set (or from the browser). I also don't like that a single module makes such a large number of files, and that they are essentially just copies of the input data. Finally I didn't like that you were forcing JSON - I guess this is because the data is multiple levels deep so was being largely skipped in TSV. However, it breaks the expectation of the user (I have default output to be YAML for example). If you would like more data to be outputted, please simplify it so that it works with TSV (typically done by concatenating keys into a string) and only output non-plot data.

#### Backwards compatibility
Given the size and complexity of this MultiQC module, I'm happy to break backwards compatibility. Having read the code I agree that trying to maintain functionality would be very difficult without essentially having duplicate modules (one for each version of the log files). I'd rather not do this, so I think it's fine to just support the newer versions and gracefully degrades for old versions. I've added notes to the Biscuit docs and change log to reflect this. To fit the "graceful degradation" pattern, I've added in checks at a couple of points to make sure that there is data to plot before adding a report section. This means that when running the new module code on old BISCUIT data you don't get warnings and empty report sections.

#### In summary
I will merge this now, as I want to get on with the v1.9 release. Please have a look over my changes though (see the [partial diff](https://github.com/ewels/MultiQC/pull/1167/files/5d0d03e40871960fa5e4e3ff0c552c032cb20f4d..abbd6fbffc1b7fd4ebac4739961442916b7b3559)). Let me know if I broke anything or made any changes that you disagree with. Also feel free to submit new pull-requests with further changes.

Some general points that could be nice in a future PR, that I thought were outside the scope of my review:

* % Duplicate reads seems to be calculated twice - once for _General Statistics_ and then again for _Duplicate Rates_
    * Generally I prefer to calculate new values in the log parsing code rather than the plotting code. Then it is available for all downstream steps.
* I think that the _Duplicate Rates_ and _Cytosine Retention_ tables would be better presented as bar plots instead of tables.
    * Both have a single value (% dups, % BA CpG retention) that the majority of users will be interested in, and it's much easier to scan and interpret a bar plot instead of a table.

These are minor though and can either be ignored or addressed in a later PR.

Thanks for your work on this!

Phil

jamorrison:
Hi @ewels 

I took a look through the changes you made and I'm content with what you did to the code. I did make note of the two general points you had for a future pull request and am more than willing to make those changes.

**Data Files**

I don't have any problems with you removing that code. Originally, I was trying to make sure the output files weren't empty, but, in hindsight, the way I had it set up wasn't the best way.

**Backwards Compatibility and Report Generation**

I ran your code over a number of samples that had QC files generated with both the old and new versions of BISCUITqc and was able to generate reports with no errors or warnings. This was done with both the data found in MultiQC_TestData and some of my own data I've generated in the past.

Cheers,
Jacob

ewels:
Great! Thanks for checking 👍

PR #614, opened 2017-10-23
adding Atropos module

tanaes:
Addresses issue #447. @jdidion has migrated the code from the separate module structure to the built-in format, and updated the log-file discovery data. I've also tested on my system + added test data in a PR to MultiQC_TestData.

PR #1182, opened 2020-05-07
Atropos

jdidion:

 - [ ] `README.md` is updated

I do not see what I am supposed to update in the main README.


ewels:
_x-ref: See https://github.com/ewels/MultiQC/pull/614 for former discussion._

> I do not see what I am supposed to update in the main README.

You're completely correct @jdidion - this is out of date now, there used to be a huge table listing all supported tools but it was getting ridiculous and I removed it a few days ago. Need to update the PR template text 👍

jdidion:
The build seems to be failing for a reason unrelated to my PR - it looks like it takes a really long time to lint the docs, and that's causing the build to exceed the maximum allowed time of 6h.

ewels:
I'm not sure what's going on with GitHub actions - this has happened a few times recently. Manually restarting the jobs has fixed it every time. They normally only take a few minutes to run, so I'm not sure why they are hanging randomly :/

ewels:
Ok, figured out the problem with GHA (it was my fault, sorry) - now fixed ✅

jdidion:
@ewels great! What are the remaining steps to get this PR approved?

ewels:
I just need to find time, sorry. I should get to you this week - was hoping to wrap everything up last week but some other things came up.

ewels:
Hi @jdidion,

I've just had my first look at this new PR. Many thanks again for picking it up and getting it to run.

The main things that need additional work as I see it are:

* Duplicated plots
   * Generally there should only be one plot per section.
   * I think the majority of these duplicated plots can be merged, instead using data tab buttons above the plot to move through the different datasets. See the [docs](https://multiqc.info/docs/#switching-datasets-1).
* One module instead of three
  * It looks like the module currently runs as if it is three - pre trim, trim and post trim.
  * Each module should create only one report module section. Not following this pattern can break a lot of the MultiQC customisation options.
* Code duplication
  * At the risk of doing a u-turn on previous comments, I'd prefer to avoid copying too much code / files as it makes maintenance and bug fixes tricky - it's ok to pull in files from other modules (eg. the theoretical GC files).
  * If you need to edit eg. the FastQC code to move stuff into functions that you can then import and use, that's fine.
* Javascript errors
  * The custom javascript seems to have quite a few problems. Lots of warnings and errors are flagged in the browser console.
  * The sequence heatmap single-sample view doesn't work.
  * The status popover click events don't work (hide / highlighting samples)
* Small stuff
  * Table headings are too long
  * Double `%%` signs in the General Stats table
  * Order of table columns is backwards (should go from start of analysis on the right to end of analysis on the left)
  * Order of modules is backwards (should go from start of analysis at the bottom to end of analysis at the top) - but need to merge into one module anyway.
  * Spiky GC plots in the example data?
  * Quite a lot of commented-out code that needs deleting
* Code consistency with the rest of MultiQC
  * Following the same coding pattern is important, as it's typically me that bugfixes everything in MultiQC. As a result I try to make sure that all modules are written in broadly the same way. The code here is a very different style though which makes it difficult for me to work on.
  * Basically - using a custom `Status` class that is passed around may be elegant, but it's something I have to figure out every time I look at the code. If the values are simply hardcoded next to the plot function call it's obvious and fast to edit.

The final point is a tricky line for me to tread as a tool maintainer. I've learned by hard experience that if I merge in modules that use a radically different code style it can cause a lot of of headaches down the line. Typically this happens when the MultiQC code is inefficient or excessively verbose, so using subclasses and object-orientated code reuse within the module makes sense. However, much of the required maintenance typically involves tweaking a single value here or there for a plot. With a lot of code reuse this can be (a) difficult to track down and (b) difficult to isolate. By keeping the code style the same as the rest of MultiQC you allow anyone to work on it.

Let me know what you think of the above. I appreciate that Atropos has been waiting for MultiQC support for a long time now - if I manage to find some time I'll try to have a stab at working on this myself.

Phil

jdidion:
Thanks for the review, @ewels. I would love to get this module into MultiQC, but there's a lot of revision required here and I just don't have the bandwidth.

@ekopylova @tanaes @nservant @antgonza you had asked about this PR previously - perhaps you could help out?

ewels:
Yup, I understand. The FastQC module that much of this code stemmed from took several years of continuous development to get to where it is today, it's a bit of a beast (and Atropos gives even more data to report).

I'll try to take a look at this when I can.

Phil

jdidion:
Thanks. And I found looking at the FastQC module now (3 years after I wrote the initial version of the MultiQC module) that it has changed substantially, mostly in the JavaScript, and my JS skills are pretty basic.

vladsavelyev:
Would it make sense to write a very simple version of the module first, just to get started? That would include only basic types of plots, or those plots that can be reused from FastQC. 

After we merge a basic, clean and robust version, we can create issues and PRs for more advanced features.

Issue #1077, opened 2019-12-09
MultiQC is not plotting Mean and Median coverage for Qualimap


CuriusScientist:
**Description of the bug:**

MultiQC is not plotting Mean and Median coverage for Qualimap, which is surely present as shown by the Qualimap report for one of the samples

**MultiQC run details (please complete the following):**
 - Command used to run MultiQC:  `multiqc ./`
 - MultiQC Version:  MultiQC v1.8
 - Quyalimap Version:  v.2.2.2-dev
 - Operating System: Ubuntu
 - Python Version: Python 3.7.3
 - Method of MultiQC installation: Conda

 **Additional context**
![M - 1](https://user-images.githubusercontent.com/9574057/70427528-2dcb1080-1a75-11ea-9626-63112fc3c33e.jpeg)
![Q - 1](https://user-images.githubusercontent.com/9574057/70427530-2dcb1080-1a75-11ea-96c2-477c25fcae09.jpeg)

ewels:
Thanks for this @CuriusScientist - I think that there is something wrong with this code. See also https://github.com/ewels/MultiQC/issues/1082

I'm away for Christmas holidays now but @apeltzer (who wrote the new code for the mean coverage calculation) is hoping to have a look. Otherwise I'll look in to this as soon as I get a chance.

Is it possible for you to attach one of these Qualimap reports? The more examples we have where this is failing the better.

Thanks,

Phil

apeltzer:
Hi @CuriusScientist ! I'll communicate the bugfixing in the other thread - if you want /  can share some data, please do 👍

apeltzer:
Please see #1085 for a proposed fix for this.

CuriusScientist:
@ewels I see that you have fixed and merged the branch "apeltzer:fix-qualimap" #1085

In order to get the latest version, I followed the following steps

-  Downloaded MultiQC from github

- Ran successfully "pip install MultiQC-master.zip"
Successfully built multiqc
Installing collected packages: multiqc
Successfully installed multiqc-1.9.dev0

- multiqc ./
[INFO ] multiqc : This is MultiQC v1.9.dev0
[INFO ] multiqc : Template : default
[WARNING] multiqc : You are running MultiQC with Python 2.7.16
[WARNING] multiqc : Please upgrade! MultiQC will soon drop support for Python < 3.6
[...]

PR #1190, opened 2020-05-13
Fix QualiMap Regex for lowcov sampels

apeltzer:

Fix for: https://github.com/ewels/MultiQC/issues/1077#issuecomment-628216381 

New Testdata for that case: https://github.com/ewels/MultiQC_TestData/pull/151

### DIFF
--- a/multiqc/modules/qualimap/QM_BamQC.py
+++ b/multiqc/modules/qualimap/QM_BamQC.py
@@ -87,7 +87,7 @@ def parse_genome_results(self, f):
         'median_insert_size': r"median insert size = ([\d,\.]+)",
         'mean_mapping_quality': r"mean mapping quality = ([\d,\.]+)",
         'general_error_rate': r"general error rate = ([\d,\.]+)",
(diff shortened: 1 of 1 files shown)


multiqc/modules/qualimap/QM_BamQC.py
```diff
@@ -87,7 +87,7 @@ def parse_genome_results(self, f):
         'median_insert_size': r"median insert size = ([\d,\.]+)",
         'mean_mapping_quality': r"mean mapping quality = ([\d,\.]+)",
         'general_error_rate': r"general error rate = ([\d,\.]+)",
-        'mean_coverage': r"mean coverageData = ([\d,]+)"
+        'mean_coverage': r"mean coverageData = ([\d]+\.[\d]+)"
```
ewels, line 90: You've dropped the comma here, which will break it for average coverages > 1000X. I think it's better to just use the same regex as above, with the additional `.` allowed?

```suggestion
        'mean_coverage': r"mean coverageData = ([\d,\.]+)"
```

PR #1196, opened 2020-05-20
Replace deprecated logger.warn with logger.warning (Fixes #1138)

### LINKED ISSUE #1138: Use log.warning instead of log.warn
New global warn log seems to be coming from the `logging` module, see https://github.com/ewels/MultiQC/issues/1126:

```
DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
  logger.warn('MultiQC Version {} now available!'.format(remote_version))
```

Need to so some find+replace to switch out all occurrences of `warn`.

cpavanrun:
Fixes #1138 via:

```bash
grep -irl 'logger.warn(' * | xargs sed -i 's/logger.warn(/logger.warning(/g'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,7 @@ to break. If you haven't already, **you need to switch to Python 3 now**.
 * Plot config `logswitch_active` now works as advertised
 * When running MultiQC modules several times, multiple data files are now created instead of overwriting one another ([#1175](https://github.com/ewels/MultiQC/issues/1175))
 * Fixed minor bug where tables could report negative numbers of columns in their header text
+* Replaced deprecated 'warn()' with 'warning()' of the logging module
 
 ## [MultiQC v1.8](https://github.com/ewels/MultiQC/releases/tag/v1.8) - 2019-11-20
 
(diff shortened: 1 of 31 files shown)


ewels:
Ah nice, thank you! Been meaning to do this for a while.

I think you're missing quite a lot of `log.warn()` statements as well (versus `logger.warn()`)

Most of the modules import the logger as `log`:

```python
# Initialise the logger
```

ewels:
Yeah, 13 > 48 changes with those additional ones added in 👍  Hopefully that's all of them..

ewels:
My approach was a little more low-skilled 😅 

image

ewels:
Any ideas how we can test for this to avoid people (me) accidentally using `log.warn()` again in the future?

cpavanrun:
@ewels, why bother testing it? warn() will become unusable anyway =). Deprecation warnings are in general a pre-empt. Which are nice to pick up on but I wouldn't have tests fail on it. Eventually a package will force you to fix it anyway.

ewels:
Fair enough, good point!

PR #1206, opened 2020-05-28
Add kaiju module

noirot:
 - [ ] `README.md` is updated -> no more need 

First plot for kaiju: barplot at each taxonomic rank found in logs. 
I'm wandering if it would be better to configure which rank I want to report in general table or if I report every ranks and hide most of them... Don't know what is the best (see lines 88-89)
Next step will by the heatmap. Then l'll see, if I add the parsing of kaiju2krona output.


ewels:
This is awesome stuff @noirot - I could find almost nothing to change in your PR 👏 😅 

Made it into v1.9 release purely because the code was so polished 👍 

Thank you!

Phil

noirot:
Thanks !

PR #1207, opened 2020-05-28
Add MALT module

maxibor:
Adding [MALT](https://uni-tuebingen.de/fakultaeten/mathematisch-naturwissenschaftliche-fakultaet/fachbereiche/informatik/lehrstuehle/algorithms-in-bioinformatics/software/malt/) Metagenomic aligner module

 - [x] There is example tool output for tools in the https://github.com/ewels/MultiQC_TestData repository: [#154](https://github.com/ewels/MultiQC_TestData/pull/154)

### DIFF
--- a/multiqc/modules/malt/malt.py
+++ b/multiqc/modules/malt/malt.py
@@ -1,0 +1,174 @@
+from collections import OrderedDict
+from multiqc import config
+from multiqc.plots import bargraph, beeswarm, table
+from multiqc.modules.base_module import BaseMultiqcModule
+
+import logging
+log = logging.getLogger(__name__)
+
+
+class MultiqcModule(BaseMultiqcModule):
+    """Malt Module"""
+
+    def __init__(self):
+
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__(
+            name='MALT',
+            anchor='malt',
+            href='https://uni-tuebingen.de/fakultaeten/mathematisch-naturwissenschaftliche-fakultaet/fachbereiche/informatik/lehrstuehle/algorithms-in-bioinformatics/software/malt/',
+            info='performs alignment of metagenomic reads against a database of reference sequences (such as NR, GenBank or Silva) and produces a MEGAN RMA file as output.'
+        )
+
+        # Find and load Malt reports
+        self.malt_raw_data = dict()
+
+        for f in self.find_log_files('malt', filehandles=True):
+            self.parse_logs(f)
+
+        if len(self.malt_raw_data) == 0:
+            raise UserWarning
+
+        log.info("Found {} reports".format(len(self.malt_raw_data)))
+
+        self.malt_summary_table()
+        self.mappability_barplot()
+        self.taxonomic_assignation_barplot()
+
+    def parse_logs(self, f):
+        """Parses a Malt log file"""
+        data = {}
+        mappability = {}
+        taxo_success = {}
+        reading = False
+        for line in f['f']:
+            line = line.rstrip()
+            if line.startswith("+++++ Aligning file:") and reading == False:
+                reading = True
+                sample = line.split()[-1]
+                sample = self.clean_s_name(sample, f['root'])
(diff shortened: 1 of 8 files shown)


maxibor:
@ewels This should be ready to be shipped with 1.9 !

ewels:
I'll give it to you - you _might_ just make it by the skin of your teeth 😅 

Almost a clean run, but I managed to find a few things to tweak in 997be7b still. I will release tomorrow morning so will merge prior to that - please shout if you see any problems before then.

Phil

ewels:
Arg, sorry - I got a bit carried away having fun in 6742334 and ended up refactoring quite a bit of the code. I'm not sure if it's any better like this, just made more sense to me with this layout (calculate all derivative data in the parsing function instead of plotting functions, made parsing functionality more reusable, removed a little unused cruft). That's what happens when you review code at midnight with a beer in hand! 🍺

multiqc/modules/malt/malt.py
```diff
@@ -0,0 +1,151 @@
...
+
+        Mappability = (Total reads / Num. of queries) * 100
+        """
+        cats = OrderedDict()
+        cats['Total reads'] = {'name': 'Mapped reads'}
+        cats['Non mapped'] = {'name': 'Non Mapped reads'}
+        config = {
+            'id': 'malt-mappability-plot',
+            'title': 'MALT: Metagenomic Mappability',
+            'ylab': 'Samples',
```
ewels, line 88: **Question** - are these y-axis labels really meant to be `Samples` and not `Number of reads`? I didn't feel confident enough to change this one.

ewels:
Ah, I'll merge now I think so it's done. Please review my changes and comments (and the one question) anyway though, then we can fix before release if needed 👍 

Thanks again for this! I confess I was planning to let this go to v1.10 until I skim read the code on my phone and figured it was basically good to go..

multiqc/modules/malt/malt.py
```diff
@@ -0,0 +1,151 @@
...
+            elif reading:
+                for k in keys:
+                    if line.startswith(k):
+                        self.malt_data[s_name][k] = int(line.split()[-1].replace(",", ""))
+                        if k == 'Num. alignments':
+                            try:
+                                self.malt_data[s_name]['Non mapped'] = self.malt_data[s_name]["Num. of queries"] - self.malt_data[s_name]['Total reads']
+                                self.malt_data[s_name]['No Assig. Taxonomy'] = self.malt_data[s_name]["Total reads"] - self.malt_data[s_name]['Assig. Taxonomy']
+                                self.malt_data[s_name]['Mappability'] = (float(self.malt_data[s_name]['Total reads']) / float(self.malt_data[s_name]["Num. of queries"]))*100.0
+                                self.malt_data[s_name]['Taxonomic assignment success'] = (float(self.malt_data[s_name]['Assig. Taxonomy']) / float(self.malt_data[s_name]['Total reads']))*100.0
```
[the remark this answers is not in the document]
maxibor: Nice catch !

multiqc/modules/malt/malt.py
maxibor: I was also a bit confused. Since it's a "rotated barplot", does the y-axis becomes the x-axis ?
It also made more sense to me if xlab is the 'Samples'.

Screenshot 2020-05-30 at 12 58 59

maxibor:
Thanks for the review and merge @ewels !

multiqc/modules/malt/malt.py
maxibor: Oh well, that'll be for 1.20 ;)

ewels: Yes, it's not super clear and you're not the first to be caught out by this. It's because the HighCharts plotting code calls it the y-axis because bar plots are usually the other way up. As it's rotated it should be x as you say, but then the code would have been more confusing for me to write. I made this decision way back in the early days and kind of feel that it's too late to change now!

PR #1209, opened 2020-05-28
Strict search patterns

ewels:
This PR is a work-in-progress to improve run times for MultiQC.

### Concept

MultiQC finds log files for tools by recursively searching through everything you throw at it for either filenames (fast) or specific strings within the file contents (slow). In the past [some tools](https://clusterflow.io/) have frequently concatenated standard-out into a single file, so many of the default search patterns look through the whole file for a match and keep searching the same file for each module even after finding a match.

It's my suspicion that very few people still do concatenate log files in this way, so by changing the defaults to make the search patterns more strict (discard after a match, limit the number of lines that are searched) we can significantly speed up MultiQC.

### Results

Speed up by how much? Running in interactive mode (so no matplotlib slowdown) on MultiQC_TestData took me 2 mins 17s this evening. With the modified code in the PR below it was 1min 50. Not a massive difference, but this gap would grow significantly if searching many (10s, 100s) thousands of files.

For more detail, I added a new time profiler that checks how long each individual search pattern takes so that we can tune performance. In the above run when considering file search time alone, performance was approximately twice as good - **89.6s for old** versus **39.4s for new**.

This plot shows the breakdown per pattern for old (blue) and new (orange):

image

You can clearly see the two classes of search: filename (fast) and contents (slow). You can also see that nearly all searches were faster.

So as not to skew interpretation by sorting on the old times, here's the same data sorted by the new times:

image

The spike where a new pattern was faster is AdapterRemoval. I didn't change this so I can't see a reason why it would be different. I'm sure if I did this test a bunch of times and took an average the results would be more accurate. If this bothers you, please go ahead..

### Question

These changes will probably break MultiQC for some people - that is, runs that previously found files will stop finding those files. It will be possible to rescue this by customising the configuration, but this isn't super easy.

So, the question is: **does the performance increase for the majority of users warrant the inconvenience for the minority of users?**

Feedback welcome! Please comment below.

### Technical TODOs

The code here isn't finished yet. Comparing the MultiQC logs from the old and new runs I can see that differing numbers of reports are found, so some logs in MultiQC_TestData are now being missed. Also, Picard is now picking up something it shouldn't and crashing.

- [ ] Edit new search patterns so that numbers of reported samples are unchanged
- [ ] Fix Picard error

### DIFF
--- a/multiqc/modules/bcl2fastq/bcl2fastq.py
+++ b/multiqc/modules/bcl2fastq/bcl2fastq.py
@@ -145,7 +145,12 @@ class MultiqcModule(BaseMultiqcModule):
         except ValueError:
             log.warning('Could not parse file as json: {}'.format(myfile["fn"]))
             return
-        runId = content["RunId"]
+        try:
+            runId = content["RunId"]
+        except KeyError:
+            # Probably not a bcl2fastq file
+            return
+
         if runId not in self.bcl2fastq_data:
             self.bcl2fastq_data[runId] = dict()
         run_data = self.bcl2fastq_data[runId]
--- a/multiqc/utils/report.py
+++ b/multiqc/utils/report.py
@@ -206,7 +206,7 @@ def get_filelist(run_module_names):
         for key in sorted(search_times, key=search_times.get, reverse=True):
             print('{} - {}'.format(search_times[key], key))
             total_search_time += search_times[key]
-            print("==========\nTOTAL: {}\n===========".format(total_search_time))
+        print("==========\nTOTAL: {}\n===========".format(total_search_time))
 
 
 def search_file (pattern, f, module_key):
--- a/multiqc/utils/report.py
+++ b/multiqc/utils/report.py
@@ -14,6 +14,7 @@ import inspect
 import lzstring
 import mimetypes
 import os
+import time
 import re
 import yaml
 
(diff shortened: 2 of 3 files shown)


ewels:
NB: These changes will probably make relatively little difference if you're using `-m` to filter which modules run and which search patterns are used. Unless you're using Picard / Samtools or one of the other modules which is heavily modified here, anyway. Which is probably most people..

ewels:
I did a little [twitter poll](https://twitter.com/tallphil/status/1266124457741029387) asking people how often they concatenate their log files. My assumption was that hardly anyone does this, but so far there is a surprising number of responses saying that people do concatenate:

image

As such, I'm inclined to leave the searches as they are and instead make it easier for people to optimise MultiQC run time themselves. For this, I've written some [new documentation](https://multiqc.info/docs/#optimising-run-time) and added a new command line option `--profile-runtime` (a119aec82f8d14e211eaa4facaad64ce1f3415de) which gives stats about performance to help identify where a bit of tuning can help.

The latter gives some nice summary statistics at the end of the run:

```
[INFO   ]         multiqc : Run took 35.28 seconds
[INFO   ]         multiqc :  - 31.01s: Searching files
[INFO   ]         multiqc :  - 1.75s: Running modules
[INFO   ]         multiqc :  - 0.96s: Compressing report data
[INFO   ]         multiqc : For more information, see the 'Run Time' section in multiqc_report.html
```

and a special section at the end of the MultiQC report with some graphs:

image (2)

Hopefully this is helpful, let me know if anyone has any feedback.

Phil

PR #1210, opened 2020-05-29
Test --ignore-samples

ewels:
I realised that it should be fairly easy to test whether modules are properly using `self.ignore_samples()` to filter out samples with `--ignore-samples`. So this adds a new GitHub action test to ignore _all_ samples with a universal glob: `--ignore-samples '*'`

The downside is that this of course finds lots of MultiQC modules that aren't doing this properly. So now I need to go through them all to make the new test pass.


ewels:
Manually tested all of the adjusted modules, can't see any changes since before modification.

Issue #1217, opened 2020-06-04
Kaiju module works only at phylum level


jkibilds:
**Description of bug:**
When I run MultiQC on table reports from Kaiju, the Kaiju module breaks if the input files have taxonomic classification at any other level than phylum. 

**MultiQC Error log:**
```
[INFO   ]         multiqc : This is MultiQC v1.9
[INFO   ]         multiqc : Template    : default
[INFO   ]         multiqc : Searching   : /mnt/home/juriski/projects/scratch/MiSeq_run_200529/results/kaiju1/taxa/genus
[INFO   ]           kaiju : Found 1 reports
[ERROR  ]         multiqc : Oops! The 'kaiju' MultiQC module broke...
  Please copy the following traceback and report it at https://github.com/ewels/MultiQC/issues
  If possible, please include a log file that triggers the error - the last file found was:
    genus/Li-2020-14538-3.kaiju.genus.tsv
============================================================
Module kaiju raised an exception: Traceback (most recent call last):
  File "/home/juriski/.conda/envs/multiqc/lib/python3.6/site-packages/multiqc/multiqc.py", line 569, in run
    output = mod()
  File "/home/juriski/.conda/envs/multiqc/lib/python3.6/site-packages/multiqc/modules/kaiju/kaiju.py", line 64, in __init__
    self.kaiju_stats_table()
  File "/home/juriski/.conda/envs/multiqc/lib/python3.6/site-packages/multiqc/modules/kaiju/kaiju.py", line 105, in kaiju_stats_table
    general_data = self.kaiju_data[self.kaiju_data.keys().sorted()[0]]
AttributeError: 'dict_keys' object has no attribute 'sorted'
============================================================
[WARNING]         multiqc : No analysis results found. Cleaning up..
[INFO   ]         multiqc : MultiQC complete
```

**File that triggers the error:**
[Li-2020-14538-3.kaiju.genus.txt](https://github.com/ewels/MultiQC/files/4727866/Li-2020-14538-3.kaiju.genus.txt)
(I had to rename the file from .tsv to .txt because GitHub allows only certain extensions)

**MultiQC run details:**
 - Command used to run MultiQC: `multiqc -o genus/mqcout genus/`
 - MultiQC Version: MultiQC v1.9
 - Operating System: CentOS 7
 - Python Version: Python 3.6.10
 - Method of MultiQC installation: conda

 **Additional context**
[...]

PR #1218, opened 2020-06-04
Correct bug #1217

noirot:
Correct bug  #1217


ewels:
I think that you can still sort the keys, you just need to convert the `.keys()` iterator to a `list()` first, or maybe just sort on the dict directly without the `.keys()` call.

See http://www.java2s.com/example/python-book/sorting-dictionary-keys-in-python-3-x.html

noirot:
yes indeed, I thougth it wasn't necessary to sort but, in fact yes.

noirot:
I'll check it next week at my return from hollidays.

PR #1239, opened 2020-06-29
Sentieon

chrispyatt:
[NA12878.AlignmentStat_metrics.txt](https://github.com/ewels/MultiQC/files/4848069/NA12878.AlignmentStat_metrics.txt)
[NA12878.GCBias_metrics.txt](https://github.com/ewels/MultiQC/files/4848070/NA12878.GCBias_metrics.txt)
[NA12878.InsertSize_metrics.txt](https://github.com/ewels/MultiQC/files/4848071/NA12878.InsertSize_metrics.txt)

 - [x] This comment contains a description of changes (with reason) - New Sentieon module added to catch output from Sentieon DNASEQ tools (the output has a modified header which is not caught by the picard module even with appropriate config). 

 - [x] Code is tested and works locally (including with `--lint` flag) -- throws warning about sentieon being found in installed modules but not README.md, but it definitely is in README.md so I can't work out the source of this error.


ewels (approved):

ewels:
Many thanks for this! I added a couple of minor tweaks and fixed the merge conflict. But generally I thought that the code looked great and seems to work well 👍🏻 

Sorry for the delay in reviewing & merging, thanks for the new module! 🚀

Issue #1233, opened 2020-06-20
Split happy section into two; enable different conditional formatting for SNPs & indels


woook:
**Is your feature request related to a problem? Please describe.**
As SNP recall is usually higher than for indels, it would be useful to have separate conditional formatting rules for these two types of variants.

**Describe the solution you'd like**
Split the happy section into two tables, one for SNPs and another for indels.

**Describe alternatives you've considered**
A helper script to take happy output and split it into two output files.

**Additional context**
Can submit a PR if this seems a sensible approach.

@sophie22

ewels:
Added in #1241 👍🏻

PR #1241, opened 2020-06-30
Split happy data into INDEL and SNP tables

sophie22:

This is an attempt to resolve issue #1233 
Data from the happy file is split into indel_data and snp_data and presented in separate tables in the report to allow for different thresholds to be set during conditional formatting set in the config file. Column headers now have an '_indel' or '_snp' suffix, so that they can be referred to in the config file.
Potential future improvement could include an option set in the config file whether or not to split the happy values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
     * Fix issue where missing out fields could crash the module ([#1223](https://github.com/ewels/MultiQC/issues/1223))
 * **featureCounts**
     * Add support for output from [Rsubread](https://bioconductor.org/packages/release/bioc/html/Rsubread.html) ([#1022](https://github.com/ewels/MultiQC/issues/1022))
+* **hap.py**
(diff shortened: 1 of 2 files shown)


sophie22:
It's not failing the automated tests but it also doesn't show the data in the actual report for the split tables right now

sophie22:
Stuck on trying to figure out what 'headers' mean in the table.plot(data, header, config)
and how to have the data column headers the same (so that data is correctly put into both tables after reading in the file content) but the HTML header tags named differently for indel and snp so that they can be referred to using different col_header names in the config file.

sophie22:
had to change the headers when storing the file data into separate dicts so that later on they are different and generate unique HTML tags (which can be used in the config file for conditional formatting)
might not be the neatest way but it works now and passes the automated tests as well

ewels (approved):
Looks great! Tested locally and working perfectly 🚀 

Thanks for the contribution, and sorry for the slow review / merge.

PR #1248, opened 2020-07-10
Add PURPLE module

vladsaveliev:
Adding a module for [PURPLE](https://github.com/hartwigmedical/hmftools/tree/master/purity-ploidy-estimator) -  a purity, ploidy and copy number caller for whole genome tumor data.

In terms of QC, we mainly use several metrics that the tool reports in two files, and they fit into a table very well. The table ended up having 10 columns, but they are very different and readable to my taste, so might be not a big deal that it's longer than 6 cols, I hope:

Screen Shot 2020-07-11 at 00 12 44
 
PURPLE also generates [some plots](https://github.com/hartwigmedical/hmftools/tree/master/purity-ploidy-estimator#charts) that might be tricky to reimplement with highcharts though (and computationally intensive), so not sure if it's needed.


ohofmann:
cc @jonbaber @p-priestley

ewels:
Hi @vladsaveliev - sorry I'm being slow here, been a busy time recently with @nf-core and conferences etc. I'm about to go on holiday, but when I get back MultiQC is at the top of the pile again, so stay tuned.

Phil

ewels (approved):
A few minor tweaks pushed to the PR, but all looking great now!

Thanks for the contribution, sorry for the time it took me to get to it.

ohofmann:
Yay! Thanks Vlad and Phil :)

PR #1273, opened 2020-08-18
JCVI genome annotation

abretaud:

 - [ ] There is example tool output for tools in the https://github.com/ewels/MultiQC_TestData repository or attached to this PR

Here's a new module parsing the output of the genome annotation statistics generated by the [JVCI python module](https://github.com/tanghaibao/jcvi).
The JVCI stats can be generated using Galaxy, and are used in an [annotation tutorial from the GTN](https://training.galaxyproject.org/training-material/topics/genome-annotation/tutorials/annotation-with-maker/tutorial.html).

I will push the test data to a PR on https://github.com/ewels/MultiQC_TestData


ewels:
Hi @abretaud,

I get an error when trying your module on the test data you supplied:

```python
Module jcvi raised an exception: Traceback (most recent call last):
  File "/Users/phil.ewels/GitHub/MultiQC/MultiQC/multiqc/multiqc.py", line 569, in run
    output = mod()
  File "/Users/phil.ewels/GitHub/MultiQC/MultiQC/multiqc/modules/jcvi/jcvi.py", line 26, in __init__
    self.parse_jcvi(f)
  File "/Users/phil.ewels/GitHub/MultiQC/MultiQC/multiqc/modules/jcvi/jcvi.py", line 280, in parse_jcvi
    exon_count = self.parse_hists(os.path.join(f['root'], 'Exon_Count', s_name + '.txt'))
  File "/Users/phil.ewels/GitHub/MultiQC/MultiQC/multiqc/modules/jcvi/jcvi.py", line 314, in parse_hists
    for required_val in range(bin_by / 2, max(values), bin_by):
TypeError: 'float' object cannot be interpreted as an integer
```

I guess tests were passing because they ran before the test data PR was merged. I just fixed a merge conflict in this PR and I suspect that the new tests there will now fail with the above error.

Phil

ewels:
Hi @abretaud,

I fixed the error and pushed one or two other tiny tweaks. When I generate a report, the histogram plots look kind of exciting though:

jcvi_exon_count_plot

All of the integers on the x-axis have a count of `0` and all the plots with counts are on `.5` of a value. This looks to me a bit like a problem with the way that the histograms are being generated. Could you take a look and confirm that there isn't anything funky happening please?

Thanks! And sorry that it's taken me so long to get to this.

Phil

abretaud:
Hi,
Sorry for the delay!
The problem was that I only tested with python 2, which was producing better looking plots.
I've rewritten the binning code, it should be more reliable now, and produce correct plots on python 3 (2 too by the way)

ewels:
Hi @abretaud,

I'm finally getting back to this PR and taking a look. I have just pushed a few commits:

* Fix merge conflicts with upstream code
* Ran [Black](https://black.readthedocs.io/en/stable/) on the code as I am now using this to standardise formatting
* Refactored some of the parsing code to be a little more succinct (looping over a dict of regexes instead of loads of separate statements)

As far as I can tell, the output of the code is unaffected.

I'm going to have a quick look at collapsing some of these plots if that's ok, as I think that quite a lot could be grouped together and accessible through tabbed buttons. I'll let you know what I do.

Phil

ewels:
Ok, I've just compressed the sections so that now we have fewer sections but the plots have tabbed buttons above them so you can select the feature type.

This makes the report shorter and easier to navigate in my mind. All of the plot data is still there unchanged.

If you get a chance, please take a look and make sure that you're happy with these changes!

Thanks,

Phil

abretaud:
Thanks! Yep, it looks better like this :+1:

PR #1283, opened 2020-09-02
Support `--collapse` on ErrorRateByReadPosition

sstadick:
This now supports the old format output that only has columns 0:11, and
the new format which includes another 7 columns that may or may not have
values.

This change supports the PR found here: https://github.com/fulcrumgenomics/fgbio/pull/608 that adds the option to not collapse positions.

The original test data for this works still.

### DIFF
--- a/multiqc/modules/fgbio/ErrorRateByReadPosition.py
+++ b/multiqc/modules/fgbio/ErrorRateByReadPosition.py
@@ -21,11 +21,16 @@ def parse_reports(self):
     error_rates = dict()
     y_max = 0.01  # default to 1%
     collapse = True  # same as the `--collapse` option on `ErrorRateByReadPosition`
+    is_new_format = False  # Test if this is new or old format
     for f in self.find_log_files('fgbio/errorratebyreadposition', filehandles=True):
         fh = f['f']
         header = fh.readline().rstrip('\r\n').split('\t')
         if not header or header[0] != 'read_number':
             continue
+
+        # Check if this is the new style
+        if 'collapsed' in header:
+            is_new_format = True
 
         # slurp in the data for this sample
         s_name = f['s_name']
@@ -34,10 +39,10 @@ def parse_reports(self):
         errors = 0
         for line in fh:
             fields = line.rstrip('\r\n').split('\t')
+            assert len(fields) == len(header), "Missing fields in line: `{}`".format(line)
             fields[1:4] = [int(field) for field in fields[1:4]]
             fields[4:11] = [float(field) for field in fields[4:11]]
-            # Check if this is the new style
-            if len(fields) > 11:
+            if is_new_format:
                 # Check if collapse was true or false
                 fields[-1] = bool(strtobool(fields[-1]))
                 collapse = fields[-1]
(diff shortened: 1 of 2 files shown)


multiqc/modules/fgbio/ErrorRateByReadPosition.py
```diff
@@ -32,14 +35,24 @@ def parse_reports(self):
         for line in fh:
             fields = line.rstrip('\r\n').split('\t')
             fields[1:4] = [int(field) for field in fields[1:4]]
-            fields[4:] = [float(field) for field in fields[4:]]
+            fields[4:11] = [float(field) for field in fields[4:11]]
+            # Check if this is the new style
+            if len(fields) > 11:
```
nh13, line 40: I'd first check in the header (i.e. `"collapse" in header` or whatever) to get the expected # of columns.  Then each row, I'd get the # of columns to keep and make sure the # of columns is the same across all rows.

multiqc/modules/fgbio/ErrorRateByReadPosition.py
```diff
@@ -73,33 +85,48 @@ def parse_reports(self):
...
-            {'name': 'C > A', 'ylab': 'C to A error rate'},
-            {'name': 'C > G', 'ylab': 'C to G error rate'},
-            {'name': 'C > T', 'ylab': 'C to T error rate'},
-        ]
+            {'name': 'A > C', 'ylab': 'A to C error rate (and T to G when collapsed)'},
+            {'name': 'A > G', 'ylab': 'A to G error rate (and T to C when collapsed)'},
+            {'name': 'A > T', 'ylab': 'A to T error rate (and T to A when collapsed)'},
+            {'name': 'C > A', 'ylab': 'C to A error rate (and G to T when collapsed)'},
+            {'name': 'C > G', 'ylab': 'C to G error rate (and G to C when collapsed)'},
+            {'name': 'C > T', 'ylab': 'C to T error rate (and G to A when collapsed)'},
```
nh13, line 93: I think the added text clutters the legend.  Can you add this information to the title or elsewhere?

multiqc/modules/fgbio/ErrorRateByReadPosition.py
```diff
@@ -73,33 +85,48 @@ def parse_reports(self):
...
-        description = 'Error rate by read position. Plot tabs show the error rates for specific substitution types.',
-        helptext = '''
-        The error rate by read position. Substitution types are collapsed based on the
-        reference or expected base, with only six substitution types being reported:
-        `A>C`, `A>G`, `A>T`, `C>A`, `C>G` and `C>T`.
-        For example, `T>G` is grouped in with `A>C`.
+    self.add_section(
+        name='Error Rate by Read Position',
+        anchor='fgbio-error-rate-by-read-position',
+        description='Error rate by read position. Plot tabs show the error rates for specific substitution types.',
```
nh13, line 123: Perhaps add "collapse=true/false" somwhere here instead of in the legend?  It's clear in the legend if it's collapsed or not given the # of lines.

nh13 (changes requested):

multiqc/modules/fgbio/ErrorRateByReadPosition.py
sstadick: I kind of think the help text covers it anyways, so I've just removed it.

multiqc/modules/fgbio/ErrorRateByReadPosition.py
sstadick: I added the `collapse` status to the description

multiqc/modules/fgbio/ErrorRateByReadPosition.py
sstadick: But the number of columns to keep depends on whether or not collapsed is True or not which we can't know until we actually start parsing the data OR implement something more complicated to peek ahead. 

I've changed it so that I check the header to establish which format type this is, and then assert that all lines match the header length. I do like getting the format from the header instead of nebulously checking each rows length. 

LMK if the new code is closer to what you are thinking.

nh13 (approved):
LGTM, @ewels this is ready from our end to merge.  Care to take a look?

sstadick:
@ewels, just wanted to check in and see if you've had a chance to look at this!

ewels (approved):

ewels:
Merged https://github.com/ewels/MultiQC/pull/1304 first which cause a conflict sorry. I assume that #1304 takes priority as it's a more recent PR..? But I'm not 100% sure so I will leave it to you guys to fix this just in case.

Thanks again, and sorry for the delay.

Phil

sstadick:
@ewels, you were correct on the priority order!

I've fixed the conflicts and this should be good to go. Thanks again!

ewels (approved):
Untested locally, but limited scope of changes so merging ✅

ewels:
Thanks!

nh13:
@ewels can this be part of a release?

ewels:
Yes, you're right I should just push a release. My previous pattern has been to do a sprint of work to get all (most) open PRs merged and bug issues fixed before each release. But I'm not able to keep up with them so this just hasn't happened over the past months.

I will push a release with the current `master` this week 👍🏻

ewels:
Apologies, kids fell sick and had to be looked after plus a grant deadline and having to run some job interviews. The release gets a stay of execution for a few more days. I'm moving house on Monday but will try to get to it later next week 😅

sstadick:
Thanks @ewels! Your work is appreciated!

PR #1288, opened 2020-09-09
New module: CCS

Redmar-van-den-Berg:
Thanks for the documentation on writing your own module, the whole process was easy to follow!

I've created a module for CCS, which is a tool to work with PacBio data. I'm not sure if it is ready to be merged yet but I do have a working version, so I was wondering if you have any feedback.

I've also added some test data (https://github.com/Redmar-van-den-Berg/MultiQC_TestData/tree/ccs) but I ended up using pytest to create more fine-grained tests, since the PacBio output files are not so easy to parse.


ewels:
Hi @Redmar-van-den-Berg,

Sorry it's taking me so long to get to these PRs. I've just fixed your merge conflicts and run the new code formatters to get the CI passing again, so please pull my commits before you do any more yourself.

I've quickly run your code on the (now-merged) TestData and am mostly pretty happy with it. There are a couple of things I wanted to mention though:

1. I don't love that the bar graph is not stacked and using tabs for different datasets. This doesn't scale well past a handful of samples and means that you can't easily compare different samples - these are two pretty core features that I try to keep MultiQC focussed on. Basically I expect users to dig for details in original underlying reports but want them to be able to spot outlier samples quickly.
    * Would you be ok with reverting to a regular stacked bar plot in this instance?
2. There is nothing added to the General Statistics table
    * Most modules usually add one or two key stats here, and I think that something like `ZMWs input` and `% ZMW pass filters` could be good candidates here? Just to stack up against columns from other modules to try to make sense of patterns across different steps in the analysis.
3. The v4 and v5 data end up with identical sample names in my testing (both called `ccs`) and overwrite one another randomly. Is the filename _always_ `ccs.report.txt`? If so we need a different way to get a sample name that is unique and more useful.

Let me know what you think about the above and I'll come back for a more thorough code review (though I think it basically looks fine already).

Apologies again for the slow review time.

Phil

ewels:
https://ccs.how/faq/reports-aux-files.html

> By default, each CCS run generates a `<outputPrefix>.ccs_report.txt` file.

Struggling to navigate the PB docs to find out how `outputPrefix` is decided upon (eg. optional or has a default so is always present).

ewels:
@Redmar-van-den-Berg - I'm quite keen to get your new MultiQC modules merged and generally available. I'm going to start implementing my suggestions in your PRs if that's ok. If you get a chance to check-in and respond that would be fab 👍🏻  Otherwise we can always take the discussion afterwards and create follow-up PRs if needed.

ewels:
Ok, I made a few changes and am happy now:

* Bar graph is now stacked
    * I calculate the total count of each category across all samples and then sort by this. This way, the smallest categories come first which works well if you want to use the log-scale
* I added total count, PF and %PF to the General Stats table
* I added `ccs` to the list of suffixes to trim off
    * If there is no output prefix, this then gives an empty string. MultiQC responds to this by giving the full filename
    * If combined with `-d` to get directory prefixes, the empty string is allowed and you just get the directory names
    * If the filenames have an output prefix, then you should get nice sample names.

Redmar-van-den-Berg:
Thanks for your fixes and for merging this module!

ewels:
Thanks for contributing! 🎉

PR #1292, opened 2020-09-11
Add preliminary boxplot support to MultiQC

bnbowman:
This PR adds preliminary support for generating box-plots, utilizing a slide-show style method similar to javascript-disabled lineplots to visualize multiple datasets in a single window.

This enables MultiQC to generate a more FastQC-like plot to visualize the range and distribution of per-base quality metrics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@
 * General Stats custom content now gives a log message
 * If `id` is not set in `JSON` or `YAML` files, it defaults to the sample name instead of just `custom_content`
 * Data from `JSON` or `YAML` now has `data` keys (sample names) run through the `clean_s_name()` function to apply sample name cleanup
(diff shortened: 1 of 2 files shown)


ewels:
Looking good @bnbowman - thanks! Any thoughts about adding an interactive version with HighCharts as well?

Aashna01:
Hi @bnbowman and @ewels : Is this boxplot ready to use for other datasets? I tried using this branch to run box plots(I have been looking into generating boxplots from MultiQC since quite a long time). However, when I used  this commit, I am facing this error constantly:
**Error - custom content plot type 'boxplot' not recognised for content ID custom_data_boxplot**
Also, Could you please provide an example input file for this boxplot if possible?
It would be a great help to see the format of the file needed for this boxplot.
Thanks in advance.

Issue #1293, opened 2020-09-14
Expose additional statistics in Picard Alignment Summary


Redmar-van-den-Berg:
**Is your feature request related to a problem? Please describe.**
The Picard Alignment Summary currently only shows the number of aligned reads in the barchart. This makes sense for Illumina data, where all reads are the same size. However, for long read sequencing, the number of aligned reads is not so relevant, and the number of aligned bases or the mean read size is more relevant.

**Describe the solution you'd like**
Add additional datasets to the picard Alignment Summary section. This way, the default behaviour will be the same (show the number of aligned reads), but users can switch over to number of aligned bases or mean read length, if this is more relevant for them.

**Describe alternatives you've considered**
I was also considering adding additional Read based statistics to the graph but making them invisible by default (like they are when you click on the legend in the HTML report), but I could not find a way to do this in the documentation.

ewels:
There could also be a config option for this to allow it to be set as default when running routinely with this kind of data.

Redmar-van-den-Berg:
I'm not sure if a configuration setting to change the default is a good solution. Even for long reads, you still want to have an idea of the number of aligned reads. Having buttons to switch to the additional data is also easier to use then having a config option. 

Depending on the experimental setup, having the mean read length and total number of aligned bases might also be relevant for people working with Illumina data, so it would be nice to have this data available for all users.

PR #1294, opened 2020-09-14
Add additional datasets to Picard Alignment Summary

Redmar-van-den-Berg:
This branch exposes additional datasets for the Picard Alignment Summary in the HTML report. See the [corresponding issue](https://github.com/ewels/MultiQC/issues/1293) for details.


ewels (approved):
Thanks for this, looks good! 👍🏻 

I did a bit of refactoring to make the code a bit shorter where possible. I also split the read length plot into its own section as I felt that it was a bit wrong to be labelled under Alignment summary somehow. It also meant that I could drop the _Percentages_ button for that bar plot (which makes no sense when there is a single category).

Shame that Picard doesn't report total bases so that we can show the unaligned bases!

PR #1300, opened 2020-09-25
Add eigenstratdatabasetools module.

TCLamnidis:
Added a module for the SNP coverage calculation script of [EigenStratDatabaseTools](https://github.com/TCLamnidis/EigenStratDatabaseTools). The module parses the json output of `eigenstrat_snp_coverage.py` and adds it to the general Stats table. Hoping to add this to eager reports, and preferred a non-custom content approach.


TCLamnidis:
Test data PR: https://github.com/ewels/MultiQC_TestData/pull/170

TCLamnidis:
The module only adds 2 table columns. There is no reason in my view to add a section and/or plot for it. The main column of interest (`#Covered_Snps`) follows a colour scale that, and closely emulates what a bar plot would look like anyhow.

@ewels Is there a way to set the `'max'` attribute of a column from another column of the data? I'd like to be able to read the value of the `"#Total SNPs"` column and use that as the `'max'` of the `#Covered_Snps` column if possible.

ewels:
Looking good, thanks for this! Apologies for taking so long to get to the PR.

> Is there a way to set the 'max' attribute of a column from another column of the data?

There are a few ways - the most obvious is probably to just find the maximum data value yourself and then set `max` to that variable. However, once you know about it, using the `shared_key` config option is best. This does exactly what you're looking for and will also work together with any other module columns using the same key (eg. `multivcfanalyzer` and `sexdeterrmine` in this case).

I've refactored slightly to put this in place, plus print a log of how many samples were found and one or two other bits 👍🏻

PR #1302, opened 2020-09-26
extend PATH variable including directory containing multiqc binary. A…

matzmz:
extended PATH env variable including the directory that contains multiqc binary.

The docker container will work correctly even if entrypoint will be overwrite to default value.

```sh
docker run -it --rm --entrypoint="" -v `pwd`:`pwd` -w `pwd` ewels/multiqc multiqc . --title "My amazing report" -b "This was made with docker"
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,8 @@ WORKDIR /usr/src/multiqc
 # Install MultiQC
 RUN python -m pip install .
 
+ENV PATH "$PATH:/usr/src/multiqc"
+
 # Set up entrypoint and cmd for easy docker usage
 ENTRYPOINT [ "multiqc" ]
 CMD [ "." ]


Dockerfile
```diff
@@ -14,6 +14,8 @@ WORKDIR /usr/src/multiqc
 # Install MultiQC
 RUN python -m pip install .
 
+ENV PATH "$PATH:/usr/src/multiqc"
```
ewels, line 17: Should this not be `/usr/src/` if it's the directory that contains the `multiqc`binary? This looks like the full path to the binary to me..?

ewels (commented):
Nice idea, thanks! One minor question but that's all (hoping that the failing tests were a fluke, forced a retry just now).

ewels:
Hi @matzmz, is there a reason that you closed this PR?

PR #1306, opened 2020-09-29
Fix broken link

marcelm:


docs/config.md
```diff
@@ -376,7 +376,7 @@ know you have files for. MultiQC supports a _lot_ of different tools and searche
 for matching files for all of them every time you run it.
 
 You can do this with the `-m` / `--module` flag (can be repeated) or in a MultiQC
-config file by using `config.module_order`. See [Order of modules](/#order-of-modules).
+config file by using `config.module_order`. See [Order of modules](https://multiqc.info/docs/#order-of-modules).
```
ewels, line 379: Ah, it was the leading slash that caused the problem I think. Thanks! 👍🏻 

```suggestion
config file by using `config.module_order`. See [Order of modules](#order-of-modules).
```

ewels (approved):

docs/config.md
marcelm: Hm, that is the version I thought of initially, but then I checked some of the other links in the documentation and they used the full URL. Which I thought make sense because if you just use the anchor, the link won’t work when viewing the Markdown document on GitHub itself because the anchor is in a different file. That is, now the link does not work within https://github.com/ewels/MultiQC/blob/master/docs/config.md . (I prefer your way, just pointing out that it’s inconsistent at the moment.)

PR #1309, opened 2020-10-01
Document the name of the custom_content module

marcelm:
I spent some time using `custom_data` as module name within `run_modules` (because that’s the name of the key one needs to write) until I found how to do it in the examples.


ewels (approved):
Great stuff, thanks!

PR #1313, opened 2020-10-07
Renaming description for duplication rate

matmu:
Duplication rate is calculated on the original/unfiltered reads according to https://github.com/OpenGene/fastp/issues/290


ewels (approved):
Thanks for spotting and fixing this!

PR #1316, opened 2020-10-08
Fix CHIP value detection

sstadick:
Previously the `verifyBamID` module was checking for an exact match of the string "CHIP" in a list [headers[i]], which would always be false. I've updated it to what I believe was the original intent, checking if the substring "CHIP" starts the header in question. 

This will now allow for CHIP values to show up in the tables.


CHANGELOG.md
```diff
@@ -30,6 +30,8 @@
     * Fix `ErrorRateByReadPosition` plotted line names to no longer concatenate multiple read identifiers and no longer have off-by-one read numbering (ex. `Sample1_R2_R3` -> `Sample1_R2`) ([#[1304](https://github.com/ewels/MultiQC/pull/1304))
 * **GATK**
   * Add support for the creation of a "Reported vs Empirical Quality" graph to the Base Recalibration module.
+* **verifyBamID**
+  * Fix "CHIP" value detection to notice when there are "CHIP" values to display. ([#1316](https://github.com/ewels/MultiQC/pull/1316))
```
nh13, line 34:
```suggestion
  * Fix "CHIP" value detection to notice when there are "CHIP" values to display. Previously, if any sample does not have a "CHIP" value, no samples with "CHIP" values are displayed.  ([#1316](https://github.com/ewels/MultiQC/pull/1316))
```

nh13 (approved):

CHANGELOG.md
sstadick: While your edit is strictly true, it was more generally true the previously NO "CHIP" value would ever be shown even if any sample had "CHIP" data

sstadick, line 34:
```suggestion
  * Fix "CHIP" value detection to notice when there are "CHIP" values to display. ([#1316](https://github.com/ewels/MultiQC/pull/1316)). Previously no "CHIP" values would be displayed even if there were "CHIP" values present.
```

ewels (approved):
Thanks both!

PR #1326, opened 2020-10-23
Added the %Occupied metric

berguner:
Added the %Occupied metric to run_metrics_details_table which is reported for NovaSeq and iSeq platforms. It's very simple, just a few lines of code and no need to change the documentation. 


ewels:
Sorry for taking so long to get to such a small PR @berguner - I guess the test data in ewels/MultiQC_TestData doesn't have this field as I can't see any difference when I test before and after the change. But it's minor enough that I'm happy to merge anyway 👍🏻 

Thanks!

PR #1327, opened 2020-10-23
Feature/picard crosscheck fingerprints

sstadick:
This add support for picard CrosscheckFingerprints output. A column is added to General Stats for each sample that indicates True/False (colored Green/Red respectively) as to whether or not all comparisons for that sample return Expected relationships.

There is also a CrosscheckFingerprints table that includes the pairwise comparison, the categorical result, the LOD_SCORE, and the threshold used, to be referenced in the event that a `False` is observed in the General Statistics.

 - [X] There is example tool output for tools in the https://github.com/ewels/MultiQC_TestData repository or attached to this PR

For test data see https://github.com/ewels/MultiQC_TestData/pull/172


sstadick:
Example of report on the test data:
image

sstadick:
Example report from the linked test data
image

docs/modules/picard.md
```diff
@@ -13,54 +13,24 @@ sequencing data.
 
 Supported commands:
 
-* `MarkDuplicates`
-* `InsertSizeMetrics`
+* `AlignmentSummaryMetrics`
```
sstadick, line 16: I reordered these to be alphabetical and also fixed the order of sections below this to match the order in this list

docs/modules/picard.md
```diff
@@ -84,15 +54,24 @@ picard_config:
...
+    CrosscheckFingerprints_table_cols:
+        - RESULT
+        - LOD_SCORE
+    CrosscheckFingerprints_table_cols_hidden:
+        - LEFT_LANE
+        - RIGHT_LANE
 ```
 
+Note that if `CALCULATE_TUMOR_AWARE_RESULTS` was set to true on the CLI for any of the CrosscheckFingerprints result files, then the `LOD_SCORE_TUMOR_NORMAL` and `LOD_SCORE_NORMAL_TUMOR` will be displayed.
+
```
sstadick, line 74: This is the new section

docs/modules/picard.md
```diff
@@ -13,54 +13,24 @@ sequencing data.
 
 Supported commands:
 
-* `MarkDuplicates`
-* `InsertSizeMetrics`
+* `AlignmentSummaryMetrics`
+* `BaseDistributionByCycle`
```
sstadick, line 17: Fixed spelling of this item

multiqc/modules/picard/CrosscheckFingerprints.py
```diff
@@ -0,0 +1,236 @@
...
+            name = _get_name(row)
+
+            # Set the cli options of interest for this file
+            row["LOD_THRESHOLD"] = lod_threshold
+            row["TUMOR_AWARENESS"] = tumor_awareness
+            self.picard_CrosscheckFingerprints_data[name] = row
+
+    # Only add sections if we found data
+    if len(self.picard_CrosscheckFingerprints_data) > 0:
+        # For each sample, flag if any comparisons don't start with "Expected"
```
sstadick, line 62:
```suggestion
        # For each sample, flag if any comparisons that don't start with "Expected"
```

multiqc/modules/picard/CrosscheckFingerprints.py
```diff
@@ -0,0 +1,236 @@
...
+        )
+
+        # Add a table section to the report
+        self.add_section(
+            name="Crosscheck Fingerprints",
+            anchor="picard-crosscheckfingerprints",
+            description="Pairwise identity checking betwen samples and groups.",
+            helptext="""
+            Checks that all data in the set of input files comes from the same individual, based on the selected group granularity.
+
```
sstadick, line 83:
```suggestion
```

multiqc/modules/picard/CrosscheckFingerprints.py
```diff
@@ -0,0 +1,236 @@
...
+    """Look at the LEFT_SAMPLE fields and determine if there are any pairs for that samples
+    that don't have a RESULT that startswith EXPECTED.
+    """
+    out_data = dict()
+    flattened = (row for row in in_data.values())
+    sorted_by_left_sample = sorted(flattened, key=lambda r: r["LEFT_SAMPLE"])
+
+    for group, values in groupby(sorted_by_left_sample, key=lambda r: r["LEFT_SAMPLE"]):
+        out_data[group] = {
+            # NB: Must coerce bool to str or else plot.table turns it into a float
```
nh13, line 230: lol

multiqc/modules/picard/CrosscheckFingerprints.py
```diff
@@ -0,0 +1,236 @@
...
+            ]
+
+        headers = OrderedDict()
+        for h in FIELD_DESCRIPTIONS:
+            # Skip anything not set to visible
+            if h not in crosscheckfingerprints_table_cols:
+                continue
+
+            # Set up the configuration for the column
+            h_title = h.replace("_", " ").strip().lower().capitalize()
```
nh13, line 199: answer is here for if the configuration key can be lowercase.  Can you update the doc?

nh13 (approved):
Any chance you can screenshot what this looks like with https://github.com/ewels/MultiQC_TestData/pull/172?

yfarjoun:
Do note that the "Expected" algorithm doesn't actually know if samples are from the same individual, and so LOD is  "expected" to be positive is the sample ID is the same (from VCF or bam header) or if tool was run with "EXPECT_ALL_GROUPS_TO_MATCH=true". I've been toying with the idea of having a concept of "Individual", but that's not for here to discuss....

nh13:
@sstadick I see that the `Sample Name` column contains both the left and right sample names.  Any chance we could have those also in two separate columns so they're sortable?  What do you think?

sstadick:
@nh13 I don't think so. The input data for the table must come as a dictionary with the sample names as the key, so you can't have multiple rows with the same sample name. 

Also of note, if someone were to run CrosscheckFingerprints and use ReadGroups as the level of granularity of the checking, I expand the names out to `<left-sample>/<left-group> - <right-sample>/<right-group>`.

My overall feeling is that `plots.table` just isn't made for  this and it's a force fit for sure. The main "feature" here is in the general stats table with the red/green color of all expected.

sstadick:
Latest screenshot
image

nh13:
@sstadick that's pretty!

@ewels I think this is ready to go!

ewels:
Thanks all! Test data merged, will take a look at the PR ASAP.

I think that there was only one sample in the test data PR? If you fancy adding a few more then that makes testing a bit easier as I can see how a typical report would look.. (But not critical).

Phil

nh13:
@ewels I think this is ready to go.  The only issue I see is that when the sample names are long, the columns will overlap, but I am assuming this is an issue for all tables?  If the columns were fit better, or re-sizeable, I think that would help.

matthdsm:
Hi @ewels,

We're looking forward to using this module, can this be merged?

Thanks
M

docs/modules/picard.md
ewels: Great! Except `R` comes after `Q` in the alphabet usually 😅

ewels:
Hi all,

Thanks for this! I've gone over the code and done a fair bit of refactoring, so please check that it still works as you expect.

I actually found a nasty little bug in the table code during this (maybe the reason why you disabled the colour schemes for `LOD`?) which was good to fix. I also added a new table config option to treat values as absolute for the bar width which works really nicely here to highlight the extreme LODs properly.

Along with a few other fixes and tweaks, I think that it looks even better now:

image

Note that the zero-point for this new feature is hardcoded to `0` for now. Surely someone will ask for this to be customisable but that can be rewritten in a future PR if so 😀 

I'll merge this now but please give it a try and shout if I messed anything up 👍🏻 

Phil

PR #1340, opened 2020-11-11
New module: odgi

Zethson:
Hi!,

test data PR: https://github.com/ewels/MultiQC_TestData/pull/174
Should be merged with this one.

 - [X] There is example tool output for tools in the https://github.com/ewels/MultiQC_TestData repository or attached to this PR

The odgi module is part of the nf-core pangenome pipeline.


ewels:
Looks like tests are failing for you - I think it’s this that’s the problem: `odgi : Found 0 reports`

You’re not raising a `UserWarning` if you don’t find any samples, which means that every single MultiQC report anyone generates will have an odgi section, even if they run it on an empty directory or ignore all files 😃

multiqc/modules/odgi/odgi.py
```diff
@@ -0,0 +1,199 @@
...
+        super(MultiqcModule, self).__init__(name='Odgi', anchor='odgi',
+                                            href='https://github.com/vgteam/odgi',
+                                            info='is an optimized dynamic graph/genome implementation.')
+
+        self.odgi_stats_map = dict()
+
+        # Find and load any odgi stats file
+        log_files = self.find_log_files('odgi')
+        if not log_files:
+            raise UserWarning
```
Zethson, line 32: We are raising one here @ewels ?

Zethson:
@ewels but we are raising a UserException when no reports were found.
Are we raising it under the wrong condition?

multiqc/modules/odgi/odgi.py
ewels: You need to use `self.ignore_samples()` too: https://multiqc.info/docs/#filtering-by-parsed-sample-names

ewels: See Kallisto for an example of this: https://github.com/ewels/MultiQC/blob/735f0bc57f408d0147a81cad1e93a711a84328e9/multiqc/modules/kallisto/kallisto.py#L31-L42

ewels:
Thanks all! I did a bunch of nasty code formatting stuff - implementing Black etc just recently so now you have a tonne of merge conflicts. If you are comfortable doing a rebase of this PR then that would be fab. If in doubt leave it though and I will do it when I come to review.

Phil

Zethson:
We'll take care of this.

Zethson:
@subwaystation @Imipenem @ewels  rebased (hopefully correctly).

multiqc/modules/odgi/odgi.py
```diff
@@ -0,0 +1,202 @@
...
+        path    in_node_space    in_nucleotide_space    nodes    nucleotides    num_penalties    num_penalties_different_orientation
+        all_paths    20.0686     19.5609    977     51365     90    0
+        """
+        lines = f['f'].splitlines()
+        # parse the generals stats
+        stats = lines[1].split('\t')
+        # parse the mean links length
+        mean_links_length = lines[4].split('\t')
+        # parse sum of path nodes distances
+        sum_of_path_nodes_distances = lines[7].split('\t')
```
ewels, line 72: I generally try to avoid addressing specific line numbers if at all possible. It is surprisingly common for things like HPC schedulers etc to insert unexpected lines into the top of files, which can offset everything and cause chaos.

If possible, it would give more robust code to parse by line contents instead.

multiqc/modules/odgi/odgi.py
```diff
@@ -0,0 +1,202 @@
...
+            return 'seqwish'
+
+    def compress_stats_data(self, sample_name, stats, mean_links_length, sum_of_path_nodes_distances) -> dict:
+        """
+        Compress odgi stats into a single dictionary to visualize.
+        """
+        return {
+            sample_name: {
+                'General stats {}'.format(sample_name): {
+                    'Length': float(stats[0]),
```
ewels, line 181: Again this is very brittle. Can we not use the column headers in the file instead?

ewels (commented):
Hi all,

Your PR made it to the top of the queue at last! I fixed some merge conflicts and now I'm getting some CI errors though (locally too):

`./cerevisiae.pan.fa.gz.pggb-s10000-p85-n10-a85-K23-k23-w10000-j5000-e5000.consensus@10000.gfa.og.stats`
```python
  File "/Users/ewels/GitHub/MultiQC/MultiQC/multiqc/multiqc.py", line 604, in run
    output = mod()
  File "/Users/ewels/GitHub/MultiQC/MultiQC/multiqc/modules/odgi/odgi.py", line 51, in __init__
    self.plot_odgi_metrics()
  File "/Users/ewels/GitHub/MultiQC/MultiQC/multiqc/modules/odgi/odgi.py", line 118, in plot_odgi_metrics
    odgi_stats_file_names = sorted(self.odgi_stats_map.keys())
TypeError: '<' not supported between instances of 'str' and 'NoneType'
```

Code looks good, my main concern is that the parsing is very inflexible and brittle. Is it possible to rewrite it to make use of the column headers in the log files instead?

Do you guys maintain this tool? Maybe it's possible to generate file outputs in a nicer format instead?

Phil

subwaystation:
Hi @ewels 
thanks for the feedback!
I am one of the maintainers of `odgi`. Any suggestions how to reformat the output file?

I agree, the code is brittle. Using the column headers will be unavoidable in the future, I will change that.

Regarding the small test case: I can't see right now, where the problem is, but I will take a look.

subwaystation:
Reworking `odgi stats` to emit a YAML instead of a pseudo TSV file.

ewels:
Ok fantastic! Definitely the best way forward if possible I think 👍🏻  Keeps the code here nice and simple and more robust ✨ 

Thanks! Let me know when it's ready - I'll put this PR on hold until then.

Phil

subwaystation:
@ewels Do you have some modules in mind which use YAML as input? So I can learn from non-brittle code ;)

ewels:
Sure! It's basically just `yaml.load()`  (which is why I like it).

https://github.com/ewels/MultiQC/blob/c8528fb8347ceeb02466342562371358b8c82b17/multiqc/modules/pycoqc/pycoqc.py#L56-L62

https://github.com/ewels/MultiQC/blob/c8528fb8347ceeb02466342562371358b8c82b17/multiqc/modules/snpsplit/snpsplit.py#L58-L59

There are also loads using `json.loads()` which is basically the same.

Once it's parsed, it's just a case of reshaping the data to fit the various MultiQC plotting functions etc. But I think that you probably have most of this in place already.

apeltzer:
I've added JSON output for multiple of my tools and found it (Java tools) slightly easier than YAML infact - e.g. [here](https://github.com/apeltzer/DeDup/blob/master/src/main/java/RMDupper.java#L259-L291) 

Shouldn't be that different doing this in C/C++ too, using a simple JSON library for that.

subwaystation:
Got it. As I already implemented the YAML output, I will leave it as it is. But good to know next time!

subwaystation:
@ewels Could you please trigger the CI? Thanks!

subwaystation:
@ewels Seems like the `rsem` module is failing? https://github.com/Imipenem/MultiQC/runs/2960964259?check_suite_focus=true.
I suspect this is not a problem of my code, but of MultiQC? Could it be fixed my merging master into this branch?

subwaystation:
Works :)

subwaystation:
I would still like to add some more basic statistics to the statistics panel. Like number of bases, number of disconnected components. But if you have it urgent with the new release, this is the minimum setting that makes sense.
Anyhow, I will push for these new features, latest Friday they should all be in.

subwaystation:
@ewels Is it possible to give the `Components` column less space, so that the columns of the bases get enough space? I couldn't find anything in https://multiqc.info/docs/#step-3-adding-to-the-general-statistics-table. Thanks!

image

subwaystation:
The same goes for the `Sample Name` column.

multiqc/modules/odgi/odgi.py
```diff
@@ -0,0 +1,277 @@
...
+        """
+        Plot the odgi stats by adding them to the general statistics table.
+        """
+        data = dict()
+        headers = OrderedDict()
+        headers["Length"] = {"title": "Length", "description": "Graph length in nucleotides.", "scale": "BuPu"}
+        headers["Nodes"] = {
+            "title": "Nodes",
+            "description": "Number of nodes in the graph.",
+            "scale": "OrRd",
```
apeltzer, line 118: As the number of nodes will always be an integer I assume, you could do a format here and remove the `.0` - same for the other columns. 

`'format': '{:,.0f}',` should do the trick here 👍🏻

apeltzer, line 118:
```suggestion
            "scale": "OrRd",
            'format': '{:,.0f}', 
```

multiqc/modules/odgi/odgi.py
```diff
@@ -0,0 +1,277 @@
...
+        }
+        headers["Edges"] = {"title": "Edges", "description": "Number of edges in the graph.", "scale": "PuBu"}
```
apeltzer, line 120:
```suggestion
        headers["Edges"] = {
        "title": "Edges", 
        "description": "Number of edges in the graph.", 
        "scale": "PuBu",
        'format': '{:,.0f}' }
```

multiqc/modules/odgi/odgi.py
```diff
@@ -0,0 +1,277 @@
...
+        headers["Paths"] = {"title": "Paths", "description": "Number of paths in the graph.", "scale": "Greens"}
```
apeltzer, line 121:
```suggestion
        headers["Paths"] = {
        "title": "Paths", 
        "description": "Number of paths in the graph.", 
        "scale": "Greens",
        'format': '{:,.0f}'}
```

multiqc/modules/odgi/odgi.py
```diff
@@ -0,0 +1,277 @@
...
+        headers["Components"] = {
```
apeltzer, line 122: Same here

multiqc/modules/odgi/odgi.py
```diff
@@ -0,0 +1,277 @@
...
+            "title": "Components",
+            "description": "Number of weakly connected components in the " "graph.",
+            "scale": "Oranges",
+        }
+        headers["A"] = {"title": "A", "description": "Number of adenine bases in the graph.", "scale": "Spectral"}
```
apeltzer, line 127: and here

multiqc/modules/odgi/odgi.py
```diff
@@ -0,0 +1,277 @@
...
+        headers["C"] = {"title": "C", "description": "Number of cytosine bases in the graph.", "scale": "Greys"}
```
apeltzer, line 128: and here

multiqc/modules/odgi/odgi.py
```diff
@@ -0,0 +1,277 @@
...
+        headers["T"] = {"title": "T", "description": "Number of thymine bases in the graph.", "scale": "Blues"}
```
apeltzer, line 129: see above

multiqc/modules/odgi/odgi.py
```diff
@@ -0,0 +1,277 @@
...
+        headers["G"] = {"title": "G", "description": "Number of guanine bases in the graph.", "scale": "BrBG"}
```
apeltzer, line 130: and ^^

multiqc/modules/odgi/odgi.py
```diff
@@ -0,0 +1,277 @@
...
+        headers["N"] = {"title": "N", "description": "Number of `N` basis in the graph.", "scale": "BrBG"}
```
apeltzer, line 131: :top:

multiqc/modules/odgi/odgi.py
```diff
@@ -0,0 +1,277 @@
...
+        headers["Total"] = {
```
apeltzer, line 132: Also an integer -> no decimals

multiqc/modules/odgi/odgi.py
```diff
@@ -0,0 +1,277 @@
...
+            "title": "Nodes With Self Loops - Total",
+            "description": "Total number of nodes having self loops in the " "graph.",
+            "scale": "Set1",
+            "hidden": True,
+        }
+        headers["Unique"] = {
```
apeltzer, line 138: see above

apeltzer (commented):
Some comments, mostly for formatting

multiqc/modules/odgi/odgi.py
subwaystation: I was already wondering how to fix this, thanks for your suggestions!

subwaystation:
Thanks again @apeltzer. I think I fixed everything. 

This is ready for review @ewels. I already incorporated all features I wanted to have :)

ewels:
Ok great! I ran late reviewing this today, but maybe that was a good thing 😁 I have this and one other PR to review + merge, then I will release. Hoping to get to that ASAP now.

I'll probably just tweak stuff directly instead of asking for feedback to keep the loop fast, but I'll ping you if I change anything so that you can take a look and make sure that you agree 😅

multiqc/modules/odgi/odgi.py
ewels: Worth mentioning that generally I try to make big numbers small. It's tough to compare `18392833` against `1839283` but it's easy to compare `18.3` against `1.8`. So if you expect a column to have large numbers in it's usually better to keep the decimals and use the `modify` key to reduce the order.

Docs: https://multiqc.info/docs/#step-3-adding-to-the-general-statistics-table

https://github.com/ewels/MultiQC/blob/21875074c6cbcf8084ea7017522f7cde04b269b5/docs/modules.md#L675-L684

ewels:
> @ewels Is it possible to give the `Components` column less space, so that the columns of the bases get enough space? I couldn't find anything in https://multiqc.info/docs/#step-3-adding-to-the-general-statistics-table. Thanks!

Column widths are done automatically by the browser based on the length of the column headers. This usually works well, but you have columns here with headers that are 1 character long, so that's skewing everything else..

I'm guessing that I'm going to ask you to drop some of these columns as it looks like it's too many for General Stats (I usually say aim for 1-3 per module), and I think you can probably just replace with a single `%GC` column? So probably not worth worrying about too much 😁

ewels:
Hi all,

I've had a bit of a play with the code and made some changes. I'll try to explain:

* Minor code refactor without changing output - 1ae5b58
    - Use `self.clean_s_name()` function ([docs](https://multiqc.info/docs/#custom-sample-names))
    - Flattened the main data dict instead of using nested keys
        - This simplifies the module code quite a bit, as we can just pass the dict directly to the plotting / table functions
        - The default format for the `mutliqc_data` format is `tsv`, which does not support nested data structures. So this change makes the data export work for most people.
- Make the table column with nucleotides use base_pairs
    - See other comment - using a `modify` lambda function to show a smaller number.
    - Also using `shared_key` so that the bar scale is shared with any other columns in the table showing base pairs (including from other modules)
    - Other minor tidying
* Reduce number of columns in general stats - 4f0f929
    - Calculate %GC and %N and show these instead of all of the bases
        - Fixes the odd column widths and reduces the number of columns in General Statistics
        - I'd kind of prefer to have fewer still columns, but wasn't sure where to cut next so left it at that.
    - Code tidying: make data keys lower case
* Replace line graph with 2x bar graphs, each with 2x datasets - e1a76f0
    - You had a line graph where the x-axis was samples and the lines were different data types. Some other modules do this, but it's very much a last resort. All of the MultiQC code assumes that lines are samples, so stuff like hiding samples, highlighting, renaming etc will not work. Also, it breaks the common assumption by people used to MultiQC. I found it really confusing until I figured out what was going on.
    - I replaced with bar graphs, which are a bit boring and lose the linkage between categories. However, they scale better with large numbers of samples and don't suffer from the problems mentioned above.

Hoping to release tomorrow lunch time. I'll leave this PR open until then so that you can check out my changes and let me know what you think 👍🏻 

Thanks again for the new module!

Phil

multiqc/modules/odgi/odgi.py
```diff
@@ -27,7 +27,7 @@ def __init__(self):
             name="ODGI",
             anchor="odgi",
             href="https://github.com/pangenome/odgi",
-            info="is an optimized dynamic graph/genome implementation.",
+            info="is an optimized dynamic graph/genome implementation, allowing efficient computation and transformation of pangenomes with minimal overheads.",
```
subwaystation, line 30: I would not formulate it like this. Rather:
`, for efficient analysis and manipulation of pangenome graphs structured in the variation graph model`.
I will push something here.

multiqc/modules/odgi/odgi.py
```diff
@@ -112,84 +120,70 @@ def parse_odgi_stats_report(self, f):
...
             "title": "Components",
-            "description": "Number of weakly connected components in the " "graph.",
+            "description": "Number of weakly connected components in the graph.",
             "scale": "Oranges",
             "format": "{:,.0f}",
         }
-        headers["A"] = {
-            "title": "A",
-            "description": "Number of adenine bases in the graph.",
+        headers["pct_gc"] = {
```
subwaystation, line 162: I understand the need to have less columns. But for debugging and comparing in detail across several different graphs, I would still like to keep a column for each base. Small example:

image

Here, we see that the total number of `N`s changed a lot (problem of PGGB), so I want to see this in the report. If I see percentages, it does not really help me, because the number of bases generally changed.

I will be a bad guy and change this back because of the above. If you don't agree, then the percentage wins, obviously.

subwaystation:
Thanks for the refactoring! https://github.com/ewels/MultiQC/commit/e1a76f0544f443503b8d22d1029d85ce9a45c0c0 is really cool. Great idea.

multiqc/modules/odgi/odgi.py
subwaystation: In the end, the exact number of bases helps to debug and validate the outcome of the PGGB pipeline.

multiqc/modules/odgi/odgi.py
subwaystation: I understand, but the numbers for different kind of versions of the graphs might be different by a huge factor. Or a tiny factor. And for validation and debugging, I wanted to see both.

multiqc/modules/odgi/odgi.py
apeltzer: Couldn't you have a separate table for the very same purpose that holds all your columns of interest? I get what Phil means that otherwise the general stats table gets very large when running more than 2-3 tools that MultiQC supports in a pipeline.

subwaystation: Ah, now I get it. I was not aware somehow that this general stats table incorporates statistics from all ran MultiQC tools.

We need an extra table, then. So for the general table I can imagine `length`, `nodes`, `edges`, `paths`.
All other statistics can go into an extra table. However, would it be fine, if all of these 4 basic statistics are there, again?
It would be better to have all of these in one table, to get the full picture of what is going on.

subwaystation: What would be your suggestion @apeltzer @ewels ?

multiqc/modules/odgi/odgi.py
subwaystation: Done.

multiqc/modules/odgi/odgi.py
apeltzer: Suggestion would be to only keep the 2-3 most interesting cols in the `general_stats` table and add all the other interesting bits to your own table https://multiqc.info/docs/#creating-a-table

apeltzer: Yeah was confusing for me too in the beginning but that's why we're all here to help ;-) 

If you think that `length, nodes, edges, paths` are the most interesting then have these in general stats and add the rest to your own custom table (and maybe have those four in your custom table too - not sure about that bit  @ewels knows more about it)

subwaystation: Will do an additional table. In a meeting now, will push something as soon as I can.

ewels: Yup, exactly right 👍🏻 Fine to keep the columns if helpful but split into a module-level table.

Don't _have_ to have columns in General Stats either - only if it's helpful to have these stats alongside other analysis. If it makes sense to keep them all in one table then can just be in a module section. If only running with this module it'll basically look the same as now.

subwaystation:
Alright, now that I don't have a general stats table anymore, the included pictures via `*_mqc.png` appear before my cool new ODGI table. Can I change that?
Furthermore, the following PNG is not added into the report:

cerevisiae pan fa 6ab2ce3 c28ecf8 da2dc96 smooth og lay draw_mqc
File name: `cerevisiae.pan.fa.6ab2ce3.c28ecf8.da2dc96.smooth.og.lay.draw_mqc.png`.


Any ideas why? Is MultiQC ignoring some key words in the file name? It was working before!

subwaystation:
When linting I get:

```
|            report | LINT: >custom_content/custom_content.py< HTML ID was not clean ('cerevisiae.pan' -> 'cerevisiae_pan') ## super(MultiqcModule, self).__init__(
|            report | LINT: >custom_content/custom_content.py< HTML ID was not clean ('cerevisiae.pan' -> 'cerevisiae_pan') ## self.add_section(name=section_name, anchor=c_id, description=section_description, plot=plot, content=content)
|            report | LINT: >custom_content/custom_content.py< HTML ID was a duplicate (cerevisiae_pan-1) ## self.add_section(name=section_name, anchor=c_id, description=section_description, plot=plot, content=content)
```

subwaystation:
I suspect that the sample name of the `*_mqc.png` files are determined from the start of the file name only? Did you change this?

subwaystation:
To reproduce, please run this branch with the following data.

[multiqc_odgi_LINT_error.zip](https://github.com/ewels/MultiQC/files/6754098/multiqc_odgi_LINT_error.zip)

subwaystation:
`multiqc --lint .`

ewels (approved):
Many thanks all for this, the module is looking great! Will come back with a working Custom Content implementation for the images in a second (as discussed in Slack, using images within a MultiQC module itself is a no-go).

ewels:
> Furthermore, the following PNG is not added into the report:

This is because the "sample name" generated for these inputs is based on the filename. The filename includes `.fa` which is matches a sample name cleaning string. This ends up being `cerevisiae pan` - for _both_ of the `_mqc.png` files. So they overwrite each other as with anything else to do with MultiQC and only one ends up in the report.

> File name: `cerevisiae.pan.fa.6ab2ce3.c28ecf8.da2dc96.smooth.og.lay.draw_mqc.png`.

If you remove the `.fa` from the filename, it works fine and both images show up.

ewels:
Ok, as is often the case with Custom Content, it took me a bit of tinkering with the core MultiQC source code plus some experimentation to get a nice output for the images here (and in this case even a bugfix: b3581e9abd9522624a8bb994578e3cdbc439d9b6) but I think I got it working quite nicely now.

You need a MultiQC config file like this: 

`multiqc_config.yaml`
```yaml
# Report section config for nice titles and descriptions
custom_data:
  odgi_graph_render:
    section_name: ODGI graph render
    description: This image shows a rendering of the assembled genome graph
  odgi_og_viz:
    section_name: ODGI graph visualisation
    description: The rendering shows the linkage between each component

# Custom search patterns to find the image outputs
sp:
  odgi_graph_render:
    fn: "*og.lay.draw.png"
  odgi_og_viz:
    fn: "*og.viz.png"
ignore_images: false

# Make the custom content stuff come after the ODGI module output
module_order:
  - odgi
  - custom_content

# Set the order that the custom content plots should come in
custom_content:
  order:
    - odgi_og_viz
    - odgi_graph_render
```

Then remove the `_mqc` suffix from the image files as that will just confuse matters (and it should work without the suffix if using the config above).

Using this in the files you sent me, I get the following report: [multiqc_report.zip](https://github.com/ewels/MultiQC/files/6754518/multiqc_report.zip), which I'm pretty happy with.

Note that at first I tried to get these images nested under the main ODGI report section, but then gave up. Because of how MutliQC / Custom Content code is structured, it turns out that this is super difficult. So I kept them as their own report sections for now. Makes very little difference in practice.

image

subwaystation:
> Ok, as is often the case with Custom Content, it took me a bit of tinkering with the core MultiQC source code plus some experimentation to get a nice output for the images here (and in this case even a bugfix: [b3581e9](https://github.com/ewels/MultiQC/commit/b3581e9abd9522624a8bb994578e3cdbc439d9b6)) but I think I got it working quite nicely now.
> 
> You need a MultiQC config file like this:
> 
> `multiqc_config.yaml`
> 
> ```yaml
> # Report section config for nice titles and descriptions
> custom_data:
>   odgi_graph_render:
>     section_name: ODGI graph render
>     description: This image shows a rendering of the assembled genome graph
>   odgi_og_viz:
>     section_name: ODGI graph visualisation
>     description: The rendering shows the linkage between each component
> 
> # Custom search patterns to find the image outputs
> sp:
>   odgi_graph_render:
>     fn: "*og.lay.draw.png"
>   odgi_og_viz:
>     fn: "*og.viz.png"
> ignore_images: false
> 
> # Make the custom content stuff come after the ODGI module output
> module_order:
>   - odgi
>   - custom_content
> 
> # Set the order that the custom content plots should come in
> custom_content:
>   order:
>     - odgi_og_viz
>     - odgi_graph_render
> ```
> 
> Then remove the `_mqc` suffix from the image files as that will just confuse matters (and it should work without the suffix if using the config above).
> 
> Using this in the files you sent me, I get the following report: [multiqc_report.zip](https://github.com/ewels/MultiQC/files/6754518/multiqc_report.zip), which I'm pretty happy with.
> 
> Note that at first I tried to get these images nested under the main ODGI report section, but then gave up. Because of how MutliQC / Custom Content code is structured, it turns out that this is super difficult. So I kept them as their own report sections for now. Makes very little difference in practice.
> 
> image

This is exactly what I was looking for, thank you very much @ewels !

subwaystation:
> Ok, as is often the case with Custom Content, it took me a bit of tinkering with the core MultiQC source code plus some experimentation to get a nice output for the images here (and in this case even a bugfix: [b3581e9](https://github.com/ewels/MultiQC/commit/b3581e9abd9522624a8bb994578e3cdbc439d9b6)) but I think I got it working quite nicely now.
> 
> You need a MultiQC config file like this:
> 
> `multiqc_config.yaml`
> 
> ```yaml
> # Report section config for nice titles and descriptions
> custom_data:
>   odgi_graph_render:
>     section_name: ODGI graph render
>     description: This image shows a rendering of the assembled genome graph
>   odgi_og_viz:
>     section_name: ODGI graph visualisation
>     description: The rendering shows the linkage between each component
> 
> # Custom search patterns to find the image outputs
> sp:
>   odgi_graph_render:
>     fn: "*og.lay.draw.png"
>   odgi_og_viz:
>     fn: "*og.viz.png"
> ignore_images: false
> 
> # Make the custom content stuff come after the ODGI module output
> module_order:
>   - odgi
>   - custom_content
> 
> # Set the order that the custom content plots should come in
> custom_content:
>   order:
>     - odgi_og_viz
>     - odgi_graph_render
> ```
> 
> Then remove the `_mqc` suffix from the image files as that will just confuse matters (and it should work without the suffix if using the config above).
> 
> Using this in the files you sent me, I get the following report: [multiqc_report.zip](https://github.com/ewels/MultiQC/files/6754518/multiqc_report.zip), which I'm pretty happy with.
> 
> Note that at first I tried to get these images nested under the main ODGI report section, but then gave up. Because of how MutliQC / Custom Content code is structured, it turns out that this is super difficult. So I kept them as their own report sections for now. Makes very little difference in practice.
> 
> image

Hi @ewels !
Sorry to come back to this so late, but using your suggested config, only one PNG for each of the sections `ODGI graph visualisation` and `ODGI graph render` is taken into account, even if I have several PNGs matching the patterns `"*og.lay.draw.png"` and `"*og.viz.png"`. Maybe this was your intention? 
How can I incorporate several of these PNGs into the report? Thanks!

subwaystation:
Ah, I need to got the `_mqc` way: https://github.com/ewels/MultiQC/blob/master/docs/custom_content.md#images.

ewels:
Should work without the `_mqc` suffix, as we set `sp:` manually. I guess it's simply that the custom content module doesn't handle more than one image per section, so it takes whichever it sees last. I'm fairly happy with this though, as I said I don't really want people embedding lots of images as it will quickly crash the browser 😅 

Using `_mqc` will probably fix it, because it will be found by the regular Custom Content code and so create a new report section for each file. But I guess none of the custom config YAML above will be applied to these sections. So no nicer section title heading and stuff. Is that right?

subwaystation:
Exactly, using the `_mqc` suffix, the sections are not neatly aligned and the name of the section simply reflects the name of the input file.

You are right, there is no nice section title heading and stuff. Is there a way to fix this? Is this something for the future of MultiQC?

ewels:
Having multiple images in a single report section? Not any time soon sorry as I have been actively avoiding this for years (I only added support for single images under duress 😅). Sorry for being a grumpy bugger on this issue.

You should be able to add a new section for each image though. Same concept as my example above, just doing a new section for each file. Not ideal but probably better than nothing.

In the distant future I hope to do a v2.0 rewrite of MultiQC where all modules will essentially create custom-content files (or something that looks a bit like them). Then they should get a lot more powerful. But if / when this happens is anyone's guess.

PR #1341, opened 2020-11-12
New module: Bustools (inspect)

Maarten-vd-Sande:

Added bustools (inspect) module. Bustools generates a convenient single-cell format. It can also generate QC statistics, and this should now be supported by MultiQC.

**I added a lot of stuff to the general statistics table, is it too much?**


Maarten-vd-Sande:
I'll make sure all boxes can be ticked today

apeltzer:
Ah, perfect for the scrnaseq pipeline 😄

ewels:
Many thanks for this @Maarten-vd-Sande - finally got around to taking a quick look at this. I fixed the merge conflicts and pulled in upstream changes, then got the tests to pass.

> I added a lot of stuff to the general statistics table, is it too much?

Yeah it is a bit on the large side. Might be worth making this into a bustools-specific table and then pulling out just one or two columns to put into General Stats - what do you think?

Maarten-vd-Sande:
@ewels thanks for picking this up, I kinda forgot about it :see_no_evil: 

I will change it into a separate table.

Maarten-vd-Sande:
@apeltzer @FloWuenne and @KevinMenden seeing as you guys are excited about this PR, any preference for what information stays at the general table, and what I will move to the bustools section? Any requests for plots? Otherwise I will keep the barplots as is, move everything to a table in the bustools section, and keep `Bus records`, `mean reads per barcode`, `distinct UMIs`, `2x depth`, and `% whitelisted barcodes`

see:
http://ocimum.science.ru.nl/bustools.html

apeltzer:
That sounds reasonable to me :-)

KevinMenden:
Yes that sounds good :-) thanks!

Maarten-vd-Sande:
@ewels this is now ready for review. Thanks a lot for fixing the merge conflicts before :pray:

ewels:
@Maarten-vd-Sande - looking good, thanks! I've just pushed a couple of commits to your PR where I fiddled with a few things, please take a look and check that you agree with everything I've done. In brief:

* Tried to use lambdas to make the numbers more manageable in the tables (see [docs](https://multiqc.info/docs/#number-base-multiplier))
* Added some `shared_key` attributes so that the background bars share a scale in the tables
* Made the General Stats table even more lean by hiding some columns by default
* Removed some config keys which were set to default values (eg. `"hidden": False`).
* Moved some section text from `description` into `helptext`
* Other minor text changes

Maarten-vd-Sande:
@ewels looks good!

ewels (approved):

ewels:
Thanks for this @Maarten-vd-Sande!

Maarten-vd-Sande:
@ewels thanks for multiqc!

Issue #692, opened 2018-02-27
Default renaming profile on report opening


rfenouil:
Dear Phil,

I find the bulk renaming options very useful to reformat ugly sample name to something more useful.
However when I open the report, the original (ugly) names are showing by default.
Would it be possible to add an option specifying which naming scheme to apply on opening ?

Thank you !

ewels:
Hi @rfenouil,

Revisiting an ancient issue here. I have never implemented your solution because (a) it would be technically difficult but more importantly, (b) I consider the original sample names to be the "true" references for each sample, and don't really want these hidden by default.

However, it's not all bad news. I just added a new core option `--replace-names` in abb8798b2d37c019e800b4043ae107a1dba4d0b0. Documentation is here: https://multiqc.info/docs/#sample-name-replacement

This does a hard swap of substrings / uses regular expressions _during report generation_. This means that the original ugly names never make it to the HTML report, and the replaced names are treated as the "true" sample name. It's 3 years late and it's not what you asked for, but hopefully this is useful for you! 😅 

This is now generally available in v1.11dev of MultiQC and will be released very soon. Let me know if you get a chance to play with it and have any suggestions.

Thanks,

Phil

PR #1342, opened 2020-11-16
New module: Lima

Redmar-van-den-Berg:
Lima, the PacBio barcode demultiplexer, is the standard tool to identify barcode sequences in PacBio single-molecule sequencing data. Starting in SMRT Link v5.1.0, it is the tool that powers the Demultiplex Barcodes GUI-based analysis application.

This module also adds an optional argument `--lima-barcodes` to specify a file that contains the mapping from the barcodes (which we can find in the Lima output) to sample names, which are not part of the Lima output. This way, the Lima statistics can be grouped together with the other per sample statistics, instead of being unique in the general statistics table.


multiqc/modules/lima/__init__.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import absolute_import
+
+import click
+
+from .lima import MultiqcModule
+
+lima_barcodes = click.option(
```
ewels, line 7: I'm not sure that we can do this sorry - this adds a new cli option for all MultiQC users for every run. I made it possible to add cli options like this purely for use in [MultiQC plugins](https://multiqc.info/docs/#multiqc-plugins). I don't really want to set a precent for module-specific cli flags in the core tool.

Is it possible to do this via a MultiQC config instead? For example:

```yaml
lima:
    barcodes: /path/to/barcodes.csv
```

Or even the barcode information itself in the config file directly if that makes more sense.

This method is less intrusive and follows existing MultiQC patterns. I appreciate that it's less user friendly.

ewels: Thinking more - as a compromise we could potentially have this cli option but only if it's enabled via a MultiQC config?

```yaml
lima:
    barcodes-cli-flag: True
```
(`False` by default).

This way, people running Lima regularly could set this in their [home directory config](https://multiqc.info/docs/#configuring-multiqc) for example, to have the cli flag for easy routine use. However it would not change the cli interface for the majority of MultiQC users who will never run this module.

I'm not sure whether this is worth the effort, but it could be a way forward if you think that the cli option is indispensable.

multiqc/modules/lima/lima.py
```diff
@@ -0,0 +1,365 @@
...
+
+        # Plot configuration
+        config = {
+            "id": "lima-filter-graph",
+            "title": "Lima: ZMW results",
+            "ylab": "Number of ZMWs",
+            "xlab": "Lima summary file",
+            "cpswitch": False,
+            "logswitch": True,
+            "stacking": None,
```
ewels, line 172: I generally say that we should not go for this approach - no stacking and using tabs for samples.

Firstly - it doesn't scale well. If there are samples with long names, or if there are a large number of samples then the report quickly becomes unusable (I try to plan every module as if someone will run 1000 samples).

Secondly - the main purpose of MultiQC is to allow people to get a fast summary of their project and spot any outlier samples that may need closer inspection. By using tabs, the end user has to click through every tab to see every sample and we lose this overview.

As a result, stacked bars showing all samples at once is the preferable display.

multiqc/modules/lima/__init__.py
ewels: Ok, now looking at a report and thinking a bit more about what this file is doing. I wonder if this functionality is basically already supported by core MultiQC? See [Bulk Sample Renaming](https://multiqc.info/docs/#bulk-sample-renaming) in the documentation.

This takes a tsv mapping original sample name (here: barcode) to arbitrary new identifiers (here: sample names) using the `--sample-names` cli flag.

The main difference is that it doesn't do a hard swap of identifier throughout the report - it introduces buttons at the top of the report where the end user can select which set to use. For an example of this in action, see [this report](https://multiqc.info/examples/ngi-rna/Test_NGI_Project_multiqc_report.html): 

image

ewels (commented):
Hi @Redmar-van-den-Berg,

Many thanks for writing this module! Great to get some more PacBio support into MultiQC 🚀 

I've done a quick first-pass over the code and report and have a few comments. Let me know what you think and I'll come back to it.

Cheers,

Phil

multiqc/modules/lima/__init__.py
Redmar-van-den-Berg: The reason I added this flag is because there is no way to know the Lima output names, since it depends on both the content of the barcode fasta file, and the content of the PacBio subreads file. I wanted to make sure that the same names are used everywhere, so that the General Statistics table doesn't have these alternating empty spots that you get when MultiQC derives different sample names from different log files.

The reason you cannot simple name the barcodes in the fasta file "sampel1, sample2" etc is that PacBio can also run in multi-barcode mode, so "Barcode1-Barcode1" is different from "Barcode1-Barcode2", which is different from "Barcode2-Barcode1".

I agree that adding custom command line flags for individual modules probably isn't the way to go, I'll have to look at some of the output files in more detail to see if there is a nice solution that doesn't require fundamental changes in MultiQC.

ewels: Oof, sounds messy. Yeah the sample naming problem sounds simple enough but is continually one of the trickiest things to get right in MultiQC...

I have a feeling that my proposed solution above with `--sample-names` won't avoid gappy General Stats tables unfortunately as it's post report generation 🤔 I wonder if I could add a generic version that runs during sample name cleanup. Basically the same thing that you're doing here, but for all modules and all sample names.

I have a vague memory of this which could mean that it's come up as a suggestion before (eg. this one from way back in 2018 https://github.com/ewels/MultiQC/issues/692) or could even mean that I've already implemented it 😆

ewels: Ok, I just added a new core option `--replace-names` in abb8798b2d37c019e800b4043ae107a1dba4d0b0

Documentation here: https://multiqc.info/docs/#sample-name-replacement

It's not exactly what you want here, as it's a simple find + replace instead of 3 columns with 2 barcodes. However it should be generally useful for users across all modules. Is this sufficient for what you need?

Redmar-van-den-Berg: `--replace-names` should work nicely, thanks. I'll have to do some pre-processing to go from the 3 column tsv file that I'm using now to the format that MultiQC expects, but that can easily be added to my pipeline.

However, this is going to be tricky for users to discover. Should we add some information on how to rename the Lima barcodes to the Lima section of the report itself? Since we know that nobody will get the expected sample names in their MultiQC report for Lima.

Redmar-van-den-Berg: Also, `--replace-names` is going to be really great for people writing pipelines. It can be quite tricky to organize your pipeline output files in such a way that MultiQC correctly parses out the sample names. Simply adding a `--replace-names` tsv file once your pipeline is done means you don't have to refactor your pipeline to get a nice MultiQC report.

ewels: Yes, my thoughts exactly 🎉 I was thinking earlier this morning that I should add one more config flag whereby any partial substring match would replace the entire sample name. eg. `123` -> `abc` would convert `sample_123` to `abc` instead of the default `sample_abc`. Should again help pipeline authors "clean up" sample names with less effort.

ewels: And yes, would be good to document this for Lima. Can do it both in the module docs and also in the section help text. Maybe worth putting a short note in the description telling people to click help to get a more in-depth help about how to do it, if that makes sense..

Could even be clever and check `config.sample_names_replace` in the module and show a warning above the plot if nothing is set using the `comment` argument in `add_section()`.

ewels:
Hi @Redmar-van-den-Berg,

I've done a bit more work on this module, again just applying my suggestions directly.

- Stripped out the command-line flag sample name replacement code
- Switched the bar plot to be stacked, as I did with your other module
- Switched the per-sample counts bar plots to a table.

The final point is very unusual for me, I'm usually doing the opposite in people's PRs 😅 However, these were two plots showing quite different things in tabs, and both had a single value. I figured that as they were distinct but related, it would be useful to be able to see the two together in context, and a table is the easiest way to achieve that.

I was tempted to put these columns in _General Statistics_ instead, but I don't know how the sample names will play alongside other analysis, so I left it as a module-level table for now.

I haven't put in any new documentation about the barcode stuff that we discussed above on this thread yet, as I understand that you maybe need to play with some new barcode file formats and things? But I have added a link in the table description to the MultiQC website. The website updates with changes on `master` (not waiting for releases), so we can add to this at any time.

I'm pretty happy with this module now, so I will merge so that it doesn't miss the upcoming v1.11 release. Please do look over my changes and test etc though, if you spot anything hopefully we can squeeze in some last minute PRs.

Phil

PR #1343, opened 2020-11-16
New module: WhatsHap

Redmar-van-den-Berg:
WhatsHap is a software for phasing genomic variants using DNA sequencing reads,     also called read-based phasing or haplotype assembly. It is especially suitable for long reads, but works also well with short reads.


marcelm:
Can you please document that this parses the results generated by the `whatshap stats` subcommand (and not `whatshap phase`, which I expected when I saw the title of this PR).

ewels:
Ah good spot @marcelm - yes if this tool can do many things, then it the module should be structured like Picard / RSeQC / Samtools etc where each subtool has its own file that is imported by the main module file. Then we can add support for other subtools one by one whilst keeping them relatively independent from one another.

Redmar-van-den-Berg:
I've added `whatshap stats` to the readme for the module. I currently have no plans to add support for the `whatshap phase` command.

ewels:
Thanks! It would still be nice to structure the code so that others can add other tools in the future though. We did this the other day for `sambamba` which also only has one subcommand supported currently: https://github.com/ewels/MultiQC/tree/master/multiqc/modules/sambamba

It shouldn't be too huge a change to the codebase hopefully. If you're unsure then I can have a stab at it when I get a chance.

Redmar-van-den-Berg:
@ewels Can this new module be merged? I think it makes more sense to restructure the code once it is needed, as part of the effort to add support for the different `whatshap` commands.

multiqc/modules/whatshap/whatshap.py
```diff
@@ -0,0 +1,319 @@
...
+                ),
+                (
+                    "phased",
+                    {
+                        "id": "phased",
+                        "title": "Phased Variants",
+                        "description": "The number of biallelic, heterozygous variants that "
+                        "are marked as phased in the input VCF. This is "
+                        "a subset of heterozygous_variants. Also, phased + "
+                        "unphased = heterozygous_variants.",
```
marcelm, line 191: I’ve just fixed this in the documentation:
```suggestion
                        "unphased + singletons = heterozygous_variants.",
```

docs/modules/whatshap.md
```diff
@@ -0,0 +1,12 @@
...
+URL: https://whatshap.readthedocs.io
+Description: >
+  WhatsHap is a software for phasing genomic variants using DNA sequencing reads,
+  also called read-based phasing or haplotype assembly. It is especially
+  suitable for long reads, but works also well with short reads.  ---
+---
+
+The WhatsHap module parses results generated by
+[WhatsHap](https://whatshap.readthedocs.io), and is currently restricted to the
+output from `WhatsHap stats`.
```
marcelm, line 12:
```suggestion
output from `whatshap stats --tsv`.
```

multiqc/modules/whatshap/whatshap.py
Redmar-van-den-Berg: According to the whatshap documentation, "phased + unphased = heterozygous_variants". Where did you find that singletons should be included as well?

marcelm: The commit where I changed it is at https://github.com/whatshap/whatshap/commit/60de9f1f0a8cb252e9077d0dbf03b2457ae808e9. The documentation did not reflect this change until a moment ago due to a build failure on Read the Docs, sorry about that.

I made that change because I was working on the `stats` subcommand for a different reason and realized that the description in the documentation was incorrect.

ewels:
Hi @Redmar-van-den-Berg @marcelm,

Apologies for being slow in reviewing, thank you for continuing to work on the code. I've just merged in the upstream changes so make sure you `git pull` before you do anything else.

Could I ask you to please double check the [test data](https://github.com/ewels/MultiQC_TestData/tree/master/data/modules/whatshap) that we have for the module? When I run on this I get the warning `bargraph | Tried to make bar plot, but had no data: multiqc_whatshap_phased_bp_plot` and the report looks pretty sparse:

image

I might do a bit of pedantic code style nitpicking and I'll do another code-read review now. I think that we're basically ready to merge, I just want to run on some more realistic data before I hit the button.

Also if missing the plot data is a real-life situation then it would be good to catch that and handle it without the empty report section.

Cheers,

Phil

multiqc/modules/whatshap/whatshap.py
```diff
@@ -0,0 +1,321 @@
...
+            # Insert the current line under chromosome
+            results[chromosome] = data
+
+        # If we were parsing an empty file, we return dummy data (basically all
+        # zero) so the sample statistics are consistent, which eases parsing
+        # later on. WhatsHap uses the fictional 'ALL' chromsome to store the
+        # summary for all results, so we re-use that to store all zero's
+        if not results:
+            results = dict()
+            results["ALL"] = defaultdict(int)
```
ewels, line 109: Why not just ignore the log file completely? Usually I would return `False` here, then catch it above and spit out a `log.error()` message and otherwise skip the file.

multiqc/modules/whatshap/whatshap.py
```diff
@@ -0,0 +1,321 @@
...
+            summary_field = "ALL"
+        return summary_field
+
+    def whatshap_add_general_stats(self):
+        """Add WhatsHap stats to the general statistics table"""
+
+        # Store the configuration for each column, this is also used to
+        # determine which columns to add to the general statistics table
+        # https://whatshap.readthedocs.io/en/latest/guide.html#the-tsv-statistics-format
+        general_stats_headers = OrderedDict(
```
ewels, line 137: This is a lot of columns for General Statistics, even if many are hidden by default. Is it worth moving most of these to a dedicated table (where they need not be hidden by default), leaving just one or two here which would be useful to compare to results from other tools?

Redmar-van-den-Berg:
> Could I ask you to please double check the [test data](https://github.com/ewels/MultiQC_TestData/tree/master/data/modules/whatshap) that we have for the module? When I run on this I get the warning `bargraph | Tried to make bar plot, but had no data: multiqc_whatshap_phased_bp_plot` and the report looks pretty sparse:
> 
> image

I had a look at this, and the module actually does pass the following data to the bargraph:
```{'Sample_1_3p': {'Phased Base Pairs': 0}, 'sample1_pre_opt0_5x': {'Phased Base Pairs': 0}, 'Sample_3_3p': {'Phased Base Pairs': 0}}```

For this module, MultiQC should allow a bar graph where all the data is zero, since that is what is present in the log files.

File: multiqc/modules/whatshap/whatshap.py (hunk 3 above)
Comment (Redmar-van-den-Berg) [reply multiqc/modules/whatshap/whatshap.py:109]: If I recall correctly, Whatshap can actually output a file that only contains the headers as a 'valid' output. If this is the case, we create the summary category 'ALL', and set that to all zero for the values we report (e.g. number of phased variants)

File: multiqc/modules/whatshap/whatshap.py (hunk 4 above)
Comment (Redmar-van-den-Berg) [reply multiqc/modules/whatshap/whatshap.py:137]: I'm not sure which columns would be most useful for users. I put them in the general statistics table in case people want to easily compare the results with other tools, for example to relate the number of phased variants to the % of duplicates from picard markduplicates.

File: multiqc/modules/whatshap/whatshap.py (hunk 3 above)
Comment (ewels) [reply multiqc/modules/whatshap/whatshap.py:109]: Ah ok, so this is not a completely empty file, but rather a file that only contains headers? Ok, then if it's a valid sample output then that makes sense 👍🏻

**Comment by ewels** [comment]
> For this module, MultiQC should allow a bar graph where all the data is zero, since that is what is present in the log files.

What I have done in other modules is show a warning above the plot saying that some samples have been ignored as they were all zero. That might be a cleaner solution.

I'm not entirely sure what to expect with this plot, so it might not even be necessary to skip the files if they're all zero _and_ there are other non-zero samples. In which case, we could just have the warning if _all_ samples are zero and skip the plot.

To do the warning, just prefix something like this to the result of the plotting function: `<div class="alert alert-warning">Warning! Warning! Danger Will Robinson!</div>` (or instead of the plotting function, if all samples are zero).

If possible, please add some more test data so that I can properly review the PR / do later bugfixes.

File: multiqc/modules/whatshap/whatshap.py (hunk 4 above)
Comment (Redmar-van-den-Berg) [reply multiqc/modules/whatshap/whatshap.py:137]: @marcelm Do you have any suggestions on which fields would be most useful for users to have in the General Statistics table? In our [PacBio pipeline](https://github.com/LUMC/PacBio-variantcalling), we are using `Input Variants`, `Unphased Variants`, `Phased Variants`, `Total Phased bp` and `Avg bp per Block`. Even though they are all correlated to how "good" the phasing is.

**Comment by Redmar-van-den-Berg** [comment]
I added some logic to skip the WhatsHap section if all samples are zero, and I also added a 'real' output file to the MultiQC_Testdata branch for whatshap.

File: multiqc/modules/whatshap/whatshap.py [hunk 5]
```diff
@@ -0,0 +1,321 @@
...
+                        "description": "The number of blocks that contain exactly one " "variant.",
+                        "hidden": True,
+                    },
+                ),
+                (
+                    "bp_per_block_sum",
+                    {
+                        "id": "bp_per_block_sum",
+                        "title": "Total Phased bp",
+                        "description": "Description of the distribution of non-singleton "
```
marcelm: That sentence (copied from the WhatsHap docs) was intended to refer to all the `bp_per_block_*` values. Since here the description now only refers to the `bp_per_block_sum` value, this should be: "The sum of the lengths of all non-singleton blocks, where the length of a block ..."

(It’s formatted a bit strangely in the docs; I’ll need to fix that.)

ewels:
> I also added a 'real' output file to the MultiQC_Testdata branch for whatshap.

I don't understand sorry, you mean on https://github.com/ewels/MultiQC_TestData somewhere?

multiqc/modules/whatshap/whatshap.py
```diff
@@ -95,7 +95,7 @@ def process_data(data):
             data = {field: value for field, value in zip(header, spline)}
             process_data(data)
             # Remove the sample and chromsome from the data
-            sample = self.clean_s_name(data.pop("sample"), logfile)
+            data.pop("sample")
```
ewels, line 98: Any reason to revert this change? I think I added the `self.clean_s_name()` function here because otherwise it's not used for the plot and sample renaming / `-d` `-s` etc flags don't work here..

multiqc/modules/whatshap/whatshap.py
marcelm: (Disclaimer: I don’t actually use `whatshap stats` so much.)

If you were to force me to pick only two columns, I would probably choose
- the percentage of phased variants, something like `phased / heterozygous_variants`
- and something describing the block sizes in terms of bp such as median bp per block (`bp_per_block_median`) or the block NG50.

multiqc/modules/whatshap/whatshap.py
Redmar-van-den-Berg: The problem was that the sample name was getting cleaned twice, once inside the loop and once outside. I've now move the cleaning to just before we return the sample name.

multiqc/modules/whatshap/whatshap.py
Redmar-van-den-Berg: I've updated the description, thanks.

Redmar-van-den-Berg:
I see now that the original test data was merged already, sorry. I've created a new pull request here: https://github.com/ewels/MultiQC_TestData/pull/225

multiqc/modules/whatshap/whatshap.py
Redmar-van-den-Berg: What about the following three columns for the general statistics table?
1. `Input Variants`, which at least for my data differs wildly between samples, so that is useful context to have.
2. `Phased Variants`
3. `bp_per_block_avg`. 
I'll also add an additional table to the WhatsHap section of the MultiQC report that holds all statistics. I'll hide the stats that are present in the general statistics table by default to keep the table clean

multiqc/modules/whatshap/whatshap.py
ewels: Ah sorry, that was my mistake when I added this then 👍🏻

ewels:
Looking good @Redmar-van-den-Berg! I just updated the table config to hide decimal places on nearly all columns.

I did wonder about adding base pair multipliers ([docs](https://multiqc.info/docs/#number-base-multiplier)) but I decided against it as I guess these phase blocks could be quite short fairly often, so maybe defaulting to Mbp is not a great idea.

I'll just wait for the updated tables now, but then I think we're good to merge 👍🏻

ewels:
To clarify (mostly for myself when checking this PR), just waiting for this: https://github.com/ewels/MultiQC/pull/1343#discussion_r796550297

multiqc/modules/whatshap/whatshap.py
marcelm: Not sure if the question was directed at me. I think it is your call as you have more experience looking at these statistics.

My thought was that it would be relevant to see how many of the variants that can in principle be phased (heterozygous variants) were actually phased. If there’s only `Phased variants` and `Input variants`, then I cannot say this because `Input variants` includes the homozygous variants. And my reason for suggesting a percentage was that I tend to estimate the ratio in my head anyway when I see two absolute numbers of which one describes a subset of the other.
On the other hand, if the coverage is sufficiently high to cover most of the heterozygous variants, WhatsHap will phase almost all of them and the ratio `phased / heterozygous_variants` is probably close to 100% most of the time and won’t be helpful.

One could also argue that `whatshap stats` columns should only show information relevant to phasing and therefore the variant count should exclude the homozygous variants. But it is of course convenient to see the full variant count.

`bp_per_block_avg` versus `_median` is probably a matter of taste.

As I said, I’m not the one who has to look at this in practice, so I’m fine with whichever way you think is best.

pontushojer: Hi, I just though I would chime in here since I do use `whatshap stats` and think this would be a really nice addition to MultiQC. 

For me the  `phased / heterozygous_variants` ratio (i.e. % phased) would be most valuable for comparison. I also use`phased_snvs / heterozygous_snvs` to get the ratio for SNVs which is usually higher in my case. 

As for block size comparison I mostly rely on the `NG50` measure. 

For my part I don't mind having a lot of columns in the General Statistics table, I just have a config which selects the ones I am interested in. It would still nice to have a separate table for all `whatshap stats` statistics but maybe this could be a separate PR.

Redmar-van-den-Berg: Thanks for the feedback @pontushojer and @marcelm! I opted to include `% Phased Variants` and `Avg bp per Block` in the general statistics.  `NG50` is hidden by default, since it is not always set in the .tsv output file. I also added all statistics to a separate statistics table in the WhatsHap section of the report.

ewels (approved):
Great stuff 👍🏻 Thanks both!

marcelm:
Thanks all! I’ll update the WhatsHap docs to recommend using MultiQC :smiley:

Redmar-van-den-Berg:
Awesome, thanks for merging!

Issue #1074, opened 2019-12-04
Module for Nanoplot


FranBonath:
**Tool Details**
 - Name of tool: Nanoplot
 - Tool homepage: https://github.com/wdecoster/NanoPlot
 - Tool description: Plotting tool for long read sequencing data and alignments
 - Data suitable for MultiQC plot(s): plots and histograms
 - Most interesting data for General Stats table: read length N50, mean read length and mean read quality, number of reads

**Tool output:**
 - Log filename pattern: `NanoStats.txt`
 - File contents: [NanoStats.txt](https://github.com/ewels/MultiQC/files/3922479/NanoStats.txt)

ewels:
Nice! And you are writing this module, right?

FranBonath:
Yes, with time...

hoelzer:
Hi, thanks for the effort to integrate Nanoplot into MultiQC! It would be great to have this, for example, also to summarize multiple Nanoplot runs for different barcodes.

RxLoutre:
Hello everyone ! I might be late but I would still be very interested by Nanoplot being added in multi_qc :) . Is there any progress ? Thank you for your work ! If help is needed, maybe I could give a hand :)

ewels:
Hi all,

NanoStat is now supported thanks to the new module written by @kpalin, just merged in #1344 🎉 It will be available in the upcoming v1.11 MultiQC release.

@FranBonath - in your original issue you named NanoPlot, but the module is NanoStat. However, the file you attached looks identical, so I'm assuming that running NanoPlot also runs NanoStat internally and generates the same file. Hopefully this is clear enough.

Note that some of the plots in NanoPlot could be nice to reproduce in MultiQC, for example line graphs showing read length histograms per sample especially (also read qualities, etc). I'm not sure if NanoPlot dumps any of these into a text file that we can add support for, if it does then that would be great 👍🏻  If anyone reading this has such a file, please create a new issue for it and we can look into adding support.

Thanks all,

Phil

MarieLataretu:
Hi all,

I'm very happy to see a NanoStats module - thanks to all the involved people!

---

As we integrated NanoPlot into our pipeline, I tried to include the `NanoStats.txt` file (produced by NanoPlot) in the MultiQC report.
[...]

PR #1344, opened 2020-11-16
New module for NanoStat 

kpalin:
New module to process nanostat output

Using test data at (https://github.com/ewels/MultiQC_TestData/pull/176)

 Might close ewels/MultiQC#1074

    * _so-so_
    * _only the general summary table_


ewels:
Many thanks for writing this @kpalin - it's looking good! You have a few functions / patterns in there that I should really use more often (eg. I love your use of `setdefault()` - I always forget this function!)

I've just pushed a few nitpick tidying changes but it's generally looking good. The only thing that really needs changing is the General Stats table at the top. You left the checkboxes on the PR unticked, and they need ticking 😅  I'll see if I can whip together some changes to address this now..

ewels:
Ok, I'm happy with this now - report looking good and hopefully you agree with all of my modifications 🤞🏻 

Only thing I'm not sure about is whether to add one or two columns back to the General Stats table, for comparison with other modules. For example, `# Reads` could be useful. But I'm going to leave it for now I think.

I am planning to release MultiQC imminently, so I will merge this right away. It should be available in the v1.11 release. Please do look over my changes and give it a test though, shout if you spot anything wrong or that you disagree with and we can follow up.

Thanks again for writing this @kpalin ! Great to get it supported 😁 

Phil

PR #1352, opened 2020-12-04
mosdepth: display contig name in coverage per contig plot 

pontushojer:

Tooltip in _Coverage per contig_ plot showed x-index which is much less useful than the actual contig name. Y-value formatted as before.


nservant:
Hi,
I just find a small issue in the Average Coverage per Contig plot in MultiQC with the labels display on hover.
The chromosome name is in 'X', and the mean depth in '%'.
For instance, chr1 at 110X is labelled "0X: 110%"
Just wondering if this is fixed in the pull request too ?
Thanks
Nicolas

pontushojer:
Hi @nservant! I have not seen exact the same labels as you mention for the Average Coverage per Contig plot. For me labels were (using the same example as you): `0: 110X`. With this PR the `0` is replaced for the actual contig-name making the label: `chr1: 110X` in this case.

nservant:
ok. Thanks @pontushojer 
I will test the new version and see how it goes. Thanks

ewels:
Thanks both! Very minor PR so have merged quickly without much testing 👀 

@nservant if you could install the dev version and confirm that this change is also what you're looking for, that would be brilliant. Thanks!

PR #1357, opened 2020-12-16
New module: pbmarkdup

Redmar-van-den-Berg:
pbmarkdup takes one or multiple sequencing chips of an amplified PacBio libray as HiFi reads and marks or removes duplicates.

This module adds **% Unique Molecules** and **%Duplicate Reads** to the General Statistics table.

This module assumes that pbmarkdup has been run for every sample, and that the pbmarkup log file is named `sample.pbmarkdup.log`.


docs/README.md
```diff
@@ -31,6 +31,8 @@ MultiQC Modules:
     leeHom: modules/leehom.md
     Lima: modules/lima.md
     minionqc: modules/minionqc.md
+    pbmarkdup: modules/pbmarkdup.md
+    pycoQC: modules/pycoqc.md
```
ewels, line 35:
```suggestion
```

Redmar-van-den-Berg:
@ewels I've fixed the last errors for this new module, can you merge it?

docs/modules/pbmarkdup.md
```diff
@@ -0,0 +1,15 @@
...
+  HiFi reads and marks or removes duplicates.
+---
+
+The pbmarkdup module parses results generated by
+[pbmarkdup](https://github.com/PacificBiosciences/pbmarkdup), and adds the
+**% Unique Molecules** and **%Duplicate Reads** to the General Statistics
+table.
+
+This module assumes that pbmarkdup has been run for every sample, and that
+the pbmarkup log file is named `sample.pbmarkdup.log`.
```
ewels, line 15: Is this a fair assumption? ie. Is it the default from the tool or something? If the filename is always determined by the user then it's usually better to search file contents instead. If possible I try not to force users to run upstream tools in a specific manner.

ewels: Looking at the [docs](https://github.com/PacificBiosciences/pbmarkdup#can-i-print-summary-information) I think that this is probably a stdout thing. So maybe better to search for file content instead.

multiqc/utils/search_patterns.yaml
```diff
@@ -469,6 +469,8 @@ somalier/pairs:
   fn: "*.pairs.tsv"
   contents: "hom_concordance"
   num_lines: 5
+pbmarkdup:
+  fn: "*.pbmarkdup.log"
```
ewels, line 473: See other comment above, better to search file contents if we don't have a standard filename from the tool.

```suggestion
  contents_re: "LIBRARY\s+READS\s+UNIQUE MOLECULES\s+DUPLICATE READS"
  num_lines: 5
```

multiqc/modules/pbmarkdup/pbmarkdup.py
```diff
@@ -0,0 +1,148 @@
...
+        For now, we just assume that the data fields are always in the same
+        order, without any missing values.
+        """
+
+        file_content = logfile["f"]
+
+        # The number of spaces in the header can vary, based on the length of
+        # the name of the library. We therefore need a regex to match the
+        # header.
+        header_pattern = "LIBRARY *READS *UNIQUE MOLECULES *DUPLICATE READS"
```
ewels, line 70: I can't imagine  this would cause a problem, but on principle we want 1 or more spaces, not 0 or more 😉 

```suggestion
        header_pattern = "LIBRARY +READS +UNIQUE MOLECULES +DUPLICATE READS"
```

multiqc/modules/pbmarkdup/pbmarkdup.py
```diff
@@ -0,0 +1,148 @@
...
+        pattern = re.compile(header_pattern)
+
+        # The file header
+        file_header = next(file_content).strip()
+
+        # Error if the header doesn't match the expected pattern
+        if not re.match(pattern, file_header):
+            raise RuntimeError(f'Unknown header: "{header}"')
```
ewels, line 78: Please don't raise, just log and skip

```suggestion
            log.error(f"Can't parse file '{logfile["fn"]}', unknown header: '{header}'")
            return False
```

ewels: Ugh, quotes mess there but you get the idea

multiqc/modules/pbmarkdup/pbmarkdup.py
```diff
@@ -0,0 +1,148 @@
...
+                 """
+            ),
+            # Can't find a DOI // doi=
+        )
+
+        self.pbmarkdup = dict()
+
+        for logfile in self.find_log_files("pbmarkdup", filehandles=True):
+            self.pbmarkdup[logfile["s_name"]] = self.parse_logfile(logfile)
+            self.add_data_source(logfile)
```
ewels, line 39: If we have a `return False` in the code for dodgy files (see comment below):

```suggestion
            data = self.parse_logfile(logfile)
            if data:
                self.pbmarkdup[logfile["s_name"]] = data
                self.add_data_source(logfile)
```

multiqc/modules/pbmarkdup/pbmarkdup.py
```diff
@@ -0,0 +1,148 @@
...
+            # Lines which start with --- denote separators in the file, and do
+            # not need to be parsed
+            if line.startswith("-"):
+                continue
+
+            # Not very nice, we assume that all fields are always present
+            name, reads, unique_mol_count, unique_mol_perc, duplicate_count, duplicate_perc = line.split()
+
+            # We are only interested in the counts, not the percentages
+            data[name] = {
```
ewels, line 94:
```suggestion
            s_name = self.clean_s_name(name, logfile)
            data[s_name] = {
```

ewels (commented):
Taking a look now @Redmar-van-den-Berg, thanks for this!

Main thought is: do we need both columns? They're basically showing the same data right, the numbers always sum to 100%? My gut feeling is to just show _% Duplicates_, and maybe as a second (hidden?) column have the number of duplicate reads. Potentially also the total read count?

Few other more minor comments, see review.

docs/modules/pbmarkdup.md
ewels, line 15: Removed now that you updated the search criteria.

```suggestion
This module assumes that pbmarkdup has been run for every sample.
```

I don't totally understand what you mean by being run for every sample. It should handle summaries that contain more than one sample in them, right?

Redmar-van-den-Berg: I've removed this from the documentation, as we are now using the file header

multiqc/modules/pbmarkdup/pbmarkdup.py
Redmar-van-den-Berg: Added

multiqc/modules/pbmarkdup/pbmarkdup.py
Redmar-van-den-Berg: Added

docs/modules/pbmarkdup.md
Redmar-van-den-Berg: I've removed this line, since I'm not sure why I added it. pbmarkdup runs on single samples.

Redmar-van-den-Berg:
It looks like the tests are failing due to some unrelated module.

ewels:
Seems to be this module, no?

```
╭─────────────── Oops! The 'pbmarkdup' MultiQC module broke... ────────────────╮
84
│Please copy this log and report it at │
85
│
https://github.com/ewels/MultiQC/issues
│
86
│Please attach a file that triggers the error. The last file found was: │
87
│test_data/data/modules/pbmarkdup/1.0/Sample_3_3p.pbmarkdup.log│
88
││
89
│Traceback(mostrecentcalllast):│
90
│File"/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/m│
91
│output=mod()│
92
│File"/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/m│
93
│self.pbmarkdup_add_general_stats()│
94
│File"/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/m│
95
│total=sample_stats["TOTAL"]│
96
│KeyError:'TOTAL'│
97
││
98
╰─���────────────────────────────────────────────────────────────────────────────╯
```

ewels:
Can't replicate locally, re-running the CI tests in case it was a random intermittent network error etc..

ewels:
Ok I take it back, replicated locally when using the flag `--fn_as_s_name`. This must be an error that comes up due to sample name cleaning, fits with the `KeyError`..

ewels (approved):
Looking great! Thanks for the work on this.

I still wonder if having absolute read counts in the table instead of the _% Duplicates_ would be more useful (see https://github.com/ewels/MultiQC/pull/1357#pullrequestreview-867720606) but I'm not too fussed and am keen to get this merged in. Happy to accept a fast / minor follow-on PR if you're interested in changing that.

Thanks again for your patience and speed in responding to my reviews!

Phil

PR #1358, opened 2020-12-17
Adding module for ARTIC (sars-cov-2 consensus genome pipeline)

will-rowe:
This PR adds a module for [ARTIC](https://github.com/artic-network/fieldbioinformatics) - a viral consensus genome pipeline widely used for SARS-CoV-2 work. It plots amplicon coverage and reports amplicons with dropout, as well as flagging variants with low support in overlap regions. 

 - [x] There is example tool output [here](https://github.com/ewels/MultiQC_TestData/pull/181)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,8 @@
   * Reference-free QC of Hi-C sequencing data
 * [**Sentieon**](https://www.sentieon.com/products/)
   * Submodules added to catch Picard-based QC metrics files
-
+* [**ARTIC**](https://github.com/artic-network/fieldbioinformatics)
+  * Viral consensus genome pipeline for nanopore amplicon sequencing data
 
 #### Module updates
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -51,6 +51,7 @@ MultiQC Modules:
     DRAGEN: modules/dragen.md
     MALT: modules/malt.md
   Post-alignment:
+    ARTIC: modules/artic.md
     Bamtools: modules/bamtools.md
     Bcftools: modules/bcftools.md
     biobambam2: modules/biobambam2.md
--- a/docs/modules/artic.md
+++ b/docs/modules/artic.md
@@ -1,0 +1,9 @@
+---
+Name: ARTIC
+URL: https://github.com/artic-network/fieldbioinformatics
+Description: A bioinformatics pipeline for working with virus sequencing data sequenced with nanopore.
+---
+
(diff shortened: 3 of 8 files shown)


will-rowe:
Hi @ewels 

Just picking this up again. I've had a go at the Custom Content suggestion you made over at https://github.com/ewels/MultiQC_TestData/pull/181 - I'm happy to do this instead as it is less overhead for all involved? I've added a couple of *_mqc.json outputs to the pipeline and the plots are working nicely this way.

Happy to close this PR if you are

ewels:
Hi @will-rowe,

Sure - that's fine by me. Generally I think it's better when MultiQC is executed within the pipeline alongside those outputs anyway. You don't have to wait for me to get around to reviewing and releasing any code if nothing else!

What about your comment about old versions of the pipeline though? https://github.com/ewels/MultiQC_TestData/pull/181#issuecomment-754022950

> The reason I went for a plugin is so that we can run MultiQC on outputs from old versions of the pipeline (as the summary file being used has been around and stable for a while). The pipeline itself has been used as part of COG-UK for around a year and we are unlikely to re-process old samples with new versions of the pipeline. There is also a big lag-time for pipeline updates and adoption...!

I'll leave the decision up to you anyway. If this PR is still here when I get to it on the list then I will review and merge it 😉 

Phil

will-rowe:
Oh yeah - there was a bit I forgot to mention which explains why it turns out the backward compatibility thing isn't realistic.

I was trying this PR on some "real" data and the report files being used are too big and are often excluded from MultiQC. So not sure how I can make this backward compatible. I think it makes most sense to do the custom output (which version 1.2.1 of artic now does) and will work with MultiQC out the box.

Thanks for the great software and your help with me getting the artic pipeline up to scratch for reporting!

ewels:
Ah, that's quite easy to modify in the config: https://multiqc.info/docs/#big-log-files

> You can configure the threshold and parse your files by changing the `log_filesize_limit` config option. For example, to parse files up to 2GB in size, add the following to your MultiQC config file:

```yaml
log_filesize_limit: 2000000000
```

ewels:
But agreed - if the files are that big then that makes it a hassle for the new MultiQC module - so custom config is probably the way to go anyway. You can always keep the code you've written here as a 3rd party plugin _([docs](https://multiqc.info/docs/#core-modules--plugins), [example plugin](https://github.com/MultiQC/example-plugin) - plugins can also change config defaults like the one above)_ to run for old reports and use custom content going forward. Then it's the best of both worlds 😄

PR #1104, opened 2020-02-07
New module: ngsderive

claymcleod:
New module: **ngsderive**

 - [x] There is example tool output for tools in the https://github.com/ewels/MultiQC_TestData repository
     - Link to PR: https://github.com/ewels/MultiQC_TestData/pull/137/files
 - [x] Code is tested and works locally (including with `--lint` flag)
 - [x] `CHANGELOG.md` is updated
 - [x] `README.md` is updated
 - [x] `docs/README.md` is updated with link to below
 - [x] `docs/modulename.md` is created
 - [x] Everything that can be represented with a plot instead of a table is a plot
 - [x] Report sections have a description and help text (with `self.add_section`)
 - [x] There aren't any huge tables with > 6 columns (explain reasoning if so)
     - I did add a few columns to the summary table, but that is mainly because the output of `ngsderive` is categorical in nature.
 - [x] Each table column has a different colour scale to its neighbour, which relates to the data (eg. if high numbers are bad, they're red)
 - [x] Module does not do any significant computational work

PR #1360, opened 2020-12-30
Ngsderive

ewels:
Hi @claymcleod and @a-frantz

I started tweaking the code in https://github.com/ewels/MultiQC/pull/1104 for the review but can't push the changes to your fork as normal. So I've created a branch and a new PR here to continue the review.

Phil


ewels:
You can see the tweaks that I made to the code in https://github.com/ewels/MultiQC/pull/1360/commits/eaafe70edc316fe15935f82ff01532e1e8db73d5 - I don't think that any of these should have a significant impact on the functionality of the code. They're mostly simplification and making text more concise.

ewels:
Ok, tests passing and PR now up to date. So should be good to go for the review properly now.

I ran the test data locally and it's working fine. Main question for @claymcleod and @a-frantz - did you use heat maps for a specific reason? I'd prefer to use line graphs here I think. They are typically easier and more accurate to interpret I think, especially for a lot of samples. They are also easier to render for the browser and are less data to embed in the report.

It's late here now so I need to stop, but I will come back to this and merge ASAP. Unless you would prefer me not to, I will switch out the heat maps for line graphs.

Phil

a-frantz:
@ewels all the changes you made look good to me. Switching to line graphs makes sense, but @claymcleod made the initial call for heatmaps. Not sure if he has a particular reason to keep it that way. Unless he thinks otherwise, I'd say go ahead 👍

claymcleod:
@a-frantz would you be able to post a picture here of the heatmap sections? It's been a minute since I've worked with `ngsderive` since you've taken over, so I can't recall why I made that decision and if I felt it was important.

a-frantz:
Screen Shot 2020-12-30 at 9 10 30 AM
Screen Shot 2020-12-30 at 9 09 52 AM
Screen Shot 2020-12-30 at 9 10 45 AM
Strandedness uses a bargraph. I agree with @ewels that the heatmaps for instrument are hard to parse, and a linegraph makes more sense for read length, particularly given the higher rendering burden.

claymcleod:
Okay, thanks.

I think that, for the read length, there is a good case to be made to make it a line graph. This would emulate the choices that have already been made on how to represent similar data (e.g. https://multiqc.info/examples/rna-seq/multiqc_report.html#cutadapt_trimmed_sequences).

On the other hand, for instrument, I don't think that a line graph would be the best way to represent this data. It is categorical in nature, so projecting it across a numerically increasing axis is confusing to me.

The chart that is closest to this in the examples appears to be something like this: https://multiqc.info/examples/rna-seq/multiqc_report.html#fastqc_status_checks. The main goal of this graph is to be able to quickly scan and see which samples within a cohort use the same instrument, so my feeling that that using a heatmap is still the best to achieve that goal. 

That said, I'm okay to be overruled if you still feel any other graph is better @ewels and/or @a-frantz.

a-frantz:
I agree that a line graph doesn't make sense for instruments, but I'm not a fan of the current heatmap implementation. I think the color scale should be based on `Confidence`: `red: low` `yellow: medium` `green: high` `white: N/A`. I think a heatmap is the most straightforward way to write that up, but wouldn't be the most efficient method considering we're only dealing with 4 possible values for each cell of the heatmap.

Also while looking over the current code I think there's a bug in the instrument heatmap generation, `ngsderive.py` line 209:
```# first pass through, start to formulate heatmap data
 for s, d in data.items():
        samples.append(s)
        data = d.get("instrument").split("/")
        samples_data.append(data)
        instruments.update(data)
```
`data` is being iterated over and overwritten.

ewels:
Ah is it not continuous data? If there are four categorical labels then could we just have a table with formatting? Either background colour, or conditional formatting labels, or both.. https://multiqc.info/docs/#conditional-formatting (might benefit from some core code refactoring to make it easier for modules to use this..)

a-frantz:
I refactored to use a table instead, but can't push here. It still needs to be color coded depending on cell value though. The way to do that would be through conditional formatting, right? From what I can see in the docs, at the module level I can color code columns, but not individual cells?

ewels:
Nice! If you could open a new PR against the branch `ngsderive` I'll merge and the changes will appear on this PR.

Yes I suspect that the formatting for the table might be tricky. I'll take a look as it's probably an opportunity to make the core table functionality better.

a-frantz:
> Nice! If you could open a new PR against the branch `ngsderive` I'll merge and the changes will appear on this PR.
> 
> Yes I suspect that the formatting for the table might be tricky. I'll take a look as it's probably an opportunity to make the core table functionality better.

Refactor opened in #1362. Asked some questions in that PR, not sure if they should be addressed here or there.

ewels:
Sounds good - I'm still on vacation for the rest of this week but I'll try to take a look either in an evening again or next week some time.

Phil

ewels:
copy of https://github.com/ewels/MultiQC/pull/1362#issue-549359572 :

> @ewels One remaining question about this is that there's no straightforward way to show `basis` in the new table format (that I can think of). I think it conveys the important information well (once color coding is added), instrument(s) and confidence.
>
> @claymcleod are we ok not displaying `basis` in this section? It can still be revealed as a default-hidden column in the general stats table. I think that's backwards from how other modules work.... In that more detailed information is present in the general stats and this dedicated `instrument` section is more of an "at-a-glance" look (because to make this categorical data "at-a-glance"-able it requires more visual space than is available in the general table).

ewels:
@a-frantz - I merged #1362 so your changes are now in this PR. Looks good! I think it's a good improvement in that I actually understand the results for the instrument guess now - before I just had those yellow boxes with `1.0` on hover which was a bit meaningless to me. The table with the text `low confidence` looks great to me.

I don't really understand why the `Instrument: Basis` column can't be moved from General Statistics to the new table?

Some other quick thoughts:

* I'd prefer to remove the `% Forward Strandedness` and `% Reverse Strandedness` columns from General Statistics, as they are a direct duplication of the bar graph below. Is that ok?
* `Read Length: % Supporting` seems like a good candidate to be hidden by default, is that ok?

That PR to the TestData repo with more reports would be great if possible @a-frantz - it's a bit of a guessing game as to what typical results could look like currently, always nice to have examples that cover a full range of possible outputs.

ewels:
Ok, I've added new core functionality to allow both background colours for categorical (string) table cell data and also the conditional formatting thing. The table are now fully formatted:

image

image

ewels:
I've also rewritten the read length heatmap as a line graph. I've given it two dataset buttons so that the data can be viewed both as percentages but also as read counts:

image

Finally I played around a bit with which table columns are hidden by default.

Example report output here: [multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/5802861/multiqc_report.html.zip)

Would be great if you could check over my changes to make sure that you're happy. I think that this is all then, and we can merge...?

a-frantz:
The updates look great! Just submitted a PR on the test data repo.

I think you're right about removing `% Forward Strandedness` and `% Reverse Strandedness` from General Statistics. And also correct that there's no reason not to include `basis` in the instrument table. I think after those two little changes, we could go ahead and merge!

Here's how the extra entry changes the table:
Screen Shot 2021-01-12 at 11 38 53 AM

@claymcleod any reservations?

ewels:
Done, report here: [multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/5804107/multiqc_report.html.zip)

Looking great now, thanks both! I'm happy with the two new formatting options for tables for all MultiQC developers too, I think that they are good additions - and not too difficult to add in the end :)

Phil

claymcleod:
Thanks again for working with us on this @ewels

ewels:
No problem - sorry it took me so long to get to! Thanks for being so responsive when I did finally get back to it..

PR #1362, opened 2021-01-05
Ngsderive table refactor

a-frantz:
continuation of #1360

@ewels One remaining question about this is that there's no straightforward way to show `basis` in the new table format (that I can think of). I think it conveys the important information well (once color coding is added), instrument(s) and confidence.

@claymcleod are we ok not displaying `basis` in this section? It can still be revealed as a default-hidden column in the general stats table. I think that's backwards from how other modules work.... In that more detailed information is present in the general stats and this dedicated `instrument` section is more of an "at-a-glance" look (because to make this categorical data "at-a-glance"-able it requires more visual space than is available in the general table).

I'll attach a screenshot momentarily.

### DIFF
--- a/multiqc/modules/ngsderive/ngsderive.py
+++ b/multiqc/modules/ngsderive/ngsderive.py
@@ -12,7 +12,7 @@ import re
 from collections import OrderedDict
 
 from multiqc.modules.base_module import BaseMultiqcModule
-from multiqc.plots import bargraph, heatmap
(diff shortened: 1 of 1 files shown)


a-frantz:
Screen Shot 2021-01-05 at 1 18 07 PM
Screen Shot 2021-01-05 at 1 18 26 PM
I changed the test data locally to show what it would look like with multiple samples with differing instruments. Haven't made a PR on the test data repo for that yet.

Issue #625, opened 2017-11-09
Multiqc - conda issue 


jweger1988:
Hello All!

I'm running anaconda on on Ubuntu 14.04. I have the python 3.6 version and I am getting an error message when I try to run MultiQC - I have never had this problem before on an older version of ubuntu. If anyone can help me with this problem it would be greatly appreciated. I've tried with both python environments version 2.7 and 3.6

Here is the error message 

```
[INFO   ]         multiqc : This is MultiQC v1.3
[INFO   ]         multiqc : Template    : default
[INFO   ]         multiqc : Searching '.'
Searching 5 files..  [####################################]  100%
[INFO   ]         bowtie2 : Found 2 reports
[INFO   ]         multiqc : Compressing plot data
[WARNING]         multiqc : Previous MultiQC output found! Adjusting filenames..
[WARNING]         multiqc : Use -f or --force to overwrite existing reports instead
[INFO   ]         multiqc : Report      : multiqc_report_4.html
[INFO   ]         multiqc : Data        : multiqc_data_4
```
```python
Traceback (most recent call last):
  File "/home/vignuzzilab/anaconda2/envs/py3-2/bin/multiqc", line 724, in <module>
    multiqc()
  File "/home/vignuzzilab/anaconda2/envs/py3-2/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/vignuzzilab/anaconda2/envs/py3-2/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/vignuzzilab/anaconda2/envs/py3-2/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/vignuzzilab/anaconda2/envs/py3-2/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/vignuzzilab/anaconda2/envs/py3-2/bin/multiqc", line 572, in multiqc
    shutil.move(fn, config.data_dir)
  File "/home/vignuzzilab/anaconda2/envs/py3-2/lib/python2.7/shutil.py", line 316, in move
    copy2(src, real_dst)
  File "/home/vignuzzilab/anaconda2/envs/py3-2/lib/python2.7/shutil.py", line 145, in copy2
    copystat(src, dst)
[...]

PR #1368, opened 2021-01-16
if copy_tree() fails, try again without preserving attributes

pmenzel:
If `copy_tree()` fails, then try to re-run it without preserving the file timestamps and modes.

This can help on mounted network shares where Python's `os.utime()` fails (e.g. #625)

CHANGELOG is not updated in this commit.


ewels:
Interesting! Given that the copy operation here typically happens within a few seconds of the original write (or sooner) I wonder if we just set `preserve_times=0, preserve_mode=0` directly in the first call? Can you think of any downsides to doing this?

pmenzel:
True! The timestamps of the files in the temp directory do not need to be preserved anyway.
I don't see a downside.

pmenzel:
Code looks much cleaner too like this. :)

ewels:
Brilliant - thanks!

PR #1370, opened 2021-02-05
feat: add `encoding` and `junction-annotation` ngsderive modules

a-frantz:
This is essentially a continuation of the previous `ngsderive` PR. We've added two modules since that PR, don't think anything new needs to go into the CHANGELOG/docs since this is still the same release.

One thing I can't figure out is how to switch datasets for a barplot. I followed the instructions in the docs by adding a `data_labels` entry to the config and supplying a `categories` list, but I don't see the button to flip the `junction-annotation` graph from "junctions" to "spliced reads". Can you point me towards what I'm missing?

### DIFF
--- a/multiqc/utils/search_patterns.yaml
+++ b/multiqc/utils/search_patterns.yaml
@@ -396,6 +396,12 @@ ngsderive/instrument:
 ngsderive/readlen:
   contents_re: "(.*)File(.*)Evidence(.*)MajorityPctDetected(.*)ConsensusReadLength"
   num_lines: 1
+ngsderive/encoding:
+  contents_re: "(.*)File(.*)Evidence(.*)ProbableEncoding"
+  num_lines: 1
+ngsderive/junction_annotation:
+  contents_re: "(.*)File(.*)total_junctions(.*)total_splice_events(.*)known_junctions(.*)partial_novel_junctions(.*)complete_novel_junctions(.*)known_spliced_reads(.*)partial_novel_spliced_reads(.*)complete_novel_spliced_reads"
+  num_lines: 1
 peddy/summary_table:
   fn: "*.peddy.ped"
 peddy/het_check:
--- a/multiqc/utils/search_patterns.yaml
+++ b/multiqc/utils/search_patterns.yaml
@@ -390,19 +390,19 @@ disambiguate:
   contents: "unique species A pairs"
   num_lines: 2
 ngsderive/strandedness:
-  contents_re: "File	TotalReads	ForwardPct	ReversePct	Predicted"
+  contents: "File	TotalReads	ForwardPct	ReversePct	Predicted"
   num_lines: 1
 ngsderive/instrument:
-  contents_re: "File	Instrument	Confidence	Basis"
+  contents: "File	Instrument	Confidence	Basis"
   num_lines: 1
 ngsderive/readlen:
-  contents_re: "File	Evidence	MajorityPctDetected	ConsensusReadLength"
+  contents: "File	Evidence	MajorityPctDetected	ConsensusReadLength"
   num_lines: 1
 ngsderive/encoding:
-  contents_re: "(.*)File(.*)Evidence(.*)ProbableEncoding"
+  contents: "File	Evidence	ProbableEncoding"
   num_lines: 1
 ngsderive/junction_annotation:
-  contents_re: "(.*)File(.*)total_junctions(.*)total_splice_events(.*)known_junctions(.*)partial_novel_junctions(.*)complete_novel_junctions(.*)known_spliced_reads(.*)partial_novel_spliced_reads(.*)complete_novel_spliced_reads"
+  contents: "File	total_junctions	total_splice_events	known_junctions	partial_novel_junctions	complete_novel_junctions	known_spliced_reads	partial_novel_spliced_reads	complete_novel_spliced_reads"
   num_lines: 1
 peddy/summary_table:
   fn: "*.peddy.ped"
--- a/docs/plots.md
+++ b/docs/plots.md
@@ -135,13 +135,14 @@ config = {
 ### Switching datasets
 
 It's possible to have single plot with buttons to switch between different
-datasets. To do this, give a list of data objects (same formats as described
-above). Also add the following config options to supply names to the buttons:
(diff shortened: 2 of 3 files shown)


a-frantz:
@claymcleod I went ahead and added the new modules here. Does it all look good to you?

claymcleod:
@a-frantz can you post a screenshot of what the results look like?

a-frantz:
Encoding only goes in the general stats table, and `evidence` is hidden by default.
Screen Shot 2021-02-05 at 5 49 58 PM

Screen Shot 2021-02-05 at 5 50 52 PM
Here's what some real data for junction annotation looks like. I want the graph to toggle between "Junctions" and "Spliced Reads", but haven't figured out how to do it yet.

Also all the junction/reads stats can be toggled for the general table, but by default I just display the three junction annotation counts.

ewels:
> I want the graph to toggle between "Junctions" and "Spliced Reads", but haven't figured out how to do it yet.

See the docs here: https://multiqc.info/docs/#switching-datasets

Shout if you have any suggestions to make this more findable 🙂

multiqc/utils/search_patterns.yaml
```diff
@@ -396,6 +396,12 @@ ngsderive/instrument:
 ngsderive/readlen:
   contents_re: "(.*)File(.*)Evidence(.*)MajorityPctDetected(.*)ConsensusReadLength"
   num_lines: 1
+ngsderive/encoding:
+  contents_re: "(.*)File(.*)Evidence(.*)ProbableEncoding"
```
ewels, line 400: Could you explain why this regular expression search is needed? It's a super slow method for finding matches, especially when you're matching any number of any characters so many times. For example, another tool could generate a huge minified JSON file on one line and this regex can take minutes to run.

Can we not just do `contents: "File	Evidence	ProbableEncoding"`? Or even `contents: "ProbableEncoding"`

Remember that the module itself can discard false-positives when it runs, so this match doesn't have to be 100% perfect. As this will run on every file for every user it does need to be fast though.

Phil

ewels: I should add that the reason I make this comment is that I was just doing the final checks before pushing a v1.10 release and I found that my final big test was going super slow / hanging indefinitely. After a little debugging (prompting an even nicer new progress bar tweak) I found that it was the `ngsderive` search patterns hanging on test data for Atropos (whose PR has been open for years but never merged). I'm not entirely sure why, but switching for a simpler search pattern fixed it.

By fixed, I mean that after the change in 6263373c86a94eaa19c832c1c0a3a27326aaf469 the  CI test run has gone from ~40 minutes total to ~3 minutes 😅 So note that `master` now has simpler search patterns for the other files, which may have broken things for you (though the report from the test data seems identical).

This is why I run tests before pushing releases...! 😀

ewels: Ping @claymcleod @a-frantz - if you could take at this ASAP to check that the new patterns are ok (or suggest an alternative) that would be fab. It's the only thing that the release is waiting for now.

I'll have a look at this PR for you now and try to sneak that in as well as a thankyou-in-advance 😄

ewels:
> > I want the graph to toggle between "Junctions" and "Spliced Reads", but haven't figured out how to do it yet.
> 
> See the docs here: https://multiqc.info/docs/#switching-datasets
> 
> Shout if you have any suggestions to make this more findable 🙂

Ok I see that you had read the docs. You were basically there, you just missed that there needs to be multiple datasets supplied. As such the data labels and additional categories were ignored. I just added a list with the data twice and it worked: 2ade628980d4eabd8c3fb106d38519a9be17d30d

You'll see that I also improved the docs a little to make this clearer.

ewels:
Also, seeing as you guys are the authors of this tool and have now seen how much of a pain it can be to find and parse tool output files like this, I may as well plug an ancient blog post that I wrote ranting about the topic a bunch of years ago 😅 http://tallphil.co.uk/writing-good-log-files/

Basically, outputting YAML file(s) with a standardised filename suffix makes MultiQC's job (and probably other tools also) much much easier 👍🏻

ewels (approved):
Looks good to me, thanks! 👍🏻

a-frantz:
Thanks @ewels! This all looks great to me. Although now I'm curious about why regex searching is supported in the first place if it presents such a large slowdown? That's a really interesting (and unexpected) slowdown.

Right now we allow users to override the default tab delimiter in `ngsderive` output, though I think in light of this (and your insightful blogpost) we should deprecate that option and just stick to TSV output. That was the reasoning for the regex, in case a user overrode that option. I'll move that discussion offline, but thanks for the advice.

ewels:
> Although now I'm curious about why regex searching is supported in the first place if it presents such a large slowdown? That's a really interesting (and unexpected) slowdown.

Yes it took me by surprise too! Once I pinned it down to those patterns and saw the improvement after switching them I stopped digging. There are other modules using `contents_re` without this level of slowdown, some even with `.*` in the pattern:

https://github.com/ewels/MultiQC/blob/071bbe84e1c2ebf652b3cb2108d69b50068f005f/multiqc/utils/search_patterns.yaml#L598-L599

I just did a bit more trial and error out of curiosity and I think the slowdown comes down to two things:

* Certain fairly-large (~100kb) minified JSON files with everything on a single line
* The fact that the regex _starts_ with `.*`

If I strip that first wildcard (eg. `File.*Evidence.*ProbableEncoding` instead of `.*File.*Evidence.*ProbableEncoding`) then my smaller test drops from 42s to 1.5s so that seems to be the killer.

[Trying this out](https://regex101.com/r/82qPk9/1) on on https://regex101.com/ with those two example patterns and one of these JSON files, the profiler says it goes from `98936` steps to `6` steps if you remove that first `.*`, which I guess makes sense, as it just tries that first string, fails and stops instead of going through every character in the file looking for that string. Still, despite this I'm still surprised it slows everything down _that much_.

Anyway, thanks for giving the 👌🏻 for the change, I will push ahead with the release 👍🏻

ewels:
ps. Just configured the CI to time out after 10 minutes to try to catch this kind of thing in the future: https://github.com/ewels/MultiQC/commit/092231d1ae92c9a2858077392cf8be8d94ac8ced

PR #1347, opened 2020-11-19
kraken module: total read counts per sample

davidealbanese:
The percentage column in Kraken reports contain max 2 decimal digits, causing #1276. Now the total read counts per sample and percentages are computed using the read count column for precise results.

 - [x] This comment contains a description of changes (with reason)
 - [x] `CHANGELOG.md` has been updated

Issue #1238, opened 2020-06-28
Module request: Bracken


angelovangel:
**Tool Details**
* Name of tool: Bracken
* Tool homepage: https://ccb.jhu.edu/software/bracken/
* Tool description: Bracken computes the abundance of species (and other taxonomic levels) in DNA sequences from a metagenomics sample, using the taxonomy labels assigned by Kraken
* Data suitable for MultiQC plot(s): Barplots, identical to the ones used for Kraken
* Most interesting data for General Stats table: % Top5 Species

**Tool output:**

[minigut_sample1_kraken2_bracken_species.txt](https://github.com/ewels/MultiQC/files/4842387/minigut_sample1_kraken2_bracken_species.txt)

_kraken2 output for the same sample, to be used as a comparison_
[minigut_sample1_kraken2.txt](https://github.com/ewels/MultiQC/files/4842390/minigut_sample1_kraken2.txt)

* Log filename pattern: specified by user

ewels:
x-ref to original Gitter thread: [link](https://gitter.im/ewels/MultiQC?at=5ef4ea4ac223cc536a360c4a)

> Hi all!
>
> While I enjoy the new kraken2 module in MultiQC and find it a very good way to drill down such data, I think it may be leading to misconceptions and improper use/interpretation of the kraken2 results. kraken2 should not (in my opinion) be used for relative abundance estimation, it assigns taxonomy to sequencing reads. See discussions DerrickWood/kraken2#32 or the kraken2 paper.
>
> It would be more appropriate to use bracken for this, which is written exactly for this - relative abundance estimation using the labels assigned by kraken. I was about to open a MultiQC issue for bracken support, but then I thought it may be actually better to “replace” kraken with bracken in the list of tools that MultiQC supports. The output of bracken is very similar to that of kraken2, yet there are some subtle differences, so it should be pretty straightforward to do.
>
> Anyway, thought it is worth discussing

dfornika:
The regex to detect a kraken file is: `^\s{1,2}(\d{1,2}\.\d{1,2})\t(\d+)\t(\d+)\t([\dUDKPCOFGS-]{1,3})\t(\d+)(\s+)(.+)`

The difference between `kraken` and the adjusted report that `bracken` outputs seems to be:

- The first column of `bracken` output is not left-padded with whitespace
- The first column of `bracken` output may be `100.00`, so there may be three digits left of the decimal
[...]

PR #1372, opened 2021-02-13
Add Bracken

dfornika:

### DIFF
--- a/multiqc/modules/bracken/bracken.py
+++ b/multiqc/modules/bracken/bracken.py
@@ -1,0 +1,299 @@
+#!/usr/bin/env python
+
+""" MultiQC module to parse output from bracken """
+
+from __future__ import print_function
+from collections import OrderedDict
+import os
+import logging
+import re
+
+from multiqc import config
+from multiqc.plots import bargraph
+from multiqc.modules.base_module import BaseMultiqcModule
+
+# Initialise the logger
+log = logging.getLogger(__name__)
+
+
+class MultiqcModule(BaseMultiqcModule):
+    """ Bracken module """
+
+    def __init__(self):
+
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__(
+            name="Bracken",
+            anchor="bracken",
+            href="https://ccb.jhu.edu/software/bracken/",
+            info="is a highly accurate statistical method that computes the abundance of species in DNA sequences from a metagenomics sample.",
+        )
+
+        self.t_ranks = OrderedDict()
+        self.t_ranks["S"] = "Species"
+        self.t_ranks["G"] = "Genus"
+        self.t_ranks["F"] = "Family"
+        self.t_ranks["O"] = "Order"
+        self.t_ranks["C"] = "Class"
+        self.t_ranks["P"] = "Phylum"
+        self.t_ranks["K"] = "Kingdom"
+        self.t_ranks["D"] = "Domain"
+        self.t_ranks["R"] = "Root"
+        # self.t_ranks['U'] = 'Unclassified'
+
+        # Find and load any bracken reports
+        self.bracken_raw_data = dict()
+        for f in self.find_log_files("bracken", filehandles=True):
+            self.parse_logs(f)
+
+        # Filter to strip out ignored sample names
+        self.bracken_raw_data = self.ignore_samples(self.bracken_raw_data)
+
+        if len(self.bracken_raw_data) == 0:
+            raise UserWarning
+
+        log.info("Found {} reports".format(len(self.bracken_raw_data)))
+
+        # Sum counts across all samples, so that we can pick top 5
+        self.bracken_total_pct = dict()
+        self.bracken_total_counts = dict()
+        self.bracken_sample_total_readcounts = dict()
+        self.sum_sample_counts()
+
+        self.general_stats_cols()
+        self.top_five_barplot()
+
+    def parse_logs(self, f):
+        """
+        Parses a bracken report output file
+
+        1. Percentage of fragments covered by the clade rooted at this taxon
+        2. Number of fragments covered by the clade rooted at this taxon
+        3. Number of fragments assigned directly to this taxon
+        4. A rank code, indicating:
+            * (U)nclassified
+            * (R)oot
+            * (D)omain
+            * (K)ingdom
+            * (P)hylum
+            * (C)lass
+            * (O)rder
+            * (F)amily
+            * (G)enus
+            * (S)pecies
+           Taxa that are not at any of these 10 ranks have a rank code that is
+           formed by using the rank code of the closest ancestor rank with
+           a number indicating the distance from that rank.  E.g., "G2" is a
+           rank code indicating a taxon is between genus and species and the
+           grandparent taxon is at the genus rank.
+        5. NCBI taxonomic ID number
+        6. Indented scientific name
+        """
+
+        # Search regexes for stats
+        br_regex = re.compile(r"^(\d{1,3}\.\d{1,2})\t(\d+)\t(\d+)\t([\dRUDKPCOFGS-]{1,3})\t(\d+)(\s+)(.+)")
+        data = []
+        for l in f["f"]:
+            match = br_regex.search(l)
+            if match:
+                row = {
+                    "percent": float(match.group(1)),
+                    "counts_rooted": int(match.group(2)),
+                    "counts_direct": int(match.group(3)),
+                    "rank_code": match.group(4),
+                    "tax_id": int(match.group(5)),
+                    "num_spaces": len(match.group(6)),
+                    "classif": match.group(7),
+                }
+                data.append(row)
+
+        self.bracken_raw_data[f["s_name"]] = data
+
+    def sum_sample_counts(self):
+        """ Sum counts across all samples for bracken data """
+
+        # Sum the percentages for each taxa across all samples
+        # Allows us to pick top-5 for each rank
+        # Use percentages instead of counts so that deeply-sequences samples
+        # are not unfairly over-represented
+        for s_name, data in self.bracken_raw_data.items():
+            total_guess_count = None
+            for row in data:
+
+                # Convenience vars that are easier to read
+                rank_code = row["rank_code"]
+                classif = row["classif"]
+
+                # Skip anything that doesn't exactly fit a tax rank level
+                if row["rank_code"] == "-" or any(c.isdigit() for c in row["rank_code"]):
+                    continue
+
+                # Calculate the total read count using percentages
+                if rank_code == "R":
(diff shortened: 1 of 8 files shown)


dfornika:
Todo before mergins: Incorporate changes from #1347 for more accurate read counts.

dfornika:
General Statistics section does not seem to be picking up values correctly:

Screenshot from 2021-02-13 11-44-05

Screenshot from 2021-02-13 11-45-59

Results generated using dataset from: https://github.com/ewels/MultiQC_TestData/blob/master/data/modules/bracken/v2.6.0/bracken_report.txt

dfornika:
Fixed!

Screenshot from 2021-02-13 12-04-03

Screenshot from 2021-02-13 12-04-27

multiqc/modules/bracken/bracken.py
```diff
@@ -0,0 +1,306 @@
...
+            "max": 100,
+            "scale": "PuBu",
+        }
+        headers["% Unclassified"] = {
+            "title": "% Unclassified",
+            "description": "Percentage of reads that were unclassified",
+            "suffix": "%",
+            "max": 100,
+            "scale": "OrRd",
+        }
```
ewels, line 201: Can this be removed?

multiqc/modules/bracken/bracken.py
```diff
@@ -0,0 +1,306 @@
...
+
+        # Get table data
+        tdata = {}
+        for s_name, d in self.bracken_raw_data.items():
+            tdata[s_name] = {}
+            for row in d:
+                percent = (row["counts_rooted"] / self.bracken_sample_total_readcounts[s_name]) * 100
+                if row["rank_code"] == "U":
+                    tdata[s_name]["% Unclassified"] = percent
```
ewels, line 210: More unclassified stuff

multiqc/modules/bracken/bracken.py
```diff
@@ -0,0 +1,306 @@
...
+                                    rank_data[s_name][classif] = 0
+                                rank_data[s_name][classif] += row["counts_rooted"]
+                                counts_shown[s_name] += row["counts_rooted"]
+
+            # Add in unclassified reads and "other" - we presume from other species etc.
+            for s_name, d in self.bracken_raw_data.items():
+                for row in d:
+                    if row["rank_code"] == "U":
+                        rank_data[s_name]["U"] = row["counts_rooted"]
+                        counts_shown[s_name] += row["counts_rooted"]
```
ewels, line 269: Can we remove this?

multiqc/modules/bracken/bracken.py
```diff
@@ -0,0 +1,306 @@
...
+                if rank_data[s_name]["other"] < 0:
+                    log.debug(
+                        "Found negative 'other' count for {} ({}): {}".format(
+                            s_name, self.t_ranks[rank_code], rank_data[s_name]["other"]
+                        )
+                    )
+                    rank_data[s_name]["other"] = 0
+
+            rank_cats["other"] = {"name": "Other", "color": "#cccccc"}
+            rank_cats["U"] = {"name": "Unclassified", "color": "#d4949c"}
```
ewels, line 282: And this

multiqc/modules/bracken/bracken.py
```diff
@@ -0,0 +1,306 @@
...
+            anchor="bracken-topfive",
+            description="The number of reads falling into the top 5 taxa across different ranks.",
+            helptext="""
+                To make this plot, the percentage of each sample assigned to a given taxa is summed across all samples.
+                The counts for these top five taxa are then plotted for each of the 9 different taxa ranks.
+                The unclassified count is always shown across all taxa ranks.
+
+                The total number of reads is approximated by dividing the number of `unclassified` reads by the percentage of
+                the library that they account for.
+                Note that this is only an approximation, and that bracken percentages don't always add to exactly 100%.
```
ewels, line 298: More mention of unclassified

ewels (commented):
Sorry I've been so incredibly slow to get to this @dfornika - keen to crack on and get it merged in now.

Just doing a little bit of testing, I have a couple of questions:

* In https://github.com/ewels/MultiQC/issues/1238#issuecomment-778547890 you say that _The taxonomic 'rank code' of `bracken` does not include the character `U`._ however this code has loads of mentions of `Unclassified` still
    * Is this just copy+paste carry-over from the Kraken module? Can it all be deleted?
* When I run on test data from Kraken, the Bracken module picks up a bunch of Kraken outputs. I'm a bit worries that it is double-reporting samples.
    * Specifically, [this one](https://github.com/ewels/MultiQC_TestData/blob/master/data/modules/kraken/issue_1428/GSR-SWIFT-2021-5-04-FF08511340.kraken2.report.txt) and [this one](https://github.com/ewels/MultiQC_TestData/blob/master/data/modules/kraken/issue_1440/test.kraken2.report.txt).
    * This is especially bad as currently the search patterns are set up to not be shared - so if Bracken finds the logs then they are not searched for Kraken and will be skipped from that module.
    * Can we exclude any samples where we find Unclassified counts? If that's basically the only difference in the outputs?

dfornika:
Hi @ewels, no apologies necessary. I know there are a lot of modules in this project so lots to review & manage. It's been a while since I looked at this so I'll have to refresh my memory a bit. I've build a few Galaxy & Nextflow iterations of Kraken2->Bracken pipelines, and the issue of handling Unclassified reads is something that has come up often. We may need to ping Jen, the author of Bracken to ask about how to handle them best. It seems to me that the Unclassified reads are lost (unaccounted for) when going from Kraken to Bracken outputs, but they probably should be included if we want to account for all reads. But the straightforward Bracken output does not include them, so is it appropriate for MultiQC to add them in by some other method? I'm not sure.

ewels:
Right, that's a slightly different question - accounting for unclassified reads that are unreported. I was assuming that they are out of the picture and was mostly thinking about how to differentiate between the two tools. Otherwise I don't know how we can do it, they look identical to me.

If we're in touch with the author, it could be nice to ask for a header line / comment / _some_ kind of additional content that will allow us to distinguish output from the tools?

dfornika:
I did touch base with Jen, just to double-check that not including Unclassified reads in the report was the intended behaviour for bracken. She confirmed that it was, and that if an analysis wanted to take those reads into account then they should be taken from the kraken report.

I think I skimmed your initial questions too quickly earlier and didn't see your point about distinguishing kraken reports from bracken reports. I do think that looking for Unclassified reads in the report is the way to distinguish the outputs. We could consider submitting an issue to the Bracken repo to suggest a header or something but I know that it can be difficult to make changes to output formats when other tools/pipelines are built around the current format and may make some assumptions that could be broken by a change.

dfornika:
Hi @ewels, I've done my best to incorporate your suggestions. I'm not well set up to test this thoroughly right now. If anyone might be available to help push this forward they may be able to do so faster than I can.

Happy to continue pushing it along as much as I can, may need your guidance to finish it up.

vladsavelyev:
@dfornika, getting back to this PR, hopefully are still keen to get it merged!

Speaking about distinguishing Kraken and Bracken reports, currently the search patterns are basically identical (except for the right-aligning the first column in Kraken), so the reports are not distinguished at all in most cases.

Is it fair to say though that Kraken reports would always contain an `unclassified` line, and Bracken would always lack it and start with `root` instead? Can we use those assumption to distinguish the outputs, i.e. modify the search patterns to:

```
kraken:
  contents_re: '^\s{0,2}(\d{1,3}\.\d{1,2})\t(\d+)\t(\d+)\t([\dUDKRPCOFGS-]{1,3})\t(\d+)\s+unclassified'
  num_lines: 1
bracken:
  contents_re: '^(\d{1,3}\.\d{1,2})\t(\d+)\t(\d+)\t([\dRUDKPCOFGS-]{1,3})\t(\d+)\s+root'
  num_lines: 1
````

File: multiqc/modules/bracken/bracken.py [hunk 6]
```diff
@@ -0,0 +1,283 @@
...
+        self.t_ranks["S"] = "Species"
+        self.t_ranks["G"] = "Genus"
+        self.t_ranks["F"] = "Family"
+        self.t_ranks["O"] = "Order"
+        self.t_ranks["C"] = "Class"
+        self.t_ranks["P"] = "Phylum"
+        self.t_ranks["K"] = "Kingdom"
+        self.t_ranks["D"] = "Domain"
+        self.t_ranks["R"] = "Root"
+        # self.t_ranks['U'] = 'Unclassified'
```
vladsavelyev, line 41:
```suggestion
```

multiqc/modules/bracken/bracken.py
```diff
@@ -0,0 +1,283 @@
...
+                        counts_shown[s_name] = 0
+                    for row in d:
+                        if row["rank_code"] == rank_code:
+                            if row["classif"] == classif:
+                                if classif not in rank_data[s_name]:
+                                    rank_data[s_name][classif] = 0
+                                rank_data[s_name][classif] += row["counts_rooted"]
+                                counts_shown[s_name] += row["counts_rooted"]
+
+            # Add in unclassified reads and "other" - we presume from other species etc.
```
vladsavelyev, line 251: Another mentioned of "unclassified" that should be removed?

multiqc/modules/bracken/bracken.py
```diff
@@ -0,0 +1,283 @@
...
+        self.add_section(
+            name="Top taxa",
+            anchor="bracken-topfive",
+            description="The number of reads falling into the top 5 taxa across different ranks.",
+            helptext="""
+                To make this plot, the percentage of each sample assigned to a given taxa is summed across all samples.
+                The counts for these top five taxa are then plotted for each of the 9 different taxa ranks.
+
+                The category _"Other"_ shows the difference between the above total read count and the sum of the read counts
+                in the top 5 taxa shown + unclassified. This should cover all taxa _not_ in the top 5, +/- any rounding errors.
```
vladsavelyev, line 278: And here :)

ewels (approved):
Spoke to @vladsavelyev - happy to merge once suggested changes are applied 👍🏻

vladsavelyev:
After speaking with Phil, I came to conclusion that the modules are really identical in functionality, and even though we are usually in favour of code duplication over dependencies between modules, in this case the tools are so related that it makes sense for Bracken to reuse the code from Kraken. So I modified Bracken to inherit from Kraken. After we refactor modules to avoid running functionality on initialisation, it can be simplified even further.

vladsavelyev:
Merge after https://github.com/ewels/MultiQC/pull/2030

vladsavelyev (approved):

dfornika:
Hi @vladsavelyev, sorry for not getting back to you on this. And thanks to you and @ewels for pushing it over the finish line!

PR #1381, opened 2021-02-24
New BclConvert Module for MultiQC

andrei-seleznev:
This is a module for BclConvert: a tool that converts BCL files that come off Illumina sequences into FASTQ.
BclConvert is a sort of "successor" of bcl2fastq, and this module is based on the bcl2fastq MultiQC module.

https://sapac.support.illumina.com/sequencing/sequencing_software/bcl-convert.html

    - **Pull Request Submitted to MultiQC_TestData repo**
    - **There's one table with 7 or 8 columns (depending if `genome_size` is set in `multiqc_config`) but they are useful!**


ewels:
x-ref https://github.com/ewels/MultiQC_TestData/pull/186

ewels:
@andrei-seleznev - many thanks for this module, it looks great!

I've just been running it locally and have pushed a few commits with some small tweaks. The biggest one was to move the sample-level stats out of the General Statistics table and into a dedicated table in the module. I agree that all of the columns are useful, but I think it's a bit much for General Stats if many other modules are run. I think it makes more logical sense to have it together with the lane-level stats, and if only this module is run then the output looks almost identical anyway.

Anyway, please have a look over my changes and let me know what you think!

multiqc/modules/bclconvert/bclconvert.py
```diff
@@ -0,0 +1,696 @@
...
+    def _set_run_details(self):
+        # we might have several runinfo.xmls, but they all ought to be the same run and cluster length
+        for runinfo_file in self.find_log_files("bclconvert/runinfo", filehandles=True):
+            runinfo = self._parse_single_runinfo_file(runinfo_file)
+
+            if self.run_id and runinfo["run_id"] != self.run_id:
+                log.error(
+                    "Input data belongs to multiple sequencer runs (detected multiple run ids). This is not supported by this module: please run MultiQC on each run seperately to generate seperate reports."
+                )
+                raise UserWarning
```
ewels, line 236: This is a bit of a shame - the previous `bcl2fastq` module doesn't have this limitation does it? Is there any way that we can get the module to work with multiple sequencer runs?

One thing is that this effectively means that the CI test don't test the module 😅 But also I wonder if it would be something that people may want to do fairly frequently..?

ohofmann: Hey @ewels - thanks for your feedback! We're working with @andrei-seleznev to test this as a replacement for our previous bcl2fastq QC steps and I never even realized it had multi-run support. We normally run MultiQC again over the individual outputs in the rare case that we need to compare between runs (with the option of MegaQC for any systematic comparison).

How does this mess with the CI testing? Wondering if there's a way to get that to work at least.

ewels: Ah nice! It doesn't break the CI as such, it's just that this module is skipped as there is more than one run in the test data repo. In fairness I guess half of the code in the module is running before it gets to the point of exiting..

andrei-seleznev: I believe the current `bcl2fastq` module in the main repo doesn't do undetermined stats recalculation, and this recalculation is the reason for the one-seqrun-only limitation.

The function __recalculate_undetermined()_ in bclconvert.py will recalcualte unknown-read-counds when parsing more than one "bclconvert run" from the same "sequencing run"  (the usecase for this is something like -  take a sequencing run, split the samplesheet by different OverrideCycles settings and run bclconvert on the resulting split, then run MultiQC on all the output - so you have MultiQC parsing "multiple bclconvert runs" from "one sequencing run"). You have to recalcualte the undertermined counts afterwards (which in BclConvert are given _per-lane_) because e.g. some read listed as "undetermined" in one BclConvert-run-outputis actually from another BclConvert-run-output . I think the recalculation of stats doent make sense if the bclconvert runs correspond to different sequencer runs - hence the limitation!

andrei-seleznev:
Thanks heaps for the feedback & for your time going over this module @ewels ! All your changes look great to me.

multiqc/modules/bclconvert/bclconvert.py
ewels: Right, makes sense! Sounds like a really nice addition 😀

Can we just disable this calculation if there are multiple runs? Ideally showing the reported undetermined reads if there is one one bclconvert run per sequencing run and hiding them otherwise? Or just skipping undermined completely in these cases.

I can imagine use cases where comparing multiple sequencing runs will be useful for people. For example, having a pool of samples sequenced across multiple runs and trying to compare outputs / pooling and so on.

andrei-seleznev: Yeah true, simply disabling the calculation does sound like a good idea! I'll go implement that :)

ewels: Brilliant - thank you! 👍🏻

ewels: @andrei-seleznev - do you think that you'll have time to get this written in the next day or two? I'm hoping to get the MultiQC version 1.11 release out this week if possible. If it's too much work then I can merge this PR to get the new module included in the release and we can do a follow-up PR whenever you have time (but if possible I'd prefer to get it all done in one of course).

Cheers,

Phil

andrei-seleznev: Yeah I reckon I can get this done in the next day or two, no worries!

ewels: Awesome 🚀

andrei-seleznev: pushed a commit which should resolve this. the module will 'combine' samples across sequencing runs (the user still sees one big table with all the samples) , the per-lane stats are provided per-seqrun-per-lane, and undetermined results are suppressed

ewels:
Hi @andrei-seleznev,

Many thanks for your speedy changes, they look great! I've just pushed a commit with a few tweaks - I remembered how to add extra HTML without creating a new section. I made the messages a little less verbose (especially on the command line) and I split them between the top of the module and just above the affected plot.

I'm super happy with this module now so will merge. Shout if there's anything I messed up or you have anything else you want to add / change. I'm hoping to push a release tomorrow lunchtime if all goes to plan (European time zone).

Phil

ohofmann:
Thanks so much, Phil and Andrei!

andrei-seleznev:
Looks great, stoked hear the module's going in, thanks for your time and work on it @ewels !

multiqc/utils/search_patterns.yaml
```diff
@@ -80,6 +80,18 @@ bcl2fastq:
   fn: "Stats.json"
   contents: "DemuxResults"
   num_lines: 300
+bclconvert/runinfo:
+  fn: "RunInfo.xml"
+  contents: "<?xml"
```
ewels, line 85: @andrei-seleznev: Post-merge comment - I'm going to remove these `contents:` search criteria. Partly because the (unescaped) tag here breaks the MultiQC documentation rendering 😆 But also the filenames are unique enough that they shouldn't need any additional search criteria. Also looking into the file contents will slow down the file search a little.

Let me know if you know of other tools making files with identical filenames, where the `contents` search criteria here would be needed to disambiguate the outputs 👍🏻

Issue #1379, opened 2021-02-23
Kraken reports including minimizer count information rise an exception


aeu79:
**Description of bug:**

Kraken2 reports that include minimizers information (KrakenUniq-like) generate and exception in MultiQC. These reports are generated using the `--report-minimizer-data` flag. 

**Workaround:**

As a workaround, the reports can be converted back to the standard format (8 to 6 columns) with:
```bash

for file in /path/to/your/reports/*.report; do
echo $file
cut -f1-3,6-8 $file > "${file}_standard.report"
done

```
**MultiQC run details (please complete the following):**

Sorry that I'm skipping the details and context, but I went through the code and I already have a fix. I'll polish a little the code and then I'll push to my fork so you can have a look.

ewels:
Sounds good, I’ll wait for a PR then 👌🏻

If you’re able to attach some example data that would be great, then we can add that to ewels/MultiQC_TestData to ensure that it stays fixed in the future 👍🏻

apeltzer:
Looks like #1380 really fixes it + visualizes the new output 👍🏻

aeu79:
Ups, created the PR. Well, this was the first part only, so I'm glad I saw #1380 right now. Still, having a quick view to the code there, it doesn't solve the "original" problem. 
Although the regular expression seemed quite specific, the 8 columns report managed to go through security :guardsman: anyhow.
Have a look to this:
[...]

PR #1380, opened 2021-02-23
Add distinct minimizer heatmap for Kraken2

maxibor:
Since version 2.1.0, Kraken2 integrated the distinct minimizer count ability of [KrakenUniq](https://genomebiology.biomedcentral.com/articles/10.1186/s13059-018-1568-0)
This new function comes with a slightly altered report (see [Kraken2 docs](https://github.com/DerrickWood/kraken2/wiki/Manual#distinct-minimizer-count-information)) format.

This PR adds a heatmap of the duplication rate of the minimizers (i.e. kmers) which is very useful to detect read stacking, and false positive hits.

- When encountering a "old" report, it behaves as before
- When encountering a "new report" (with distinct minimizer count) it adds the duplication heatmap.

PR checklist:

- [x] This comment contains a description of changes (with reason)
- [x] `CHANGELOG.md` has been updated

PR #1383, opened 2021-02-24
Kraken: Report minimizers

aeu79:
_Phil edit_: x-ref https://github.com/ewels/MultiQC/issues/1379

### DIFF
--- a/multiqc/modules/kraken/kraken.py
+++ b/multiqc/modules/kraken/kraken.py
@@ -93,26 +93,22 @@ class MultiqcModule(BaseMultiqcModule):
         # Search regexes for stats
         k2_regex = re.compile(r"^\s{1,2}(\d{1,2}\.\d{1,2})\t(\d+)\t(\d+)\t([UDKPCOFGS-]\d{0,2})\t(\d+)(\s+)(.+)")
         data = []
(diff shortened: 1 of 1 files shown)


aeu79:
Uf, pushed the wrong branch! I'm sorry! (fixing it soon)

ewels:
@aeu79 do I understand correctly that this PR is a duplicate of https://github.com/ewels/MultiQC/pull/1380 and can be closed without merging?

aeu79:
> @aeu79 do I understand correctly that this PR is a duplicate of #1380 and can be closed without merging?

Yes, well, kind of. #1380 solves the issue described in #1379. But the underlying problem is still there, the one with the reg exp that I described [there](https://github.com/ewels/MultiQC/issues/1379#issuecomment-785014177). If you agree, I could add an issue and once the PR #1380 is merged, create a new PR to make the reg exp more stringent. That would avoid (or at least try) to have new crashes whenever the number of columns is modified in kraken reports. Actually, in the manual for kraken2 this feature is described as still experimental: 
>At present, this functionality is an optional experimental feature -- meaning that we may later alter it in a way that is not backwards compatible with previous versions of the feature.

PR #1385, opened 2021-03-04
Picard wgsmetrics cov histo

ewels:
Turns out that if you run Picard with `CAP_COVERAGE` set to 1 million, the histogram code in MultiQC takes hours to run (if at all) and the report is so big that it crashes the browser.

This PR:

* Smooths the histogram plot by default to a maximum of 1000 data points
* Adds config options to customise this smoothing (or disable it)
* Adds config options to tweak how the long tail of coverage is cut
* Adds a config option to completely skip the histogram


ewels:
Tested with example data from @drpatelh - without this PR my laptop sounds like it's about to take off and I cancelled it after waiting a few minutes. After the PR MultiQC runs in a couple of minutes and even seems to generate a report with a plot that looks vaguely useful. So I think that the default plot smoothing probably does the job 🚀 

Extra config options are never a bad idea though.

Issue #1005, opened 2019-08-15
AdapterRemoval output confusing vs FLASH


olgabot:
**Description of bug:**

We tested using [AdapterRemoval](https://github.com/czbiohub/nf-core-crisprvar/pull/15) and [FLASH](https://github.com/czbiohub/nf-core-crisprvar/pull/12) to merge reads for a CRISPR analysis pipeline and generated MultiQC outputs for both. It was confusing to see them side-by-side as AdapterRemoval looked like it had 2x the reads, but really I think some of the numbers weren't getting subtracted from one another.

### AdapterRemoval output
![Screen Shot 2019-08-15 at 4 54 17 AM](https://user-images.githubusercontent.com/806256/63093159-f1e5cb80-bf18-11e9-8d12-8a5c44c48434.png)

It seems that AdapterRemoval is showing total reads when it should be showing:

- Full-length collapsed pairs (as is)
- uncollapsed pairs = Retained read pairs - Full length collapsed pairs

### FLASH output (on same samples)
![Screen Shot 2019-08-15 at 4 54 12 AM](https://user-images.githubusercontent.com/806256/63093161-f1e5cb80-bf18-11e9-9e16-5a9100052c72.png)

**File that triggers the error:**
Please drag and drop (and upload to the GitHub issue) an input file that I can use to replicate the error.

Here are some log files:

[adapterremoval_logs.zip](https://github.com/ewels/MultiQC/files/3505442/adapterremoval_logs.zip)
[flash_logs.zip](https://github.com/ewels/MultiQC/files/3505443/flash_logs.zip)

**MultiQC run details (please complete the following):**
 - Command used to run MultiQC: `multiqc -f $rtitle $rfilename --config $multiqc_config . -m fastqc -m adapterRemoval -m flash`
 - MultiQC Version: MultiQC v1.7
 - Operating System: Ubuntu
 - Python Version: 
```
(base) root@671e64af0899:/# /opt/conda/envs/nf-core-crisprvar-1.0dev/bin/python --version
Python 2.7.15
```

 - Method of MultiQC installation: conda, this is the [environment.yml](https://github.com/czbiohub/nf-core-crisprvar/blob/master/environment.yml)

ewels:
Are you sure that this isn't a nomenclature thing? _i.e._ one read pair is two reads?

ewels:
Ok, so if I understand you correctly, we should just not plot `Retained read pairs`? As this is already a sum of the other categories?

[...]

PR #1386, opened 2021-03-05
Fix AdapterRemoval v2 

apeltzer:
This builds on top of work by @alexhbnr and follows the discussion in #1005 to get things fixed. I'll try to get to the points we discussed there tomorrow then ideally to get this hopefully in a useful state. Will get back to what Mikkel said in the PR and help him getting a JSON output for his rewrite but having a (backwards) compatible solution is I think a good idea - can then later have a switch in this module to read either the current log output or the JSON and prefer the JSON over this parser ...? 

@jfy133 to let you know about it :-)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,8 @@ For further information, please see the [documentation](https://multiqc.info/doc
   - Add support for the creation of a "Reported vs Empirical Quality" graph to the Base Recalibration module.
 - **hap.py**
   - Updated module to plot both SNP and INDEL stats ([#1241](https://github.com/ewels/MultiQC/issues/1241))
+- **indexcov**
+  - Fixed bug when making the PED file plots ([#1265](https://github.com/ewels/MultiQC/issues/1265))
 - **Kaiju**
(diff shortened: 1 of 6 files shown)


ewels:
Please see https://github.com/ewels/MultiQC/blob/master/CHANGELOG.md#major-change-code-linting for failing tests 😄

fgvieira:
Since there is an open PR to fix `AdapterRemoval`, I though I'd bring this here.

In the `Adapter Removal` module there is a section for `Length Distribution Paired End Collapsed`, but it then allows o select for singletons, discarded, etc...

image

If it is PE, why allow for reads that are not PE? Or is it the title that is wrong?

jfy133:
Hi @fgvieira - actually there is a lot of problems with the current implementation, to follow the saga please see here: https://github.com/ewels/MultiQC/issues/1005

If you happen to be able to restrucutre the whole module based on the info in that thread that would be awesome!

ewels:
Solved by https://github.com/ewels/MultiQC/pull/1647

PR #1387, opened 2021-03-06
New module: OptiType

apeltzer:
- [x] There is example tool output for tools in the https://github.com/ewels/MultiQC_TestData repository or attached to this PR (see https://github.com/ewels/MultiQC_TestData/pull/190) 
    - Hid three minor alleles, but major A1, B1, C1 are shown in main table now  


apeltzer:
Not sure I can really make that a plot (its just stating the HLA-(A,B,C)(1,2) allele(s)

multiqc/utils/search_patterns.yaml
```diff
@@ -398,6 +398,9 @@ ngsderive/instrument:
 ngsderive/readlen:
   contents_re: "(.*)File(.*)Evidence(.*)MajorityPctDetected(.*)ConsensusReadLength"
   num_lines: 1
+optitype:
+  contents_re: "(.*)A1(.*)A2(.*)B1(.*)B2(.*)C1(.*)C2(.*)Reads(.*)Objective"
```
ewels, line 402: This style of search pattern is apparently evil in terms of run time (see https://github.com/ewels/MultiQC/pull/1370#discussion_r588935271). Please just use `contents` and a pasted string if possible.

apeltzer: Also added to use `contents` now 👍🏻

ewels (approved):
Looks good, thanks 👍🏻 

Don't love that it's table only and without colours, but I agree that tough to do anything about it. _Could_ maybe do some kind of [conditional formatting](https://multiqc.info/docs/#conditional-formatting) thing but would probably look rubbish. Better would be to assign a number and then display a background colour based on that, but would require some additional stuff to be written to the MultiQC core for that. No time for that now, but can come back to it another time perhaps.

Might be nice to have a stacked bar plot to show the number of samples with each allele type perhaps? eg. a bar for A1 / A2 / B1 etc broken into sub-bars for each type, with the width being the number of samples with that type.

Phil

ewels:
Ok I couldn't resist, I added a plot showing a summary of subtype usage. I hope it's ok / useful 🤞🏻 😅 

Only one sample in the example datasets, so a little tricky to really test it to destruction. But should be ok I think.

ewels:
Just found this: https://multiqc.info/docs/#custom-cell-background-colours

Turns out I already wrote the feature to give custom background colours to table cells! So wouldn't be so hard to add after all..

PR #1402, opened 2021-03-26
solved breakdown bcftools

BiscuitCandy:
modified '/multiqc/modules/bcftools/stats.py'

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,12 @@
 #### New MultiQC Features
 
 #### New Modules
+-->
 
 #### Module updates
--->
+
+- **bcftools**
+  - Fixed bug where `QUAL` value `.` would crash MultiQC ([#1400](https://github.com/ewels/MultiQC/issues/1400))
 
 #### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


ewels (changes requested):
Could you explain your logic here please?

I don't think that this will work as it stands, as it will replace integers with `0.0`, which is wrong.

BiscuitCandy:
> Could you explain your logic here please?
> 
> I don't think that this will work as it stands, as it will replace integers with `0.0`, which is wrong.

About that,

if the given parameter contains "." *(i.e decimal point)* then it adds a `0 at the front and a 0 at the end`
that means if the parameter is "5.2" then it becomes "05.20" and if the parameter is "." then it becomes "0.0"
**And it works if the parameter has a decimal point**

BiscuitCandy:
> The latest commit is valid for natural numbers too
> Here it just adds a *0* at the start thus leaving the value unchanged 

 `let the parameter be 548 then it becomes float("0548") i.e 548.0` 
 `Or if the parameter is "." then it becomes float("0.")`

ewels:
Ok nice, the update code looks good! A more elegant approach than just capturing the error as I had expected.

If you can just format with Black to get the tests to pass then I think we're good to go. Docs: https://multiqc.info/docs/#black

BiscuitCandy:
`It's Done`

ewels (approved):
Perfect, thanks! 👍🏻

PR #1406, opened 2021-04-01
MultiQC fails to parse plotProfile when TSS values is missing

ftabaro:

MultiQC fails to parse `plotProfile` text output if the TSS bin is not found (eg. unannotated peaks). I fixed the issue by catching the error.


ewels:
Ok nice, thanks! It seems that the `plotProfile` files can be quite variable 😅  If you have an example file that looks like this, that would be great - I can add it to the [ewels/MultiQC_TestData](https://github.com/ewels/MultiQC_TestData/tree/master/data/modules/deeptools) repository for future testing.

PR #1417, opened 2021-04-26
New module: Sambamba Markdup 

gartician:

This is a new MultiQC module for [sambamba markdup](https://lomereiter.github.io/sambamba/docs/sambamba-markdup.html) to show the extent of read duplication in a sequencing library. Duplicate reads and unique reads are extracted out of the log files, and duplicate rate is calculated and displayed to the user in the general statistics table (columns = Duplicate Reads and Duplicate Rate) along with a bar graph section. This is my first time submitting a PR to a public project, so please let me know if I can improve on something!


ewels:
Thanks for this! As Sambamba is the main tool and Markdup is a subcommand, it would be nice to structure this like module like the code for Samtools / Picard / VCFTools etc - see https://github.com/ewels/MultiQC/tree/master/multiqc/modules/samtools

Although we will only have one subcommand this then future-proofs us for adding more later.

Shout if you have any questions or need any help with this.

Thanks!

Phil

gartician:
I figured out how to run `markdup.py` as a sub-command of `sambamba`!

multiqc/modules/sambamba/markdup.py
```diff
@@ -0,0 +1,131 @@
...
+        return len(self.markdup_data_dict)
+
+    def markdup_clean_s_name(self, s_name):
+
+        """
+        Clean up sample name from 'markdup_sample_1' to 'sample_1'.
+        Is there a better way to do this?
+        """
+
+        return s_name.replace("markdup_", "")
```
ewels, line 59: Not really - there is `fn_clean_exts` and `fn_clean_trim` in the config but they both work with suffixes. I don't think that there is any equivalent for prefixes.. 🤔 

I think that your approach is fine, but I would (a) not use a custom function just for a `.replace()` call 😅 and (b) you need to use `self.clean_s_name()` - see https://multiqc.info/docs/#custom-sample-names

I'll leave a suggested change above to give an example of what I mean.

multiqc/modules/sambamba/markdup.py
```diff
@@ -0,0 +1,131 @@
...
+        # Clean sample name from 'markdup_sample_1' to 'sample_1'
+        # Find and load sambamba logs to markdup_data_dict.
+        # Regex for key phrases and calculate duplicate rate.
+        # Give user warning if redundant samples are found.
+
+        self.markdup_data_dict = dict()
+
+        for f in self.find_log_files("sambamba/markdup"):
+
+            sample_name = self.markdup_clean_s_name(f["s_name"])
```
ewels, line 29:
```suggestion
            sample_name = self.clean_s_name(f["s_name"].replace("markdup_", ""), f['root'])
```

See comment below for discussion..

multiqc/utils/search_patterns.yaml
```diff
@@ -363,6 +361,9 @@ longranger/invocation:
   max_filesize: 2048
 macs2:
   fn: "*_peaks.xls"
+sambamba/markdup:
+  fn: "markdup_*"
+  contents: "finding positions of the duplicate reads in the file"
```
ewels, line 366: If this is always at the top of the file, then please set `num_lines: 5` or something to stop MultiQC crawling through huge logs.

setup.py
```diff
@@ -101,7 +101,6 @@
             "bowtie1 = multiqc.modules.bowtie1:MultiqcModule",
             "bowtie2 = multiqc.modules.bowtie2:MultiqcModule",
             "busco = multiqc.modules.busco:MultiqcModule",
-            "bustools = multiqc.modules.bustools:MultiqcModule",
```
ewels, line 104: Was this an accident? (looks like it)

docs/README.md
```diff
@@ -56,7 +56,6 @@ MultiQC Modules:
     Bcftools: modules/bcftools.md
     biobambam2: modules/biobambam2.md
     BUSCO: modules/busco.md
-    Bustools: modules/bustools.md
```
ewels, line 59: Accidental?

multiqc/utils/config_defaults.yaml
```diff
@@ -352,9 +351,6 @@ module_order:
   - busco:
       module_tag:
         - Denovo
-  - bustools:
-      module_tag:
-        - RNA
```
ewels, line 357: Is there a reason that you're stripping bustools everywhere? 😅

ewels (commented):
Nice, that was speedy! Quick read through of the code looks good, couple of comments. Will try to get to testing the code for a final review + merge when I can.

multiqc/modules/sambamba/markdup.py
```diff
@@ -0,0 +1,131 @@
...
+            self.markdup_data_dict[sample_name] = self.get_markdup_stats(f["f"])
+
+            self.add_data_source(f)
+
+            if f["s_name"] in self.markdup_data_dict:
+                log.debug("Duplicate sample name found! Overwriting: {}".format(f["s_name"]))
+
+        if len(self.markdup_data_dict) == 0:
+            raise UserWarning
```
ewels, line 38:
```suggestion
            return 0
```

ewels: I _think_ this may fix the tests. They're failing currently because you're generating a report when there should not be one.

ewels: Ah it's not this sorry (though I think you do want to return `0` anyway, else other submodules won't run in the future).

It's because you're not filtering sample names: https://multiqc.info/docs/#filtering-by-parsed-sample-names

setup.py
gartician: These were accidental, I'm gonna re-add them back.

multiqc/modules/sambamba/markdup.py
gartician: Agreed, I added this so sambamba markdup can also honor the `--ignore-samples` flag from the user.

multiqc/modules/sambamba/markdup.py
gartician: This worked great.

gartician:
I'm going to submit a new PR since I hard-reset my fork to get even with `ewels/MultiQC`. I initially tried to git pull from `ewels/MultiQC` (upstream) to the fork but that didn't work out in practice.

ewels:
ok! No need to close and reopen in the future, the PR updates dynamically. You can see that the commits and diffs here are now identical to https://github.com/ewels/MultiQC/pull/1421 (but our comments + review are lost). Doesn't matter though, fine to continue with the new PR 👍🏻

PR #1419, opened 2021-04-27
New module: Ensembl VEP

berguner:
The "vep" module parses the summary statistics of Ensembl VEP variant annotation tool. It can parse both HTML and plain text stats.

- [x] Example tool output: [multiqc_vep_test.zip](https://github.com/ewels/MultiQC/files/6385135/multiqc_vep_test.zip)
    * VEP General statistics table has 8 columns


ewels:
Wow, parsing HTML reports - you're a brave man! I usually flat out refuse to do this as it can be such a mess 😅 

Test data added in ewels/MultiQC_TestData@7cc212fac1bf87906926cc8e6bcac3c7c586e8d9

multiqc/modules/vep/vep.py
```diff
@@ -0,0 +1,221 @@
...
+                name="General Statistics",
+                anchor="vep-general-statistics",
+                description="VEP General statistics",
+                helptext="Table showing general statistics of VEP annotaion run",
+                plot=table.plot(plot_data, plot_cats, table_config),
+            )
+
+    def generate_bar_graphs(self):
+        """Add bar graphs from the given VEP stats"""
+        titles = [
```
ewels, line 190: Whilst it's less code and cleaner (and DRY) to write the bar graphs in a loop like this, I prefer not to do this for MultiQC modules. The best MultiQC modules have well polished output, which basically comes down to fine-tuning how each plot looks and reads. This kind of customisation is very awkward to do if in a loop.

The two immediate examples here are that the descriptions are pretty unhelpful, and I want to customise the colours on one bar plot. This super tricky straight away.

ewels (commented):
Thanks for this! Speed look at the code and report looks generally good. I can see a bit of tidying up that I would like to do so might have a speed stab at that now and push to your PR. If it gets too late I'll leave a review with requests instead though.

Phil

multiqc/modules/vep/vep.py
```diff
@@ -0,0 +1,221 @@
...
+            "sortRows": True,  # Whether to sort rows alphabetically
+            "only_defined_headers": False,  # Only show columns that are defined in the headers config
+            "col1_header": "Sample Name",  # The header used for the first column
+            "no_beeswarm": True,  # Force a table to always be plotted (beeswarm by default if many rows)
+        }
+        title = "General statistics"
+        plot_cats = OrderedDict()
+        plot_data = OrderedDict()
+        color_list = ["Oranges", "Blues", "Reds", "Greens"]
+        for sample_name in self.vep_data:
```
ewels, line 170: Yeah same here - this is _too_ efficient. A guideline for MultiQC colours is that they should match the type of data that they represent. So reds = things that are bad when there are lots. Greens = good, and so on.

Here you have meaningful colours assigned randomly. We also can't customise single columns, for example to hide specific ones and so on.

multiqc/modules/vep/vep.py
```diff
@@ -0,0 +1,365 @@
...
+            "Overlapped genes",
+            "Existing variants",
+            "Novel variants",
+            "Variants filtered out",
+            "Variants processed",
+            "Lines of input read",
+        ]
+        # Set up the base config for each column
+        table_cats = OrderedDict()
+        color_list = ["Oranges", "Reds", "Blues", "Greens"]
```
ewels, line 186: In the end after a bit of order change the colours ended up being ok so I dropped my whinging on this 😅 Main thing is to make it easy to scan / differentiate between columns and this does that 👍🏻

multiqc/utils/config_defaults.yaml
```diff
@@ -271,10 +274,11 @@ fn_ignore_files:
   - "*.gtf"
   - "*.bed"
   - "*.vcf"
+  - "*.tbi"
   - "*.txt.gz"
   - "*.pdf"
-  - "*[!_][!m][!q][!c].html"
   - "*.md5"
+  - "*[!_\\.][!mv][!qe][!cp].html"
```
ewels, line 281: Adding on something I was thinking about the other day whilst we're here..

```suggestion
  - "*[!_\\.][!mv][!qe][!cp].html" # Allow _mqc.html and _vep.html files
  - "*.py[cod]" # Compiled python files, eg. MultiQC installation directory
```

docs/modules/vep.md
```diff
@@ -0,0 +1,19 @@
...
+Description: Ensembl VEP determines the effect of your variants on genes, transcripts and protein sequences, as well as regulatory regions.
+---
+
+The VEP module parses the summary statistics generated by
+[Ensembl VEP](https://www.ensembl.org/info/docs/tools/vep/index.html).
+
+MultiQC parses the VEP summary statistics stored in either HTML or plain text format. Note that
+you must run VEP with one of the options below to be able to use this module:
+
+- `--stats_file [SAMPLE_NAME].vep.html`
```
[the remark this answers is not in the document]
ewels: Ah it's because MultiQC will ignore the `*.html` file otherwise, gotcha 👍🏻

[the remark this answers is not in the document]
ewels:
```suggestion
- `--stats_file [SAMPLE_NAME].vep.html` _(without the `.vep` suffix, MultiQC will ignore the HTML file)_
```

ewels (approved):
Ok great, looking good - thank you for writing this! Been meaning to write a module for VEP for literally years and years now..! 😅 

As you can see I've refactored your front-end code quite a bit, making it much more verbose but also a little more customised and hopefully pretty. Please check it and make sure that you agree with all of my changes 👍🏻

docs/modules/vep.md
```diff
@@ -0,0 +1,19 @@
...
+- `--stats_file [SAMPLE_NAME]_vep.html`
+- `--stats_text --stats_file [SAMPLE_NAME].vep.txt`
```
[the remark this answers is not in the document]
berguner: Yes, unfortunately. That's why I decided to parse both formats.

berguner:
Thanks for reviewing the code so quickly. I tested your changes and everything works as intended and the speed wasn't affected at all. The report looks much better now. I really liked how you grouped the consequences graphs and the general color scheme. I am looking forward to see this module in the main repo 😊

a-frantz:
I'm really excited to try this! But I've found two issues. My summary files match the default `fn_ignore_files` pattern: `- "*[!_\\.][!mv][!qe][!cp].html" # Allow _mqc.html and _vep.html files`. For the record, my pipeline names the result files `[sample].annotated.somatic.vcf_summary.html`.

Once I removed that ignore pattern, I hit a runtime error:
```
│ Traceback (most recent call last):                                                                                       │
│   File "/Users/afrantz/MultiQC/multiqc/multiqc.py", line 604, in run                                                     │
│     output = mod()                                                                                                       │
│   File "/Users/afrantz/MultiQC/multiqc/modules/vep/vep.py", line 55, in __init__                                         │
│     self.bar_graph_consequences()                                                                                        │
│   File "/Users/afrantz/MultiQC/multiqc/modules/vep/vep.py", line 227, in bar_graph_consequences                          │
│     plot_data, plot_cats, plot_config = self._prep_bar_graph(title)                                                      │
│   File "/Users/afrantz/MultiQC/multiqc/modules/vep/vep.py", line 347, in _prep_bar_graph                                 │
│     plot_data[s_name] = self.vep_data[s_name][title]                                                                     │
│ KeyError: 'Coding consequences'
```
I believe MultiQC parses files alphabetically? In which case it failed about 1/3 of the way through 78 files. I can upload the offending HTML if it helps with debugging.

berguner:
Thanks for testing it Andrew. It is probably because the report didn't have a section called "Coding consequences". All of the reports that I used for testing had this section, so I couldn't catch it. Can you please share the VEP report that caused the issue?

a-frantz:
Sure, I just opened it up the HTML in my browser and there is a "Coding consequences" pie chart. Interesting, maybe the HTML is a bit funky for this one?
[SJETV093_D_G.annotated.somatic.vcf_summary.html.gz](https://github.com/ewels/MultiQC/files/6392856/SJETV093_D_G.annotated.somatic.vcf_summary.html.gz)

berguner:
Strangely this file didn't cause any problem in my test setup. I simply changed the suffix to `.vep.html` and put this among the other test files and it was able to parse all of them. Here is the generated report.
[multiqc_report.zip](https://github.com/ewels/MultiQC/files/6392924/multiqc_report.zip)

VEP 100 is quite recent, do you have any reports from older VEP versions which might be causing the problem?

a-frantz:
Sorry, all I have are v100 (but I have a few thousand I can test with).

This is strange. Running `grep -c 'Coding consequences' *` shows that all 78 of my test files have 3 occurrences of "Coding consequences". When I removed the one I just uploaded, it failed on a different file (that also has "Coding consequences" *somewhere* in the file). I repeated that process about 5 times, each time it failed with the same error on a new file.

Could this be a more tricky bug caused by something less obvious? It doesn't seem to be a simple matter of "Coding consequences" being missing.

berguner:
I put a checkpoint for empty bar graphs. I tried to imitate a problem where some of the piecharts were missing or empty and tested it and it was able to handle missing data. Can you please also test the latest version?

a-frantz:
Weird, I didn't think that would work, but it did. None of my samples have missing data in any of the tables, but they're all being processed now without error. Not sure what's going on, but it works, so I'm happy.

This looks great! I've been hoping this would be added for over a year now.

My last suggestion is that `_summary.html` be excluded from `fn_ignore_files`. That's the extension VEP v100 gives my summary files, which is currently filtered out by the default configuration.

Thanks for adding this.

ewels:
Thanks for the testing both! @a-frantz is the summary prefix automatically done or are you specifying that on the command line? My understanding was that you have to give the full html file name when you run VEP..

a-frantz:
I believe in VEP v100 it's automatically generated with that suffix. It is in my pipeline at least, been over a year since I ran it manually/looked at the arguments. It's possible something I'm specifying causes that behavior (VEP has some strange behaviors if I remember correctly). This is how I run it:
```
VEP_ARGS = (
    "--offline --everything --cache --format vcf --vcf --fork 4 --force_overwrite"
)
...
    annotated_vcf_path = vcf_path.replace(".somatic", ".annotated.somatic")
    exec_list = ["vep", "--dir", cache, "-i", vcf_path, "-o", annotated_vcf_path]
    exec_list += consts.VEP_ARGS.split()
    _ = util.run_cmd(exec_list)
```
It could auto-generate that suffix because of the 2 VCF arguments?

berguner:
`_summary.html` is the default suffix. I took the liberty and set it as `[._]vep.html` because that's how I name my files. Also I wasn't sure how to exclude `_summary.html` from `fn_ignore_files` while keeping `*.html`.

ewels:
ok thanks both. I am tempted to just drop HTML files from the file patterns that MultiQC skips. I think that this is the first module that actually parses HTML files so until now it hasn't been a problem. It looks like it was added 4 years ago as part of the Quast module: https://github.com/ewels/MultiQC/commit/ac28c2544592ee6b149a05411d7a7703d65d0737 (https://github.com/ewels/MultiQC/pull/491) - I _think_ just to speed up the search, so hopefully removing it won't break anything.. 😟 

Then we can have `_summary.html` as the default search pattern which of course you could overwrite @berguner - I much prefer using tool defaults if at all possible.

ewels:
> Sure, I just opened it up the HTML in my browser and there is a "Coding consequences" pie chart. Interesting, maybe the HTML is a bit funky for this one?
> [SJETV093_D_G.annotated.somatic.vcf_summary.html.gz](https://github.com/ewels/MultiQC/files/6392856/SJETV093_D_G.annotated.somatic.vcf_summary.html.gz)
> 
> Generated with VEP v100.2 btw.

I tried this file locally and it worked fine for me too.. 🤔

a-frantz:
> > Sure, I just opened it up the HTML in my browser and there is a "Coding consequences" pie chart. Interesting, maybe the HTML is a bit funky for this one?
> > [SJETV093_D_G.annotated.somatic.vcf_summary.html.gz](https://github.com/ewels/MultiQC/files/6392856/SJETV093_D_G.annotated.somatic.vcf_summary.html.gz)
> > Generated with VEP v100.2 btw.
> 
> I tried this file locally and it worked fine for me too.. 🤔 (before pulling the latest commit)

That is very weird. I have no idea what could've been wrong with my local setup to cause an error like that. But seems to have resolved itself, so I think we can ignore it.

berguner:
It was easier than I thought actually, I managed to exclude `_summary.html` suffix from `fn_ignore_files` while keeping `*.html`. I hope this works for everyone.

multiqc/utils/config_defaults.yaml
```diff
@@ -232,6 +232,9 @@ fn_clean_exts:
   - ".mpileup"
   - ".primer_trim"
   - ".mapped"
+  - ".vep"
+  - "_vep"
+  - ".txt_summary"
```
ewels, line 237: We already have `".txt"` further up, so I don't think that we need this one - please correct me if I'm wrong.

```suggestion
```

ewels:
> It was easier than I thought actually, I managed to exclude `_summary.html` suffix from `fn_ignore_files` while keeping `*.html`. I hope this works for everyone.

Nice! That's a monster of a pattern 😆 Works for me though, so happy to stick with this for now (until we get some more HTML reports that need parsing!). Thanks for figuring it out 👍🏻

multiqc/utils/config_defaults.yaml
berguner: I wasn't sure if `".txt"` would clean up `".txt_summary"` so I added this to make sure. If you say `".txt"` would clean it we can remove it.

ewels:
Brilliant stuff - thanks all! 🎉

multiqc/utils/config_defaults.yaml
ewels: Yeah, `fn_clean_exts` removes the string and everything after it, wherever it appears in the sample name. `fn_clean_trim` only removes the string if it's at the end. We have `.txt` in the former so it should strip the `_summary` as well.

I removed this before merge 👍🏻

PR #1421, opened 2021-05-04
New module: Sambamba Markdup (Round 2)

gartician:

I did a hard-rest in my fork to pull in the newest updates, then re-added sambamba markdup. 

This is a new MultiQC module for [sambamba markdup](https://lomereiter.github.io/sambamba/docs/sambamba-markdup.html) to show the extent of read duplication in a sequencing library. Duplicate reads and unique reads are extracted out of the log files, and duplicate rate is calculated and displayed to the user in the general statistics table (columns = Duplicate Reads and Duplicate Rate) along with a bar graph section. This is my first time submitting a PR to a public project, so please let me know if I can improve on something!


gartician:
Thank you for being flexible with the new PR as I'm still figuring out how to properly submit to public projects. I'm sure the next time will be smoother! 
I've re-ran everything locally with the `--lint` and `--ignore` flags, and everything seemed to work okay. It's ready for testing

gartician:
Are there additional things I need to complete this PR?

ewels:
Great! Now it's just a case of me testing the code locally to complete my review. This usually results in me either asking you to change some stuff or these days I often just change it myself and ask you to check that you agree with what I did. I will get to this asap.

gartician:
Thank you @ewels for accepting the new module, I'm going to close this PR now!

ewels:
Hi @gartician, did I miss something?

I've had this tab open in my browser for a couple of weeks now (I've been away looking after children with colds for 10 days unexpectedly, sorry). But it's not merged yet, so not available in MultiQC yet.. Why close it now?

Phil

gartician:
Sorry, what are the steps left to do?

ewels:
I need to finish reviewing and merge the pull-request into the MultiQC repository 😄

gartician:
I git pulled all the new updates since last month and applied multiQC with lint to MultiQC_TestData. It seemed like vep, qc3c, kraken, and fastp modules broke along with other warning messages. I re-ran multiQC with sambamba and it continued to worked just fine `multiqc --lint MultiQC_TestData/data/modules/sambamba/ --filename sambamba.html`. How should we approach this? Should I wait until the other modules are fixed 🤔

ewels:
Yeah, sorry about that. I merged in test data to replicate other errors before the fixes were ready and you were unlucky to run the tests before those PRs were merged. They were all merged last night though, so if you pull in updates again now, the tests should pass.

gartician:
I was able to pull in the new changes to fix the previously broken modules. Everything ran smoothly on my local computer, and the new update should be ready to test again.

gartician:
Is there anything else I can add to push these changes?

ewels:
Not yet - I just need to finish the review. This is a top priority PR now and will be included before the next release. Hopefully within the next week or two.

Once I have done the review I will probably push a bunch of my own changes to this PR which I will want you to check over and confirm that you agree to. I may also ask for changes from you, but not always.

Phil

ewels:
Hi @gartician,

Just having my first go at reviewing this and wanted to check over the numbers with you. I wonder if we need to do some additional calculation with the numbers..

Looking at this bar plot:

image

Which corresponds to this log:

```

sambamba 0.6.8 by Artem Tarasov and Pjotr Prins (C) 2012-2018
    LDC 1.11.0 / DMD v2.081.2 / LLVM6.0.1 / bootstrap LDC - the LLVM D compiler (0.17.6git-0156298)

finding positions of the duplicate reads in the file...
  sorted 22892630 end pairs
     and 205923 single ends (among them 682 unmatched pairs)
  collecting indices of duplicate reads...   done in 4177 ms
  found 12288581 duplicates
collected list of positions in 0 min 47 sec
removing duplicates...
collected list of positions in 3 min 59 sec
```

So we are showing the raw read counts in this plot:

* `Sorted End Pairs`
* `Single Ends`
* `Single Unmatched Pairs`
* `Duplicate Reads`

But looking at the log, I think that `Single Unmatched Pairs` is a subset of `Single Ends`, right? And `Duplicate Reads` also a subset. So I wonder if the total read count we are showing in the bar plot sums to greater than the total number of reads for that sample...

It's a little tricky to be certain from the log file alone, without knowing more about the files used to generate them. Would you mind looking into this and checking over the numbers to make sure that they do sum to 100% please?

Thanks!

Phil

multiqc/modules/sambamba/markdup.py
```diff
@@ -0,0 +1,132 @@
...
+        # Find and load sambamba logs to markdup_data_dict.
+        # Regex for key phrases and calculate duplicate rate.
+        # Give user warning if redundant samples are found.
+
+        self.markdup_data_dict = dict()
+
+        for f in self.find_log_files("sambamba/markdup"):
+
+            # define sample name
+            sample_name = self.clean_s_name(f["s_name"].replace("markdup_", ""), f["root"])
```
ewels, line 31: This requires the user to specify a log filename that matches this format, right? It's not a standardised log filename from the tool itself?

I generally try to avoid doing this if at all possible, as users sometimes don't have any choice about the filenames that are generated to capture this kind of output.

Having said this, the log here looks pretty horrible with no mention of input filename. Maybe we just take `f["s_name"]` and use that, whatever it is. We should definitely remove the `replace` bit though and use the central MultiQC sample name cleaning configs to do this for us.

ewels:
Another point - all of the test data is currently for paired-end files. It would be great if you could generate some example data for SE datasets too. Presumably the log output will look different, so it would be good to confirm that the module works with both types.

Thanks!

Phil

multiqc/modules/sambamba/markdup.py
ewels: Ah ignore me, I already suggested that and then remembered that it's suffixes only for the config cleaning: https://github.com/ewels/MultiQC/pull/1417#discussion_r622517152

Sorry! But yeah, I think that we can probably still remove this completely if the filename convention is just something you're doing yourself manually and not coming from the tool.

gartician:
These are great feedback, I'll take a closer look in the next couple days.

ewels:
Sounds good 👍🏻  I posted on the Sambamba google groups mailing list too, but no reply yet: https://groups.google.com/u/1/g/sambamba-discussion/c/f1VfbxXp1ik

multiqc/modules/sambamba/markdup.py
gartician: For Sambamba Markdup, the user can freely name the log file. I removed it completely so now it's just whatever the user puts down: 

`sample_name = f["s_name"]`

This may ultimately come down to preference, but I'm wondering if it's easier to just drop the `markdup_` prefix in general? This format is not a standard by any means. I think dropping the prefix (and relying on the content search pattern) will better allow the end user to use and even "discover" this module more easily. Does this make sense?

gartician:
Single and paired-end reads in the new update will be supported!

I looked through the original sequencing files, and the reads in the bar plot sum to >100% because I forgot how proportions work for a second. I've updated the bar graph section to reflect the actual composition of a sample for both SE and PE data. So now the duplicate read counts in the bar graph matches the duplicate rate entry in the general statistics table. 

Regarding the Google groups comment, I also believe that "duplicates" is a subset of "end pairs", and "unmatched pairs" a subset of "single ends". It's the only way the logs add up to the original input BAM file. Oddly enough, all the read types don't sum to 100% but to 98-99%. This happens in both in-house and public data, as if Sambamba Markdup wasn't able to recognize the entire file 😕

ewels:
Sounds great! Thank you for investigating and figuring this out, much appreciated.. Remember that you can just commit + push and your updates will be appended to this pull request 😀 

Could the 98-99% thing be a rounding error? Or is this from the integer read counts? 🤔 Sounds a little odd, but hopefully won't be enough of a problem to cause issues for people. We can always write a note in the code and someone might figure it out in the future! I've done that before 😅

ewels:
Hi @gartician - just to confirm, I can't see any changes to the Python code in your most recent commits. Just the docs markdown.

I'm only mentioning this because the commit message in 29ec220f1c186a192f9adf3f98edbdaff62891cd sounds like maybe you thought it contained more changes than this..? 😀

multiqc/modules/sambamba/markdup.py
```diff
@@ -63,27 +64,91 @@ def get_markdup_stats(self, f):
...
+        Print debug msg if a sample neither fits singleEnds or PE scheme.
+        Output duplicate rate.
+        """
+
+        if (sortedEndPairs > 0) and (singleUnmatchedPairs > 0):
+            duplicate_rate = duplicateReads / (sortedEndPairs * 2 + singleEnds - singleUnmatchedPairs) * 100
+            duplicate_rate = round(duplicate_rate, 2)
+        elif (sortedEndPairs == 0) and (singleUnmatchedPairs == 0):
+            duplicate_rate = duplicateReads / singleEnds * 100
+            duplicate_rate = round(duplicate_rate, 2)
```
ewels: This code is the same as the above if `sortedEndPairs` and `singleUnmatchedPairs` are both `0` (`singleEnds - 0` is `singleEnds`, `sortedEndPairs * 2` is `0` if `sortedEndPairs` is `0`).

I think we can get rid of this `if` block..

ewels (commented):
Results are looking good! I started going through the code and ended up refactoring quite a bit to simplify it. The output should be identical, though please double check that you agree with my changes 👍🏻

ewels:
Many thanks for your work on this @gartician - great to get it supported by MultiQC! 🎉

gartician:
Thank you for guiding me through this process @ewels! I sure got a lot of practice out of it 😄 
If I were to contribute another module in the future, would I have permission to execute the tests?

ewels:
Great! Yes, the actions authorisation thing is new, to try to stop people using GitHub Actions to mine bitcoin. As you now have a PR merged it shouldn't affect you again on this repo.

PR #1423, opened 2021-05-05
AdapterRemoval: fixed bug when showing number of read types

fgvieira:

I believe the number of reads displayed is misleading.
As an example, after adapter removal trimming, a sample ends up with these reads on the output files:
```
reads.pair1.truncated.gz: 72,571,435
reads.pair2.truncated.gz: 72,571,435
reads.singleton.truncated.gz: 56,678 (55,102 + 1,576)
reads.collapsed.gz: 8,093,391
reads.collapsed.truncated.gz: 1,457,443
reads.discarded.gz: 3,486,928 (1,716,701 + 1,770,227)
```

However, the module currently shows:

adapt_removal

I believe that the first number should be `R1 + R2` (145,142,870).
The issue is that Adapter removal counts retained reads as the sum of `R1 + R2 + Singleton R1 + Singleton R2 + Full length Collapsed + Truncated Collapsed`, but the module is currently only doing `retained_reads = retained - Singleton R1 - Singleton R2`.
I think it would make more sense to also remove the number of collapsed reads from `retained_reads` and call it `Paired reads`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@
 - **Picard**
   - `RnaSeqMetrics` - fix assignment barplot labels to say bases instead of reads ([#1408](https://github.com/ewels/MultiQC/issues/1408))
   - `CrosscheckFingerprints` - fix bug where LOD threshold was not detected when invoked with "new" picard cli style. fixed formatting bug ([#1414](https://github.com/ewels/MultiQC/issues/1414))
+- **AdapterRemoval**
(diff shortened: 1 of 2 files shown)


fgvieira:
Just noticed that #1386 tried to fix the same thing, so I guess I'll close it...

ewels:
Yes, also a long issue in #1005 where we tried to figure this out. From memory, things are not quite as simple as they seem at first.

We sort of ground to a halt on that discussion and PR. Would be great if you could provide a fresh set of eyes 👀

PR #1439, opened 2021-05-27
Update figures, bug fix, and improve efficiency in biscuit module

jamorrison:

This pull request is to address comments made in PR #1167.
- *% Duplicate reads seems to be calculated twice*: I looked through the duplicate rate code and I couldn't seem to find where this was calculated twice. However, the percent aligned was calculated twice. I've updated the code to only calculate the various percentages needed once (occurs during log file parsing). Let me know if there was a particular part you had in mind for the duplicate rate being calculated twice.
- *I think that the Duplicate Rates and Cytosine Retention tables would be better presented as bar plots*: These two charts have been updated to be bar graphs, rather than tables.

I also updated how the cytosine retention log files were parsed to avoid a bug where values in scientific notation wouldn't be properly read from the file and turned into percentages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@
 
 #### Module updates
 
+- **biscuit**
+  - Duplicate Rate and Cytosine Retention tables are now bargraphs.
+  - Refactor code to only calculate alignment statistics once.
+  - Fixed bug where cytosine retentions values would not be properly read if in scientific notation.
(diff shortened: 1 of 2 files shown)


ewels (approved):
Looks great! I usually tell people not to turn off stacking for bar plots (I've just rewritten two other module PRs that did that), but in this case I think it works really well 😁 

No comments or changes from me, many thanks 👍🏻

PR #1452, opened 2021-06-15
Fixed typo causing wrong multipel of contigs from QUAST

### LINKED ISSUE #1442: Quast L75 is not in divided by 1000 (K) in the html result.
Hello,
I just ran the multiqc in some quast outputs and I noticed the L75 displayed in the html is the actual number of contigs instead of the number of thousands of contigs (K). This is not a serious issue, but seems easy to correct.

The quast version used is QUAST v5.0.2
And the multiqc version is MultiQC v1.10.1
Python version is 3.8
Installed with conda
running on a ubuntu server

ErikDanielsson:
Should close #1442. 


ewels (approved):
Great spot 🤦🏻 

I've thought for a while that maybe I should add in some checks for expected config keys in these functions to warn about this kind of thing. Never quite got around to it though..

Issue #1454, opened 2021-06-16
Add Pangolin support


drpatelh:
**Tool Details**

- Name of tool: `Pangolin`
- Tool homepage: `https://github.com/cov-lineages/pangolin`
- Tool description: `Software package for assigning SARS-CoV-2 genome sequences to global lineages.`
- Data suitable for MultiQC plot(s): `table`
- Most interesting data for General Stats table: `lineage`

**Tool output:**

- Log filename pattern: ~~`*.pangolin.*`~~ _[none]_
- File contents: Full example output files can be obtained from [here](https://nf-co.re/viralrecon/results#viralrecon/results-f0171324a60d1759ca7f5d02351372d97d22cc12/platform_illumina/variants/ivar/pangolin/)

ewels:
[SAMPLE_01.pangolin.csv](https://github.com/ewels/MultiQC/files/6663938/SAMPLE_01.pangolin.csv)

@drpatelh - can there be more than one sample in these files sometimes? (generally, not just within viralrecon). Or is it always just two lines?

And is it the tool that has the `*.pangolin.*` filename, or is that the pipeline? ie. can we be sure that all pangolin users will have these filenames for the outputs?

drpatelh:
There can indeed be more than one sample in that file - we just so happen to run it on a fasta file with a single entry. Sorry should have mentioned.

The output file can be called whatever you want too as indicated [here](https://github.com/nf-core/modules/blob/c8168bc3516d6f8c3dc6d47c5142073cee82934e/software/pangolin/main.nf#L34) but I put `*.pangolin.*` as a sensible default to sanitise maybe?

ewels:
If there is no standard, then I usually try to find files based on their file contents instead. eg. here probably could do:

```yaml
pangolin:
    contents: pangolin_version
    num_lines: 1
```

ewels:
@ErikDanielsson - fancy taking a crack at this?

ewels:
Merged - will be part of the next release 👍🏻

PR #1458, opened 2021-06-18
New Pangolin module

ErikDanielsson:
Wrote new module Pangolin module, according to specification given in #1454. It might be that there are too many columns in the generated table, since I didn't know what was essential and non-essential to the report.  

---


multiqc/modules/pangolin/pangolin.py
```diff
@@ -0,0 +1,157 @@
...
+            anchor="pangolin",
+            href="https://github.com/cov-lineages/pangolin",
+            info="is a software package for assigning SARS-CoV-2 genome sequences to global lineages.",
+        )
+
+        # Find and parse the sample files
+        self.pangolin_data = dict()
+        for f in self.find_log_files("pangolin", filehandles=True):
+            self.parse_pangolin_log(f)
+
```
ewels, line 35: You're missing the sample filtering and stop-point for zero samples here. So MultiQC will _always_ generate a report and a Pangolin section, even if there are no samples found.

This a pretty common mistake, there's a CI test for it and that's what's failing. It's easy to fix though, I'll push the required code now.

Docs: 

* https://multiqc.info/docs/#filtering-by-parsed-sample-names
* https://multiqc.info/docs/#no-files-found

multiqc/modules/pangolin/pangolin.py
```diff
@@ -0,0 +1,164 @@
...
+        self.add_section(
+            name="Run table",
+            anchor="pangolin-run-table",
+            description="Statistics gathered from the input pangolin files",
+            plot=self.pangolin_table(),
+        )
+
+    def parse_pangolin_log(self, fh):
+        contents = csv.DictReader(fh["f"])
```
ewels, line 53: This is in a core Python library? I don't think I've ever used this before 🤯 

Learn something new every day! 😆

multiqc/modules/pangolin/pangolin.py
```diff
@@ -0,0 +1,164 @@
...
+        for row in contents:
+            taxon_name = row["taxon"]
+            row.pop("taxon")
+            self.pangolin_data[taxon_name] = row
```
ewels, line 57: This is nice, but you're far too trusting. MultiQC is used by enough people in enough weird situations that we need to expect people to throw all kinds of random inputs at it.

```console
$ cat > not_pangolin.csv
This looks like it's from Pangolin (pangolin_version) but it's not! It's a lie!
Not really a csv
No data.
Should break horribly.

$ multiqc -f .

  /// MultiQC 🔍 | v1.11.dev0 (9b5a1d1)

|           multiqc | Search path : /Users/ewels/GitHub/MultiQC/TestData/data/modules/pangolin
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 7/7
╭────────────────────────────────── Oops! The 'pangolin' MultiQC module broke... ───────────────────────────────────╮
│ Please copy this log and report it at https://github.com/ewels/MultiQC/issues                                     │
│ Please attach a file that triggers the error. The last file found was: ./not_pangolin.csv                         │
│                                                                                                                   │
│ Traceback (most recent call last):                                                                                │
│   File "/Users/ewels/GitHub/MultiQC/MultiQC/multiqc/multiqc.py", line 604, in run                                 │
│     output = mod()                                                                                                │
│   File "/Users/ewels/GitHub/MultiQC/MultiQC/multiqc/modules/pangolin/pangolin.py", line 34, in __init__           │
│     self.parse_pangolin_log(f)                                                                                    │
│   File "/Users/ewels/GitHub/MultiQC/MultiQC/multiqc/modules/pangolin/pangolin.py", line 55, in parse_pangolin_log │
│     taxon_name = row["taxon"]                                                                                     │
│ KeyError: 'taxon'                                                                                                 │
│                                                                                                                   │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
|           multiqc | No analysis results found. Cleaning up..
```

ewels: You also need to clean the sample name, as described here: https://multiqc.info/docs/#custom-sample-names

And also log about identical sample names, as described here: https://multiqc.info/docs/#identical-sample-names

Fixing all of this in a commit I'm about to push 👍🏻

ErikDanielsson:
Thanks for all the tips! I'll have to look through all the changes you've made and give the docs a more thorough read.

PR #1459, opened 2021-06-18
Fix failing qc3C tests

ErikDanielsson:
Fixed failing qc3C tests due to different field names in the older version of the files:
- `unobs_fraction` was `unobs_frac` in the older version
- `separation_bins:vs_accepted` was `separation_bins:zvs_accepted`
For reference:  [new](https://github.com/ewels/MultiQC_TestData/blob/master/data/modules/qc3C/bam_out/DRR177157/report.qc3C.json) and [old](https://github.com/ewels/MultiQC_TestData/blob/master/data/modules/qc3C/v0.4a2/bam_out/DRR177157/report.qc3C.json)

---

### DIFF
--- a/multiqc/modules/qc3C/qc3C.py
+++ b/multiqc/modules/qc3C/qc3C.py
@@ -965,20 +965,8 @@ class MultiqcModule(BaseMultiqcModule):
                 )
             parsed = parsed[-1]
             # Check if fields in json file are named as in the old version
-            if "unobs_fraction" not in parsed:
-                parsed["unobs_fraction"] = parsed["unobs_frac"]
-            if "vs_accepted" not in parsed["separation_bins"]:
-                parsed["separation_bins"]["vs_accepted"] = parsed["separation_bins"]["zvs_accepted"]
-
         except json.JSONDecodeError:
             log.warning("Could not parse qc3C JSON: '{}'".format(f["fn"]))
-            return
-        except KeyError as ex:
-            log.error(
-                "The entry {} was not found in the qc3C JSON file '{}'".format(
-                    str(ex), os.path.join(f["root"], f["fn"])
-                )
-            )
             return
 
         s_name = self.clean_s_name(os.path.basename(f["root"]), os.path.dirname(f["root"]))
@@ -987,8 +975,12 @@ class MultiqcModule(BaseMultiqcModule):
 
         try:
             analysis_mode = parsed["mode"]
+            if "unobs_fraction" not in parsed:
+                parsed["unobs_fraction"] = parsed["unobs_frac"]
 
             if analysis_mode == "bam":
+                if "vs_accepted" not in parsed["separation_bins"]:
+                    parsed["separation_bins"]["vs_accepted"] = parsed["separation_bins"]["zvs_accepted"]
 
                 # set some variables to shorten the lines below
                 inf = parsed["classification"]["informative"]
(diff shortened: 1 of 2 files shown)


multiqc/modules/qc3C/qc3C.py
```diff
@@ -964,9 +964,22 @@ def _none_to(x, y):
...
+            if "unobs_fraction" not in parsed:
+                parsed["unobs_fraction"] = parsed["unobs_frac"]
+            if "vs_accepted" not in parsed["separation_bins"]:
+                parsed["separation_bins"]["vs_accepted"] = parsed["separation_bins"]["zvs_accepted"]
+
         except json.JSONDecodeError:
             log.warning("Could not parse qc3C JSON: '{}'".format(f["fn"]))
             return
+        except KeyError as ex:
+            log.error(
```
ewels, line 977: Why / when does this happen? Does it need to be an error log - can it be fixed?

If it's a version thing for qc3c then I would typically make it a debug message.

I would also quite like to drop some of the other log messages in this module down to debug. It's a bit noisy currently..

ErikDanielsson: It happens if the input json file contain neither the old nor the new key, I think it should be an error message rather than a debug message therefore.

ewels: Ok - any reason it's coming up so much from the files in https://github.com/ewels/MultiQC_TestData ?

ErikDanielsson: I hadn't realised that module was parsing two separate formats, of which one didn't have one of the field names I was checking for. This should be fixed with the latest commit.

ewels: Great! Thanks for checking 👍🏻

ewels:
Looking good - whilst we have the PR open, please could we move the `Enabled junction frequency plot for non-trivial digest` log to debug? That's the other one that pollutes the MultiQC log output a bit.

Then we're good to merge I think 🚀

ErikDanielsson:
Should be done now, and ready for merge.

multiqc/modules/qc3C/qc3C.py
```diff
@@ -964,6 +964,7 @@ def _none_to(x, y):
                     )
                 )
             parsed = parsed[-1]
+            # Check if fields in json file are named as in the old version
```
ewels, line 967: Not any more, right?

```suggestion
```

multiqc/modules/qc3C/qc3C.py
```diff
@@ -974,8 +975,12 @@ def _none_to(x, y):
 
         try:
             analysis_mode = parsed["mode"]
+            if "unobs_fraction" not in parsed:
+                parsed["unobs_fraction"] = parsed["unobs_frac"]
```
ewels, line 979:
```suggestion
            if "unobs_fraction" not in parsed and "unobs_frac" in parsed:
                parsed["unobs_fraction"] = parsed["unobs_frac"]
```

multiqc/modules/qc3C/qc3C.py
```diff
@@ -974,8 +975,12 @@ def _none_to(x, y):
...
 
             if analysis_mode == "bam":
+                if "vs_accepted" not in parsed["separation_bins"]:
```
ewels, line 982:
```suggestion
                if "vs_accepted" not in parsed["separation_bins"] and "zvs_accepted " in parsed["separation_bins"]:
```

ewels (approved):
Minor suggestions, feel free to ignore if you don't agree 👍🏻

Issue #1374, opened 2021-02-17
MultiQC Error on Dragen with Single-end reads


pchitsutti:
**Description of bug:**

Multiqc error upon reading Dragen metric file.
The samples there are single-end reads.

**MultiQC Error log:**

```

[INFO   ]         multiqc : This is MultiQC v1.10.dev0
[INFO   ]         multiqc : Template    : default
[INFO   ]         multiqc : Searching   : /gpfs/gpfs1/dls/apex/uat_data/analysis/apex/SINGLEREAD/haib20TEST01/SRqconly/batch_1/results
[ERROR  ]         multiqc : Oops! The 'dragen' MultiQC module broke... 
  Please copy the following traceback and report it at https://github.com/ewels/MultiQC/issues 
  If possible, please include a log file that triggers the error - the last file found was:
    /gpfs/gpfs1/dls/apex/uat_data/analysis/apex/SINGLEREAD/haib20TEST01/SRqconly/batch_1/results/6405-JL-0001_111/metrics/6405-JL-0001_111.mapping_metrics.csv
============================================================
Module dragen raised an exception: Traceback (most recent call last):
  File "/gpfs/gpfs1/dls/home/apex/.local/lib/python3.6/site-packages/multiqc/multiqc.py", line 579, in run
    output = mod()
  File "/gpfs/gpfs1/dls/home/apex/.local/lib/python3.6/site-packages/multiqc/modules/dragen/dragen.py", line 78, in __init__
    samples_found |= self.add_mapping_metrics()
  File "/gpfs/gpfs1/dls/home/apex/.local/lib/python3.6/site-packages/multiqc/modules/dragen/mapping_metrics.py", line 56, in add_mapping_metrics
    self.report_mapping_metrics(data_by_sample, data_by_rg_by_sample)
  File "/gpfs/gpfs1/dls/home/apex/.local/lib/python3.6/site-packages/multiqc/modules/dragen/mapping_metrics.py", line 87, in report_mapping_metrics
    plot=table.plot(data_by_rg, own_tabl_headers, pconfig={"namespace": NAMESPACE}),
  File "/gpfs/gpfs1/dls/home/apex/.local/lib/python3.6/site-packages/multiqc/plots/table.py", line 53, in plot
    return make_table(dt)
  File "/gpfs/gpfs1/dls/home/apex/.local/lib/python3.6/site-packages/multiqc/plots/table.py", line 157, in make_table
    val = header["modify"](val)
[...]

PR #1465, opened 2021-06-28
Fixed dragen generating errors when run on single-end output 

ErikDanielsson:
Handled the `dragen` module generating errors when run on single-end output due to numbers being replaced by `NA`, as suggested in #1374. Also added check that data is present before generation of bar graph in `mapping_metrics.py`.


ewels:
Thank you @ErikDanielsson !

Issue #1329, opened 2020-10-26
support linked adapters in cutadapt


qifei9:
**Is your feature request related to a problem? Please describe.**

For [linked adpaters](https://cutadapt.readthedocs.io/en/latest/guide.html#linked-adapters-combined-5-and-3-adapter) in cutadapt like:
```
cutadapt -a ^ADAPTER1...ADAPTER2 -o out.fastq.gz in.fastq.gz
```

the output will report the removed sequences from 5' and 3' end separately:

```
=== First read: Adapter background ===

Sequence: GGGGGGGGGGATATGCGATCGC...GTGTCTCAACGGTATCCGCAACT; Type: linked; Length: 22+23; 5' trimmed: 1629783 times; 3' trimmed: 115720 times


No. of allowed errors:
0-9 bp: 0; 10-19 bp: 1; 20-22 bp: 2

No. of allowed errors:
0-9 bp: 0; 10-19 bp: 1; 20-23 bp: 2

Overview of removed sequences at 5' end
length  count   expect  max.err error counts
20      60597   0.0     2       0 0 60597
21      389603  0.0     2       0 231655 157948
22      1043405 0.0     2       321979 609400 112026
23      126665  0.0     2       0 67428 59237
24      9513    0.0     2       0 0 9513



Overview of removed sequences at 3' end
length  count   expect  max.err error counts
3       28945   141088.9        0       28945
4       10134   35272.2 0       10134
5       4431    8818.1  0       4431
6       3462    2204.5  0       3462
7       3057    551.1   0       3057
8       2775    137.8   0       2775
9       2650    34.4    0       2596 54
10      2768    8.6     1       2535 233
11      2582    2.2     1       2385 197
...
```

however, multiQC only report the 3' end result in the Cutadapt -> Trimmed Sequence Lengths section.

**Describe the solution you'd like**

Perhaps make 2 plots, one for the 5' end result and one for the 3' end result?

**Describe alternatives you've considered**

**Additional context**

ewels:
Great spot @qifei9 - again agree that this should be fixed. Good work on finding these edge cases! 😅 

As with the other issue, if you have a complete example file that I can use then that would be ideal.

Phil

ewels:
Support for linked reads has been added by @ErikDanielsson in #1469 - two plots are now shown, with trimming lengths for 5' and 3'. This will be included in the upcoming v1.11 release.

Thanks for reporting @qifei9!

Phil

PR #1469, opened 2021-06-30
Added support for linked adapters in Cutadapt

ErikDanielsson:
Added support for Cutadapt linked adapter data, in accordance with #1329. If linked adapters are found, the plots for the two ends are displayed as windows of the trimmed length plot. 


ewels:
Nice! Did a high-speed test locally and the plot looks good. The axis doesn't change for `5' end Obs/Exp` though, maybe that needs a double-check?

Seeing the plot, I'm tempted to go back on myself a little too (sorry!). Maybe having two plots is good, each with two dataset buttons. So that both 5' and 3' plots are visible on report load. If it's not too big a change would that be ok?

Also - the `Filtered Reads` plot still isn't rendering (`Tried to make bar plot, but had no data`) and there are no columns in the General Stats table. So these two things need investigating too 👍🏻 

Thanks!

multiqc/modules/cutadapt/cutadapt.py
```diff
@@ -142,31 +146,44 @@ def parse_cutadapt_logs(self, f):
...
+                            self.cutadapt_length_exp[end][plot_sname][a_len] = float(r_seqs.group(3))
                             if float(r_seqs.group(3)) > 0:
-                                self.cutadapt_length_obsexp[plot_sname][a_len] = float(r_seqs.group(2)) / float(
+                                self.cutadapt_length_obsexp[end][plot_sname][a_len] = float(r_seqs.group(2)) / float(
                                     r_seqs.group(3)
                                 )
                             else:
                                 # Cheating, I know. Infinity is difficult to plot.
-                                self.cutadapt_length_obsexp[plot_sname][a_len] = float(r_seqs.group(2))
+                                self.cutadapt_length_obsexp[end][plot_sname][a_len] = float(r_seqs.group(2))
```
ErikDanielsson, line 184: I was puzzled by the axis as well at first, but it is because when the expected value is zero Obs/Exp is equal to Counts

ErikDanielsson:
I was puzzled by the axis also at first, but this is due to the expected value being zero: https://github.com/ewels/MultiQC/blob/21962696eecd707d13210ae387c1ead6c25d4956/multiqc/modules/cutadapt/cutadapt.py#L160-L166.

After investigating a bit I believe the general stats table and the bar graph isn't generated because the log file we are supplying isn't complete since it doesn't contain the summary section. I could that add to the test file from another file if it is desirable.

ewels:
Same case as the other PR - works fine when running with just the one example file, but breaks when a mixture of outputs is used..

ErikDanielsson:
That was a choice rather than a mistake, since I thought we only want to parse one format of Cutadapt data at once. It should be easy to fix though.

ewels:
Enough people use MultiQC that you can almost guarantee that _someone_ will do this, even if it's a bit of an edge case 😅 

I'm having a play with a quick refactor to fix it now.

ewels:
Ok, I think this should work now - end up with three potential sections: 5', 3' or not specified. No longer crashes with a mixture of samples and it should also pick up the 3' or 5' end for regular trimming from one of the lines further up in the log.

Still a couple of things to do though - the `% BP Trimmed` column in the _General Statistics_ table is not filled and the _Filtered Reads_ barplot is empty when using linked adapters. So I guess this needs some special handling to parse / calculate the totals needed in these outputs.

ewels:
Weird - CI tests are failing but works fine for me locally. Probably a sign that I should go to bed..

ewels:
Weird, I wonder what is special about `?` as a key? Especially as it worked for me locally but not on Actions.. 🤔

ewels:
ahhh, `--lint`. That explains it, thanks 👍🏻

ErikDanielsson:
The original issue though, was that the dictionaries weren't initialised with `?` as a key.

ewels (approved):
Awesome work @ErikDanielsson - this one ended up being quite tricky in the end!

Issue #1281, opened 2020-09-01
Handle disabled / missing FastQC sections.


schorlton:
**Description of bug:**

I configured FastQC using its `limits.txt` configuration file, located in the `Configuration` folder. I specifically disabled the `overrepresented_sequences` module by setting it to `1`. FastQC ran fine, MultiQC failed as was expecting this module. I believe MultiQC should run regardless of any FastQC modules missing or customized.

**MultiQC Error log:**
 

```python
 Module fastqc raised an exception: Traceback (most recent call last):
    File "/usr/local/lib/python3.8/site-packages/multiqc/modules/fastqc/fastqc.py", line 824, in overrepresented_sequences
      max_pcnt   = max( [ float(d['percentage']) for d in self.fastqc_data[s_name]['overrepresented_sequences']] )
  KeyError: 'overrepresented_sequences'
  
  During handling of the above exception, another exception occurred:
  
  Traceback (most recent call last):
    File "/usr/local/lib/python3.8/site-packages/multiqc/multiqc.py", line 569, in run
      output = mod()
    File "/usr/local/lib/python3.8/site-packages/multiqc/modules/fastqc/fastqc.py", line 117, in __init__
      self.overrepresented_sequences()
    File "/usr/local/lib/python3.8/site-packages/multiqc/modules/fastqc/fastqc.py", line 830, in overrepresented_sequences
      if self.fastqc_data[s_name]['statuses']['overrepresented_sequences'] == 'pass':
  KeyError: 'overrepresented_sequences'


```

**File that triggers the error:**

[test.fastq_barcode01_fastqc.zip](https://github.com/ewels/MultiQC/files/5156569/test.fastq_barcode01_fastqc.zip)

**MultiQC run details (please complete the following):**
* Command used to run MultiQC: 
* MultiQC Version: 1.9
* Operating System: Docker container: quay.io/biocontainers/multiqc:1.9--pyh9f0ad1d_0
* Python Version:  Python 3.8.2
* Method of MultiQC installation: conda->biocontainer

**Additional context**

ewels:
Nice catch, thanks! I'm kind of amazed that this hasn't been reported before. Should be easy to fix.

ewels:
Ok, so turns out that the module can already handle any FastQC section being skipped, _except_ `overrepresented_sequences`. So that probably explains why this hasn't come up before.

Fixed by @ErikDanielsson in #1472 - will be included in the upcoming v1.11 release.

[...]

PR #1472, opened 2021-06-30
Added handling of missing section in FastQC

ErikDanielsson:
Added check in `fastqc` to skip  `overrepresented_sequences`  section if it is missing from reports, to catch bug reported in #1281.


ewels (approved):
Great! I simplified a little but otherwise fab 👍🏻  Thanks!

ewels:
❤️ CI testing.. Fix works fine if only running on samples that are missing the section. But if you run on a mixture of samples where some do and some don't, then you get a new error..

PR #1483, opened 2021-07-05
Lima rename

Redmar-van-den-Berg:
Add some parsing magic for the Lima module, in combination with the `--replace-names` option. I like the behavior of putting the raw `barcode1--barcode2` sample names from Lima in their own table, to keep the General Statistics table clean. However, once the samples are renamed (presumably to match the rest of the MultiQC report), it is nicer to include them in the General Statistics table. This pull request adds this functionality.

1. If `--replace-names` is not used, all Lima count results will be put in their own table in the Lima section, using `barcode1--barcode2` naming.
2. Any sample that has been renamed using `--replace-names` is moved to the General Statistics table
3. If all Lima samples have been renamed, the Lima -> _Per sample count data_ section is not included in the report at all.
4. If there are Lima samples that have not been renamed, the Lima section points to the MultiQC Lima documentation on how to rename the samples.
5. The Lima documentation has been updated with instructions to use `--replace-names`.


ewels (approved):
Looks great! Haven't tested locally as it's a bit of a fiddly one, but the code changes are simple and the CI tests are passing so I'm happy 👍🏻 

I updated your branch (to be sure with the CI tests) and added mention of this to the changelog ✅

PR #1501, opened 2021-07-19
ran prettier on the changes

KyleStiers:


KyleStiers:
👍

ewels (commented):
Many thanks for the contribution here, but I'm afraid I'm going to reject this one.

There is already documentation about configuring plots if people don't like the default colours, and having the percentage selected by default is a fairly personal preference (plus clicking "Percentage" in the report is quite easy). None of this is really module specific.

Apologies for the long wait in reviewing!

PR #1503, opened 2021-07-23
Seqwho

### LINKED ISSUE #1388: seqwho
**Tool Details**

- Name of tool: seqwho
- Tool homepage: https://git.biohpc.swmed.edu/s181649/seqwho
- Tool description: Type Sequence files for species and assay
- Data suitable for MultiQC plot(s): fastqc like plots
- Most interesting data for General Stats table: Inferred Sequence Type, Inferred Species

**Tool output:**
TSV file of stats
PDF file of figures

- Log filename pattern: 
- File contents:

vsmalladi:

Close #1388


vsmalladi:
@ewels any ideas when this can be merged in?

ewels:
I'm finally back to working 100% now, so soon hopefully! Hoping to find time to work through all of the open PRs in the next few weeks if all goes well.

multiqc/utils/search_patterns.yaml
```diff
@@ -767,6 +767,8 @@ samtools/rmdup:
   shared: true
 sargasso:
   fn: "overall_filtering_summary.txt"
+seqwho:
+  fn: "SeqWho_call.json"
```
vladsavelyev, line 771: Looking at the SeqWho code, the output is written to stdout and thus can have any name. I wonder if we can detect a file based on its contents?

vladsavelyev (commented):
Merged with main and clean up - the code looks good now! Thanks a lot for the contribution, @vsmalladi.

One question about the search pattern.

multiqc/utils/search_patterns.yaml
vladsavelyev: Perhaps this would be unique enough:

```yaml
seqwho:
  contents: '  "Per Base Seq": ['
```

vsmalladi: ya that should work.

vladsavelyev (approved):

vsmalladi:
Thanks for getting this in

ewels (approved):
LGTM - the only bit I don't like is the `"stacking": None,`. This means the plots scale badly, as if you have 100 samples there are now 400 bars which gets unreadable very fast. However, I understand that it's wrong to stack percentages, so I'm not sure I have a better solution.

We've had similar situations before and talked about using a Parallel Coordinates plot. This isn't supported by MultiQC currently (beyond a hacky version with a line plot), but we may be able to add it in the future with Plotly now. That could be something to think about if we want to come back to this in the future.

No better solution that I can think of for now, so happy to merge as-is. Thanks @vsmalladi !

Issue #1461, opened 2021-06-26
Investigate parallelisation


ewels:
I've always dodged the issue of parallelisation due to the fact that MultiQC is relatively fast. If it only takes 15 seconds to run, what's the point of going to the effort of messing around with this? However, as the number of supported tools increases and the size of genomic datasets expands, MultiQC runs are starting to take longer in some cases. When you start getting 15-30 minute runs then this becomes more relevant.

I can think of two potential places where we may be able to parallelise. Thinking out loud:

- File search per module
    - Currently there is some dependency between patterns: if a file is found by one it will be skipped by the other patterns. Need to think about how to handle this.
    - May need some testing to see what the bottleneck is here. If it is disk access, then parallelisation may not have much effect.
- File search per file
    - As an alternative, could potentially batch up the discovered files and run all module searches on each batch? This would avoid cross-dependencies between module searches.
    - The added bonus of this is that it will also help users running a small number of modules on a large number of files.
    - Will probably need to refactor to do a simple `os.walk` to get the complete file list first and then run the searches in parallel once we have this
        - The filename-only patterns can probably be done during this initial walk still, as they should be lightning fast.
- Running the modules
    - Need to ensure order of final outputs
    - There _shouldn't_ be any cross-dependencies between modules, but someone will probably have some somewhere. May need to think about a mechanism to allow some kind of sequential running? But hopefully not.

Would be good to do some benchmarking early on to check the additional complexity is worth it.

PR #1508, opened 2021-07-30
Make file search parallel

ErikDanielsson:
Made MultiQC file search run in parallel. One can specify the number of processes that MultiQC can start with the `--cpu` flag. If not specified, it defaults to `1` and the file search is thus run serially. Specifying `0` cpus makes MultiQC autodetect the number of cores in the computer. Partially solves #1461. 

### Overview of implementation
The parallel file search works, roughly, in the following way. 
1. When a `--cpus` is set to a number other than `1` the `add_file` function in `report.py` creates as many processes as specified with the `--cpus` flag.
2. After the processes have been started, a thread for updating the progress bars is created in the main process.
3. Running file search:
  -  The main process divvies up the file paths found during the filename level search into chunks based on file size. A chunk is then given to each child process
  - The child processes run the necessary functions on the files in the current chunk and send their progress to the progress tracker thread continually. This updates the process' progress bar. 
  - When a child process is done with chunk, it asks the main process for another, and tells the progress tracker thread to update the main progress bar with the completed chunk.

4. When there are no more file path chunks left,  the main process sends exit signals to the child processes which in turn tell the progress tracker thread that they are finished. 
5. The main thread gathers the data from the child processes. 
6. The child processes and the progress tracking thread are joined with the main process.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### MultiQC updates
 
+Added `--cpu` flag for running file search in parallel ([#1461](https://github.com/ewels/MultiQC/issues/1461))
+
 ### New Modules
 
 ### Module updates
--- a/multiqc/multiqc.py
+++ b/multiqc/multiqc.py
@@ -10,6 +10,7 @@ Imported by __init__.py so available as multiqc.run()
 from __future__ import print_function
 
 import base64
+import multiprocessing
 import click
 from distutils import version
 from distutils.dir_util import copy_tree
@@ -175,6 +176,7 @@ logger = config.logger
 @click.option("-v", "--verbose", count=True, default=0, help="Increase output verbosity.")
 @click.option("-q", "--quiet", is_flag=True, help="Only show log warnings")
 @click.option("--profile-runtime", is_flag=True, help="Add analysis of how long MultiQC takes to run to the report")
+@click.option("--cpus", "cpus", type=int, default=1, help="Let MultiQC use this many processes")
 @click.option("--no-ansi", is_flag=True, help="Disable coloured log output")
 @click.version_option(config.version, prog_name="multiqc")
 def run_cli(
@@ -214,6 +216,7 @@ def run_cli(
     verbose,
     quiet,
     profile_runtime,
+    cpus,
     no_ansi,
     **kwargs,
 ):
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### MultiQC updates
 
+Added `--cpu` flag for running file search in parallel ([#1461](https://github.com/ewels/MultiQC/issues/1461))
```
ErikDanielsson, line 7:
```suggestion
Added `--cpus` flag for running file search in parallel ([#1461](https://github.com/ewels/MultiQC/issues/1461))
```

ewels:
I'm pretty sure that this worked for me when I tried it out over the summer, but now I get the following if I try to run with anything over 1 cpu:

```python
  File "/Users/phil/GitHub/MultiQC/MultiQC/multiqc/utils/report.py", line 363, in get_filelist
    worker_proc.start()
  File "/opt/miniconda3/envs/py3.8/lib/python3.8/multiprocessing/process.py", line 121, in start
    self._popen = self._Popen(self)
  File "/opt/miniconda3/envs/py3.8/lib/python3.8/multiprocessing/context.py", line 224, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "/opt/miniconda3/envs/py3.8/lib/python3.8/multiprocessing/context.py", line 284, in _Popen
    return Popen(process_obj)
  File "/opt/miniconda3/envs/py3.8/lib/python3.8/multiprocessing/popen_spawn_posix.py", line 32, in __init__
    super().__init__(process_obj)
  File "/opt/miniconda3/envs/py3.8/lib/python3.8/multiprocessing/popen_fork.py", line 19, in __init__
    self._launch(process_obj)
  File "/opt/miniconda3/envs/py3.8/lib/python3.8/multiprocessing/popen_spawn_posix.py", line 47, in _launch
    reduction.dump(process_obj, fp)
  File "/opt/miniconda3/envs/py3.8/lib/python3.8/multiprocessing/reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
AttributeError: Can't pickle local object 'get_filelist.<locals>.FileAdderProcess'
```

I guess that it's related to this SO post: https://stackoverflow.com/questions/52265120/python-multiprocessing-pool-attributeerror

ewels:
@ErikDanielsson thinks that the pickling stuff could be difficult to resolve. May be better to close this PR and try to reimplement using a different method.

vladsavelyev:
The file search is much less of a bottleneck now, ut's heavily optimized and we try to avoid reading large chunks of files. So perhaps this is not a priority now!

PR #1514, opened 2021-08-09
Add PCT_SELECTED_BASES to HsMetrics output

nh13:
The `PCT_SELECTED_BASES` is probably the most useful output of this metric (as per the original author @tfenne), so not making it hidden by default.


ewels (approved):
Sounds good, thanks 👍🏻

PR #1524, opened 2021-08-16
Add citations

ErikDanielsson:
Added `doi` attribute to the `BaseMultiQCModule` for collecting citations of the invidual tools. If set, it tries to fetch the citation which is then added in the intro section for the module. If the request fails, the doi link is added instead. 


ewels:
Whilst this code works really nicely to get the full citation with a request, I have just stripped this bit of code. I'm a little paranoid about doing any web requests as they can be a bit unpredictable and slow. Considering that we hope to collect DOIs for most modules, this could start to get called a lot of times in a single run and add a fair bit of run time.

Instead, I think we should be able to dynamically pull the same info client-side in the web browser. This will come at no cost to report generation time and give effectively the same output.

ewels:
ok, implemented javascript solution which gives a popover when clicked:

image

Not super well tested, but seems to be working ok so far..

ewels:
Todo:

- [ ] Take an array of DOIs, not just one
- [ ] Test with no internet
- [ ] Test with broken DOI
- [ ] Test with other weird inputs
- [ ] Test with weird papers (eg. zenodo / url only etc)

PR #1526, opened 2021-08-17
Fix path filters

### LINKED ISSUE #1308: `path_filters` requires leading `./` in paths
MultiQC version: 1.19

I want to restrict the logs used by a module to those found in a certain directory. I’m using a configuration like this:

```yaml
module_order:
  - cutadapt:
      path_filters:
        - "log/step2/*.log"
```
With this, the log file will not be found when running `multiqc -c config.yaml .`.

If I specify the filter as `./log/step2/*.log`, that is, with a leading `./`, it will work.

Digging into this, it appears that the `fnmatch.fnmatch` at 
https://github.com/ewels/MultiQC/blob/7584e64e10885f38367628a5a4a9033d48e82011/multiqc/modules/base_module.py#L110 fails because the first parameter `report.last_found_file` is `./log/step2`. This in turn comes from the `os.walk` call at https://github.com/ewels/MultiQC/blob/7584e64e10885f38367628a5a4a9033d48e82011/multiqc/utils/report.py#L192 , where the `root` is `.` (because I ran with `multiqc .` I assume).

I believe a solution could be to wrap (some of?) the calls to `os.path.join(root, something)` with an `os.path.normpath(...)`, which will convert `./log/file.log` to just `log/file.log`. For backwards compatibility, that same transformation should probably be applied also to the pattern.

I spent some time trying to understand the code so I could prepare a patch, but the function is quite long and I’ve given up for the moment (some refactoring would be good here).

ErikDanielsson:
Made path filters without complete paths work correctly by prefixing analysis directories when filtering. Should close #1308. 


ewels (approved):
image

PR #1527, opened 2021-08-17
Add table level conditional formatting

### LINKED ISSUE #761: New feature request: Conditional formatting for whole tables
Hi Phil,
as discussed on Gitter, just requesting a new feature to allow for conditional formatting at the **table level** (instead of per-column).

More specifically, I would like to have all values greater than X in my custom table highlighted in red (which I can do per-column using the `table_cond_formatting_rules` in the config file). However in my use case, the column names will be different for every new run, so I can't use that.

From our chat:
> _Phil:_
> Conditional formatting for whole table - I don’t think that this is possible currently.. Will need to add it as a new feature.
> Should be doable - just need a special case “column name” (eg. `all_columns`)

Thanks a lot in advance!
Roxane

ErikDanielsson:
Added table level conditional formatting, should close. For example:
```yaml
table_cond_formatting_rules:
    table_name:
        all_columns:
            pass:
                - lt: 0.5
            warn:
                 - gt: 6
             fail:
                 - gt: 8
```

Should close #761.


ewels:
Thanks for this! I had a look and it works exactly as you say. However, I don't think we really need the `all_columns` thing in this case and I came up with a simpler implementation in 575f39a58bd037c98209fd224d09bc86ed1c3144, so I'll go for that instead if that's ok.

Phil

PR #1528, opened 2021-08-17
Always save custom content data to file

### LINKED ISSUE #1194: Generate custom content files for all plots, not just tables
Would it be possible to add the section `anchor` or `id` to name the custom content files in `multiqc_data/`

e.g. `mqc_mqc_mplplot_kgrbwiulmq_1.txt` could use [this](https://github.com/nf-core/viralrecon/blob/56e7d84b62f80575b14e3c0d1ccdbb8fd51ff73e/assets/headers/ivar_variants_header_mqc.txt#L1) and be called `mqc_mqc_mplplot_ivar_variants.txt` instead.

ErikDanielsson:
Updated `custom_content` to always write the data to file. I couldn't figure out how to handle heatmap data however, since it is given as a nested list rather than a dictionary. Should close #1194 eventually.


ewels:
> I couldn't figure out how to handle heatmap data however, since it is given as a nested list rather than a dictionary

Yeah I've hit this before with heatmaps. Also some modules create more complicated data structures that don't fit into `.tsv` files. I think it's fine - the export works fine if using `-k yaml` or `-k json` and I've thought about saving all of these output formats anyway.

ewels:
Ahh, it actually crashes with the heatmap with tsv files.. Ok then we just skip those then.

ewels:
Ok and whilst this PR fixes the second part of the issue - saving the files even when not exporting the plots - it doesn't solve the first part. The random filenames. Now we get two versions of all the data when running with `--export`, a random filename and a nice filename..

ewels:
Ok, done I think. No duplicate files for custom content now, and plot IDs set so no more random alphanumeric sequences.

PR #1536, opened 2021-08-19
Run CI tests on single module on single module update

ErikDanielsson:
Tried to rewrite GitHub actions to run only the changed module if a single module is changed by the PR. The logic for this is shown in this boolean:
```Python
base_module_unchanged and nbr_changed_modules == 1 and 
(nbr_changed_files == 1 or nbr_changed_files == 2 and changelog_changed)
```
So we basically add the `-m` flag if only one module has changed, and the modules is not the base module. I'm not sure that this will work -- I'm finding it a bit difficult testing gh actions. 


ewels (approved):
I think this looks awesome!

ewels:
Hmm, maybe spoke too soon. Can't get it to work, seems to always run without `-m`.

I thought it could have been because of a typo, but #1759 doesn't seem to have fixed it.

Needs more work, too late for me to figure it out now. Will come back to it.

PR #1232, opened 2020-06-19
Add new module to support DRAGEN-FastQC

bnbowman:
- [X] This comment contains a description of changes (within reason)
 - [X] `CHANGELOG.md` has been updated

Illumina is in the process of updating it's DRAGEN Bio-It platform that, among other things, adds a new feature to generate hardware-accelerated primary analysis statistics similar to the widely used FastQC tool.  This PR creates a new module that parses the outputs from that new feature and uses it to replicate the plots that a user would expect to see from FastQC, or from MultiQC's existing FastQC module.

NOTE #1: This module doesn't yet properly support sample-quality flags, but it needs to before it's merged.  But I wanted to get this PR up to start the discussion before the weekend, while I'm still working on that.

NOTE #2:  As part of supporting full graphical parity with FastQC, this PR also includes new plotting functionality to create box-and-whisker plots.   I imagine these need a fair amount of work still, particularly for things like pconfig support, but I think some code-review would help differentiate the must-haves from nice-to-haves

 - [X] There is example tool output for tools in the https://github.com/ewels/MultiQC_TestData repository or attached to this PR
 - [X] Code is tested and works locally (including with `--lint` flag)
 - [X] `docs/README.md` is updated with link to below
 - [X] `docs/modulename.md` is created
 - [X] Everything that can be represented with a plot instead of a table is a plot
 - [X] Report sections have a description and help text (with `self.add_section`)
 - [X] There aren't any huge tables with > 6 columns (explain reasoning if so)
 - [X] Each table column has a different colour scale to its neighbour, which relates to the data (eg. if high numbers are bad, they're red)
 - [X] Module does not do any significant computational work

PR #1541, opened 2021-09-04
Dockerfile upgrade

alexiswl:
  * Use alpine:3.13 as base image for multi-arch build of dockerfile
    * Multiarch images can be built through `docker buildx` like so on a docker swarm running on multiple platforms:
      ```bash
      docker buildx build --platform linux/amd64,linux/arm64 --progress plain --push --tag ewels/multiqc:latest .
      ```  
    *  See [docker buildx](https://docs.docker.com/buildx/working-with-buildx/) for more info
  * Maintain *-dev apk packages through installation of multiqc (for packages without pre-compiled binaries (i.e kiwisolver)
    * Although not an issue in the existing master branch of multiqc, the dragen-fastqc pr (#1232) has [kiwisolver](https://pypi.org/project/kiwisolver/) as a requirement in [setup.py](https://github.com/ewels/MultiQC/pull/1232/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7). Kiwisolver currently has the following [bug](https://github.com/nucleic/kiwi/issues/102) that means the build apk packages must be present at installation time of multiqc.  
  * Added non-root USER for safer container invocation
    * See more info [here](https://engineering.bitnami.com/articles/why-non-root-containers-are-important-for-security.html)
  * Moved from ENTRYPOINT to CMD for CWL conformance.
    * See more info [here](https://www.commonwl.org/v1.1/CommandLineTool.html#DockerRequirement)


Dockerfile
```diff
@@ -1,19 +1,85 @@
...
+    ln -fs /usr/include/locale.h /usr/include/xlocale.h && \
+    ln -fs /usr/bin/python3 /usr/local/bin/python && \
+    ln -fs /usr/bin/pip3 /usr/local/bin/pip && \
+    echo "Install deps through pip" 1>&2 && \
+    pip3 install -v --no-cache-dir --upgrade \
+      pip && \
+    pip3 install -v --no-cache-dir \
+      numpy && \
+    pip3 install -v --no-cache-dir \
+      matplotlib=="$MATPLOTLIB_VERSION" && \
```
ewels, line 49: Do we need numpy and matplotlib to be installed explicitly here? Can this not be handled by the main MultiQC requirements?

Dockerfile
```diff
@@ -1,19 +1,85 @@
...
+# Add the MultiQC source files to the container
+USER "${USER_NAME}"
+
+# Set default workdir to user home
+WORKDIR "/home/${USER_NAME}"
+  
 # Set up entrypoint and cmd for easy docker usage
-ENTRYPOINT [ "multiqc" ]
-CMD [ "." ]
+CMD [ "multiqc" ]
```
ewels, line 84: I think that the docs will need updating to reflect new usage here.

ewels (approved):
Thanks for this - looks good, couple of minor questions but otherwise no problems.

Out of curiosity, does the final image size end up being about the same?

Would also be nice to check for a couple of other bonus packages if we're building everything ourselves. For example, `ps` for Nextflow if it's not already there..

Dockerfile
alexiswl: > Out of curiosity, does the final image size end up being about the same?

Latest image on ghcr.io for multiqc is 206 Mb, this one is about 266 Mb

> Do we need numpy and matplotlib to be installed explicitly here? Can this not be handled by the main MultiQC requirements?

Tested and looks like we can drop these. I had them in explicitly as I'd thought since you'd set the base image as a specific version as matplotlib, you'd wish to do the same here. 

When matplotlib is just a dependency of multiqc, it appears to compile (which reflects the docs here https://matplotlib.org/stable/users/installing/index.html).  I'm not sure I've seen this when I've specified it on its own pip install line.

Dockerfile
alexiswl: This bit can be dropped since we try and use [biocontainers](https://quay.io/repository/biocontainers/multiqc) for our CWL workflows anyway

ewels:
* Removed the matplotlib and numpy installations
* Removed the `CMD` completely
* Updated the docs

ewels (changes requested):
Hmm, I can't get it to build locally - not sure if I broke something or if it was problematic from the start (I'm on a MacBook M1, so docker can have issues sometimes).

Could try adding `--prefer-binary` from the [docs](https://matplotlib.org/stable/users/installing/index.html) you linked maybe? Bedtime for me..

ewels:
Ok, it builds ok on GitPod 🎉 

I's a fair bit bigger though, half the size again:

```
gitpod /workspace/MultiQC (dockerfile-update) $ docker image ls
REPOSITORY      TAG       IMAGE ID       CREATED          SIZE
multiqc/new     latest    c07c030d8fe4   24 seconds ago   336MB
ewels/multiqc   latest    6444a526efaa   3 months ago     204MB
```

Would be nice to try to trim it down a little more, though it's not super important.

It would be cool to set up automated multi-arch builds! This [blog post](https://dev.to/cloudx/multi-arch-docker-images-the-easy-way-with-github-actions-4k54) makes it sound very doable..

ewels:
It would also be nice to use a multi-stage build instead of the huge chained run commands.

ewels:
ok, I messed around with a few ideas and ended up with https://github.com/ewels/MultiQC/pull/1836 which uses one of the official Python base images.

This still ends up bigger than the existing image, but by a smaller margin:

```
gitpod /workspace/MultiQC (dockerfile_python_base) $ docker image ls
REPOSITORY      TAG           IMAGE ID       CREATED         SIZE
ewels/multiqc   python-slim   ce6cc07ddaf2   7 minutes ago   277MB
ewels/multiqc   latest        6444a526efaa   3 months ago    204MB
```

I think I prefer this as going directly from Python feels a bit simpler and hopefully more robust. It's still a multi-arch base image, so can hopefully still set up multi-arch MultiQC images.

What do you think @alexiswl?

alexiswl:
This looks great!

ewels:
Multi-arch effort started in https://github.com/ewels/MultiQC/pull/1541

ewels:
Multi-arch effort started in https://github.com/ewels/MultiQC/pull/1837

Issue #1551, opened 2021-09-17
Add CheckQC module


maleasy:
### Name of the tool

CheckQC

### Tool homepage

https://github.com/Molmed/checkQC

### Tool description

CheckQC is a program designed to check a set of quality criteria against an Illumina runfolder

### Tool output

CheckQC's `--json` flag prints JSON formatted results to the stdout, see this example:
[checkqc.json.zip](https://github.com/ewels/MultiQC/files/7184631/checkqc.json.zip)

### Data suitable for MultiQC plot(s)

CheckQC includes up to six QC checks (handlers), and only lists the samples/lanes that to not pass a check. There could be one plot per check

### Most interesting data for the General Stats table

Samples with too few reads, listed in the JSON output in the "ReadsPerSampleHandler" section:

```json
  "ReadsPerSampleHandler": [
    {
      "type": "error",
      "message": "Number of reads for sample 4 was too low on lane 1, it was: 17.456 M",
      "data": {
        "lane": 1,
        "number_of_samples": 6,
        "sample_name": "4",
        "sample_reads": 17.455795,
        "threshold": 22.5
      }
    }
  ]
```

maleasy:
Created a pull request for such a module #1552

ewels:
PR merged. Thanks!

maleasy:
👍

PR #1552, opened 2021-09-17
Checkqc module

maleasy:
Initial version of a module for CheckQC (see https://github.com/ewels/MultiQC/issues/1551#issue-999166331)


maleasy:
The *MultiQC - Linux / Linux -Python 3.x* checks fail in subtest *Test for missing CSPs* with this output result:



```
Run python test/print_missing_csp.py --report full_report.html --whitelist CSP.txt
The following scripts are missing from CSP.txt
  'sha256-TUYHIjQsABv4n1G4GdIAz0ZljvuXPEvQ9T20Sok7TvE=' # ////////////////////////////////////////////////// Base JS for MultiQC Reports//
  'sha256-teMLGfDW72TRam2f0Fnj53uchjf3tQR9N0xo+hR2PXU=' # ////////////////////////////////////////////////// MultiQC Table code///////////
  'sha256-AEcY37NW7iIjmlzDOPZvuAorKHxXZxsECPnuSP52fB8=' # ////////////////////////////////////////////////// HighCharts Plotting Code/////
  'sha256-iPIReyQyAoheerXqjtulr8WN7lPMWacMVU4awy5CTJg=' # ////////////////////////////////////////////////// Static MatPlotLib Plots Javas
  'sha256-KGgqQTL/PWbF29mNXwFnepbDYBEnTKdMtejlNhEiXfs=' # ////////////////////////////////////////////////// MultiQC Report Toolbox Code//
  'sha256-RO0nmf6TEJtrp+R4JF8N4UrF+A2Hfkb8BKxyS7nM5go=' # // Return JS code required for plotting a single sample// RSeQC plot. Attempt to
  'sha256-kH3IklvmfvBiZprdCgEZqGuZjhiGEF5V3tHncoRciQI=' # // Javascript for the FastQC MultiQC Mod///////////////// Per Base Sequence Cont
```

I am not sure where the problem is. Can you please advise?

Thanks!

ewels:
> I am not sure where the problem is. Can you please advise?

In short, it wasn't your fault - was failing for all MultiQC tests. It's since been fixed, I just pulled in updates from `master` and hopefully tests will pass now..

ewels:
x-ref test data PR: https://github.com/ewels/MultiQC_TestData/pull/218

ewels:
Thanks for this @maleasy - will be great to get a CheckQC module in! The tool was written by colleagues of mine so it's been on my radar to add for ages.. Maybe @matrulda you could cast your eyes over this PR if you get a few minutes?

I've just had a quick stab at getting the CI tests to pass and have moved them on a little. But looks like it's still catching some issues, such as the module not [returning a `UserWarning`](https://multiqc.info/docs/#no-files-found) when no files are found (or possibly not filtering all samples out with `--ignore-samples`). Now that the CSP tests are fixed hopefully you can push this PR on now to get the little green tick @maleasy ✅ 

Phil

matrulda:
Wow! Cool, I will have a look and try it out 😄 👍

maleasy:
@ewels @matrulda I refactored the code primarily to make sure that all samples are filtered out on `--ignore-samples` before adding content. Passes all tests now.

matrulda:
@maleasy Just wanted to give you a life sign. This week has been packed with meetings, I've started trying the module out and will hopefully have time to finish the review at the start of next week.

multiqc/modules/checkqc/checkqc.py
```diff
@@ -0,0 +1,571 @@
...
+            if len(self.log_files) == 1:
+                sample = self.clean_s_name(lane, f)
+            else:
+                sample = self.clean_s_name(f"{lane} ({run})", f)
+
+            if self.is_ignore_sample(sample):
+                continue
+            self.add_data_source(f, sample)
+            p_undetermined = issue["data"]["percentage_undetermined"]
+            threshold = issue["data"]["threshold"]
```
matrulda, line 414: CheckQC [reports](https://github.com/Molmed/checkQC/blob/36070f4d0227fec42ec6001a6d46bc9db198a856/checkQC/handlers/undetermined_percentage_handler.py#L56) that percentage undetermined couldn't be computed if `yield == 0`, in that case `data` only contains `lane` and `percentage_underermined` (which is `N/A`).  It would be nice if that case was handled.

matrulda (commented):
Great work and thanks for doing this! 👏 Left some comments that you can have a look at.

multiqc/modules/checkqc/checkqc.py
```diff
@@ -0,0 +1,571 @@
...
+        if "ErrorRateHandler" in content:
+            self.get_error_rate_data(content["ErrorRateHandler"], run, f)
+        if "UndeterminedPercentageHandler" in content:
+            self.get_undetermined_percentage_data(content["UndeterminedPercentageHandler"], run, f)
+        if "UnidentifiedIndexHandler" in content:
+            self.get_unidentified_index_data(content["UnidentifiedIndexHandler"], run, f)
+        return general_stats
+
+    def _get_unique_runname(self, content):
+        base_run_name = content["run_summary"]["instrument_and_reagent_type"]
```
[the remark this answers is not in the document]
maleasy: I agree that the both solutions, filename or instrument type are not ideal. With the filename I see a potential problem when the same filename is present in several subdirectories. But I guess we could do the same thing as now and add suffixes `_1`, `_2`, and so forth. I do not have a strong preference either way.

multiqc/modules/checkqc/checkqc.py
```diff
@@ -0,0 +1,571 @@
...
+        self.log_files = list(self.find_log_files("checkqc"))
+
+        general_stats = dict()
+        for f in self.log_files:
+            raw_content = json.loads(f["f"])
+            genstats = self.parse_checkqc_json(raw_content, f)
+            if genstats:
+                general_stats.update(genstats)
+            self.add_data_source(f)
+            # self.add_general_stats(f)
```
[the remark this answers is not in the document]
maleasy: Done in ead255e

multiqc/modules/checkqc/checkqc.py
```diff
@@ -0,0 +1,571 @@
...
+            self.get_undetermined_percentage_data(content["UndeterminedPercentageHandler"], run, f)
+        if "UnidentifiedIndexHandler" in content:
+            self.get_unidentified_index_data(content["UnidentifiedIndexHandler"], run, f)
+        return general_stats
+
+    def _get_unique_runname(self, content):
+        base_run_name = content["run_summary"]["instrument_and_reagent_type"]
+        run_name = base_run_name
+        i = 1
+        while run_name in self.checkqc_data:
```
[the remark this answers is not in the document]
maleasy: Fixed in ead255e

multiqc/modules/checkqc/checkqc.py
```diff
@@ -0,0 +1,571 @@
...
+        Args:
+            issues (dict): JSON dict from CheckQC containing UnidentifiedIndexHandler results
+            run (str): name of sequencing run
+            f (dict): MultiQC log file dict
+        """
+        idx_to_lane_to_rep = {}
+        threshold = None
+
+        for issue in issues:
+            # is_error = issue["type"] == "error"
```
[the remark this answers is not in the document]
maleasy: Done in ead255e

multiqc/modules/checkqc/checkqc.py
maleasy: Thanks for the pointer, fixed in ead255e, this case does not break parsing the checkqc JSON now. I added it to the plot as an empty entry so the user can at least see that something is up with the respective sample. Not sure how to transport more info, e.g. that yield is zero, because no legend is shown for a bar of zero length.

multiqc/modules/checkqc/checkqc.py
```diff
@@ -0,0 +1,571 @@
...
+            m = re.match(
+                r"Index: ([ATGC+]+) on lane: (\d+) was significantly "
+                r"overrepresented \(([0-9.]+)%\) at significance "
+                r"threshold of: ([0-9.]+)%\.",
+                msg,
+            )
+            if not m:
+                # TODO: Could add handling of other message types
+                continue
+            index, lane, overrep, threshold = m.groups()
```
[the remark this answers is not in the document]
maleasy: you are right, removed in ee684b4

multiqc/modules/checkqc/checkqc.py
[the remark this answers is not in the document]
matrulda: Yeah, that's true. I think the current solution is fine, just wanted to mention it.

multiqc/modules/checkqc/checkqc.py
matrulda: checkqc_module_no_yield
I tried it out and it looks like this. I think this is fine, but it would nice if the message you wrote ("Yield is 0, no undetermined percentage computation possible") was communicated. I'm not super familiar with how MultiQC modules work, so can't really suggest how to best implement it. Would it be helpful if I send you the json file I used?

multiqc/modules/checkqc/checkqc.py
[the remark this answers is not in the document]
maleasy: Ok, then we'll leave it with the instrument type for the time being

multiqc/modules/checkqc/checkqc.py
maleasy: In your test, was the yield 0 in all lanes? I had not thought of this possibility. The corresponding JSON file would be helpful, thanks! Below is a plot where only one lane has yield 0 ( (lane 8 on run hiseq2500_rapidhighoutput_v4_1), I was hoping the legend would include the "Yield is 0" message, but it does not for a value of 0.
I have to see how to do this best, I was hoping to somehow include it in the barplot, but probably a separate plot or section is needed. 

checkqc_undetermined-percentage-plot

matrulda: Yeah, It's basically a failed MiSeq run, so only one lane. It's quite clear from the other handlers that something is wrong, so I don't see it as super necessary to fix, just a nice to have. I can't upload the file here, so send me an e-mail (you can find my address on my profile) if you want me to send the file. But like I said, I'm also fine with approving this as is.

maleasy: I created a dummy JSON with an entry like this for testing, I think that should cover it (correct me if I'm wrong):

```json
    "UndeterminedPercentageHandler": [
        {
            "type": "error",
            "message": "The percentage of undetermined indexes was to high on lane 8, it was: N/A",
            "data": {
              "lane": 1,
              "percentage_undetermined": "N/A"
            }
        }
    ],
```

In 1d28323 I added an additional section for cases like this, it contains a table listing the lanes with yield zero.

multiqc/modules/checkqc/checkqc.py
```diff
@@ -472,6 +480,30 @@ def add_undetermined_percentage_section(self):
...
+            run (str): name of sequencing run
+        """
+        data = self.checkqc_data["UndeterminedPercentageHandler_ZeroYield"]
+
+        pconfig = {"id": "checkqc_zero-yield-table", "table_title": "CheckQC: Lanes with Yield 0", "scale": "Reds"}
+
+        headers = OrderedDict()
+        headers["lane"] = {"title": "Lane", "description": "Sequencing lane", "format": "{:,.0f}"}
+        if self.onerun:
+            headers["run"] = {"title": "Run", "description": "Sequencing run"}
```
matrulda, line 498: Shouldn't this be set if `not self.onerun`?

multiqc/modules/checkqc/checkqc.py
```diff
@@ -35,6 +35,7 @@ def __init__(self):
         self.runs = set()
 
         self.log_files = list(self.find_log_files("checkqc"))
+        self.onerun = len(self.log_files) == 1
```
matrulda, line 38: Nice :+1:

multiqc/modules/checkqc/checkqc.py
```diff
@@ -472,6 +480,30 @@ def add_undetermined_percentage_section(self):
...
+        """Add a section for lanes with zero yield
+
+        Creates a table with the lane names with zero yield
+
+        Args:
+            run (str): name of sequencing run
+        """
+        data = self.checkqc_data["UndeterminedPercentageHandler_ZeroYield"]
+
+        pconfig = {"id": "checkqc_zero-yield-table", "table_title": "CheckQC: Lanes with Yield 0", "scale": "Reds"}
```
matrulda, line 493: I think it would be less confusing to set `col1_header` to something else than the default "Sample Name". Maybe `Lane Number`? It's a bit overkill since the other column also is specifying lane. I guess only a list is needed really. Is it possible to make a table with only the first column? I've only made lists in MultiQC reports by adding a HTML list using the custom content module before.

multiqc/modules/checkqc/checkqc.py
matrulda: Yeah, exactly, only the message is a bit different, but it isn't used in the code as far as I can see.
```
    "UndeterminedPercentageHandler": [
        {
            "type": "error",
            "message": "Yield for lane: {} was 0. No undetermined percentage could be computed.",
            "data": {
                "lane": 1,
                "percentage_undetermined": "N/A"
            }
        }
    ],
```
(Oops, we have a :bug: {})

multiqc/modules/checkqc/checkqc.py
maleasy: You are right, the table columns are somewhat redundant. I don't think it is possible to only have the first column, because this column is created automatically by MultiQC using the sample names. I used the lane names because I don't have any other useful info to put in the cell.
I don't find documentation about HTML content in a module, I'm not sure if that is easily possible as with custom content. It would be possible to simply add the list of failed lanes to the section text separated by commas, and have no plot/table at all.

matrulda: I would suggest to use "run" as "sample" (line 149) instead of "lane", and `col1_header="Run"`. That also removes the need for the special case of one run. I tried it out:
zero_yield_table
What do you think?

matrulda: The changes I tested: https://github.com/matrulda/MultiQC/commit/9051fa6b4a271706c34a3cc163c1a01b1914b6f4

multiqc/modules/checkqc/checkqc.py
```diff
@@ -428,6 +435,11 @@ def get_undetermined_percentage_data(self, issues, run, f):
                 self.checkqc_data["UndeterminedPercentageHandler"] = data
             else:
                 self.checkqc_data["UndeterminedPercentageHandler"].update(data)
+        if zero_yield_data:
+            if "UndeterminedPercentageHandler_ZeroYield" not in self.checkqc_data:
+                self.checkqc_data["UndeterminedPercentageHandler_ZeroYield"] = zero_yield_data
+            else:
+                self.checkqc_data["UndeterminedPercentageHandler_ZeroYield"] += zero_yield_data
```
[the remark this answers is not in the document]
maleasy: You are right, fixed in your commit matrulda@9051fa6

multiqc/modules/checkqc/checkqc.py
maleasy: Fixed in matrulda@9051fa6

multiqc/modules/checkqc/checkqc.py
maleasy: Thats a good solution! I pulled your commit

matrulda: Nice, thanks :smiley:

matrulda (approved):
Looks good to me, approved! :heavy_check_mark: Again, great work @maleasy ! 

@ewels I leave the rest to you now. :)

apeltzer:
Thank you all for making this possible 💯 🥳

maleasy:
Great, thank you all!
@matrulda: Thanks for the helpful comments and code bits

matrulda:
No problemat all! I've shown this module to my colleagues and we look forward to start using it at our facility :partying_face:

ewels:
Brilliant stuff - thank you for the extensive review @matrulda and thanks for being responsive and implementing all the changes @maleasy!

I'll try to do one final check myself soon and merge when I get a moment 👍🏻

ewels:
Thanks for this @maleasy! Just skim read the code and I can't see any issues on that side 👍🏻  (just a minor thing - missing a log output when something is found, which I've just added and pushed to the PR).

Looking at the report (based only on the file I have in the test data repo), I have a few suggestions for changes:

* General stats table
    * I think that we can remove these columns.. They seem to exactly mimic the bar plot underneath so I'm not sure that they really add anything.
* I would have to have two heatmaps showing all samples / lanes with pass / warn / fail for each test type. I'm not sure if this is actually possible (samples with no warnings are not reported, right @matrulda?) and I know some are read-specific. But it would be nice to flatten where applicable and have a summary table showing which samples fail across the board.. (as we do for FastQC)
* Too few reads per sample
    * Can we alphabetically sort the sample names please.. (or maybe lanes first?)
* Cluster PF too low
    * Can we avoid the PF acronym wherever we're not short on space? eg. Title, first category chart legend
* Graph sample names
    * I wonder if we drop the axis title and use a more verbose `Lane 3` instead of just `3`? Lane - Read could definitely do with this. eg. `Lane 3 - R2` instead of `3 -2`. Prefixes and numbers should always be short so I don't think that space is an issue.
* Descriptions / help text could be improved
    * _Error rate too high_ followed by _Some lanes have too high error rate._ is not super helpful 😉 What error rate is this? How is it calculated? What does it mean? Same for most sections.
    * I think also that the descriptions should be rephrased to make it clear that we are only showing the errors / warnings. For example, instead of _Some samples have too few reads_ we could have _The following samples reported an error because they did not meet their minimum read threshold._ etc...

All fairly minor stuff to just polish the output and make it easier for newcomers. When reviewing I generally try to imagine that I'm a lab scientist who has been emailed a MultiQC report like this and I'm trying to figure it out from scratch.

I hope this is ok! Let me know if you have any thoughts on any of the above 👍🏻 

Phil

maleasy:
Hi Phil,
I won't be able to work on this in the coming weeks. I'll check out your suggestions when I find the time.

ewels:
Ok, thanks for letting me know 👍🏻 Maybe I can wrap it up with help from @matrulda before then, depending on how I go with the other PRs that need handling.

ewels:
Ok, I've just pushed some changes to try to tweak the report output a little. Hopefully you guys agree that these are improvements, please let me know if you disagree with anything.

Thanks for this!

apeltzer:
Looks all good to me 👍🏻

matrulda:
Hi! (over a year later :sweat: ). Lost track of this, but super happy that this got merged. Thank you very much for all your work @maleasy @ewels @apeltzer  :tada:

ewels:
Wow, some serious inbox diving to find this @matrulda! 😂 I hope it's useful!

PR #1564, opened 2021-10-20
Mosdepth Median -> Coverage in General Statistics

schorlton:
We had a user confused by what `Median` was in the General Statistics table:
image

This field came from `mosdepth` and was median coverage. Sure, they could hover their mouse, but they suggested the field title be changed to something clearer, eg. `Coverage` or `Median Coverage`. This seems to make sense! I chose `Coverage` here for brevity.


schorlton:
@fgvieira any chance you could take another look at this?

ewels:
Thanks for this! The change has been superseded in a larger pull-request now though, see #1570

Closing, please let me know if I missed something.

Phil

PR #1567, opened 2021-10-23
Clarify length field in general statistics

schorlton:

When both FastQC and QUAST reports are included in MultiQC, both tools include a field with the header `Length` in the general statistics table. On first glance, this is very confusing for users. Here we just clarify which length they're presenting. Also fixed a bug where QUAST size suffixes can change.

### DIFF
--- a/multiqc/modules/quast/quast.py
+++ b/multiqc/modules/quast/quast.py
@@ -138,8 +138,8 @@ class MultiqcModule(BaseMultiqcModule):
             "modify": lambda x: x * self.contig_length_multiplier,
         }
         headers["Total length"] = {
-            "title": "Length ({})".format(self.total_length_suffix),
-            "description": "The total number of bases in the assembly (mega base pairs).",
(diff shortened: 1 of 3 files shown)


multiqc/modules/quast/quast.py
```diff
@@ -138,8 +138,8 @@ def quast_general_stats_table(self):
             "modify": lambda x: x * self.contig_length_multiplier,
         }
         headers["Total length"] = {
-            "title": "Length ({})".format(self.total_length_suffix),
-            "description": "The total number of bases in the assembly (mega base pairs).",
+            "title": "Assembly Length ({})".format(self.total_length_suffix),
+            "description": "The total number of bases in the assembly ({}).".format(self.total_length_suffix),
```
ewels, line 142: Nice spot on the suffix 👍🏻

ewels (approved):
Yup, agreed - in hindsight probably a little too aggressive on the short column titles. Agree that this is clearer 👍🏻 

Thanks!

Issue #1499, opened 2021-07-19
`fastp` module crashes with empty samples (zero reads)


telatin:
### Description of bug

Automatic processing a sequencing run that included a negative control that manage not to have any reads failed (division by zero). A division check can make the module safer when automatically processing raw outputs.

### File that triggers the error

Fastp JSON report of an empty sample:

 📦  [PCRblank.fastp.json.zip](https://github.com/ewels/MultiQC/files/6839517/PCRblank.fastp.json.zip)

### MultiQC Error log

```console
$ multiqc -o report .

  /// MultiQC 🔍 | v1.11

|           multiqc | Search path : /Volumes/Informatics/GMH/projects/758-emily/control-run/filt
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 32/32  
|             fastp | FastQ input file has zero raw reads
╭───────────────────────────────────────── Oops! The 'fastp' MultiQC module broke... ──────────────────────────────────────────╮
│ Please copy this log and report it at https://github.com/ewels/MultiQC/issues                                                │
│ Please attach a file that triggers the error. The last file found was: ./PCRblank.fastp.json                                 │
│                                                                                                                              │
│ Traceback (most recent call last):                                                                                           │
│   File "/Users/telatina/miniconda3/lib/python3.9/site-packages/multiqc/multiqc.py", line 624, in run                         │
│     output = mod()                                                                                                           │
│   File "/Users/telatina/miniconda3/lib/python3.9/site-packages/multiqc/modules/fastp/fastp.py", line 47, in __init__         │
│     self.parse_fastp_log(f)                                                                                                  │
│   File "/Users/telatina/miniconda3/lib/python3.9/site-packages/multiqc/modules/fastp/fastp.py", line 230, in parse_fastp_log │
│     self.fastp_data[s_name]["adapter_cutting_adapter_trimmed_reads"]                                                         │
│ ZeroDivisionError: float division by zero                                                                                    │
│                                                                                                                              │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
|           multiqc | No analysis results found. Cleaning up..
|           multiqc | MultiQC complete
```

ewels:
Thanks @telatin 👍🏻  Fixed in #1489, tested and seems to solve the issue ✨

PR #1489, opened 2021-07-10
Fix for 0 division error in fastp module.

bjohnnyd:
Fixes issue where a division by 0 occurs if there are 0 reads in input fastq. 

Additional fix on top of #1451 which updates `CHANGELOG.md` in 797674f.

fixes #1466
fixes #1444

- [x] This comment contains a description of changes (with reason)
- [ ] `CHANGELOG.md` has been updated

Issue #1466, opened 2021-06-28
ZeroDivisionError in 'fastp' MultiQC module for dataset with 0 reads


tbukur:
### Description of bug

The 'fastp' MultiQC module broke while running many samples. As it turned out, one of the samples had zero reads.

### MultiQC Error log

```console
============================================================
Module fastp raised an exception: Traceback (most recent call last):
  File "/code/Anaconda/3/2020.07/lib/python3.8/site-packages/multiqc/multiqc.py", line 594, in run
    output = mod()
  File "/code/Anaconda/3/2020.07/lib/python3.8/site-packages/multiqc/modules/fastp/fastp.py", line 47, in __init__
    self.parse_fastp_log(f)
  File "/code/Anaconda/3/2020.07/lib/python3.8/site-packages/multiqc/modules/fastp/fastp.py", line 201, in parse_fastp_log
    self.fastp_data[s_name]["after_filtering_total_reads"]
ZeroDivisionError: float division by zero
============================================================
```

### File that triggers the error

[SRR12380136.fastp_stats.json.txt](https://github.com/ewels/MultiQC/files/6726407/SRR12380136.fastp_stats.json.txt)

### Command used to run MultiQC

multiqc -l fastp_json_files.txt -c .my_multiqc.yaml

### MultiQC Version

multiqc, version 1.10.1

### Operating System

CentOS Linux 7

### Python Version

Python 3.8.3

ErikDanielsson:
This is solved as of #1451, which is included in the current development version: v1.11.dev.

bjohnnyd:
@ErikDanielsson  and @ewels  I am hitting this same error in fastp module in samples with zero reads using the current 1.11 release which includes the #1451.

I am using version `multiqc-1.11-pyhdfd78af_0`  and the error still occurs but I can see that this `ZeroDivisionError` catch form #1451 is not covering all the cases in `fastp.py` module.

The `ZeroDivsionError` should also be caught here:

https://github.com/ewels/MultiQC/blob/d6d1332a947670274ad489dac8ba96da0a848e92/multiqc/modules/fastp/fastp.py#L228-L235

and I can do a quick PR if needed.

Issue #1444, opened 2021-06-09
ZeroDivisionError - fastp module breaks when FastQ has 0 raw reads


drpatelh:
I tested with MultiQC `v1.10`. Creating a fresh folder with just the file below and running `multiqc .` should reproduce the error below:

```console
[WARNING]         multiqc : MultiQC Version v1.10.1 now available!
[INFO   ]         multiqc : This is MultiQC v1.10
[INFO   ]         multiqc : Template    : default
[INFO   ]         multiqc : Searching   : viralrecon/testing
Searching   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 1/1
[ERROR  ]         multiqc : Oops! The 'fastp' MultiQC module broke...
  Please copy the following traceback and report it at https://github.com/ewels/MultiQC/issues
  If possible, please include a log file that triggers the error - the last file found was:
    ./test1.fastp.json
============================================================
Module fastp raised an exception: Traceback (most recent call last):
  File "conda/envs/multiqc-1.10/lib/python3.9/site-packages/multiqc/multiqc.py", line 579, in run
    output = mod()
  File "conda/envs/multiqc-1.10/lib/python3.9/site-packages/multiqc/modules/fastp/fastp.py", line 47, in __init__
    self.parse_fastp_log(f)
  File "conda/envs/multiqc-1.10/lib/python3.9/site-packages/multiqc/modules/fastp/fastp.py", line 201, in parse_fastp_log
    self.fastp_data[s_name]["after_filtering_total_reads"]
ZeroDivisionError: float division by zero
============================================================
[WARNING]         multiqc : No analysis results found. Cleaning up..
[INFO   ]         multiqc : MultiQC complete

```

Added `.txt` extension to be able to upload here.

[test1.fastp.json.txt](https://github.com/ewels/MultiQC/files/6625100/test1.fastp.json.txt)

ewels:
_@drpatelh, pushing the limits of every MultiQC module by running with zero reads_ 😆 

Feel free to take a stab at it - these are usually fairly easy to fix with a `try` / `except ZeroDivisionError`... Otherwise will get to it when I can 👍🏻

drpatelh:
Fix one and find another one. Sorry! No rush.

ewels:
Fixed by @ErikDanielsson and me in #1451 👍🏻

PR #1569, opened 2021-10-27
catch zerodivisionerror

sparktx-nathan-johnson:
I had the experience of encountering a ZeroDivisionError in the fastp module when a number of reads was zero. There was already a try: except around the block, so I just updated it to catch the ZeroDivisionError as well. I'd be grateful if you would consider this change!

Thank you for making this tool available. 

### DIFF
--- a/multiqc/modules/fastp/fastp.py
+++ b/multiqc/modules/fastp/fastp.py
@@ -230,7 +230,7 @@ class MultiqcModule(BaseMultiqcModule):
                 self.fastp_data[s_name]["adapter_cutting_adapter_trimmed_reads"]
                 / self.fastp_data[s_name]["before_filtering_total_reads"]
             ) * 100.0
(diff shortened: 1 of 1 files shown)


ewels:
Hi @sparktx-nathan-johnson,

> To my current ability to tell, my proposed change is not responsible for the test failure.

Correct, the failing CI tests were something unrelated (just fixed in https://github.com/ewels/MultiQC/pull/1578)

> I had the experience of encountering a ZeroDivisionError in the fastp module when a number of reads was zero.

Many thanks for this fix! This has actually been reported a bunch of times (#1466, #1444, #1499) and already fixed in PR #1489.

I merged the #1489 pull-request earlier this evening, which applies the same fix that you propose here, so I will close your pull-request. Thanks anyway for your contribution though, and let me know if you hit any problems.

Cheers,

Phil

PR #1575, opened 2021-11-06
use errno instead of os.errno

jchorl:
Most `os.errno` calls were migrated to `errno`. Looks like this one was missed. It errored on me, just migrating it too.


ewels (approved):
Good spot. Thanks!

PR #1576, opened 2021-11-06
upsize label sizing in bar/line charts

jchorl:
Just upsizes some label sizing in the bar/line charts. I understand there is no optimal size for everyone. However I find 8 point font _very_ small, and can't imagine many folks would complain with slightly bigger font sizing. Thoughts?

### DIFF
--- a/docs/plots.md
+++ b/docs/plots.md
@@ -109,6 +109,7 @@ config = {
     'title': None,                          # Plot title - should be in format "Module Name: Plot Title"
     'xlab': None,                           # X axis label
     'ylab': None,                           # Y axis label
+    'labelSize': None,                      # Axis label size - should be an integer (default: 8)
     'ymax': None,                           # Max y limit
     'ymin': None,                           # Min y limit
     'yCeiling': None,                       # Maximum value for automatic axis limit (good for percentages)
@@ -236,6 +237,7 @@ config = {
     'title': None,               # Plot title - should be in format "Module Name: Plot Title"
     'xlab': None,                # X axis label
     'ylab': None,                # Y axis label
+    'labelSize': None,           # Axis label size - should be an integer (default: 8)
     'xCeiling': None,            # Maximum value for automatic axis limit (good for percentages)
     'xFloor': None,              # Minimum value for automatic axis limit
     'xMinRange': None,           # Minimum range for axis
--- a/docs/plots.md
+++ b/docs/plots.md
@@ -109,7 +109,7 @@ config = {
     'title': None,                          # Plot title - should be in format "Module Name: Plot Title"
     'xlab': None,                           # X axis label
     'ylab': None,                           # Y axis label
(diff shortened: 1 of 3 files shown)


ewels:
This one might be more complex than it first seems. It was a looong time ago that I wrote this code, but I seem to remember that I spent quite a long time fighting with margins and balancing font size versus spacing for long sample names.

It's very common for sample names to be huge in MultiQC reports, so this needs to work in those cases.

jchorl:
Any fud with shoving it in the `pconfig` so we can e.g. `pconfig.get("labelSize", 8)`?

ewels:
Not at all - that follows my general ethos which is "sensible defaults with ability to customise".

ewels:
Testing your PR and comparing before / after, it could also be as simple as me trying to match the matplotlib font sizes to the normal interactive plot font sizes / HTML text font sizes.

I'd be willing to meet you half way and set the defaults to 10, how does that sound? Also making it configurable sounds like a sensible thing to add as well.

jchorl:
I'll just make it configurable - then we don't have to worry about catering to personal prefs in defaults

jchorl:
Tested using https://raw.githubusercontent.com/ewels/MultiQC_TestData/master/data/custom_content/embedded_config/test_bargraph_mqc.txt

With/without:
```
# pconfig:
#    omitted fields...
#    labelSize: 18
```

Screenshot 2021-11-07 at 22-00-27 MultiQC Report
Screenshot 2021-11-07 at 21-59-53 MultiQC Report

I didn't touch the js plotly stuff. That code looks a little hairy. Let me know if that's fine, and also if this config should be documented somewhere. I couldn't find an enumeration of all the possible `pconfig` overrides to add this to.

ewels:
Nice, thanks! If you could just add some docs about this, then good to merge.. 👍🏻

jchorl:
Wasn't sure where to put docs. Let me know if another place is better!

docs/plots.md
```diff
@@ -109,6 +109,7 @@ config = {
     'title': None,                          # Plot title - should be in format "Module Name: Plot Title"
     'xlab': None,                           # X axis label
     'ylab': None,                           # Y axis label
+    'labelSize': None,                      # Axis label size - should be an integer (default: 8)
```
ewels, line 112:
```suggestion
    'labelSize': 8,                         # Axis label font size
```

docs/plots.md
```diff
@@ -236,6 +237,7 @@ config = {
     'title': None,               # Plot title - should be in format "Module Name: Plot Title"
     'xlab': None,                # X axis label
     'ylab': None,                # Y axis label
+    'labelSize': None,           # Axis label size - should be an integer (default: 8)
```
ewels, line 240:
```suggestion
    'labelSize': 8,              # Axis label font size
```

ewels (approved):
Looks good 👍🏻  Changelog too, but I can do that myself in a second..

PR #1577, opened 2021-11-06
class/styling improvements

jchorl:
I'd recommend reviewing without whitespace, because one file was largely indented. GH has that option, add `&w=1` at the end of the URL.

Few minor styling tweaks here:
1. Add `!important` to some background colors. This is critical to get chrome to print those background colors in PDFs. Right now, chrome still won't print those background colors, that's a CSS config that can be made separately.
2. Add a special `mqc-module-section-first` css class to the first mqc section. This is important for print styling - it allows you to utilize the CSS [`break-after`](https://developer.mozilla.org/en-US/docs/Web/CSS/break-after) property to keep the title/first section together. It makes printed pages look substantially better. Again, this doesn't include the CSS, it just enables the customization.
3. Add a `custom_logo` css class to the custom logo. This will allow for customizing size of the logo when printing.

Note that there is a [css media query](https://developer.mozilla.org/en-US/docs/Web/CSS/@media) for printing, so adding styling later for printing will not impact multiqc styling as-is today.

### DIFF
--- a/multiqc/templates/default/content.html
+++ b/multiqc/templates/default/content.html
@@ -10,41 +10,46 @@ the output from each module and print it in sections.
 {% for m in report.modules_output %}
   {% if m.sections | length > 0 %}
   <div id="mqc-module-section-{{ m.anchor }}" class="mqc-module-section">
-    <h2 id="{{ m.anchor }}">{{ m.name }}</h2>
-    {{ m.intro if m.intro }}
-    {% if m['comment'] %}<blockquote class="mqc-section-comment">{{ m['comment'] }}</blockquote>{% endif %}
-    {% for s in m.sections %}
-      {% if s['print_section'] %}
-        {% if (s['name'] is none or s['name'] | length == 0) and s['helptext'] is not none and s['helptext'] | length > 0 %}
-          <button class="btn btn-default btn-sm pull-right" type="button" data-toggle="collapse" data-target="#{{ s['anchor'] }}_helptext" aria-expanded="false" aria-controls="{{ s['anchor'] }}_helptext">
-            <span class="glyphicon glyphicon-question-sign" aria-hidden="true"></span>
-            Help
-          </button>
+    <div class="mqc-module-section-first">
+      <h2 id="{{ m.anchor }}">{{ m.name }}</h2>
+      {{ m.intro if m.intro }}
+      {% if m['comment'] %}<blockquote class="mqc-section-comment">{{ m['comment'] }}</blockquote>{% endif %}
+      {% for s in m.sections %}
+        {% if s['print_section'] %}
+          {% if (s['name'] is none or s['name'] | length == 0) and s['helptext'] is not none and s['helptext'] | length > 0 %}
+            <button class="btn btn-default btn-sm pull-right btn-help" type="button" data-toggle="collapse" data-target="#{{ s['anchor'] }}_helptext" aria-expanded="false" aria-controls="{{ s['anchor'] }}_helptext">
+              <span class="glyphicon glyphicon-question-sign" aria-hidden="true"></span>
+              Help
+            </button>
+          {% endif %}
+          <div class="mqc-section mqc-section-{{ m.anchor }}">
+            {% if s['name'] is not none and s['name'] | length > 0 %}
+              <h3 id="{{ s['anchor'] }}">
+                  {{ s['name'] }}
+                  {% if s['helptext'] is not none and s['helptext'] | length > 0 %}
+                    <button class="btn btn-default btn-sm pull-right btn-help" type="button" data-toggle="collapse" data-target="#{{ s['anchor'] }}_helptext" aria-expanded="false" aria-controls="{{ s['anchor'] }}_helptext">
+                      <span class="glyphicon glyphicon-question-sign" aria-hidden="true"></span>
+                      Help
+                    </button>
+                  {% endif %}
+              </h3>
+            {% endif %}
+            {% if s['description'] is not none and s['description'] | length > 0 %}<div class="mqc-section-description">{{ s['description'] }}</div>{% endif %}
+            {% if s['comment'] is not none and s['comment'] | length > 0 %}<blockquote class="mqc-section-comment">{{ s['comment'] }}</blockquote>{% endif %}
+            {% if s['helptext'] is not none and s['helptext'] | length > 0 %}
+              <div class="collapse mqc-section-helptext " id="{{ s['anchor'] }}_helptext">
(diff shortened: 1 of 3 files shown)


multiqc/templates/default/content.html
```diff
@@ -10,41 +10,46 @@
...
-                {% endif %}
-            </h3>
-          {% endif %}
-          {% if s['description'] is not none and s['description'] | length > 0 %}<div class="mqc-section-description">{{ s['description'] }}</div>{% endif %}
-          {% if s['comment'] is not none and s['comment'] | length > 0 %}<blockquote class="mqc-section-comment">{{ s['comment'] }}</blockquote>{% endif %}
-          {% if s['helptext'] is not none and s['helptext'] | length > 0 %}
-            <div class="collapse mqc-section-helptext " id="{{ s['anchor'] }}_helptext">
-              <div class="well">{{ s['helptext'] }}</div>
-            </div>
+    <div class="mqc-module-section-first">
```
ewels, line 13: Do we not need a conditional here as well, to mirror the closing `</div>`?

```suggestion
    {% if loop.first %}
    <div class="mqc-module-section-first">
    {% endif %}
```

ewels (approved):
Great! Just the one comment but I agree with the additions. Can't say that printing is something I've ever spent time looking at (clearly!) but welcome the improvements.

Any reason that we can't add some print styles to the CSS right away?

multiqc/templates/default/content.html
jchorl: No. This is a bit hacky, but we're actually grouping the `h2` and first loop iteration into a single `div`. This means that the `h2` and the section right after wouldn't be split across a page break (with the right css classes).

If you'd prefer, I could move:
```
      <h2 id="{{ m.anchor }}">{{ m.name }}</h2>
     {% if m['comment'] %}<blockquote class="mqc-section-comment">{{ m['comment'] }}</blockquote>{% endif %}
```

Right into the second for, and we could guard them with `if loop.first`. I think that's slightly cleaner. Thoughts?

jchorl:
> Any reason that we can't add some print styles to the CSS right away?

There's a whole whack of em. I'd recommend leaving that to a different PR. Note my css-overrides pr ;)

I'll get those merged in, but I feel like best to keep separate to keep things simple.

multiqc/templates/default/content.html
ewels: Ah right, sorry I missed that this was outside the loop. Yeah, might be nicer if the loop start is moved up above this chunk as you say. Might help future-me some pain 😅

jchorl:
Done!

Here's some sample inputs:
```
test_2_bargraph_mqc.txt:
# id: custom_bargraph_tsv_2
# section_name: 'Other section'
# description: 'This output is described in the file header. Any MultiQC installation will understand it without prior configuration.'
# format: 'tsv'
# plot_type: 'bargraph'
# pconfig:
#    id: 'custom_bargraph_w_header_2'
#    title: 'Custom Content Bar Plot 2'
#    ylab: 'Number of things'
Category_1      374
Category_2      229
Category_3      39
Category_4      253

test_3_bargraph_mqc.txt:
# parent_id: sect_one
# parent_name: sect_one
# id: custom_bargraph_tsv_3
# section_name: 'TSV Custom Content 3'
# description: 'This output is described in the file header. Any MultiQC installation will understand it without prior configuration.'
# format: 'tsv'
# plot_type: 'bargraph'
# pconfig:
#    id: 'custom_bargraph_w_header_3'
#    title: 'Custom Content Bar Plot 3'
#    ylab: 'Number of things'
Category_1      374
Category_2      229
Category_3      39
Category_4      253

test_bargraph_mqc.txt:
# parent_id: sect_one
# parent_name: sect_one
# id: custom_bargraph_tsv
# section_name: 'TSV Custom Content'
# description: 'This output is described in the file header. Any MultiQC installation will understand it without prior configuration.'
# format: 'tsv'
# plot_type: 'bargraph'
# pconfig:
#    id: 'custom_bargraph_w_header'
#    title: 'Custom Content Bar Plot'
#    ylab: 'Number of things'
Category_1      374
Category_2      229
Category_3      39
Category_4      253
```

And here's the result:
- https://gist.github.com/jchorl/03695790fd9f37af6fb9b60b60cd563f

Looks like it did the right thing!

jchorl:
btw I'll send you a before/after with all the print styling when I polish up the css. it's pretty :fire:

multiqc/templates/default/content.html
```diff
@@ -10,41 +10,48 @@
...
-          {% if s['helptext'] is not none and s['helptext'] | length > 0 %}
-            <div class="collapse mqc-section-helptext " id="{{ s['anchor'] }}_helptext">
-              <div class="well">{{ s['helptext'] }}</div>
-            </div>
+    {% if loop.first %}
+    <div class="mqc-module-section-first">
+      <h2 id="{{ m.anchor }}">{{ m.name }}</h2>
+      {{ m.intro if m.intro }}
+      {% if m['comment'] %}<blockquote class="mqc-section-comment">{{ m['comment'] }}</blockquote>{% endif %}
+    {% endif %}
```
ewels, line 19:
```suggestion
      {% if loop.first %}
      <div class="mqc-module-section-first">
        <h2 id="{{ m.anchor }}">{{ m.name }}</h2>
        {{ m.intro if m.intro }}
        {% if m['comment'] %}<blockquote class="mqc-section-comment">{{ m['comment'] }}</blockquote>{% endif %}
      {% endif %}
```

ewels:
Running Chrome on MacOS I honestly can't see any difference after this PR - tables still don't have any cell background fill when printed and worse seem to have white text 🙈 

image

Also I was wondering why you need a class on the first section, rather than just setting `break-after` on every section div (`.mqc-module-section`)? What's special about the first section in this regard?

Anyway, the PR doesn't make stuff any _worse_ for me so I'm still happy to merge 😅

jchorl:
> Also I was wondering why you need a class on the first section, rather than just setting break-after on every section div (.mqc-module-section)? What's special about the first section in this regard?

When I hacked this all up locally, I had the `h2` heading at the bottom of a page, but the contents on the next page. It looked a little odd. With this change, we're guaranteed to always get the section header with the first part of the section - a minor improvement.

Don't worry about chasing down the white text and bg, I chased all that down. I'll cut another PR soon ;)

ewels:
Right, sorry my question was a result of me misreading the code again (good evidence of why you shouldn't review PRs at 1am when already sleep deprived 😅). I thought that the code was checking the loop for the first module. But in the fresh light of a new day I see that it's the loop for the first section of each module. Now everything makes a bit more sense 😄

PR #1579, opened 2021-11-07
make printing palatable

jchorl:
There's a fair amount of hacked up css surgery going on here. I added comments for what each part does. I'm super flexible on where this code should go, how we should organize it, etc. This is just a first pass. Please provide feedback!

Here's an old & new printed report:
[old.pdf](https://github.com/ewels/MultiQC/files/7493373/old.pdf)
[new.pdf](https://github.com/ewels/MultiQC/files/7493374/new.pdf)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 - Added support for `--custom-css-file` / `config.custom_css_files` option to include custom CSS in the final report ([#1573](https://github.com/ewels/MultiQC/pull/1573))
 - MultiQC now skips modules for which no files were found - gives a small performance boost ([#1463](https://github.com/ewels/MultiQC/issues/1463))
 - New option to customise font size for axis labels in flat MatPlotLib charts ([#1576](https://github.com/ewels/MultiQC/pull/1576))
-- Improvements to printed report formatting for reports generated with the simple template ([#1579](https://github.com/ewels/MultiQC/pull/1579))
+- CSS Improvements to make printed reports more attractive / readable ([#1579](https://github.com/ewels/MultiQC/pull/1579))
 
 ### New Modules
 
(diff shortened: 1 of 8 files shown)


jchorl:
@ewels I _think_ this is the last of my PRs :P

You can see from the examples what it does. Np if you want to play around with it, it feels extremely hacky. Fwiw, these were the configs I was using locally. Now that we have css overrides, I can just apply these myself, so no rush ;)

It mostly feels bad that it makes multiqc much more fragile. If the html changes, the print styling will almost certainly break...

I wouldn't be offended if you didn't want to merge.

ewels (approved):
Love this! I have always neglected PDF exports and sort of as an extension, printing. Now that I see what you've done here I'm wondering why.. Very simple addition and makes a huge improvement to the output 🖨️ ✨ 

> If the html changes, the print styling will almost certainly break...

Yeah, but I can't see that this would be any worse than the current state of not having any print styling.. So not really a big issue as far as I can see (also the template rarely changes these days).

ewels:
I was a bit confused at first as I couldn't see some of the effects when testing locally. Then I realised that not all the classes had been propagated through all of the templates. I guess you're mostly using `-t simple`, but I see no reason why the default template can't also look nice when printed. So I ported (most of) the new class names and markup over to all templates.

I've also had a bit of a play and added a bunch of additional CSS, mostly for stuff in the default interactive template. For example, making tables dumb and entirely visible and so on. Looking great now!

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Added support for `--custom-css-file` / `config.custom_css_files` option to include custom CSS in the final report ([#1573](https://github.com/ewels/MultiQC/pull/1573))
 - MultiQC now skips modules for which no files were found - gives a small performance boost ([#1463](https://github.com/ewels/MultiQC/issues/1463))
 - New option to customise font size for axis labels in flat MatPlotLib charts ([#1576](https://github.com/ewels/MultiQC/pull/1576))
+- Improvements to printed report formatting for reports generated with the simple template ([#1579](https://github.com/ewels/MultiQC/pull/1579))
```
ewels, line 14: Aha, missed this before!

```suggestion
- CSS Improvements to make printed reports more attractive / readable ([#1579](https://github.com/ewels/MultiQC/pull/1579))
```

ewels:
Black CI is failing but seems to be due to an issue with the docker container. Passes without issue locally ✅

Issue #1558, opened 2021-10-05
`--ignore` does not work properly


nikalfa11:
### Description of bug

Hi,
in the directory containing FastQC outputs where I am running multiqc there are 2 files called:

* `Undetermined_S0_R1_001_fastqc.zip`
* `Undetermined_S0_R2_001_fastqc.zip`

besides the FastQC of all other samples. I would like to run multiqc ignoring the 2 Undetermined files. So I ran inside this directory:

```console
> multiqc . --ignore Undetermined*
> multiqc . --ignore Undetermined_S0_R1_001_fastqc.zip,Undetermined_S0_R2_001_fastqc.zip
```

Any of these options worked and the Undetermined files were included. Only when I tried one file name at the time, it worked:

```console
> multiqc . --ignore Undetermined_S0_R1_001_fastqc.zip
```

MultiQC ignored the Undetermined R1 file. Why the first two commands didn't work? Shouldn't the command with the `*` work for both files? And how to ignore multiple files at once?

Thank you very much in advance

Niccolò

ewels:
My first guess is that you need to enclose the `*` in quotes:

```bash
multiqc . --ignore "Undetermined*"
```

Without this, `bash` will be expanding the glob before it gets to MultiQC. MultiQC will actually receive the following:

```bash
multiqc . --ignore Undetermined_S0_R1_001_fastqc.zip Undetermined_S0_R2_001_fastqc.zip
```

This will be treating the second filename (`Undetermined_S0_R2_001_fastqc.zip`) as an input search path. In this case that would have no effect. I suspect if you check again that _one_ of the undetermined files is always skipped?

If I'm right and the documentation needs fixing, or I'm wrong and it's something else. Either way, thanks for reporting 😀

MatthiasZepper:
@nikalfa11 Thank you for bringing this issue to our attention. 

I have now updated [the documentation](https://multiqc.info/docs/#choosing-where-to-scan) to clarify how the `--ignore` option can be used with multiple files. Subject to approval, it will in the future explicitly cover an example how to specify multiple files as the option's argument. Briefly, either use a quoted wildcard argument: 

`multiqc . --ignore "Undetermined*"` 

or specify the option multiple times: 

`multiqc . --ignore "Undetermined_S0_R1_001_fastqc.zip" --ignore "Undetermined_S0_R2_001_fastqc.zip"` 

Lycka till!

ewels:
[...]

PR #1584, opened 2021-11-12
Issue #1558: Clarify documentation for the --ignore option

MatthiasZepper:

This pull request updates `docs/usage.md` to clarify the section on the `--ignore` option. The documentation now also briefly mentions the `--ignore-symlinks` flag, which was introduced in [MultiQC v1.6](https://github.com/ewels/MultiQC/releases/tag/v1.6), but not yet covered. 

Not applicable

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -28,14 +28,16 @@ multiqc data/*_fastqc.zip
 multiqc data/sample_1*
 ```
 
-You can also ignore files using the `-x`/`--ignore` flag (can be specified multiple
-times). This takes a string which it matches using glob expansion to filenames,
-directory names and entire paths:
+If the `--ignore-symlinks` flag is set, MultiQC will ignore symlinked directories and files.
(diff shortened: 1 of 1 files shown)


ewels (approved):
Thanks! Sorry, giving you a docs task was a bit mean as I have my own idiosyncrasies that I usually stick to..

* One sentence per line
    * GitHub inserts newlines after a single line break, but most markdown rendering engines (including the MultiQC and nf-core websites) ignore them.
    * Having shorter lines makes the diffs easier to read when editing docs
    * Using sentences as natural line breaks (rather than a character width) makes editing easier and diffs clearer.
* I usually draw attention to "gotchas" like the quotes thing here with blockquotes, then use CSS to make these highlighted. This is mostly because Markdown doesn't have (easy) ways to have custom styles for this like RST (though there are extensions). It works fine.

Otherwise great!

PR #1589, opened 2021-11-17
fix string and array comp

jchorl:
Subtle bug here. The `if` is always `True`, which had weird subtle bugs downstream. Somehow this caused descriptions to be duplicated, I have no idea why.

```
$ python3
>>> a = "."
>>> a[:-1] == "."
False
```


ewels (approved):
Great spot! I also have no idea why it would have that effect 😆  But if this fixes it then I'm not going to waste too much brain power thinking about it..

Thanks!

Issue #220, opened 2016-05-25
qualimap bamqc: needed files


avilella:
How does the qualimap bamqc parser detect if there is a qualimap report?
Is it by looking at the presence of `genome_results.txt` or does it require other `.html` files to be present?

I found myself that deleting everything but the `genome_results.txt` makes MultiQC (head master, pre-0.7) not find any qualimap reports...

Thx for the clarification

ewels:
It uses a [bunch of files](https://github.com/ewels/MultiQC/blob/master/multiqc/utils/search_patterns.yaml#L58-L63):

```yaml
raw_data: raw_data
genome_results: genome_results.txt
coverage: coverage_histogram.txt
insert_size: insert_size_histogram.txt
genome_fraction: genome_fraction_coverage.txt
gc_dist: mapped_reads_gc-content_distribution.txt
```

Currently, it looks through the analysis directories and finds a directory called `raw_data` with a `genome_results.txt` inside, then grabs the other files after that. The method it uses is non-standard though and doesn't use the core function - this means that it misses several core functions. I currently have an issue open to rewrite it ( #210 ), so it will probably change in the near future.

Phil

ewels:
Hi @avilella,

Note that I've just pushed a change where I've basically rewritten the entire Qualimap module. Let me know if I've broken anything for you.

Cheers,

Phil

PR #1591, opened 2021-11-18
New module: Iso-Seq

sguizard:
A module to parse the json and csv reports produce by the program isoseq3 refine. 
The test data have been the submitted in the [pull request #220](https://github.com/ewels/MultiQC_TestData/pull/220)

- [X] There is example tool output for tools in the <https://github.com/ewels/MultiQC_TestData> repository or attached to this PR


vladsavelyev:
I refactored the module and merged in the `cluster` code from https://github.com/ewels/MultiQC/pull/1592

Fixed some potential issues and got rid of the `pandas` dependency.

However, the module needs more work. Currently, `refine` generates too many tables, which probably should be re-thought, merged together and/or represented as charts if possible. There are also a bit too many metrics in general stats, perhaps we could hide some by default.

vladsavelyev:
Moving the PR into the v2.0 milestone, as I'd like to use a boxplot to represent min/mean/median/stddev/max stats instead of a table, and I want to implement it with Plotly directly instead of first implementing it with HighCharts.

vladsavelyev:
Alright, I added a new `box` plot type: https://github.com/MultiQC/MultiQC/pull/2358. And I re-implemented the `refine` stats table using it:
Screenshot 2024-02-20 at 21 10 36

I might do the same for "Cluster size distribution", as it would look more natural than the bar plot there.

vladsavelyev:
Should we replace this plot
Screenshot 2024-02-21 at 00 46 36

With a box plot?
Screenshot 2024-02-21 at 00 46 28

The former is readable if it's always going to be important to know if the number of clusters is in that range, and ≤2 vs >2 is an important threshold. The latter generalizes better. I'm not familiar with HiFi reads enough

multiqc/modules/isoseq/isoseq.py
```diff
@@ -0,0 +1,231 @@
...
+            },
+        }
+        self.general_stats_addcols(data_by_sample, headers, namespace="refine")
+
+    def _add_refine_plot(self, values_by_metric_by_sample):
+        data_labels = {
+            "fivelen": "5' primer length",
+            "threelen": "3' primer length",
+            "insertlen": "Insert length",
+            "polyAlen": "Poly(A) length",
```
ewels, line 198: I think that these two can be their own separate plots, rather than tabbed data labels.

multiqc/modules/isoseq/isoseq.py
```diff
@@ -0,0 +1,231 @@
...
+                "label": title,
+                "title": "Iso-Seq refine: " + title,
+            }
+            for metric, title in data_labels.items()
+        ]
+
+        self.add_section(
+            name="Iso-Seq refine",
+            anchor="insert-refine-stats",
+            description="Iso-Seq refine statistics",
```
ewels, line 216: Would be nice if we could have a better description of specifically what `isoseq refine` is doing. PacBio docs don't seem to be very descriptive of this.

ewels (commented):
Couple of comments, but generally looks good 👍🏻

PR #1592, opened 2021-11-19
Add module `isoseq3 cluster`

sguizard:

- [X] There is example tool output for tools in the <https://github.com/ewels/MultiQC_TestData> repository or attached to this PR
- [X ] Each table column has a different colour scale to its neighbour, which relates to the data (eg. if high numbers are bad, they're red)


ewels:
Coming to these super late sorry. But from a quick glance, I wonder if this and https://github.com/ewels/MultiQC/pull/1591 should both be a single module, called `isoseq3`?

Then have the code split up still, done as we have for samtools / picard etc.

vladsavelyev:
I moved the code into the `refine` [PR](https://github.com/ewels/MultiQC/pull/1591), to have one single `isoseq` module with 2 submodules, so closing this PR

Issue #1563, opened 2021-10-13
BCL Convert module`KeyError`: `# of >= Q30 Bases (PF)`


ElisabettaNale:
### Description of bug

I am trying to use BCL convert output file (v.3.9.3) to generate the corresponding MultiQC report. The MultiQC software expects to find the following key `# of >= Q30 Bases (PF)` presumably in the `Demultiplex_Stats.csv`. file, but it can't find it.

### File that triggers the error

[bcl_output.zip](https://github.com/ewels/MultiQC/files/7338160/bcl_output.zip)

### MultiQC Error log

```console
multiqc . --config /hpcnfs/techunits/genomics/work/nf_elisabetta_chromium/tumgt-genomics_bclconvert/nf/multiqc_config.yaml -o .

/// MultiQC 🔍 | v1.11

|           multiqc | Search path : /hpcnfs/techunits/genomics/work/nf_elisabetta_chromium/tumgt-genomics_bclconvert/nf/work/b2/2043677ab0d4ec1b884d20da487d5a/DUAL-TEN/Reports
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 9/9  
╭──────────────────────────────────── Oops! The 'bclconvert' MultiQC module broke... ─────────────────────────────────────╮
│ Please copy this log and report it at https://github.com/ewels/MultiQC/issues                                           │
│ Please attach a file that triggers the error. The last file found was: ./RunInfo.xml                                    │
│                                                                                                                         │
│ Traceback (most recent call last):                                                                                      │
│   File "/usr/local/lib/python3.8/dist-packages/multiqc/multiqc.py", line 624, in run                                    │
│     output = mod()                                                                                                      │
│   File "/usr/local/lib/python3.8/dist-packages/multiqc/modules/bclconvert/bclconvert.py", line 44, in __init__          │
│     self.parse_demux_data(demux)                                                                                        │
│   File "/usr/local/lib/python3.8/dist-packages/multiqc/modules/bclconvert/bclconvert.py", line 347, in parse_demux_data │
│     lane["basesQ30"] += int(row["# of >= Q30 Bases (PF)"])                                                              │
│ KeyError: '# of >= Q30 Bases (PF)'                                                                                      │
│                                                                                                                         │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
|           multiqc | No analysis results found. Cleaning up..
[...]

PR #1595, opened 2021-11-23
GH1563: add defaults for some columns

yanick:
This is for #1563 

Big :warning: -> I am not using this package directly, so I'm likely to have made assumptions that are wrong.

This being said, the case reported in the issue is going :boom: because two columns are not present in a csv file. I'm assuming that is a valid schema of that file, and that it's reasonable to set the values of those columns to be '0' if not found. 

If I'm mistaken and the csv absolutely needs to have those columns, the patch will have to be changed to politely tell the user that their format is wonky. :-)  


skanwal:
@yanick a quick comment as you mentioned you aren't using this module directly - if you've time, I think this [comment](https://github.com/ewels/MultiQC/issues/1563#issuecomment-949884512) can also be tackled as part of your PR to read correct read stats column from a different file `Quality_Stats.csv`, which is very useful metrics.

ewels:
Many thanks for this @yanick!

I continued your PR on a bit, adding in parsing functionality for the new `Quality_Metrics.csv` file mentioned by @Garrett-Ng in https://github.com/ewels/MultiQC/issues/1563#issuecomment-949884512 (as suggested by @skanwal above). The module should now work with both types of output and give comparable reports, based on testing with the example data provided in #1563.

Please all give this new code a try and let me know if you hit any problems 👍🏻 

Phil

PR #1596, opened 2021-11-25
reset multiqc between runs

jchorl:
Currently, multiqc leverages global state to store info. If you run it twice, you get weird results - e.g. table columns get duplicated. This moves the `init` logic into a re-callable function, such that multiqc can reset state between runs.

I wrote a couple quick scripts to test:
```console
$ cat other.py 
def init():
    global foo
    foo = ["bar"]

$ cat test.py 
import other

other.init()
print(other.foo)
other.foo.append("baz")
print(other.foo)
other.init()
print(other.foo)

$ python3 test.py 
['bar']
['bar', 'baz']
['bar']
```

### DIFF
--- a/multiqc/utils/report.py
+++ b/multiqc/utils/report.py
@@ -32,37 +32,67 @@ try:
 except NameError:
     pass  # Python 3
 
-# Set up global variables shared across modules
-general_stats_data = list()
-general_stats_headers = list()
-general_stats_html = ""
-data_sources = defaultdict(lambda: defaultdict(lambda: defaultdict()))
-plot_data = dict()
-html_ids = list()
-lint_errors = list()
-num_hc_plots = 0
-num_mpl_plots = 0
-saved_raw_data = dict()
-last_found_file = None
-runtimes = {
-    "total": 0,
-    "total_sp": 0,
-    "total_mods": 0,
-    "total_compression": 0,
-    "sp": defaultdict(),
-    "mods": defaultdict(),
-}
-file_search_stats = {
-    "skipped_symlinks": 0,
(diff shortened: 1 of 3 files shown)


ewels:
hah, smart! Yes this has been bugging me for a while, one of those things where I was less experienced with Python when I started MultiQC and didn't know better.

I have been vaguely planning to refactor a lot of this code to use objects instead, eg. have a `multiqc.Report()` object with all of this state information. However that's a bigger job and is one of many things that I've been putting in a v2.0 rewrite in my head. This could be a nice simple workaround until then.. 👍🏻

multiqc/utils/report.py
```diff
@@ -32,37 +32,67 @@
...
-    "skipped_not_a_file": 0,
-    "skipped_ignore_pattern": 0,
-    "skipped_filesize_limit": 0,
-    "skipped_no_match": 0,
-}
-
-# Make a dict of discovered files for each seach key
-searchfiles = list()
-files = dict()
+def init():
```
ewels, line 37:
```suggestion
# Inside a function so that the global vars are reset if MultiQC is run more than once within a single session / environment
def init():
```

ewels (approved):
Not tested locally, but CI is still passing and I think if anything was wrong they would fail.

If you could please just add a small note in the changelog that'd be fab, then I'll merge 👍🏻

jchorl:
No global state would be nice :)

Updated the changelog.

ewels:
The modules already use classes, so it's mostly a report object, a template object(?), a config object and then a MultiQC run object I guess.. If you fancy a challenge 😉

PR #1607, opened 2021-12-25
Numeric filename fix

### LINKED ISSUE #1606: numeric bam files cause an error
### Description of bug

stack trace produced on some input files with a numeric file name (e.g. 0877)

_No response_

### MultiQC Error log

```python
Module qualimap raised an exception: Traceback (most recent call last):
  File "/mnt/galaxy/data/galaxy/tool-deps/_conda/envs/__multiqc@1.8/lib/python3.8/site-packages/multiqc/multiqc.py", line 546, in run
    output = mod()
  File "/mnt/galaxy/data/galaxy/tool-deps/_conda/envs/__multiqc@1.8/lib/python3.8/site-packages/multiqc/modules/qualimap/qualimap.py", line 39, in __init__
    n['BamQC'] = QM_BamQC.parse_reports(self)
  File "/mnt/galaxy/data/galaxy/tool-deps/_conda/envs/__multiqc@1.8/lib/python3.8/site-packages/multiqc/modules/qualimap/QM_BamQC.py", line 23, in parse_reports
    parse_genome_results(self, f)
  File "/mnt/galaxy/data/galaxy/tool-deps/_conda/envs/__multiqc@1.8/lib/python3.8/site-packages/multiqc/modules/qualimap/QM_BamQC.py", line 104, in parse_genome_results
    s_name = self.clean_s_name(d['bam_file'], f['root'])
  File "/mnt/galaxy/data/galaxy/tool-deps/_conda/envs/__multiqc@1.8/lib/python3.8/site-packages/multiqc/modules/base_module.py", line 209, in clean_s_name
    s_name = os.path.basename(s_name)
  File "/mnt/galaxy/data/galaxy/tool-deps/_conda/envs/__multiqc@1.8/lib/python3.8/posixpath.py", line 142, in basename
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not float
```

bwlang:
sometimes an input file will match and be converted to a number - which causes later assumption about about it being a string to fail.
fixes #1606


ewels (changes requested):
Thanks! I think I would prefer a more generic fix though, can we just coerce a string type when the basename function is called instead?

Phil

bwlang:
If one does "0729" => 729.0  even converting to a string later will mess up the name... 
Seems like a bad idea to even try to convert a name to a float.  I don't love the repetition in this patch, but did not see a better way...

ewels:
Right, I hadn't thought about that. So it's not only the `ValueError` that's the issue here.. Ok 👌🏻

ewels (approved):
LGTM

ewels:
Thanks for the fix!

Issue #1588, opened 2021-11-15
`multiqc_general_stats.txt` is empty


g-pacheco:
### Description of bug

Hello, 

I have been using version **v1.12.dev0** and I have noticed that the files `multiqc_general_stats.txt` is written but it is empty. Was this function disabled on purpose? It is a very useful file, so I am hoping that this is a bug ^^  

Thanks a lot, George.

![Screenshot (268)](https://user-images.githubusercontent.com/37849281/141846853-5ab5ba73-4e06-48cf-bb66-86eb3b9aaf92.png)

ewels:
huh, oops! Yes that's very definitely a bug.. Well spotted, thanks!

ewels:
It looks like the file is created by the regular `table.plot()` function, here: https://github.com/ewels/MultiQC/blob/5ab527b74721e44c9070032855e5019651d0fbc3/multiqc/multiqc.py#L832-L838

So if the table is showing properly in the report, and this function is working for other modules / tables (is it..?) I have no idea why the General Stats file would suddenly be empty... 🤔

g-pacheco:
Hej.

It also seems that the written files changed quite a bit. Maybe there is a connection with this change?

Best regards, George.

![NEW](https://user-images.githubusercontent.com/37849281/142849143-1971f018-f791-4980-82ba-5326812c1158.png)
![OLD](https://user-images.githubusercontent.com/37849281/142849146-dd511c5c-eb5c-42fa-8d60-90c705c39c0f.png)

g-pacheco:
Hello, 

May I ask if there has been any development here?

Best regards, George.

g-pacheco:
Hej @ewels --

Any news here?

ewels:
Not yet. I've had a speed look, just to see if I could spot anything obvious. But I haven't had a chance to properly dig into it yet.

If you're able to find the culprit then please submit a pull-request with a fix 👍

ewels:
Fixed by @MatthiasZepper in #1609 after a fairly in-depth investigation!

g-pacheco:
Hej, 

Thanks so much for fixing this. I am sorry, but I am still learning about the different versions of a repository on _GitHub_. I tried to update _MultiQC_ like this:

```
source activate WHOLE_PATH_TO_MULTIQC_CONDA_ENVIRONMENT
conda update multiqc
conda deactivate
```

Then the file was still not created. Then I tried to update it like this instead:

```
[...]

PR #1609, opened 2022-01-05
Bugfix for Issue 1588: `multiqc_general_stats.txt` is empty

MatthiasZepper:
Commit 40a4d18 / #1528 introduced a return statement to `util_functions.write_data_file` that would skip writing data to a file which is not an OrderedDict. This however also affected the _general_stats_data_, which is of type list and thus caused #1588.

This commit addresses this bug and wraps the reshape operation into a try statement. Instead of returning, it will now try and otherwise fall back to YAML export. 

Additionally, based on the false assumption that the `multiqc_general_stats.txt` would collect all modules' output, I have also come up with a rescue effort to remove aberrant items and proceed with the regular ones. However, this bloats the function quite heavily and it does not help with the heatmaps and scatterplots nonetheless. So I guess, it should be removed again, unless there is a future feature in planning that could make good use of that functionality?  

- [ ] `CHANGELOG.md` has been updated (_Not applicable since bug was fixed before release_)

### DIFF
--- a/multiqc/utils/util_functions.py
+++ b/multiqc/utils/util_functions.py
@@ -66,6 +66,31 @@ def write_data_file(data, fn, sort_cols=False, data_format=None):
                         return None
                 return json.JSONEncoder.default(self, obj)
 
+        # some metrics can't be coerced to tab-separated output (omit those and write separately to YAML)
+        if data_format not in ["json", "yaml"] and len(data) > 1:
+            aberrant_data = list()
+            for _, key in zip(range(len(data)), data):
+                if type(data[key]) is not dict and type(data[key]) is not OrderedDict:
+                    aberrant_data.append(key)
+
+            for key in aberrant_data:
+                with io.open(os.path.join(config.data_dir, fn + "_" + str(key)), "w", encoding="utf-8") as f:
+                    yaml.dump(data[key], f, default_flow_style=False)
+                    config.logger.debug(
+                        "Some metrics could not be coerced into tab-separated output and were written as YAML instead to "
+                        + fn
+                    )
+                del data[key]
+
+            if not data:  # no items left for regular export
+                return
+
+        elif len(data) == 1 and type(data) is not dict and type(data) is not OrderedDict:
+            config.logger.debug(
+                "Metrics of " + fn + "can't be saved as tab-separated output. Choose JSON or YAML output instead."
+            )
+            return
+
         # Save file
         with io.open(os.path.join(config.data_dir, fn), "w", encoding="utf-8") as f:
             if data_format == "json":
@@ -75,10 +100,7 @@ def write_data_file(data, fn, sort_cols=False, data_format=None):
                 yaml.dump(data, f, default_flow_style=False)
             else:
                 # Default - tab separated output
-                # Heatmaps and other odd things might break this, so skip if so
-                # TODO: leaves an empty file, should clean this up
-                if type(data) is not dict and type(data) is not OrderedDict:
-                    return
+
                 # Convert keys to strings
                 data = {str(k): v for k, v in data.items()}
                 # Get all headers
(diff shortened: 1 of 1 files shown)


ewels:
Nice work! I don't completely understand what you mean about removing aberrant data? Do you have an example file / situation where this happens?

MatthiasZepper:
> Nice work! I don't completely understand what you mean about removing aberrant data? Do you have an example file / situation where this happens?

No, that is exactly the point. I erred about the purpose of the `mutliqc_general_stats.txt` file, because I started out with a test directory that contained a _FastQC_ and a _Fastp_ report. I just removed the current return statement and got a nice `mutliqc_general_stats.txt` uniting the information from both modules' output. 

When adding a heatmap from the custom content to my test directory it crashed (which was predicable, because that was the reason to enforce OrderedDicts in the first place). 

Crashs also occurred, when I ran the export only on the condition `len(data) == 1`  for test purposes, so ultimately skipping writing insular data items to a file. This made me believe, that the heatmap would not only be written separately to a file, but also be one of the items within the `mutliqc_general_stats', so I came up with various attempts to spot and remove such aberrant items and introduced about 200 lines of special cases handling logic (even much more intricate than what I committed in the end) to fully fix the bug. 

This version allowed me to run it successfully on my test directory with the heatmap plus the FastQC/Fastp reports and I got a nice  `mutliqc_general_stats.txt` plus the heatmap in a separate file. What drove me crazy was the fact that running the same code on the [MultiQC_TestData](https://github.com/ewels/MultiQC_TestData) repo did not crash, but also did not yield a `mutliqc_general_stats.txt` at all. I assumed, this would be attributable to silently failing because of malformed items within the general stats.  

Therefore, I came up with the current change that would explicitly test for each item, if the TSV reshape works and otherwise fall back to YAML. Adding some temporary log statements to trace the execution however made me realize that there is indeed no file or module where this functionality is needed at all at this point. 

But since I evidently do not understand under which conditions a `mutliqc_general_stats.txt` is generated and the output from which modules is considered within that file, there might be future situations where checking the single elements of the provided data dict might be useful. **Ultimately, it is up to you to say, if the current development roadmap with regard future modules justifies retaining this somewhat complex special case handling or if acting on the maxim of lean code is preferable.** 

In the latter case, I would undo outsourcing the reshape procedure into a separate sub-function and just trigger the fallback on the except statement. This is evidently the cleanest situation that would boil it down to a single-digit number of additional lines of code.

ewels:
The general stats file is just made from the general stats table. So it should always be the same structure every time and should have no prospect of changing any time soon. Depending on which files you tested with, the file could have been absent if there was no general stats table (not all modules / custom content adds to this table).

Sounds like you've had a bit of an epic, but in this case I think I'd prefer the lean code approach if you don't mind..

MatthiasZepper:
> Sounds like you've had a bit of an epic, but in this case I think I'd prefer the lean code approach if you don't mind..

No, I don't mind at all, because I would decide in a similar fashion if I were you. I just had a glimmer of hope that not all of the efforts that I put into the special edge case handling were in fact entirely useless 🙄.   

But alas:

> _I have not failed. I’ve just found 10,000 ways that won’t work. – Thomas A. Edison_

multiqc/utils/util_functions.py
```diff
@@ -99,6 +90,23 @@ def default(self, obj):
 
                 body = "\n".join(rows)
 
+            except:
+                data_format = "yaml"
+                config.logger.warning(
+                    fn
+                    + " could not be saved as tab-separated output. Choose JSON or YAML output instead. Falling back to YAML."
+                )
```
ewels, line 98:
```suggestion
               logger.debug(f"{fn} could not be saved as tsv/csv. Falling back to YAML.")
```

ewels (approved):
Looks good! Just the tweak to the log message..

multiqc/utils/util_functions.py
ewels: I stand corrected! `config.logger` wasn't a mistake I see but intentional! Smart. Logging in this file has always been awkward (see other function where I pass the logger as an argument). This seems like quite an elegant solution though..

PR #1614, opened 2022-01-20
Update plot_indelhist.py

zjzh:
I refactored code with set comprehension which is more concise and efficient，how do think this change which has practical value？


ewels:
Thanks for the contribution, but I'm not sure that this has a significant positive effect and would prefer to leave the code in its current state.

Issue #1622, opened 2022-02-04
ZeroDivisionError in Skewer module


silvia-s:
### Description of bug

Hi, I just got the same error (`Oops! The 'skewer' MultiQC module broke...` [...] `ZeroDivisionError: float division by zero`) as issue #1586 using Skewer with MultiQC v1.12.dev0. While waiting for this to be fixed, is there perhaps any workaround or previous version of MultiQC without this bug that I can use?

Many thanks,
Silvia

ewels:
It [should already be fixed](https://github.com/ewels/MultiQC/commit/1058c590c7ca78a1738e10cb7176289535bd8066) on `master` already, so if your v1.12.dev0 installation is relatively recently installed (after 15th Nov 2021) then you shouldn't be getting this error...

If you are getting this with a fresh install then please attach the output and example files, it's possible that it's a similar but different bug..

silvia-s:
I'm afraid I can't share the input files, as it's a set of sensitive samples from a collaborator. Here's a screenshot (with light blue boxes masking the paths) of the stdout/stderr from MultiQC, hope it's helpful: 

![Screenshot_MultiQC](https://user-images.githubusercontent.com/9861211/152573696-f718c984-999a-4604-ac2f-bee6de51e292.png)

I downloaded that version today, using:
`git clone --branch zero-div-rnaseq --single-branch https://github.com/ewels/MultiQC.git`

...and my MultiQC command was the following:
`multiqc InputFolderName -n Final_QC_report.html --no-data-dir --interactive --config MyConfigFile.yaml`

Silvia

MatthiasZepper:
Thank you for reporting this issue. Technically, that is a straightforward fix. In the skewer module, the percent calculation will fail if there are no available reads:

https://github.com/ewels/MultiQC/blob/7117ab7197a074b0154202ebd22c0239c120ddaa/multiqc/modules/skewer/skewer.py#L152-L153

I have now created a PR (#1623) that should prevent this error in the future. However, you might also want to check why so many of your samples have no reads and either relax the filtering criteria or double check the barcode reference used for de-multiplexing?

Lycka till from the NGI team!
Thias

silvia-s:
[...]

Issue #1509, opened 2021-08-04
NaN value in JSON output


victorskl:
### Description of bug

There seem to be `NaN` as value when choose to dump in JSON format i.e. `--data-format json`. This is of course invalid as JSON value. Please see attached screenshot. And error condition for use case such as when trying to insert MultiQC data straight into MySQL table as JSON data type. Report itself render no issue. I can also report that this happen on SnpEff toolbox / plot data.

I also note that, in Python `json.dumps` [allowed NaN by default](https://stackoverflow.com/questions/6601812/sending-nan-in-json). For better, it should `allow_nan=False` to conform with JSON format, I suppose. 

We have post-processing step for MultiQC data ingestion. Do you recommend setting `NaN` to `null` if we handle during processing? 

Our MultiQC version is `1.10.dev0`.

Thanks

![image](https://user-images.githubusercontent.com/4586509/128121593-035ebf6a-f0a4-41da-bdfb-89750681fa88.png)

cmatKhan:
```
library(TidyMultiqc)
library(here)

df = load_multiqc(here("data/multiqc_output.json"))
```

```
Error in parse_con(txt, bigint_as_char) : 
  lexical error: invalid char in json text.
                  "mapped_failed_pct": NaN,                 "paired in
                     (right here) ------^
```

What did you do? replace NaN with some known nonsense number in the JSON?

brainstorm:
I wonder if we could somehow lure @pcingola (main dev of snpEff AFAICT) to drop the use of `NaN` in snpEff and instead use a very large number to stay within the designated schema type:

https://github.com/pcingola/SnpEff/search?p=1&q=NaN

That'd solve the issue(s) we are seeing here with Python's JSON serialization where the data type for several fields flips from `Double` to something that JSON parsers do not accept as `Double` (because it isn't).

Please note that turning `NaN` into a string is not a good option either for the aforementioned reasons ;)

/cc @reisingerf @ohofmann @andrewpatto

ewels:
> Do you recommend setting `NaN` to `null` if we handle during processing?

I guess that this is the best solution.. The Stack Overflow post you linked to has a comment saying that the `simplejson` package can do this conversion, instead of just falling over with a `ValueError`.

If it's only a limited number of modules doing this (just SnpEff?) then the easiest solution would probably be to get the module code to replace `NaN` with `None`, which will then be serialised to `null` in the JSON.

multimeric:
[...]

PR #1623, opened 2022-02-04
Issue1622: ZeroDivisionError in Skewer

MatthiasZepper:

This PR should fix the ZeroDivisionError in the _Skewer_ module reported in #1622 . Since no test data was provided, I can't say for sure, though.


MatthiasZepper:
Sorry for the force-push. Unfortunately, I had checked out from the half-finished anglerfish module branch instead of the master branch.

ewels (approved):
Good find, thanks!

Only change is to use `None` instead of `Nan`, because the latter can break JSON export. `None` _should_ be handled fairly nicely by MultiQC, eg. skipping table columns or plot data points etc where appropriate.

MatthiasZepper:
> Good find, thanks!
> 
> Only change is to use `None` instead of `Nan`, because the latter can break JSON export. `None` _should_ be handled fairly nicely by MultiQC, eg. skipping table columns or plot data points etc where appropriate.

Fair enough. I had run a quick test with a made up Skewer report (as no real test data could be provided) that worked fine (including outputting the plots with mathplotlib), but I admit that I haven't had a look in the JSON output.

ewels:
I'm referring to this issue mostly: https://github.com/ewels/MultiQC/issues/1509

Issue #1624, opened 2022-02-09
Picard Target Region Coverage hard to read


Redmar-van-den-Berg:
### Description of bug

In recent versions of `picard CollectHsMetrics` (>2.23.8), PCT_TARGET_BASES is reported up to 100000x coverage, to support high depth amplicon sequencing (see https://github.com/broadinstitute/picard/pull/1542)

However, this extends the x-axis of the MultiQC graph up to 100.000, when most 'regular' sequencing project have a coverage around 100x.

MultiQC graph for picard 2.23.5
![image](https://user-images.githubusercontent.com/15814544/153212708-5aa93740-45e0-4268-9424-5bf422ab2ed2.png)

MultiQC graph for picard 2.26.10 (same data)
![image](https://user-images.githubusercontent.com/15814544/153212842-c36129d2-9317-4d33-8db7-c351069a4459.png)

Redmar-van-den-Berg:
MultiQC graph for picard 2.26.10 (same data), with the code from https://github.com/ewels/MultiQC/pull/1626. Note that the x-axis now logarithmic. 
![image](https://user-images.githubusercontent.com/15814544/153799203-6ca6d860-b9ee-42ab-88ff-5c4da4199481.png)

ewels:
We've come across something similar with `InsertSizeMetrics` and `WgsMetrics` before and have these config options as a result:

https://multiqc.info/docs/#insertsizemetrics
https://multiqc.info/docs/#wgsmetrics

Instead of a new solution, could we mimic the same pattern for this module for consistency?

ewels:
> We've come across something similar with `InsertSizeMetrics` and `WgsMetrics` before and have these config options as a result:
> 
> https://multiqc.info/docs/#insertsizemetrics https://multiqc.info/docs/#wgsmetrics
> 
> Instead of a new solution, could we mimic the same pattern for this module for consistency?

Redmar-van-den-Berg:
I can add the config options for consistency, but I would still like to keep the default behaviour as is. That way, a regular 'naive' user will get the third plot by default, but they are free to modify it using the `fold_coverage_xmax` option. Otherwise, they will get the second plot by default, which would make MultiQC more difficult to use for anyone but power users.

ewels:
[...]

PR #1626, opened 2022-02-11
Exclude zero values, or the y-axis goes up to 100k

Redmar-van-den-Berg:
This is a bugfix for https://github.com/ewels/MultiQC/issues/1624


ewels:
Reposting https://github.com/ewels/MultiQC/issues/1624#issuecomment-1047660287

> We've come across something similar with `InsertSizeMetrics` and `WgsMetrics` before and have these config options as a result:
> 
> https://multiqc.info/docs/#insertsizemetrics https://multiqc.info/docs/#wgsmetrics
> 
> Instead of a new solution, could we mimic the same pattern for this module for consistency?

vladsavelyev:
What if we just set `xmax` to the highest coverage where all values are non-zero? See the update of the PR. Seems to work nicely.

Screenshot 2023-10-15 at 20 46 36

vladsavelyev (approved):

ewels (approved):
LGTM 👍🏻

PR #1627, opened 2022-02-12
Modify NanoStat to NanoPack with some slight fixes

schorlton:
Thanks for your ongoing support of MultiQC!

This PR proposes to change the name of NanoStat to NanoPack. As I understand it, there are multiple tools within the NanoPack family, and actually many of them can produce the `NanoStats.txt` file that this module ingests. However, some of the tools (eg. NanoPlot) can actually produce rich `NanoStats.txt` file (containing information on pore count) which NanoStat cannot. Currently, the module interprets these as having been generated from the wrong file type (eg. generating `NanoStats.txt` with `--fastq_rich` in NanoPlot makes this module think that the data was generated from the sequencing summary file), so I have removed that description text. I also proposed a couple minor changes to the language and formatting. cc @wdecoster for any comments. Thanks!

image


vladsavelyev:
Hi @schorlton - thank you for the contribution, and sorry for not responding to the PR for such a long time! Hopefully, we will be much faster with reviews from now on.

I've taken a look at the tools from the NanoPack, and it looks like the report we are parsing is generated by a `write_stats` function from the library called [nanomath](https://github.com/wdecoster/nanomath/blob/master/nanomath/nanomath.py#L189-L237) (there are 2 versions of that report, and the first thing I want to fix is to [support](https://github.com/ewels/MultiQC/issues/1995) the new version). 

In the whole pack, [as far as I can see](https://github.com/search?q=owner%3Awdecoster+nanomath+write_stats&type=code), the only tools that use that function are NanoPlot and NanoComp. There is also a tool in the pack called [Cramino](https://github.com/wdecoster/cramino/blob/master/README.md) that produced a different format, and probably needs a separate module. 

Because of the above, I'm not sure if it's fair to rename the module to NanoPack. Renaming might potentially break downstream user scripts, so I'd probably leave it called NanoPlot for now.

vladsavelyev:
I thought a bit more about it. Since the old and the new format have a one-to-one match, it was a pretty straightforward update to the current module to support both. I think for simplicity and back-compatibility I will just keep the old name, I think it should be clear to users that NanoStat refers to stats from the tools from the NanoPack family. But happy to discuss more.

wdecoster:
Well yeh, sorry about the mess in naming things :) 
The NanoStats.txt files are all created from some code in the nanomath module, with the --tsv flag from NanoPlot/NanoStat/NanoComp giving a better formatted table. The report's content can vary depending on the available data, which in turn would depend on the input file format (cram vs fasta vs sequencing_summary vs...).

Happy to clarify further.

vladsavelyev:
@wdecoster thank you a lot for the clarification!

Let me know if you have ideas about how to best name and organize the NanoPack-related MultiQC module(s). Would you go ahead with renaming this module to NanoPack, as @schorlton proposed?

Do I understand correctly that the better-formatted table enabled with `--tsv` is the [legacy table here](https://github.com/wdecoster/nanomath/blob/master/nanomath/nanomath.py#L217-L236)? Is it going to be discontinued at some point, or we should support both formats equally? (related to this [issue](https://github.com/ewels/MultiQC/issues/1995)).

wdecoster:
No, --tsv is handled [slightly higher in the code](https://github.com/wdecoster/nanomath/blob/40aa42a11bd056c268ed10a5bc25a3f99a538317/nanomath/nanomath.py#L203C1-L211C42). It is a lot less complicated :-D 
I am not against renaming it to NanoPack, as the summary format can be generated by NanoPack-family tools NanoPlot, NanoComp and NanoStat -- but they all use the same code in nanomath that I wrote too many years ago.
The stats output will be included in the HTML report of NanoPlot and NanoComp, and also written to a NanoStats.txt file (so also that name is not wrong).

I don't have strong feelings about changing or keeping the name, both make sense.

vladsavelyev:
Thank you for the explanation! 

In this case, I'd prefer to keep things simple, and would just keep the existing name NanoStat.

vladsavelyev:
@schorlton, I created a separate PR with your other changes in this PR, as we still want them in, regardless of the module name: https://github.com/ewels/MultiQC/pull/2062

Feel free to review. You also should be displayed as a commit co-author once it's merged. 

Thanks again for the contribution!

schorlton:
Thank you all! Looks great.

Issue #1550, opened 2021-09-15
Custom Content: Section name and description based on filename for `.txt` files


BELKHIR:
Hi,

Actually, the section name and description are automatically based on the filename for png, jpeg and jpg files with `_mqc` at the end of the filename.

Is it possible to extend this behavior to filenames ending with `*_mqc.(yaml|yml|json|txt|csv|tsv|log|out|png|jpg|jpeg|html)` ?

Many thanks

ewels:
Hi @BELKHIR,

Could you please provide a little more detail and ideally an example file? When I test with [one of my example `.txt` files](https://github.com/ewels/MultiQC_TestData/blob/master/data/custom_content/no_config/issue_1091_mqc.txt) it does set the section name as the filename.

Phil

BELKHIR:
Hi @ewels,

I tested with your provided txt file and the result was fine !
However, it seems that this problem is related to the lack of header line in txt files.
ex  this file of PCA eigenvalues will reproduce the problem : 
axis1	3.65933584612439
axis2	1.548253919225
axis3	0.461137974231304
axis4	0.417858586883222
axis5	0.407933212752084
axis6	0.385368911026864
axis7	0.370025550609452
axis8	0.358294662093722
 
While if we add a header line there is no more problem with section name and description generation : 
Axis	        Value
axis1	3.65933584612439
axis2	1.548253919225
axis3	0.461137974231304
axis4	0.417858586883222
axis5	0.407933212752084
axis6	0.385368911026864
axis7	0.370025550609452
axis8	0.358294662093722

Thank's a lot,

Khalid,

ewels:
Thanks, confirmed my end - I'll take a look 👍🏻

ewels:
ok, found the culprit. Should be fixed in 01457ca23ed841f4cbcdf10fc72e6fff3083eb2c

PR #1634, opened 2022-02-17
Only set id for custom content when id not set by metadata

### LINKED ISSUE #1629: DupRadar section in MultiQC report is duplicated for each sample
### Description of bug

When using MultiQC v1.12 the custom content we use for `dupRadar` in nf-core/rnaseq seems to be duplicated once for each sample. This doesn't occur with v1.11 and the content is correctly rendered in the same plot / portion of the General Stats table.

image

You should be able to reproduce by downloading the files below, untarring and running `multiqc .` in the folder.

[test.tar.gz](https://github.com/ewels/MultiQC/files/8062767/test.tar.gz)

### MultiQC Error log

```console
There is no explicit error as it appears to be a rendering issue in the report.
```

idot:

Closes #1629
https://github.com/ewels/MultiQC/issues/1550 set the id for all custom content.
Now only when id not already in metadata

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ _nothing yet.._
 
 ### Module updates
 
-_nothing yet.._
+- Only set id for custom content when id not set by metadata ([#1629](https://github.com/ewels/MultiQC/issues/1629))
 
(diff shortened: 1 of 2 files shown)


ewels (approved):
Thanks for the fix 👍🏻

PR #1644, opened 2022-03-01
Ensure logger removal is compatible with dictConfig configuration

J-81:
This includes a patch for a logger removal step bug that arose from using dictConfig to configure the root logger.


ewels (approved):
Thanks!

PR #1204, opened 2020-05-27
AdapterRemoval: Fix the calculation of retained and discarded reads when option collapse is activated

alexhbnr:
Here is some suggestion how to fix the plotting of the number of reads that were processed by AdapterRemoval2 when the option "collapse" is activated, which particularly affects the number of retained and discarded reads (see issue here: #1005)

Briefly, it corrects the number of retained and discarded reads but subtracting the number of collapsed reads from it.

I am not sure, if this fixes all problems of this plot, but it would be a solution to start working from.

 - [x] This comment contains a description of changes (with reason)
 - [ ] `CHANGELOG.md` has been updated

PR #1647, opened 2022-03-15
AdapterRemoval: fixes number of reads that are discarded/collapsed

NiemannJ:

This is following a discussion with @jfy133.
Like #1386 and #1423, this should fix issue #1005 and now display the correct number/proportion of **reads** that are collapsed/discarded/paired.

On the test dataset the output used to look like this, where "Retained Read Pairs" contains e.g. collapsed/truncated read pairs: 

ar_retained_plot (1)

With my edits the plot looks like this now, where the categories are non-overlapping and proportional to the input data, making it easy to see what proportion of the data was collapsed and discarded: 

ar_retained_plot

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,10 @@ _nothing yet.._
 
 ### Module updates
 
-- Only set id for custom content when id not set by metadata ([#1629](https://github.com/ewels/MultiQC/issues/1629))
+- **AdapterRemoval**
+  - Finally merge a fix for counts of reads that are discarded/collapsed ([#1647](https://github.com/ewels/MultiQC/issues/1647))
+- **Custom content**
+  - Only set id for custom content when id not set by metadata ([#1629](https://github.com/ewels/MultiQC/issues/1629))
 
 ## [MultiQC v1.12](https://github.com/ewels/MultiQC/releases/tag/v1.12) - 2022-02-08
 
--- a/multiqc/modules/adapterRemoval/adapterRemoval.py
+++ b/multiqc/modules/adapterRemoval/adapterRemoval.py
@@ -178,10 +178,19 @@ class MultiqcModule(BaseMultiqcModule):
         self.result_data["unaligned_total"] = unaligned_total
         self.result_data["reads_total"] = reads_total
         self.result_data["discarded_total"] = reads_total - self.result_data["retained"]
-
-        self.result_data["retained_reads"] = (
-            self.result_data["retained"] - self.result_data["singleton_m1"] - self.result_data["singleton_m2"]
-        )
+        if self.__read_type == "paired":
+          if not self.__collapsed:
+              self.result_data["paired_reads"] = (
+              self.result_data["retained"] - self.result_data["singleton_m1"] - self.result_data["singleton_m2"]
+              )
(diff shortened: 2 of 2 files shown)


jfy133:
Hi @NiemannJ I think this does indeed make much more sense! 

However I think we need to rename `Paired Reads`, and also add a help section describing what we are looking at as I think while this is now numerically correct, it maybe slightly unintitutive for people who look a bit closer due to the naming scheme when dealing with PE data.

For example, I _think_ (please correct me if I'm wrong) `Paired Reads` now represent reads that were not collapsed. So I think should be `Uncollapsed Reads` rather than just reads.

Also we need to add a text box explaining that we are still dealing with _single_ reads in the numbers, even if `Full-length Collapsed Reads` would _imply_ to a user they are look at two reads that are _now_ merged into one (if that makes sense?). I know in the AR2 settings file this would be referred to as 'Read Pairs' but I know many people aren't that familiar with the AR2 settings file.

So I think line 305 (for example) could be updated to _something_ like:

"The number of retained and discarded reads. All numbers refer to the status of the original 'uncollapsed' reads. Therefore the number of 'collapsed' reads in the output FASTQ file will be approximately half of the numbers in this plot" (but well explained/written :grimacing: ) 

Or something like that...

I attach a before/after example of the eager test data. 

@alexhbnr would you also be happy with this?

[example.zip](https://github.com/ewels/MultiQC/files/8259778/example.zip)

alexhbnr:
It looks great, @NiemannJ, thanks for putting in the work.
I agree with @jfy133 that the label "paired reads" is unintuitive and should contain something along the lines of "uncollapsed" or a synonym of it.

apeltzer:
Should we not wait until Mikkel Schubert fixes these sorts of things upstream in AR2?

jfy133:
@apeltzer ideally yes, this is a faster fix for a very long on-going issue ;). Mikkel has gone quite since Nov on AR3

jfy133:
Looks good to me now! It sucks AR doesn't report percentage of pairs merged (this would be useful for general stats....) but maybe we wait for AR3 for that.

NiemannJ:
@jfy133 Just for clarification, you mean the percentage of pairs that were collapsed, right? I can easily add that to the summary table, the numbers are already there. I personally would also like a column with the percentage of discarded reads, because I've been working with data where +50% of the reads were removed, so I can add that as well if you think it's of interest as well.

jfy133:
> @jfy133 Just for clarification, you mean the percentage of pairs that were collapsed, right? I can easily add that to the summary table, the numbers are already there. I personally would also like a column with the percentage of discarded reads, because I've been working with data where +50% of the reads were removed, so I can add that as well if you think it's of interest as well. 

Yes exactly! I'm up for both if us easy to insert

jfy133:
@NiemannJ can you post a screenshot of the final general stats and barchart?

NiemannJ:
@jfy133 Sure, here you go:
image
and
image

ewels:
Nice! Thanks for putting an end to this saga 😆 

If everyone's happy, I will merge this and close #1386, #1423, #1005 and #1204 🥵

ewels (approved):

PR #1651, opened 2022-03-21
Edit bclconvert search pattern

matthdsm:

Set bclconvert search pattern for `RunInfo.xml` to shared, as it interferes with the function of the MultiQC_SAV plugin

### DIFF
--- a/multiqc/utils/search_patterns.yaml
+++ b/multiqc/utils/search_patterns.yaml
@@ -85,6 +85,7 @@ bcl2fastq:
   num_lines: 300
 bclconvert/runinfo:
   fn: "RunInfo.xml"
+  shared: true
 bclconvert/demux:
   fn: "Demultiplex_Stats.csv"
 bclconvert/quality_metrics:


ewels:
> as it interferes with the function of the MultiQC_SAV plugin

If the problem is specific to the plugin, can you not just override the MultiQC config from within the plugin?

You should be able to use the `before_config` hook and set the config option. Docs a little patchy about this, but this is maybe best: https://multiqc.info/docs/#searching-for-files

Basically editing `multiqc.config` at that point will be after the default config is loaded but before any user-specific configs. So end-users will still be able to override whatever you set in the plugin.

You can see this in action in the bcbio plugin - [hook here](https://github.com/MultiQC/MultiQC_bcbio/blob/8c85d10df261655e7cdd3f8c256ef4149cd52aa8/setup.py#L35-L37) and [function here](https://github.com/MultiQC/MultiQC_bcbio/blob/master/multiqc_bcbio/__init__.py).

matthdsm:
Thanks for the tip! I'll do it that way.

PR #1655, opened 2022-03-30
Fix jinja2 version

schorlton:
MultiQC errors with jinja2 version >3.1.0 that it cannot import `jinja2.escape`. Here we fix the jinja2 version to temporarily fix that.
Eventually, perhaps someone could look into what the replacement is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Major spruce of the command line help, using the new [rich-click](https://github.com/ewels/rich-click) package
 - Drop some of the Python 2k compatability code (eg. custom requirements)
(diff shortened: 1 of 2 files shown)


ewels:
I can only find one usage of this call, in a single module:

https://github.com/ewels/MultiQC/blob/9e860becd880be42ba76a6cee942d5b59fd680ff/multiqc/modules/nanostat/nanostat.py#L100

I think I'd rather just fix that instead of pinning the versions...

ewels:
Attempted in https://github.com/ewels/MultiQC/pull/1659 - let's see if the tests pass there.. 🤞🏻

ewels:
Closing in favour of #1659 - thanks though! 🙏🏻

Issue #1597, opened 2021-11-26
Module VEP broke


quentin67100:
### Description of bug

I tried to use MultiQC with singularity. But it failled to add VEP stats. I also tried with the conda version of multiqc but it failled too.

### File that triggers the error

[Pologne.stat.vep.html.zip](https://github.com/ewels/MultiQC/files/7608793/Pologne.stat.vep.html.zip)
,

### MultiQC Error log

```console
/// MultiQC 🔍 | v1.11

|           multiqc | Search path : /shared/projects/gentaumix/Pologne
|          qualimap | Found 4 BamQC reports
|    MarkDuplicates | Skipping MarkDuplicates sample 'multiqc_data' as missing essential fields
|    MarkDuplicates | Skipping MarkDuplicates sample 'multiqc_data' as missing essential fields
|    MarkDuplicates | Skipping MarkDuplicates sample 'multiqc_data' as missing essential fields
|    MarkDuplicates | Skipping MarkDuplicates sample 'multiqc_data' as missing essential fields
|    MarkDuplicates | Skipping MarkDuplicates sample 'multiqc_data' as missing essential fields
|            picard | Found 4 MarkDuplicates reports
|            picard | Found 8 ValidateSamFile reports
╭────────────────── Oops! The 'vep' MultiQC module broke... ───────────────────╮
│ Please copy this log and report it at                                        │
│ https://github.com/ewels/MultiQC/issues                                      │
│ Please attach a file that triggers the error. The last file found was:       │
│ /shared/projects/gentaumix/Pologne//06_VEP/Pologne.stat.vep.html             │
│                                                                              │
│ Traceback (most recent call last):                                           │
│   File "/usr/lib/python3.8/site-packages/multiqc/multiqc.py", line 624, in r │
│     output = mod()                                                           │
│   File "/usr/lib/python3.8/site-packages/multiqc/modules/vep/vep.py", line 3 │
│     self.parse_vep_html(f)                                                   │
│   File "/usr/lib/python3.8/site-packages/multiqc/modules/vep/vep.py", line 1 │
│     existing = values[1].split("(")[0].replace(" ", "")                      │
│ IndexError: list index out of range                                          │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
|      fastq_screen | Found 8 reports
|            fastqc | Found 8 reports
|           multiqc | Compressing plot data
[...]

PR #1656, opened 2022-03-31
Fixed bug when other fields also have a "-" instead of an integer.

fgvieira:
While parsing a VEP stats file from SV calls, got an error:
```
ValueError: invalid literal for int() with base 10: '-'
```
It seems it was due to a line in the stats file that had:
```
Overlapped regulatory features	-
```

The fix was to always check if the value is `-` before converting to `int`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ _nothing yet.._
 
 - **AdapterRemoval**
   - Finally merge a fix for counts of reads that are discarded/collapsed ([#1647](https://github.com/ewels/MultiQC/issues/1647))
+  ** VEP **
+  - Fixed bug when `General Statistics` have a value of `-` ([#1656](https://github.com/ewels/MultiQC/pull/1656))
 - **Custom content**
   - Only set id for custom content when id not set by metadata ([#1629](https://github.com/ewels/MultiQC/issues/1629))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ _nothing yet.._
 
 - **AdapterRemoval**
   - Finally merge a fix for counts of reads that are discarded/collapsed ([#1647](https://github.com/ewels/MultiQC/issues/1647))
-  **VEP**
+- **VEP**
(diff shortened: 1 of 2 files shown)


fgvieira:
Failing on unrealted module `nanostat` with:
```
AttributeError: module 'jinja2' has no attribute 'escape'
```

ewels:
The existing code has a `try` / `except` block to catch this `ValueError` already though, so it should already be handled?

Could you attach a file that triggers the error? I guess it must be being raised from somewhere else in the code.

I generally try not to replace `-` with `0` in MultiQC modules, as they are not quite the same thing. Better to use `None` or just omit that metric (as is currently done in this module).

ewels:
Yeah, I think I did it that way because I was handling more than one error at once - see https://github.com/ewels/MultiQC/issues/1597 (both `IndexError` and `ValueError`). But yeah, I don't have a strong preference, I agree that specific is probably safer - as long as it also catches the other error mentioned in that issue.

I more meant that I didn't see why that chunk of code needed editing if it was already behaving in the correct way. But you answered my question there - it's the `txt` parsing code that needed the fix 👍🏻 

Please do use `None` instead of `0`. Happy to keep your change from the `try`/`except` to the specific case with the `-` if you like 👍🏻  Hopefully I added the test data from #1597 to the test data repo for the CI checks, but maybe double check with those files just in case if that's ok.

fgvieira:
Also works with example file from #1597.

ewels:
Great stuff, thanks! Just tried locally and looks good with one exception. I had thought that passing `None` would skip it from the table, but it seems that it is printed:

image

Versus before when it was a mixture of being omitted and showing as `0`:

image

What do you think about adjusting the code to simply not set these keys? Instead of setting them to `None`. Personally I think that the blank table cells look better (and are just as faithful..?).

Apologies for going back on myself, poor memory of exactly how the MultiQC internals work..

fgvieira:
Sure, as long as it is consistent. :smile: 
I feel that having both blank cells and `0` is a bit confusing.

ewels (approved):
Many thanks, looking great!

PR #1660, opened 2022-04-04
Pangolin 4.0 compatability

mattloose:
Recently pangolin has been updated to version 4.0 and this changes the output CSV file - see: https://github.com/cov-lineages/pangolin/releases/tag/v4.0

This causes the module to fail in its current state as `row['qc_status']` already exists and the current replacement triggers a key error by searching for `row['status']` which no longer exists. Thanks to @alexomics for tracking down the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,10 @@ _nothing yet.._
   - Fixed bug when `General Statistics` have a value of `-` ([#1656](https://github.com/ewels/MultiQC/pull/1656))
 - **Custom content**
   - Only set id for custom content when id not set by metadata ([#1629](https://github.com/ewels/MultiQC/issues/1629))
-- \*\*Nanostat\_\_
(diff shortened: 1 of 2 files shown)


ewels:
Thanks @mattloose! Do you have an example output file that we can add to [ewels/MultiQC_TestData](https://github.com/ewels/MultiQC_TestData/tree/master/data/modules/pangolin) please? Then I can test locally as well..

alexomics:
I've created a table [pangolin_4-0.csv](https://github.com/ewels/MultiQC/files/8409404/pangolin_4-0.csv) from the [example output table](https://cov-lineages.org/resources/pangolin/output.html)

ewels:
Great stuff, thanks!

mattloose:
I see you have managed to find a CSV before I was able to reply. Glad this is resolved. Thanks.

Issue #501, opened 2017-07-19
General Stats: incorrect sorting of columns with unit in cells 


jfy133:
When sorting the 'General Stats' column and the cells of a column contains a non-numeric symbol, sorting does not occur correctly.

For example, when sorting a 'coverage' column (with X-fold coverage) in descending order, 9.0X will incorrectly come before 300X 

An example can be found here:
[multiqc-mtcap.zip](https://github.com/ewels/MultiQC/files/1159230/multiqc-mtcap.zip)

As a general point, it is redundant to have the unit symbol both in each cell and also in the header. So having % next to every value in the '% GC' column is unnecessary; having just numbers in every column would also improve the sorting issue described above. Of course, I don't know how hard it would be to parse this, given you are pulling from lots of different programs.

ewels:
Thanks @jfy133, this definitely needs to be sorted. I remember fixing it way back in the early days in fact, so I must have broken that support somewhere along the way.

The value suffixes are added back by MultiQC so it's trivial to remove them, however the general feeling that I've had from most users is that they are welcome. It should be possible to force numeric sorting (ignoring text suffixes). But perhaps I should do a twitter poll or something to gauge support more widely :)

Phil

ewels:
> _this definitely needs to be sorted_

Apologies for the pun..

jfy133:
No need for apologies, it was appreciated ;).

It sounds good to me, I'll look out for the poll! Looking forward to start using MultiQC to present data to our supervisors.

ewels:
Hi @jfy133,

This bug should be resolved now with the latest dev version. Thanks for reporting it! Let me know if you run into any problems still.

Phil

PR #1665, opened 2022-04-08
Allow sorting of table columns with text contents

jchorl:
https://github.com/ewels/MultiQC/issues/501, and the ensuing [commit](https://github.com/ewels/MultiQC/commit/a9e9b982fd2b9c88cf0ab9d0e9f57d23c51a5194) caused table sorting to strip all non-numeric values. This is pretty big hammer to deal with values such as `300X`. Table values probably shouldn't contain suffixes/units as they make sorting ~impossible (how can you distinguish `300X` from `1a4X`, where the latter is some arbitrary string??).

Anyway, we can't really change all the table inputs. But we can do a little better to infer if the value is numeric or not. This isn't perfect, as I'm sure someone will have values shaped like e.g. `1eeee`, `1ffff`, and this will not sort them properly. But it's a start. If you have ideas on how to better infer, I'm open to them.


ewels:
No you're right, I should prioritise these as they're nice and bite-size. I'll get to it ASAP 👍🏻

ewels (approved):
LGTM, thanks for this 👍🏻

PR #1666, opened 2022-04-09
Upgrade jQuery tablesorter plugin to v2

jchorl:
Updates tablesorter plugin.

JS min file downloaded from: https://mottie.github.io/tablesorter/docs/#Download

I updated the CSS to cooporate with the new version. I don't know how best to identify incompatibilities/regressions. It looked good to me?

Why? The new version can read default sort data annotated as html data attributes on tables. See the second example [here](https://mottie.github.io/tablesorter/docs/example-option-sort-list.html). This means we can work into the multiqc table python logic to slurp configs on initial sort and annotate per-table.


vladsavelyev:
Hey @jchorl, just getting here through the backlog. Happy to merge (looks like the library wasn't updated since then, so you have still the latest version here!)

But I'm not sure I'm following the justification 🤔 Could you add more detail on what you had in mind, if you recall? Is there a use case where we might need to pass that initial sort config?

jchorl:
Yeah, for what it's worth, we've been cherry-picking these commits and using them for a while, and it's been working well.

PR #1668, opened 2022-04-15
optimize linegraph category comparison

jchorl:
Optimizes the category comparison of the linegraph.

See the issue is, `d["data"]` does not change while looping through `d["data"]` as best I can tell. So every iteration, the result will be the same. The only thing we need to check is that the values are lists. Once we verify that, we just need to check categories once per dataset.

In one case, I was comparing a list of 68000 cats, 68000 times. This is an n^2 problem, no bueno.

The fix brought multiqc from not on a dataset to finishing in ~20s.

I didn't think this was worthy of CHANGELOG, lmk if you feel otherwise

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Fixed edge-case bug in custom content where a `description` that doesn't terminate in `.` gave duplicate section descriptions.
 - Tidied the verbose log to remove some very noisy statements and add summaries for skipped files in the search
 - Add timezone to time in reports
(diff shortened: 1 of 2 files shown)


schorlton:
Hi @ewels, bumping this - thanks!

jchorl:
Updated changelog!

ewels (approved):
Oops, sorry for delaying this by 6 months and missing a release due to requesting a one-line changelog entry 🙄 

Awesome stuff as always, thank you! 👏🏻 🙏🏻

PR #1677, opened 2022-05-07
Add timezone to reports

schorlton:
The time in reports is confusing because there is no timezone attached to it. This simply adds the timezone. An alternative approach would be to use ISO format, although this is slightly less human readable. We could also consider moving to UTC time.

```python3
# Current format
from datetime import datetime
datetime.now().strftime("%Y-%m-%d, %H:%M")
'2022-05-07, 14:47'
# New format
datetime.now().astimezone().strftime("%Y-%m-%d, %H:%M %Z")
'2022-05-07, 14:47 PDT'
# Alternative suggestion in UTC/ISO
import datetime
datetime.datetime.utcnow().replace(tzinfo=datetime.timezone.utc).isoformat(' ', timespec='minutes')
'2022-05-07 21:47+00:00'
```


ewels:
Nice! Definitely like the idea 👍🏻 Will merge as soon as I get a chance to test locally.

PR #1694, opened 2022-05-30
Handle too long and low complexity in fastp

schorlton:
Fastp can filter for many reasons. MultiQC previously plotted filtering reasons, but oddly omitted reads that were too long or low complexity from this plot. This was deceiving as you could have 99% of reads filtered because they were too long, but the plot would show 100% of reads passed filtering because we didn't include this field. This fixes that.

image


ewels (approved):
Nice, LGTM - thanks!

PR #1641, opened 2022-02-25
added gfastats

gf777:
- [ ] This comment contains a description of changes (with reason)
- [ ] `CHANGELOG.md` has been updated

- [ ] There is example tool output for tools in the <https://github.com/ewels/MultiQC_TestData> repository or attached to this PR
- [ ] Code is tested and works locally (including with `--lint` flag)
- [ ] `docs/README.md` is updated with link to below
- [ ] `docs/modulename.md` is created
- [ ] Everything that can be represented with a plot instead of a table is a plot
- [ ] Report sections have a description and help text (with `self.add_section`)
- [ ] There aren't any huge tables with > 6 columns (explain reasoning if so)
- [ ] Each table column has a different colour scale to its neighbour, which relates to the data (eg. if high numbers are bad, they're red)
- [ ] Module does not do any significant computational work

PR #1699, opened 2022-06-05
added gfastats

gf777:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ Thumbs.db
 
 .idea/
 runTest.sh
+multiqc.xcworkspace/contents.xcworkspacedata
+multiqc.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
+multiqc.xcworkspace/xcuserdata/gformenti.xcuserdatad/UserInterfaceState.xcuserstate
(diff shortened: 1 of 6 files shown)


ewels (changes requested):
Just fixed up the merge conflicts and made a couple of minor tweaks.

Main thing left is the two items in the PR template:

There should be 2-3 columns added to General Stats, maximum. This table is specifically for comparing to other stats from other modules, so shouldn't be so long.

- Please remove any columns that are not interesting
- Please thing about whether any columns can be combined into a plot - eg. do the `# xx components` fields sum? Could they be a stacked bar plot instead?
- With columns that are left, you can move into a dedicated table within a module section. Then the limit on columns is not the same.
- Consider breaking into multiple tables to make it easier to read / interpret.
- The columns have no configuration or customisation at all
   - Please add colours
   - Please use [number base multipliers](https://multiqc.info/docs/#number-base-multiplier)
   - Please remove decimals where discrete data
   - Please use [`shared key`s](https://multiqc.info/docs/#creating-a-table) when columns are numerically comparable

ewels:
I'd forgotten that we'd already discussed this module in PR #1641 before it got split out here. I had said basically the same thing, and looks like you already had some ideas about how to split the table up: https://github.com/ewels/MultiQC/pull/1641#issuecomment-1146219889

Issue #1654, opened 2022-03-25
Division by Zero in Sambamba Markdup


epruesse:
### Description of bug

In file `sambamba/sambamba.py` line 93:

https://github.com/ewels/MultiQC/blob/45ce9ae0160c38e250e3e0e4a9335adc5af26b6b/multiqc/modules/sambamba/markdup.py#L93

A simple `try/except` should do it. May want to check every place in every module divisions occur. If you have a division, you'll eventually have a division by zero...

### File that triggers the error

Simple all empty (previously filtered to nothing) file:

```
finding positions of the duplicate reads in the file...
  sorted 0 end pairs
     and 0 single ends (among them 0 unmatched pairs)
  collecting indices of duplicate reads...   done in 0 ms
  found 0 duplicates
collected list of positions in 0 min 0 sec
removing duplicates...
collected list of positions in 0 min 0 sec
```

aidaanva:
Hi,
I also had the same error running nf-core/eager and affecting MALT, in this case there was one sample with 0 reads assigned. 

The error looks like this: 

`  /// MultiQC 🔍 | v1.11

|           multiqc | MultiQC Version v1.12 now available!
|           multiqc | Only using modules: adapterRemoval, bowtie2, custom_content, damageprofiler, dedup, fastp, fastqc, gatk, kraken, malt, mtnucratio, multivcfanalyzer, picard, preseq, qualimap, samtools, sexdeterrmine, hops, bcftools
|           multiqc | Search path : /mnt/archgen/MICROSCOPE/pathogen_screening/2021-03-09-keltenfuersten_BW/fullbacetvir2017/work/5c/a5b3dca9f4b85354e0b9529b211141/multiqc_config.yaml
|           multiqc | Search path : /mnt/archgen/MICROSCOPE/pathogen_screening/2021-03-09-keltenfuersten_BW/fullbacetvir2017/work/5c/a5b3dca9f4b85354e0b9529b211141
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 1006/1006  
╭────────────────── Oops! The 'malt' MultiQC module broke... ──────────────────╮
│ Please copy this log and report it at                                        │
│ https://github.com/ewels/MultiQC/issues                                      │
│ Please attach a file that triggers the error. The last file found was:       │
│ ./malt/malt.log                                                              │
│                                                                              │
│ Traceback (most recent call last):                                           │
[...]

PR #1700, opened 2022-06-07
fix zero division error in sambamba markdup module #1654

gartician:

This PR helps catch the zero-division problem in the sambamba markdup module referenced in https://github.com/ewels/MultiQC/issues/1654


ewels (approved):
Looks great, thank you!

PR #1704, opened 2022-06-13
Checkatlas

drbecavin:
[NEW MODULE] This comment contains a description of changes (with reason)
- [x] CHANGELOG.md has been updated
- [x] (push last thursday)] There is example tool output for tools in the https://github.com/ewels/MultiQC_TestData repository or attached to this PR
- [ ] NO (do not understand what to do ?) docs/README.md is updated with link to below
- [x]  `docs/modulename.md` is created
- [x] Report sections have a description and help text (with self.add_section)


drbecavin:
Alright ! Thanks! I did not know that.

PR #1707, opened 2022-06-16
New Module: Anglerfish

Juho0:
Anglerfish is a tool designed to de-multiplex Illumina libraries sequenced on Oxford Nanopore Technologies flowcells for the purpose of quality control. Assessment of pool balancing, contamination and insert sizes are currently supported.


multiqc/modules/anglerfish/anglerfish.py
```diff
@@ -0,0 +1,247 @@
...
+        ## Index for each sample and their reads in order to iterate without knowing sample names
+        index = 0
+        try:
+            for k in parsed_json["sample_stats"]:
+                key_list = k.keys()
+                for key in key_list:
+                    if key != "sample_name":
+                        self.anglerfish_data[s_name][key + "_{}".format(index)] = float(k[key])
+                    else:
+                        self.anglerfish_data[s_name]["sample_name_{}".format(index)] = k["sample_name"]
```
ewels, line 83:
```suggestion
                        self.anglerfish_data[s_name][f"sample_name_{index}"] = k["sample_name"]
```

ewels: Is this not the same output as the generic line above? Why do you need the `if key != "sample_name":`?

multiqc/modules/anglerfish/anglerfish.py
```diff
@@ -0,0 +1,247 @@
...
+
+        # Parse Sample Stats
+        ## Index for each sample and their reads in order to iterate without knowing sample names
+        index = 0
+        try:
+            for k in parsed_json["sample_stats"]:
+                key_list = k.keys()
+                for key in key_list:
+                    if key != "sample_name":
+                        self.anglerfish_data[s_name][key + "_{}".format(index)] = float(k[key])
```
ewels, line 81: This is the standard way of doing format strings in Python 2 (which MultiQC was originally written in). However, since MultiQC became Python 3+ only a while back, I now prefer to write in the newer f-string syntax:

```suggestion
                        self.anglerfish_data[s_name][f"{key}_{index}"] = float(k[key])
```

I think it's usually quite a lot easier to read..

multiqc/modules/anglerfish/anglerfish.py
```diff
@@ -0,0 +1,247 @@
...
+                    else:
+                        self.anglerfish_data[s_name]["sample_name_{}".format(index)] = k["sample_name"]
+                index += 1
+            self.anglerfish_data[s_name]["sample_stats_amount"] = index
+        except (KeyError):
```
ewels, line 86:
```suggestion
        except KeyError:
```

multiqc/modules/anglerfish/anglerfish.py
```diff
@@ -0,0 +1,247 @@
...
+        total_count = 0
+        try:
+            for k in parsed_json["undetermined"]:
+                if len(k) > 0:
+                    self.anglerfish_data[s_name]["undetermined_count_{}".format(index)] = float(k["count"])
+                    total_count += float(k["count"])
+                    self.anglerfish_data[s_name]["undetermined_index_{}".format(index)] = k["undetermined_index"]
+                    index += 1
+            self.anglerfish_data[s_name]["undetermined_amount"] = index
+        except (KeyError):
```
ewels, line 102:
```suggestion
        except KeyError:
```

multiqc/modules/anglerfish/anglerfish.py
ewels: More of these elsewhere in the file, please update throughout.

multiqc/modules/anglerfish/anglerfish.py
```diff
@@ -0,0 +1,247 @@
...
+                    data[key]["library"] = float((reads / total_read) * 100)
+                    data["undetermined"]["library"] = float((total_count / total_read) * 100)
+            except (KeyError):
+                log.debug("No general stats table generated from Anglerfish json: {}".format(s_name))
+            except (ZeroDivisionError):
+                log.debug("No library in general stats table generated from Anglerfish json: {}".format(s_name))
+
+        headers = OrderedDict()
+        headers["library"] = {
+            "namespace": "Anglerfish",
```
ewels, line 137: No need for these, it should be the default.

multiqc/modules/anglerfish/anglerfish.py
```diff
@@ -0,0 +1,247 @@
...
+        headers["#reads"] = {
+            "namespace": "Anglerfish",
+            "title": "#reads",
+            "description": "Amount of reads",
+            "min": 0,
+            "scale": "PuOr",
+        }
+        headers["mean_read_len"] = {
+            "namespace": "Anglerfish",
+            "title": "mean_read_len",
```
ewels, line 155: Title case and as easy as read as possible please (whilst keeping as short as possible).

```suggestion
            "title": "Read Length",
```

multiqc/modules/anglerfish/anglerfish.py
```diff
@@ -0,0 +1,247 @@
...
+            "description": "Amount of library used. Calculated from a samples #reads divided by total number of input reads",
+            "max": 100,
+            "min": 0,
+            "scale": "PuBu-rev",
+            "suffix": " %",
+        }
+
+        headers["#reads"] = {
+            "namespace": "Anglerfish",
+            "title": "#reads",
```
ewels, line 148:
```suggestion
            "title": "# Reads",
```

multiqc/modules/anglerfish/anglerfish.py
```diff
@@ -0,0 +1,247 @@
...
+            "description": "Amount of reads",
```
ewels, line 149: I don't love the word "Amount" used in this context.

```suggestion
            "description": "Total number of reads",
```

multiqc/modules/anglerfish/anglerfish.py
```diff
@@ -0,0 +1,247 @@
...
+            except (KeyError):
+                log.debug("No general stats table generated from Anglerfish json: {}".format(s_name))
+            except (ZeroDivisionError):
+                log.debug("No library in general stats table generated from Anglerfish json: {}".format(s_name))
+
+        headers = OrderedDict()
+        headers["library"] = {
+            "namespace": "Anglerfish",
+            "title": "library",
+            "description": "Amount of library used. Calculated from a samples #reads divided by total number of input reads",
```
ewels, line 139: I don't understand this description sorry..?

multiqc/modules/anglerfish/anglerfish.py
```diff
@@ -0,0 +1,247 @@
...
+        headers["mean_read_len"] = {
+            "namespace": "Anglerfish",
+            "title": "mean_read_len",
+            "description": "Mean read length",
+            "min": 0,
+            "scale": "RdYlGn",
+        }
+        headers["std_read_len"] = {
+            "namespace": "Anglerfish",
+            "title": "std_read_len",
```
ewels, line 162:
```suggestion
            "title": "Length StdDev",
```

multiqc/modules/anglerfish/anglerfish.py
```diff
@@ -0,0 +1,247 @@
...
+                    data["Sample: {}".format(sample_name)]["Mean"] = self.anglerfish_data[s_name][
+                        "mean_read_len_{}".format(i)
+                    ]
+                    data["Sample: {}".format(sample_name)]["Standard Deviation"] = self.anglerfish_data[s_name][
+                        "std_read_len_{}".format(i)
+                    ]
+            else:
+                # For non existing sample stat and faulty sample stat
+                debug_list.append(s_name)
+        if len(data) != 0:
```
ewels, line 193: Can be nicer to do something like:

```python
if len(data) == 0:
    return
```

Saves indenting a massive chunk of code and less likely to miss something.

multiqc/modules/anglerfish/anglerfish.py
```diff
@@ -0,0 +1,247 @@
...
+            }
+            # Plot table if less than 10 samples exist, beeswarm if more
+            p = ""
+            if total_samples < 10:
+                p = table.plot(data, None, config)
+            else:
+                p = beeswarm.plot(data, None, config)
+            self.add_section(
+                name="Read Lengths Summary",
+                anchor="Anglerfish-sample-statistics",
```
ewels, line 206: Lower case for HTML anchors generally.

```suggestion
                anchor="anglerfish-sample-statistics",
```

multiqc/modules/anglerfish/anglerfish.py
```diff
@@ -0,0 +1,247 @@
...
+        """Generate Undetermined indexes Bar Plot"""
+        data = {}
+        debug_list = []
+        for s_name in self.anglerfish_data:
+            index = self.anglerfish_data[s_name]["undetermined_amount"]
+            # Index smaller than 0 caused by KeyError from no undetermined data
+            if index > 0:
+                for i in range(index):
+                    undetermined_index = self.anglerfish_data[s_name]["undetermined_index_{}".format(i)]
+                    data["{}".format(undetermined_index)] = {}
```
ewels, line 224: No need for an f string here as you're not mixing the variable with any strings.

```suggestion
                    data[undetermined_index] = {}
```

multiqc/modules/anglerfish/anglerfish.py
```diff
@@ -0,0 +1,247 @@
...
+                    data["Sample: {}".format(sample_name)] = {}
+                    data["Sample: {}".format(sample_name)]["Mean"] = self.anglerfish_data[s_name][
+                        "mean_read_len_{}".format(i)
+                    ]
+                    data["Sample: {}".format(sample_name)]["Standard Deviation"] = self.anglerfish_data[s_name][
+                        "std_read_len_{}".format(i)
+                    ]
+            else:
+                # For non existing sample stat and faulty sample stat
+                debug_list.append(s_name)
```
ewels, line 192: This seems to only be used in a loop for log messages at the end. I think you can just log here directly and not bother with the variable?

Same pattern later as well.

ewels (commented):
Haven't tried running the code yet, but from a quick read through looks very, very good 👏🏻 

Excellent work!

Some minor suggestions to tidy up, will try to find time to actually run and test ASAP.

multiqc/modules/anglerfish/anglerfish.py
Juho0: Done!

multiqc/modules/anglerfish/anglerfish.py
Juho0: > Is this not the same output as the generic line above? Why do you need the `if key != "sample_name":`?

The values are represented as strings, but for all keys except "sample_name" I want the value as a float.

multiqc/modules/anglerfish/anglerfish.py
Juho0: Done

multiqc/modules/anglerfish/anglerfish.py
Juho0: Done

multiqc/modules/anglerfish/anglerfish.py
Juho0: Done

multiqc/modules/anglerfish/anglerfish.py
Juho0: Done!

multiqc/modules/anglerfish/anglerfish.py
Juho0: Done!

multiqc/modules/anglerfish/anglerfish.py
Juho0: Done!

multiqc/modules/anglerfish/anglerfish.py
Juho0: Done!

multiqc/modules/anglerfish/anglerfish.py
Juho0: Done

multiqc/modules/anglerfish/anglerfish.py
Juho0: Done

multiqc/modules/anglerfish/anglerfish.py
Juho0: Changed

ewels:
Additional functionality:

- Scale % library to number of reads that have a barcode (so that %ages sum to 100%)
- Add bar plot showing % reads with / without barcode

ewels:
* # Reads should be formatted without a decimal in the table `{.0f}`

multiqc/modules/anglerfish/anglerfish.py
ewels: I prefer to try to make _everything_ a float in this case, then capture the `ValueError` if it goes wrong and set as a string in those cases.

Minor personal preference, but avoids needing to specify key names which is always nice.

ewels: Done in 1a24f4d61e832736ec9df6e704814591b0acd745

ewels (approved):
Great stuff, thanks @Juho0 !

PR #1728, opened 2022-07-20
New module: Porechop

jfy133:
Adds module for [Porechop](https://github.com/rrwick/Porechop)

- [x] There is example tool output for tools in the <https://github.com/ewels/MultiQC_TestData> repository or attached to this PR: https://github.com/ewels/MultiQC_TestData/pull/235

Example output:
[multiqc_report.zip](https://github.com/ewels/MultiQC/files/9149741/multiqc_report.zip)


jfy133:
I think tests are failing possibly because it's taking too long to search for the string inside each log file... need to work out how to speed up I guess

docs/modules/porechop.md
```diff
@@ -0,0 +1,11 @@
+---
+Name: Porechop
+URL: https://github.com/rrwick/Porechop
+Description: >
+  Porechop is a tool for finding and removing adapters from Oxford Nanopore reads. Adapters on the ends of reads are trimmed off, and when a read has an adapter in its middle, it is treated as chimeric and chopped into separate reads. Porechop performs thorough alignments to effectively find adapters, even at low sequence identity.
+---
+
+The Porechop module parses the log file generated by
+[Porechop](https://github.com/rrwick/Porechop), which is a tool for demultiplexing and processing nanopore data.
```
jfy133, line 9:
```suggestion
The Porechop module parses the log file generated by [Porechop](https://github.com/rrwick/Porechop), which is a tool for demultiplexing and processing nanopore data.
```

ewels:
I'm tempted to combine the 3 plots into 1, with 3 datasets. Feels quite verbose for fairly simple stats. But I'll leave it for now as it's late and I should be in bed.

It's an easy change if you agree though: https://multiqc.info/docs/#switching-datasets

jfy133:
Thanks @ewels ! I wasn't sure what was the normal practise here, looking through other examples it looked like switching was between counts and percentages, and the percentages were calculated by the module itself and wasn't derived from the log files themselves.

Therefore in this case were the percentages were in the porechop log files, I thought it better to plot each set of raw data separately - but I have no preference either way.

ewels:
I don't mean the percentages really - that bit is fine. I more meant combining the plots of the three different metrics (start, end, split) into one section / plot. Just with tabs to move between them. Each would still have counts / percentages in the same way.

PR #1731, opened 2022-07-26
New module: DIAMOND

mjamy:
Adds module for [DIAMOND](https://github.com/bbuchfink/diamond).

- [x] There is example tool output for tools in the <https://github.com/ewels/MultiQC_TestData> repository or attached to this PR: https://github.com/ewels/MultiQC_TestData/pull/237

Example output:
[multiqc_report.zip](https://github.com/ewels/MultiQC/files/9190993/multiqc_report.zip)


ewels:
As with https://github.com/ewels/MultiQC/pull/1732 - I'm not sure that the bar plot adds anything, so I've removed it. Hope that's ok.

multiqc/utils/search_patterns.yaml
```diff
@@ -272,6 +272,9 @@ deeptools/plotPCAData:
 deeptools/plotProfile:
   contents: "bin labels"
   num_lines: 1
+diamond:
+  contents: "type: Diamond database"
+  fn: "*.log"
```
ewels, line 277: Does this module parse only the output from the  `--log` flag? https://github.com/bbuchfink/diamond/wiki/3.-Command-line-options#general-options

In which case, is the filename _always_ `diamond.log`? That's much easier (and faster) to search for..

ewels: Talked to @jfy133 and it looks like this is the case. I updated the module accordingly in https://github.com/ewels/MultiQC/commit/a48995423992146f41e1f7be423bec40259db1ab:

- Update search pattern
- Parse sample name from logged command instead of filename
- Fall back to log directory name

Note that if you're manually renaming these log files in the taxprofiler pipeline, you'll need to tweak the search pattern in the pipeline MultiQC config file for it to still find the files. See [docs](https://multiqc.info/docs/#module-search-patterns).

PR #1732, opened 2022-07-26
New module: Filtlong

sofstam:
Adds module for [Filtlong](https://github.com/rrwick/Filtlong)

- [x] There is example tool output for tools in the <https://github.com/ewels/MultiQC_TestData> repository or attached to this PR https://github.com/ewels/MultiQC_TestData/pull/236

Example output:
[multiqc_report.zip](https://github.com/ewels/MultiQC/files/9192338/multiqc_report.zip)


multiqc/modules/filtlong/filtlong.py
```diff
@@ -0,0 +1,114 @@
...
+                self.filtlong_data[s_name]["Target bases"] = {}
+                self.filtlong_data[s_name]["Target bases"] = float(l.lstrip().split(" ")[1])
+
+            elif "keeping" in l:
+                self.filtlong_data[s_name]["Keeping bases"] = {}
+                self.filtlong_data[s_name]["Keeping bases"] = float(l.lstrip().split(" ")[1])
+
+            elif "fall below" in l:
+                self.filtlong_data[s_name]["Keeping bases"] = {}
+                self.filtlong_data[s_name]["Keeping bases"] = float(0)
```
ewels, line 60: I'm a bit nervous about this. Does the log really mean that 0 bases are kept? I read it as meaning that _all_ bases were kept, rather than 0:

```
Filtering long reads
  target: 183248995 bp
  reads already fall below target after filtering

Outputting passed long reads
```

Now I'm not sure!

multiqc/modules/filtlong/filtlong.py
```diff
@@ -0,0 +1,114 @@
...
+    def parse_logs(self, logfile):
+        """Parsing Logs. Note: careful of ANSI formatting log"""
+        file_content = logfile["f"]
+        for l in file_content:
+            # Find the valid metric
+            if "target:" in l:
+                s_name = logfile["s_name"]
+                self.add_data_source(logfile, s_name=s_name)
+                self.filtlong_data[s_name] = {}
+                self.filtlong_data[s_name]["Target bases"] = {}
```
ewels, line 51: No need to define an empty dict, only to overwrite it again in the next line.. (we're not using keys within this dict as we were on the line above).

ewels (approved):
Great! I've made a few edits but I can't push directly your PR as I don't have push permissions (this is probably as it's come from a GitHub org and not your personal user account).

The main thing I've done is to remove the bar plots. They only have a single category (single number) and this duplicates what is already in the General Statistics table, so I'm not really sure that they add anything to the report.

I'm a bit nervous about setting the reads to `0` as in my review comment, but I'll merge anyway as I know you need this released. We can always create a patch if needed.

ewels:
You can see my changes here: https://github.com/ewels/MultiQC/compare/b5773d2b..6b3b8c2e

* 6b3b8c2e Filtlong: Make search pattern slightly more specific
* 82314c06 Keeping bases -> Bases kept
* a7cdb268 Filtlong: Remove bar plots
* 5b86e259 Code cleanup
* bb82a544 Run pre-commit
*   b5773d2b Fix merge conflicts

multiqc/modules/filtlong/filtlong.py
ewels: Ok, I adjusted this in 66fea63ac86316ab56fc3802b23a346b0de90234 to simply omit this value instead of setting it to `0`. Hopefully that's safer.

We can review in the future if needed.

Issue #1455, opened 2021-06-16
Add Nextclade support


drpatelh:
**Tool Details**

- Name of tool: `Nextclade`
- Tool homepage: `https://github.com/nextstrain/nextclade`
- Tool description: `Viral genome alignment, clade assignment, mutation calling, and quality checks`
- Data suitable for MultiQC plot(s): `table`
- Most interesting data for General Stats table: `clade`

**Tool output:**

- Log filename pattern: ~~`*.nextclade.*`~~ _[none]_
- File contents: Full example output files can be obtained from [here](https://nf-co.re/viralrecon/results#viralrecon/results-f0171324a60d1759ca7f5d02351372d97d22cc12/platform_illumina/variants/ivar/nextclade/).

Nextclade can output various file formats as listed [here](https://github.com/nf-core/modules/blob/c8168bc3516d6f8c3dc6d47c5142073cee82934e/software/nextclade/main.nf#L26-L30) but adding `csv` support would be a good start.

ewels:
[SAMPLE_01.csv](https://github.com/ewels/MultiQC/files/6663949/SAMPLE_01.csv)

Same questions as in https://github.com/ewels/MultiQC/issues/1454#issuecomment-862456067

> @drpatelh - can there be more than one sample in these files sometimes? (generally, not just within viralrecon). Or is it always just two lines?
> 
> And is it the tool that has the `*. nextclade.*` filename, or is that the pipeline? ie. can we be sure that all Nextclade users will have these filenames for the outputs?

drpatelh:
Same answers as https://github.com/ewels/MultiQC/issues/1454#issuecomment-862465024 :)

drpatelh:
Will try and generate a multi-line file for you to test. Ping me on Slack if I forget.

PR #1512, opened 2021-08-06
Add Nextclade module

ErikDanielsson:
Added new `Nextclade` module.  Some of the columns in the `Nextclade` table should probably removed. Will close #1455.

- [x] This comment contains a description of changes (with reason)
- [x] `CHANGELOG.md` has been updated
- [x] There is example tool output for tools in the <https://github.com/ewels/MultiQC_TestData> repository or attached to this PR
- [x] Code is tested and works locally (including with `--lint` flag)
- [x] `docs/README.md` is updated with link to below
- [x] `docs/modulename.md` is created
- [x] Everything that can be represented with a plot instead of a table is a plot
- [x] Report sections have a description and help text (with `self.add_section`)
- [ ] There aren't any huge tables with > 6 columns (explain reasoning if so)
- [x] Each table column has a different colour scale to its neighbour, which relates to the data (eg. if high numbers are bad, they're red)
- [x] Module does not do any significant computational work

PR #1740, opened 2022-08-09
Implement nextclade functionality

jchorl:
Extension on https://github.com/ewels/MultiQC/pull/1512

> Added new Nextclade module. Some of the columns in the Nextclade table should probably removed. Will close https://github.com/ewels/MultiQC/issues/1455.

I merged in those commits to my branch, then merged master, resolved conflicts, and here we are.

I'm attaching a sample report from the test data. github only supports zips, so unzip first.
[multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/9295215/multiqc_report.html.zip)

    - There is a very wide table, but most of the columns are hidden by default. There are 5 visible (6 with sample name)


ewels:
Nice! Just needs a DOI in the module setup for tests to pass.

Thanks for this!

jchorl:
[multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/9331804/multiqc_report.html.zip)

Done!

ewels (approved):
Great stuff, thanks all!

PR #1741, opened 2022-08-10
New module: PRINSEQ++

jfy133:

- [x] There is example tool output for tools in: https://github.com/ewels/MultiQC_TestData/pull/238

image


ewels:
@jfy133 - code looks good, but two questions on the report output:

1. The numbers in the General Stats table and the beeswarm look basically the same. Can we just have the 2-3 most important columns in GenStat and everything else in the plot?
2. Can the beeswarm plot instead be a stacked bar plot? Do the numbers sum up to the correct total? I think that this would be far easier to read.

jfy133:
Thanks @ewels!

1. Not really - each filter is somewhat independent from one another (they aren't strictly additive, you could have one or the other depending on which flags you provide), and there is no default.
2. They could theoretically be yes, I opted for the beeswarm as generally barcharts to me should represent fractions of the whole (in this case number of reads), whereas the log file just tells you the fractions that are removed - i.e., there is no 'input total' value to then represent the full 100% (if that makes sense). But I'm not that bothered as it's just a personal 'intutition', fine to have a barchart if that makes more sense to you.

ewels:
Ok, let's sum the values for the general stats to avoid duplication and give the key figure.

No point doing the barplot if we don't have the read count total. Can add this in the future if that ever makes it into the log file.

PR #1742, opened 2022-08-11
New module: BBDuk

jfy133:
This is a standalone module from BBMap, which partly uses in BBDuk, however BBDuk 'standalone' produces different stdout (log) files, thus the module. 

- [x] There is example tool output for tools in the <https://github.com/ewels/MultiQC_TestData> repository or attached to this PR: ewels/MultiQC_TestData/pull/239 


ewels:
Ok, I've made quite a lot of changes here so please have a thorough review if you can.

* Rewrote the log parsing to use regexes, and to parse the base numbers as well as read counts
* Take the sample name from the first input file name, rather than the log file name
* Removed most General Stats columns
* Added the filtering types to the bar plot
* Added base count as a second dataset to the bar plot
* Removed the beeswarm plot

Let me know what you think!

PR #1745, opened 2022-08-14
report median read length for fastqc

jchorl:
Uses median read length for fastqc, instead of mean. Note that **this is a breaking change**. Another option is to add an extra column, to keep both.

The median is way more suitable for sequencing reads as it doesn't assume the length is normally distributed.

There are issues on fastp to output more stats, but they don't seem to have much traction. I suppose I could also try to PR them, but they seem less responsive.

You can see a subtle difference in lengths on the testdata repo:
Old:
Screenshot 2022-08-14 at 23-02-01 MultiQC Report

New:
Screenshot 2022-08-14 at 22-59-41 MultiQC Report


ewels (changes requested):
Yeah, I wrote this code back before long reads were really a thing and it was more of a fair assumption 😅 

I'm a bit nervous about the breaking change as this is by far the most used module. As such, I'd prefer to keep the previous average and add this in as an additional value if possible. At least then the same value as before will be present in the data exports for downstream analysis.

Also, I'd like the column title to be slightly different, eg. `Median Read Length` so that there's a visible difference between reports.

multiqc/modules/fastqc/fastqc.py
```diff
@@ -195,14 +195,31 @@ def parse_fastqc_report(self, file_contents, s_name=None, f=None):
...
+
+        for d in sequence_length_distributions:
+            length_reads += d["count"]
             length_bp += d["count"] * self.avg_bp_from_range(d["length"])
-            total_count += d["count"]
+
+            if median is None and length_reads >= total_count / 2:
+                # if the distribution-entry is a range, we use the average of the range.
+                # this isn't technically correct, because we can't know what the distribution
+                # is within that range. Probably good enough though.
```
ewels, line 216: We should really ask @s-andrews to add mean / median read length as an output metric from FastQC directly instead of fudging it like this 🤔 

@jchorl - if you fancy adding an issue that would be nice 😉  https://github.com/s-andrews/FastQC

multiqc/modules/fastqc/fastqc.py
```diff
@@ -268,14 +286,23 @@ def fastqc_general_stats(self):
...
             "hidden": hide_seq_length,
         }
+        headers["median_sequence_length"] = {
+            "title": "Median Read Length",
+            "description": "Median Read Length (bp)",
+            "min": 0,
+            "suffix": " bp",
+            "scale": "RdYlGn",
+            "format": "{:,.0f}",
+            "hidden": True,
```
ewels, line 304: I'm happy for the median to be the one that is shown by default (I'd prefer this actually).

I was mostly thinking about the `multiqc_data` outputs for the backwards compatibility to be honest.

multiqc/modules/fastqc/fastqc.py
```diff
@@ -268,14 +286,23 @@ def fastqc_general_stats(self):
...
         }
         headers["avg_sequence_length"] = {
-            "title": "Read Length",
+            "title": "Average Read Length",
             "description": "Average Read Length (bp)",
             "min": 0,
             "suffix": " bp",
             "scale": "RdYlGn",
             "format": "{:,.0f}",
             "hidden": hide_seq_length,
```
ewels, line 295:
```suggestion
            "hidden": True,
```

multiqc/modules/fastqc/fastqc.py
```diff
@@ -242,9 +260,9 @@ def fastqc_general_stats(self):
                 pass
 
         # Are sequence lengths interesting?
-        seq_lengths = [x["avg_sequence_length"] for x in data.values()]
+        avg_seq_lengths = [x["avg_sequence_length"] for x in data.values()]
```
ewels, line 263:
```suggestion
        median_seq_lengths = [x["median_sequence_length"] for x in data.values()]
```

multiqc/modules/fastqc/fastqc.py
```diff
@@ -242,9 +260,9 @@ def fastqc_general_stats(self):
...
         try:
-            hide_seq_length = False if max(seq_lengths) - min(seq_lengths) > 10 else True
+            hide_seq_length = False if max(avg_seq_lengths) - min(avg_seq_lengths) > 10 else True
```
ewels, line 265:
```suggestion
            hide_seq_length = False if max(median_seq_lengths) - min(median_seq_lengths) > 10 else True
```

multiqc/modules/fastqc/fastqc.py
ewels:
```suggestion
            "hidden": hide_seq_length,
```

multiqc/modules/fastqc/fastqc.py
jchorl: This should really just be:
```
hide_seq_length = max(median_seq_lengths) - min(median_seq_lengths) <= 10
```

multiqc/modules/fastqc/fastqc.py
jchorl: Will do!

jchorl:
Suggestions incorporated and tested:
Screenshot 2022-12-16 at 21-46-50 MultiQC Report


Will cut an issue shortly on fastqc

ewels (approved):

ewels (approved):
Looks great. Thanks for the tweaks!

jchorl:
> Sometimes I don't see the optional median length column. How can I unhide it?

Have you clicked "Configure Columns"? Does it show up in that menu?

santataRU:
> > Sometimes I don't see the optional median length column. How can I unhide it?
> 
> Have you clicked "Configure Columns"? Does it show up in that menu?

Thanks for your suggestions! It works!

PR #1751, opened 2022-08-30
New module: Truvari

pontushojer:
This PR adds a new module for the tool Truvari (https://github.com/ACEnglish/truvari). Specifically the output from the command `truvari bench` is supported. Below is a screenshot of the results when run on the test data.

image

PR for test data have been submitted here (https://github.com/ewels/MultiQC_TestData/pull/241)


vladsavelyev:
@multiqc-bot changelog

multiqc/modules/truvari/bench.py
```diff
@@ -53,6 +53,9 @@ def parse_bench_stats(self):
                 self.add_data_source(f, section="bench")
                 data[f["s_name"]] = stats
 
+                if version is not None:
+                    self.add_software_version(version, f["s_name"])
```
vladsavelyev, line 57: Nice!!

multiqc/modules/truvari/bench.py
```diff
@@ -0,0 +1,234 @@
...
+
+        self.add_section(
+            name="Truvari bench",
+            anchor="truvari-bench",
+            description="Concordance statistics parsed from the output from <code>truvari bench</code>.",
+            plot=table.plot(data, keys, {"id": "truvari-bench-summary"}),
+        )
+        # Generate shuffled list of colors to label samples
+        colors = list(mcolors.CSS4_COLORS)
+        random.shuffle(colors)
```
vladsavelyev, line 197: To make it reproducible, I'd pass a seed to random.

But even better, I'd use the `mqc_colour` module, pick a scale, and pass the sample index into it. See how it's done for Freyja: https://github.com/ewels/MultiQC/blob/256fd77e4e99ad0ba11cc6a609a9e7b839b1a3f0/multiqc/modules/freyja/freyja.py#L58
https://github.com/ewels/MultiQC/blob/256fd77e4e99ad0ba11cc6a609a9e7b839b1a3f0/multiqc/modules/freyja/freyja.py#L100

multiqc/modules/truvari/bench.py
```diff
@@ -0,0 +1,234 @@
...
+
+        # Check that not too many samples for scatter plot
+        if len(data) > len(colors):
+            log.debug("Too many samples ({}) to generate scatter plot.".format(len(data)))
+            return len(data)
```
vladsavelyev, line 202: I think it's not a good idea to limit the number of samples here. If we scale up, it's okay if some samples have the same color, but we still should support the arbitrary number of samples. Probably in the future MultiQC will be smarter with the large numbers of samples and generate more efficient plots e.g. adding interactivity for only outliers.

pontushojer: Yes, not sure why I though this was a good idea. It is removed now!

multiqc/modules/truvari/bench.py
pontushojer: I think I got too hanged up on not repeating colors, but with the tooltip this is not really a problem.  Now it uses the `mqc_colour` module instead of my bespoke solution ;)

vladsavelyev (approved):
Great, looks perfect to me now!

ewels:
Thanks @pontushojer! Just had a look at the report, quick naive question:

* From the header tooltips, it kind of sounds like `TP (base)` + `FP` = `Base calls`? But that's not the case. Is that expected?
    * `Number of the base VCF calls matching the comp VCF` + `Number of the comp VCF calls non-matching the base VCF` = `Number of calls in the base VCF`
* If those numbers are meant to sum (😅) would a stacked bar plot work / be better?

ewels (commented):
Nice! A few minor comments but generally the code looks great 👍🏻 

It would be nice to add some `helptext` strings above the sections, explaining what the numbers mean and how to interpret them. It's ok to copy this from the source tool docs, if available.

multiqc/modules/truvari/bench.py
```diff
@@ -0,0 +1,228 @@
...
+            if stats:
+                # Use output directory as sample name
+                f["s_name"] = os.path.basename(f["root"])
+                f["s_name"] = self.clean_s_name(f["s_name"], f, root=os.path.dirname(f["root"]))
+
+                # Load stats
+                stats = json.loads(str(stats))
+
+                # Convert entries with NaN to 0.0
+                stats = {k: v if v != "NaN" else 0.0 for k, v in stats.items()}
```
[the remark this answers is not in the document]
pontushojer: I cannot acctually remember why I added this. As the test files works fine without it so I am just gonna remove it.

multiqc/modules/truvari/bench.py
```diff
@@ -0,0 +1,228 @@
...
+        # that, we will reuse colors.
+        color_scale = mqc_colour_scale("plot_defaults")
+
+        # Make scatter plot
+        scatter_data = {}
+        for i, (sample, sample_data) in enumerate(data.items()):
+            scatter_data[sample] = {
+                "x": sample_data["precision"] * 100.0,
+                "y": sample_data["recall"] * 100.0,
+                "color": color_scale.get_colour(i, lighten=0.9),
```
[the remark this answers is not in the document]
pontushojer: My thinking is that the different colours allows you to keep track of point if you export the plot as a flat image. With the interactive plots I can see that there may not be a need for this.

multiqc/modules/truvari/bench.py
```diff
@@ -0,0 +1,228 @@
...
+                if "Truvari version:" in line:
+                    version = line.strip().split(":")[-1]
+
+            if stats:
+                # Use output directory as sample name
+                f["s_name"] = os.path.basename(f["root"])
+                f["s_name"] = self.clean_s_name(f["s_name"], f, root=os.path.dirname(f["root"]))
+
+                # Load stats
+                stats = json.loads(str(stats))
```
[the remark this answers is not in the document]
pontushojer: Added in latest version

multiqc/modules/truvari/bench.py
```diff
@@ -0,0 +1,228 @@
...
+            "placement": 101,
+            "scale": "RdYlGn",
+        }
+        bench_headers["f1"] = {
+            "title": "F1",
+            "description": "F1 score of the SV calls. Definition: 2 * ((recall * precision) / (recall + precision))",
+            "suffix": "%",
+            "modify": lambda x: x * 100,
+            "placement": 102,
+            "scale": "RdYlGn",
```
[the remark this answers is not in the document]
pontushojer: Fixed!

pontushojer:
> Thanks @pontushojer! Just had a look at the report, quick naive question:
> 
> * From the header tooltips, it kind of sounds like `TP (base)` + `FP` = `Base calls`? But that's not the case. Is that expected?
>   
>   * `Number of the base VCF calls matching the comp VCF` + `Number of the comp VCF calls non-matching the base VCF` = `Number of calls in the base VCF`

> * If those numbers are meant to sum (😅) would a stacked bar plot work / be better?

Yeah, its actually `TP (base)` + `FN` = `Base calls` and similarly `TP (comp)` + `FP` = `Comp calls`. I have not update the descriptions and added helptext to clarify this. 

The bar plot was a nice suggestion so I also added this. This allows one to see a breakdown of classification for call from the perspective of `Comp` (the comparison VCF)

image

and perspective of `Base` (the baseline "truth" VCF).

image

I also updated several other things since there were slight changes to the truvari output from version v4.0.0. Some statistics had been renamed and version parsing also needed an update. I added a new log for this version in this PR https://github.com/ewels/MultiQC_TestData/pull/284

vladsavelyev:
The bar plot looks great! 

Though it looks like it now fully replicates the information shown in the table. And perhaps with another tab for the GT-matched version of the statistics, it can also cover the hidden columns as well, so the table could be dropped in favour of the bar plot? 🤔

pontushojer:
> The bar plot looks great!
> 
> Though it looks like it now fully replicates the information shown in the table. And perhaps with another tab for the GT-matched version of the statistics, it can also cover the hidden columns as well, so the table could be dropped in favour of the bar plot? 🤔

Yes, I was actually starting to think the same when I implemented this. Now I have added the genotype data to the bar graph as well. I also moved the "GT concordance" stat to the general stats table. Then I remove the table as its all all represented in the other plots/tables. 

Here is the current report sections

**General stats**
"GT concordance" is hidden by default.
image

**Bargraph**
image

**Scatter plot** 
I remove the colouring per sample
image

vladsavelyev:
@pontushojer, it looks great and ready to merge!

I actually agree on your point that the scatter plot looks a bit less helpful in a static version without some way to tell samples apart. Though colors probably won't help much in this case, we'd need to show labels. But that's for the core plotting code to implement, the modules shouldn't worry about this, in my opinion. Something to take a look. I created an issue for that: https://github.com/ewels/MultiQC/issues/2111

Issue #1709, opened 2022-06-16
bclconvert module - Undetermined barcodes not shown


matthdsm:
### Description of bug

Bclconvert module is missing the section to visualise the top unknown barcodes

matthdsm:
Seems to be a duplicate of #1695

matthdsm:
Turns out the module doesn't handle the Undetermined barcodes/lane, when the run is demultiplexed by lane
https://github.com/ewels/MultiQC/blob/3627b64750d94f79f44ac222a570947d1134f63d/multiqc/modules/bclconvert/bclconvert.py#L59-L72

I'll look into fixing this

matthdsm:
Hi @andrei-seleznev,

It seems the `bclconvert` module can't handle the when the same run was demultiplexed per lane. e.g. we tend to spread the demultiplex in different jobs per lane over our cluster. This results in multiple directories with stats files for the same run, but different lanes. Per lane and per sample stats seem to work, but the undetermined barcodes are nowhere to be found.

Is that something you can easily fix?

Thanks
Matthias

PR #1752, opened 2022-09-01
Update bclconvert: show undetermined barcodes for multiple runs

nvnieuwk:
- [x] Removed the check for the undetermined barcodes plot so the barcodes are also generated when there are multiple runs


nvnieuwk:
ping @matthdsm

matthdsm:
Hi @ewels ,  @andrei-seleznev 

Could you comment as to _why_ mutli lane parsing was disabled? Everything seems to run just fine...
[multiqc_report (1).html.zip](https://github.com/ewels/MultiQC/files/9468701/multiqc_report.1.html.zip)

nvnieuwk:
Fixes #1709

alexiswl:
Hi @matthdsm,  

There's a small discussion on this here: https://github.com/ewels/MultiQC/pull/1381#discussion_r660300270

I think for your use-case (each lane is being run exclusively), there's a sound argument as to why the undetermined reads should exist and will see what we can do. 

For our use case, we were not splitting by lane, but by sample type, since each sample type had various different bcl convert settings, such as adapter reads, override cycles and minimum trimmed read length.  There would often be situations where multiple bclconvert runs would process different samples on the same lane. This would mean the undetermined reads were abnormally high on both bclconvert runs, one could use some methods to try determine the number of actually undetermined reads by cross-referencing large undetermined counts with samples on other bcl-convert runs.  

From reading the comment above, the issue wasn't so much about the calculation, but ensuring that undetermined indices weren't merged over different sequencing runs? @andrei-seleznev am I right in paraphrasing that much? 

Is there a way to determine how many sequencing runs have been scanned by the multiqc as opposed to multiple bclconvert workflows over the same run?

matthdsm:
Thanks for the comment! That does clarify the reasoning behind it.

I the case of multiple runs, I agree the undetermined runs should be hidden. They are, however, very useful in our case where we split each run by lane. This is also the default through basespace if I'm not mistaken.

ewels:
> This is also the default through basespace if I'm not mistaken.

I'm tempted to follow this if this is the behaviour elsewhere.

@nvnieuwk - if you could you add a check for a MultiQC config parameter to switch this behaviour on and off, that would be great. Then people who want the current behaviour can keep it by setting a config flag.

See https://github.com/ewels/MultiQC/commit/0f04894d2e8f649679026c3655006a821fdd5929 for an implementation example.

Hopefully everyone will be happy then and we can merge? @alexiswl ?

nvnieuwk:
The option has been added!

multiqc/modules/bclconvert/bclconvert.py
```diff
@@ -182,7 +185,7 @@ def __init__(self):
         )
 
         # Add section with undetermined barcodes
-        if self.num_demux_files == 1:
+        if create_undetermined_barplots:
```
ewels, line 188: I think we still want to do it by default if only a single demux file.. I can't see a reason _not_ to do it in this case.

ewels: Addressed in  e9c0e26

ewels (approved):

ewels:
Ok, thanks everyone! Hopefully this PR is a good compromise for all. Please shout if that's not the case.

Issue #1197, opened 2020-05-21
umi-tools module


chris-cheshire:
Removal of PCR duplicates is an essential step in iCLIP analysis. We use the `dedup` function of this package but the other functions could also be useful for including in multiQC.

Happy to fork and create myself

**Tool Details**
 - Name of tool: umi-tools
 - Tool homepage: https://github.com/CGATOxford/UMI-tools
 - Tool description: A tool for dealing with UMIs
 - Data suitable for MultiQC plot(s): stacks and % showing the numbers of removed PCR dups
 - Most interesting data for General Stats table:
   - `Mean number of unique UMIs per position: 1.17`
   - `Max. number of unique UMIs per position: 9`

**Tool output:**
 - Log filename pattern:
   - `stdout` - log files start with `# UMI-tools version: 1.0.1`
 - File contents
   - I am happy to fork and do this PR myself so can upload some appropriate test data then (I have pasted an example below just to give an idea)

**Example stdout for the dedup command**
```
# UMI-tools version: 1.0.1
# output generated by dedup --umi-separator=: -I exp3.Aligned.sortedByCoord.out.bam -S exp3.dedup.bam --output-stats=exp3
# job started at Thu May 21 08:07:07 2020 on 115a66d88402 -- 857194ec-ae8e-4bbf-b6c4-7c40706f6d79
# pid: 31, system: Linux 4.19.76-linuxkit #1 SMP Fri Apr 3 15:53:26 UTC 2020 x86_64
# assigned_tag                            : None
# cell_tag                                : None
# cell_tag_delim                          : None
# cell_tag_split                          : -
# chimeric_pairs                          : use
# chrom                                   : None
# compresslevel                           : 6
# detection_method                        : None
# gene_tag                                : None
# gene_transcript_map                     : None
# get_umi_method                          : read_id
# ignore_umi                              : False
# in_sam                                  : False
# log2stderr                              : False
# loglevel                                : 1
# mapping_quality                         : 0
# method                                  : directional
# no_sort_output                          : False
# out_sam                                 : False
# output_unmapped                         : False
# paired                                  : False
# per_cell                                : False
# per_contig                              : False
# per_gene                                : False
# random_seed                             : None
# read_length                             : False
# short_help                              : None
# skip_regex                              : ^(__|Unassigned)
# soft_clip_threshold                     : 4
# spliced                                 : False
# stats                                   : exp3
[...]

PR #1769, opened 2022-09-21
New module: UMI-tools

nolanmaier:
This module parses log files produced by [UMI-tools](https://umi-tools.readthedocs.io/en/latest/). Currently only the `dedup` command is supported. This module was requested in #1197.

- [X] There is example tool output for tools in the <https://github.com/ewels/MultiQC_TestData> repository or attached to this PR: ewels/MultiQC_TestData#244

Example output:
[multiqc_report.zip](https://github.com/ewels/MultiQC/files/9630284/multiqc_report.zip)


ewels (approved):
Great stuff, thanks very much! 🎉

Issue #1771, opened 2022-09-26
Nix flake builds but binary is broken


victorwkb:
### Description of bug

As mentioned in the title, the flake builds without error when running `$ nix build github:ewels/multiqc`.

However, when running the binary, `$ ./result/bin/multiqc` does not work which gives the following error log.

I believe that the line from error log (`ModuleNotFoundError: No module named 'pkg_resources'`) is pointing at the `setuptools` library for Python projects.

### MultiQC Error log

```console
$ nix build github:ewels/multiqc
trace: 
 automatically detected pname: 'multiqc'
trace: 
 automatically detected version: '1.14.dev0'
trace: 
 automatically detected requirements of multiqc:1.14.dev0 22.05pre-git:
matplotlib>=2.1.1
networkx>=2.5.1
numpy
click
coloredlogs
future>0.14.0
jinja2>=3.0.0
lzstring
markdown
pyyaml>=4
requests
rich>=10
rich-click
simplejson
spectra>=0.0.10

$ ./result/bin/multiqc 
Traceback (most recent call last):
  File "/nix/store/byjd7sg6avjcjkfsj62i0zy79zwfazvf-python3.8-multiqc-1.14.dev0/bin/.multiqc-wrapped", line 6, in <module>
    from multiqc.__main__ import run_multiqc
  File "/nix/store/byjd7sg6avjcjkfsj62i0zy79zwfazvf-python3.8-multiqc-1.14.dev0/lib/python3.8/site-packages/multiqc/__init__.py", line 16, in <module>
    from .multiqc import run
  File "/nix/store/byjd7sg6avjcjkfsj62i0zy79zwfazvf-python3.8-multiqc-1.14.dev0/lib/python3.8/site-packages/multiqc/multiqc.py", line 30, in <module>
    from .plots import table
  File "/nix/store/byjd7sg6avjcjkfsj62i0zy79zwfazvf-python3.8-multiqc-1.14.dev0/lib/python3.8/site-packages/multiqc/plots/table.py", line 9, in <module>
    from multiqc.plots import beeswarm, table_object
  File "/nix/store/byjd7sg6avjcjkfsj62i0zy79zwfazvf-python3.8-multiqc-1.14.dev0/lib/python3.8/site-packages/multiqc/plots/beeswarm.py", line 8, in <module>
    from multiqc.plots import table_object
[...]

PR #1772, opened 2022-09-27
added setuptools to flake

IllustratedMan-code:

I just added setuptools to the flake so that it actually works (not enough testing on my part). I didn't add to the changelog because I wasn't sure it was necessary. But happy to do so if need be.


IllustratedMan-code:
closes #1771

ewels (approved):
Thanks!

PR #1777, opened 2022-10-10
Remove empty bcftools stats variant depths plot

SusiJo:
The module bcftools stats would show empty variant depth plots if no data was there, but an extra line with a sample name. This can happen when there is no INFO/DP field in a vcf file. So this PR should solve the issue and avoid showing an empty plot in the MultiQC report. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### MultiQC updates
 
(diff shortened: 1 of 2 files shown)


ewels (commented):
Great! Just moving the changelog entry, then it looks good to merge.

Phil

ewels (approved):
Great stuff, thanks!

PR #1779, opened 2022-10-13
fix duplicate heatmap for kraken

jchorl:
The kraken module takes the top 5 taxons and shows duplication across samples.

The problem is when some samples don't include those taxons. The heatmap "collapses left" because of how the dict is turned into a list:
Screenshot 2022-10-13 at 14-49-41 Screenshot 2022-10-12 at 16-20-36 MultiQC Report png (PNG Image 590 × 508 pixels)

This means the duplication amount shown on the second row is actually for the wrong taxon!

Really, we should fill in `None` (or 0?) when a taxon isn't present in a sample. After fixing:
Screenshot 2022-10-13 at 15-52-49 MultiQC Report

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Add CI testing for Python 3.10 and 3.11
 - Add new code formatting tool [isort](https://pycqa.github.io/isort/) to standardise the order and formatting of Python module imports
 - Remove Python 2-3 compatability `from __future__` imports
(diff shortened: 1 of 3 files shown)


ewels (approved):
Nice, improves the MultiQC_TestData too:

image

to this:

image

ewels:
In fact, I don't think that this heatmap needs to be square.. Tacked a minor config improvement on to your PR 👍🏻

PR #1798, opened 2022-11-11
Update Qualimap: add additional entries for qualimap when region stats present

jfy133:
The current Qualimap module does not pick up statistics relevant for when stats on  particular regions are generated via a bed file. This PR adds support for displaying these statistics in the general stats table (when the parrticular entries are present)

- [x] There is example tool output for tools in the <https://github.com/ewels/MultiQC_TestData> repository or attached to this PR: https://github.com/ewels/MultiQC_TestData/pull/246

### DIFF
--- a/multiqc/modules/qualimap/QM_BamQC.py
+++ b/multiqc/modules/qualimap/QM_BamQC.py
@@ -81,41 +81,49 @@ def parse_reports(self):
 def parse_genome_results(self, f):
     """Parse the contents of the Qualimap BamQC genome_results.txt file"""
     regexes = {
-        "bam_file": r"bam file = (.+)",
-        "total_reads": r"number of reads = ([\d,]+)",
-        "mapped_reads": r"number of mapped reads = ([\d,]+)",
-        "mapped_bases": r"number of mapped bases = ([\d,]+)",
-        "sequenced_bases": r"number of sequenced bases = ([\d,]+)",
-        "mean_insert_size": r"mean insert size = ([\d,\.]+)",
-        "median_insert_size": r"median insert size = ([\d,\.]+)",
-        "mean_mapping_quality": r"mean mapping quality = ([\d,\.]+)",
-        "general_error_rate": r"general error rate = ([\d,\.]+)",
-        "mean_coverage": r"mean coverageData = ([\d,\.]+)",
-        "regions_size": r"regions size = ([\d,\.]+)",
-    }
-    regexes_secondary = {
-        "regions_mapped_reads": r"number of mapped reads = ([\d,]+)",
+        "Input": {
+            "bam_file": r"bam file = (.+)",
+        },
+        "Globals": {
+            "total_reads": r"number of reads = ([\d,]+)",
+            "mapped_reads": r"number of mapped reads = ([\d,]+)",
+            "mapped_bases": r"number of mapped bases = ([\d,]+)",
+            "sequenced_bases": r"number of sequenced bases = ([\d,]+)",
+        },
+        "Insert size": {
+            "mean_insert_size": r"mean insert size = ([\d,\.]+)",
+            "median_insert_size": r"median insert size = ([\d,\.]+)",
+        },
+        "Mapping quality": {
+            "mean_mapping_quality": r"mean mapping quality = ([\d,\.]+)",
+        },
+        "Mismatches and indels": {
+            "general_error_rate": r"general error rate = ([\d,\.]+)",
+        },
+        "Coverage": {
+            "mean_coverage": r"mean coverageData = ([\d,\.]+)",
+        },
+        "Globals inside": {
+            "regions_size": r"regions size = ([\d,\.]+)",
+            "regions_mapped_reads": r"number of mapped reads = ([\d,]+)",  # WARNING: Same as in Globals
+        },
     }
     d = dict()
-    for k, r in regexes.items():
-        r_search = re.search(r, f["f"], re.MULTILINE)
-        if r_search:
-            if "\d" in r:
-                try:
-                    d[k] = float(r_search.group(1).replace(",", ""))
-                except ValueError:
-                    d[k] = r_search.group(1)
-            else:
-                d[k] = r_search.group(1)
-
-    for k, r in regexes_secondary.items():
-        r_search = re.findall(r, f["f"], re.MULTILINE)
-        if len(r_search) > 1:
-            if "\d" in r:
-                try:
-                    d[k] = float(r_search[1].replace(",", ""))
-                except ValueError:
-                    d[k] = r_search[1]
+    section = None
+    for line in f["f"].splitlines():
+        if line.startswith(">>>>>>>"):
+            section = line[8:]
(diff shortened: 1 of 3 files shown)


multiqc/modules/qualimap/QM_BamQC.py
```diff
@@ -103,6 +107,17 @@ def parse_genome_results(self, f):
                     d[k] = r_search.group(1)
             else:
                 d[k] = r_search.group(1)
+
+    for k, r in regexes_secondary.items():
+        r_search = re.findall(r, f["f"], re.MULTILINE)
+        print(r_search)
```
ewels, line 113:
```suggestion
```

multiqc/modules/qualimap/QM_BamQC.py
```diff
@@ -103,6 +107,17 @@ def parse_genome_results(self, f):
...
+        if len(r_search) > 1:
```
ewels, line 114: Why > 1 and not just true?

multiqc/modules/qualimap/QM_BamQC.py
```diff
@@ -103,6 +107,17 @@ def parse_genome_results(self, f):
...
+            if "\d" in r:
```
ewels, line 115: Why do you need this?

multiqc/modules/qualimap/QM_BamQC.py
```diff
@@ -91,6 +91,10 @@ def parse_genome_results(self, f):
         "mean_mapping_quality": r"mean mapping quality = ([\d,\.]+)",
         "general_error_rate": r"general error rate = ([\d,\.]+)",
         "mean_coverage": r"mean coverageData = ([\d,\.]+)",
+        "regions_size": r"regions size = ([\d,\.]+)",
+    }
+    regexes_secondary = {
+        "regions_mapped_reads": r"number of mapped reads = ([\d,]+)",
     }
```
ewels, line 98: Why does this need to be in its own dict? Can it not just be added to the previous block?

multiqc/modules/qualimap/QM_BamQC.py
```diff
@@ -592,6 +609,20 @@ def general_stats_headers(self):
         "format": "{0:.2f}",
         "hidden": True,
     }
+    self.general_stats_headers["regions_size"] = {
+        "title": "{} Region size".format(config.read_count_prefix),
+        "description": "Size of target region",
+        "suffix": " bp",
+        "scale": "RdYlGn",
```
ewels, line 616:
```suggestion
        "scale": "PuBuGn",
```

ewels (commented):
Looks good, just not sure why you need the second separate set of regexes.

multiqc/modules/qualimap/QM_BamQC.py
jfy133: If I remember correctly, it's something like the way the loop was constructed for searching each regex, it didn't allow the pick up of the 'second' line with the string (there is nothing to distinguish from the earlier line, other than a 'section' header a few lines above the second hit of this line.

jfy133: But I'm going to have to refer to @TCLamnidis for the python details (he fixed most of this for me 😅)

multiqc/modules/qualimap/QM_BamQC.py
jfy133: I think it was to make sure to only parse this if there is actually a second hit in the file (otherwise you would end up picking up the first line and including that again...) but I again, I'll leave this to Thiseas to confirm...

TCLamnidis: exactly what @jfy133 said. The output contains two values named the same in two separate sections, with the on-target section (`Globals inside`) only appearing when a bedfile is given. So the first element of the search is already picked up by the original regex, but not the second.

multiqc/modules/qualimap/QM_BamQC.py
TCLamnidis: The original regex finds only the first match, which is always present. This regex collects all matches by using `.findall` instead of `. search`.

multiqc/modules/qualimap/QM_BamQC.py
TCLamnidis: I suppose it would be possible to change the entire regex matching to `.findall` and set for all other values that we want the first element, then do a check for `len > 1` for that regex to separate the two values.
This approach seemed "safer" in that we are not changing the matching for values that work as intended. "Don't fix what ain't broken."

multiqc/modules/qualimap/QM_BamQC.py
ewels: oof 🤦🏻‍♂️  http://tallphil.co.uk/writing-good-log-files/ :shakefist:

multiqc/modules/qualimap/QM_BamQC.py
ewels: But why the `if "\d" in r:` ?

multiqc/modules/qualimap/QM_BamQC.py
jfy133: `oof` was precisely my reaction when I saw that :laughing: . Is this otherwise a reasonable solution (and can resolve the comment?)

ewels: Please leave it for now, I might have a quick play and see if I can refactor a bit. To be honest, if it's just finding the first hit and there can be repeats of things, it makes me a bit nervous that the original code could accidentally report the wrong thing in certain cases. So I'm inclined to refactor it to know which section it's in.

multiqc/modules/qualimap/QM_BamQC.py
```diff
@@ -103,6 +107,17 @@ def parse_genome_results(self, f):
...
+    for k, r in regexes_secondary.items():
+        r_search = re.findall(r, f["f"], re.MULTILINE)
+        print(r_search)
+        if len(r_search) > 1:
+            if "\d" in r:
+                try:
+                    d[k] = float(r_search[1].replace(",", ""))
+                except ValueError:
+                    d[k] = r_search[1]
+
```
TCLamnidis, line 119:
```suggestion
            try:
                d[k] = float(r_search[1].replace(",", ""))
            except ValueError:
                d[k] = r_search[1]

```

TCLamnidis: Removing that unneeded `if` and updating indentation.

multiqc/modules/qualimap/QM_BamQC.py
ewels: Ok, refactored to parse per-section instead. Hopefully the code is a bit more robust to repeat strings now 👍🏻

ewels:
Thanks both! I refactored the code a bit to make it a little cleaner / more robust. I also stumbled across a bug in the _Snippy_ module so snuck a fix for that into this PR as well.

PR #1799, opened 2022-11-13
Quality score distributions

gitMakeCoffee:
FastQC computes several distribution statistics for PHRED scores across reads, such as quartiles and percentiles, and represents them as boxplots. The current plotting in MultiQC only shows the mean quality, giving only a partial view of the actual quality metrics.

This pull request (feature branch) contains a new qualityplot.py plotting module, which represents quality scores as miniature histograms using Matplotlib (static output). The statistics represented are the median, upper and lower quartiles, and the 10th and 90th percentiles, as computed by FastQC. The result is shown in a new section, just above the mean quality line graph. The plot can be a bit crowded if too many samples are combined but can still allow for the detection of outliers.

NB: the plotting code lacks the "hidediv" and the "pidx" variables. The code may therefore require some improvement to better integrate with MultiQC's architecture.

Here is a preview for 4 samples. 
Thank you for developing MultiQC =)

download

### DIFF
--- a/multiqc/modules/fastqc/fastqc.py
+++ b/multiqc/modules/fastqc/fastqc.py
@@ -119,6 +119,9 @@ class MultiqcModule(BaseMultiqcModule):
 
         # Now add each section in order
         self.read_count_plot()
+
+        self.per_base_quality_plot()
+
         self.sequence_quality_plot()
         self.per_seq_quality_plot()
         self.sequence_content_plot()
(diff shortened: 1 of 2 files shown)


ewels:
Hi @gitMakeCoffee,

Many thanks for this! I actually had something very similar in one of the earliest versions of MultiQC, but made a deliberate decision to remove it. Although the current plot misses some data, it allows comparison of many samples. It's very common for people to run MultiQC with tens, hundreds and even thousands of samples. Having box plots, or showing the median data in any way really, does not scale to this and the plot rapidly becomes unreadable.

The general idea of MultiQC is not to faithfully replicate all available data from source, but instead to give a general overview where it's possible to spot outlier samples so that they can be investigated in more detail in the original source reports.

As such, I'm afraid I'm not going to accept your pull request. However, remember that MultiQC has an extensive plugin system, so if this type of visualisation is useful for you then you should be able to generate it fairly easily without it being in the core codebase.

Thanks again,

Phil

PR #1804, opened 2022-11-23
update bustools

Maarten-vd-Sande:

So I realized I simply forgot to add the median reads metric :facepalm:...

It's always reported in the output: https://github.com/ewels/MultiQC_TestData/blob/master/data/modules/bustools/0.40.0/inspect.json#L5 and I've tested the PR
image

### DIFF
--- a/multiqc/modules/bustools/bustools.py
+++ b/multiqc/modules/bustools/bustools.py
@@ -81,6 +81,12 @@ class MultiqcModule(BaseMultiqcModule):
             "format": "{:,.0f}",
             "shared_key": "barcodes",
         }
+        self.headers["medianReadsPerBarcode"] = {
+            "title": "Median reads per barcode",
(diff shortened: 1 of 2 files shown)


multiqc/modules/bustools/bustools.py
```diff
@@ -81,6 +81,12 @@ def prepare_data(self):
             "format": "{:,.0f}",
             "shared_key": "barcodes",
         }
+        self.headers["medianReadsPerBarcode"] = {
+            "title": "Median reads per barcode",
+            "scale": "RdYlGn",
+            "min": 0,
+            "format": "{:,.2f}",
```
ewels, line 88: I think this should be the default so you don't really need it, but no big deal..

ewels (approved):
Great stuff, thanks!

PR #1810, opened 2022-12-07
mosdepth missing values imputed

adamrtalbot:
Changes:
 - Mosdepth output has missing values in `*.{region,global}.dist.txt`
 - This change to module fills any missing values with the next value
 - e.g., if there is 100% at 100X and 80% at 80X, the value at 90X will be recorded as 80X
 - Previously, this would have been recorded at 0%.
 - This may underestimate coverage slightly but it's not clear from MosDepth docs how it should be handled.
 - See https://github.com/brentp/mosdepth/issues/190

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,8 @@
   - Make heatmap full width
 - **malt**
   - Fixed division by 0 in malt module ([#1683](https://github.com/ewels/MultiQC/issues/1683))
+- **Mosdepth**
+  - Don't pad the General Stats table with zeros for missing data ([#1810](https://github.com/ewels/MultiQC/pull/1810))
 - **Picard**
   - HsMetrics: Allow custom columns in General Stats too, with `HsMetrics_genstats_table_cols` and `HsMetrics_genstats_table_cols_hidden`
 - **RSeQC**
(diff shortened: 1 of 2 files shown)


ewels:
Thanks! Speed thought: generally I try to avoid computing new data within MultiQC (plenty of exceptions abound). I agree that reporting `0.0%` is just plain wrong, but what do you think about using `None` instead? (An empty table cell).

Won't be as pretty, but would be more true to the original data. This data is essentially the same as the plot lower in the report though, right? So prettiness shouldn't be super important? Or have I got that wrong?

mosdepth-cumcoverage-dist-id

ewels:
aha! Yes, if we remove the `0`s so that they're not there (or are `None`, might work) then the plot will not have those points and the line will be smooth.

ewels:
> Unfortunately using None instead of zero creates a coverage of `None%`.

Hmm, how annoying. Can we just remove the keys if `None` then please. Should work fine if they keys are just absent.

ewels (approved):
Great! Just moved the log statement out of the loop as I was worried that there might be a lot per sample and it could make the debug log file huge.

Thanks for working on this and accepting feedback 👍🏻

ewels:
Just before I merge - does this definitely fix the lot as well? The test data isn't affected like your screenshot above.. From a skim read of the code it now looks like it's only affecting values in the general stats table. Just want to check that the plot is good as well.

ewels:
Sorry, had a typo - I was talking about the plot. If you have some example data kicking around I can take a quick look for a fix.. I'd prefer to do both in one PR if possible, just so that we don't forget.

ewels:
Ok, let's merge - I'm hoping to get a release out in 2022 and it would be good to include this 😅

ewels:
I dropped the log message, hope that's ok. Figured that it's fairly self explanatory behaviour..

ewels:
Very sensible! 👍🏻

PR #1814, opened 2022-12-08
Improve speed of parsing output file headers

adamrtalbot:
Parsing the output data file headers was using a nested loop. This was quite slow, so I replaced it with a list comprehension. This works much, much faster but is harder to read. I've added some comments to explain what it is doing.

### DIFF
--- a/multiqc/utils/util_functions.py
+++ b/multiqc/utils/util_functions.py
@@ -73,12 +73,19 @@ def write_data_file(data, fn, sort_cols=False, data_format=None):
             try:
                 # Convert keys to strings
                 data = {str(k): v for k, v in data.items()}
-                # Get all headers
-                h = ["Sample"]
-                for sn in sorted(data.keys()):
-                    for k in data[sn].keys():
-                        if type(data[sn][k]) is not dict and k not in h:
-                            h.append(str(k))
+                # Get all headers from the data, except if data is a dictionary (i.e. has >1 dimensions)
+                # Use list -> set -> list to get only unique values
+                h = list(
+                    set(
+                        [
+                            str(data_header)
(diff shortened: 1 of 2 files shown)


multiqc/utils/util_functions.py
```diff
@@ -73,12 +73,19 @@ def default(self, obj):
...
-                # Get all headers
-                h = ["Sample"]
-                for sn in sorted(data.keys()):
-                    for k in data[sn].keys():
-                        if type(data[sn][k]) is not dict and k not in h:
-                            h.append(str(k))
+                # Get all headers from the data, except if data is a dictionary (i.e. has >1 dimensions)
+                # Use list -> set -> list to get only unique values
+                h = list(
+                    set(
```
ewels, line 79: Are `set`s ordered? I don't think so, right? In which case we lose the original order of headers with this..

adamrtalbot: No, but they get sorted again a few lines below on lines 89-90 so I assumed it was OK. Looking at it, this will add `Sample` to the rightmost column, so that's wrong. I'll fix it up a bit more.

adamrtalbot: Replaced with `list -> dict -> list` which does the same thing but preserves order! Neat!

ewels: > No, but they get sorted again a few lines below

That's only if the `sort_cols` config is set, which is usually isn't. Typically column order is preserved from the module.

Using dict keys is the way I would usually do this, sorry should have suggested that 👍🏻 Then they're unique and ordered (at least in Python 3, which MultiQC is now).

ewels (approved):
LGTM, thanks!

PR #1817, opened 2022-12-13
update busco colors

jchorl:
I've changed the colours to more closely resemble a pass/warning/fail scheme to make it clearer for users of the module that there should only be single copies of the single copy orthologs. I have also changed the order that items are plotted to reflect the severity of the warning/failure and match the colour gradient. In general, we feel that duplicated BUSCOs is a more severe error than fragmented BUSCOs, as "fragmentation" is a common finding on nanopore sequencing data and rather reflects frameshifts/erroneous stop codons, rather than duplicated BUSCOs which is usually from erroneous binning, barcode crosstalk or contamination. Feel free to suggest other colours.


ewels (approved):
LGTM, thanks!

PR #1818, opened 2022-12-14
fix: frontmatter yaml formatting issue

olehpratsko:
- [x] When parsing markdown, frontmatter for the changed files was not read correctly. Issue returned:
```
YAMLException: can not read a block mapping entry; a multiline key may not be 
  an implicit key at line 6, column 1:
```
It also can be visible in GitHub files preview. The frontmatter table is not displayed correctly.


ewels (approved):
Thanks!

PR #1689, opened 2022-05-20
Cellranger module

edg1983:
Hello MultiQC team,
I've created a new module to integrate cellranger QC metrics from VDJ or count analysis into MultiQC reports

- [x] `CHANGELOG.md` has been updated
- [X] There is example tool output for tools in the <https://github.com/ewels/MultiQC_TestData> repository or attached to this PR. I've attached 2 HTML example reports in the ZIP archive.
- [X] Code is tested and works locally (including with `--lint` flag)
    - Actually, the report is generated correclty, but using `--lint` I get these warnings
    ```
    LINT: >cellranger/vdj.py< HTML ID was a duplicate (reads-1) ## {'namespace': 'Cellranger VDJ'}
    LINT: >cellranger/vdj.py< HTML ID was a duplicate (valid_bc-1) ## {'namespace': 'Cellranger VDJ'}
    LINT: >cellranger/vdj.py< HTML ID was a duplicate (Q30_bc-1) ## {'namespace': 'Cellranger VDJ'}
    LINT: >cellranger/vdj.py< HTML ID was a duplicate (Q30_UMI-1) ## {'namespace': 'Cellranger VDJ'}
    LINT: >cellranger/vdj.py< HTML ID was a duplicate (estimated_cells-1) ## {'namespace': 'Cellranger VDJ'}
    LINT: >cellranger/vdj.py< HTML ID was a duplicate (avg_reads_cell-1) ## {'namespace': 'Cellranger VDJ'}
    LINT: >cellranger/vdj.py< HTML ID was a duplicate (reads_in_cells-1) ## {'namespace': 'Cellranger VDJ'}
    ```
- [X] `docs/README.md` is updated with link to below
- [X] `docs/modulename.md` is created
- [X] Everything that can be represented with a plot instead of a table is a plot
- [X] Report sections have a description and help text (with `self.add_section`)
- [X] There aren't any huge tables with > 6 columns (explain reasoning if so)
    - I've tried to keep visible columns to the minimum and generally tables have now 6-9 cols. The ones that are visible are those considered of main interest after discussion with colleagues routinely working with this kind of data.
- [ ] Each table column has a different colour scale to its neighbour, which relates to the data (eg. if high numbers are bad, they're red)
- [X] Module does not do any significant computational work

[example_data.zip](https://github.com/ewels/MultiQC/files/8737021/example_data.zip)

PR #1821, opened 2022-12-15
Cellranger module

ewels:
Continuation of PR #1689 as I don't have push access to HTGenomeAnalysisUnit/MultiQC.git so can't work on the PR to get it merged.


ewels:
@edg1983 I've just started looking at the outputs from this module, and I'm half wondering if it should really be two modules: `Cell Ranger` and `Cell Ranger V(D)J`. They seem quite distinct in the two sets of outputs, and also in [the](https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/latest/what-is-cell-ranger) [docs](https://support.10xgenomics.com/single-cell-vdj/software/pipelines/latest/using/vdj).

What do you think?

edg1983:
@ewels cellranger is a tool with multiple analysis modes. The most popular is count for sure, then VDJ generates a similar output in terms of the HTML report, and we use it often here so I've integrated this one as well. However, there are also other modalities like multi but the output of them is more complex to summarise. Given that the scope of the count and VDJ analysis is quite different, I agree that it makes sense to split them in separate modules. I don't have much time right now to work on this, but I can come back to it at the beginning of the next year.

ewels:
Thinking about it with a fresh set of eyes this morning I think it's probably best structured as you have done. Thought it looked a bit odd with the different outputs in one module, but I guess it'll be rare for people to have both in one MultiQC report.. (?) so probably doesn't matter.

Having it as one module with the "Cell Ranger" name is probably clearest for people and easiest to maintain 👍🏻

edg1983:
I can see a point in both approaches. 
Keeping both VDJ and count under a single cellranger module seems more in line with what you have so far for other modules, which are organized mainly by tool. One may perform a multi-modal experiment and then use multiple instances of cellranger. Still, I agree that it's not a typical scenario, and it is more likely to be analyzed using cellranger multi, which is not managed by this module anyway.

ewels:
@edg1983 one of the requirements for MultiQC modules is that table columns have customised colour scales. This is to infer whether a value is good or not (eg. red - yellow - green) and also makes it much easier to spot small numeric differences quickly. Finally, it makes it easier to quickly scan reports and recognise columns.

I've refactored your code to take a dict of colour scales (see the [docs](https://multiqc.info/docs/#table-colour-scales-1) for available names) and made a start. It's getting late though so I need to stop. Any help would be much appreciated!

ewels:
Ok, I'm happy with this now.

Still have a niggling feeling that we could be doing more with plots, but can't find anything better for now. I tried a heatmap for showing the percentage stats on the VDJ report but it was pretty awful, with one sample at least, so I didn't push that.

Many thanks @edg1983!

PR #1822, opened 2022-12-16
docs: update docs structure

olehpratsko:
- [x] Restructure docs to match gatsby docs solution requirements


ewels (changes requested):
Let's leave it as it is, if it's not too much trouble.. I think it'll be easier for people to find stuff I think. And I don't expect many more images that we have currently (they're almost all for only one of the files).

PR #1826, opened 2022-12-19
fastp.py: read and print more information

dwishsan:
In the log, read the 'read1/2_after/before_filtering' and the 'summaty-before_filtering' fields. Compute mean read length (if paired-end data). New columns in General Statistics table: read1/2 after/before filtering total reads, read1/2 after/before filtering read length and before/after filtering mean read length

### DIFF
--- a/multiqc/modules/fastp/fastp.py
+++ b/multiqc/modules/fastp/fastp.py
@@ -175,6 +175,80 @@ class MultiqcModule(BaseMultiqcModule):
             self.fastp_qual_plotdata[k][s_name] = {}
             self.fastp_gc_content_data[k][s_name] = {}
             self.fastp_n_content_data[k][s_name] = {}
+        
+        # Parse read1 - get number of read (paired for paired-end data) before
+        try:
+            for k in parsed_json["read1_before_filtering"]:
+                if type(parsed_json["read1_before_filtering"][k]) in [int, float]:
+                    self.fastp_data[s_name]["read1_before_filtering_{}".format(k)] = float(
+                        parsed_json["read1_before_filtering"][k]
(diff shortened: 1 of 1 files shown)


ewels:
I don't really understand the motivation for this pull-request. Please could you explain _why_ these metrics are useful?

On a technical level:

* Please run on the [test data](https://github.com/ewels/MultiQC_TestData/tree/master/data/modules/fastp) and fix errors
* The General stats should have minimal numbers of columns. Fastp already has 5 before this PR, which is plenty. Please report these new values in a different way
    * A table in the Fastp section, or better still a new plot.

dwishsan:
Hello,

It is just metrics that my biologist co-workers wanted to have for quality
control.
Basically, these metrics are shown with the FastQC module, but the downside
is that FastQC print one line per sample mate (for paired-end data) in the
summary table. This makes the table less readable (3 times more rows and
statistics corresponding to the same sample on different rows, eg: sampleA,
sampleA_R1 and sampleA_R2). I can manage to print only one line for the
FastQC module but, as far as I understand, this will only select one of the
two mate and print its statistics.
My collaborators wanted to have 1 line per sample with the info of the
FastQC module, and they also wanted to have access to mate information from
the FastQC module if needed.
I figure that the best way to do it was to completely hide the FastQC
module in the summary table, and print the information from the FastP
module. Basically, I turned the row corresponding to mates to FastP columns.

If this is too much column, I can simply hide the ones I added. Or reject
the commit fs you do not think it is useful.

Best,
Audric


Le jeu. 22 déc. 2022 à 01:45, Phil Ewels ***@***.***> a
écrit :

PR #1828, opened 2022-12-19
STAR: add all alignment summary metrics into a new separate table

dwishsan:
Add the percent and raw counts of mapped reads (unique+multimapped)


vladsavelyev:
Fixes https://github.com/MultiQC/MultiQC/issues/982

PR #1829, opened 2022-12-20
Fix merge conflicts for #1232

ewels:
Continuing PR #1232 as I don't have write access to push fixes there.

> Add new module to support DRAGEN-FastQC


ewels:
Ok, I think that I have got the code to a point where tests are passing and the module could be merged. I had to make a fair few changes, please see the latest list of commits above 👆🏻 

The main thing that I'm struggling with now is the box plots. In the very first iteration of MultiQC I included the box plot images from FastQC in the reports. However, I quickly found that doing so does not scale - when you have more than a couple of samples you're back to not being able to summarise all samples (which is the whole point of MultiQC). Also, the image file sizes bloat the report.

The same is true of this module. Running with one or two samples, the box plots are ok. But as soon as you start to run with bigger projects, it becomes a big problem. As a quick example, I made 100 copies of the test data and generated a report. It took over 4 minutes (MacBook M1 Pro, this kind of report should take < 10 seconds). The report file size is over 20MB for a single HTML file. I was able to load it, but the plot shows up with a huge grid of buttons which are fairly unusuable.

This is not something that can even be fixed by adding JavaScript box plots. Although that would solve the render time and file size issues, it would still be impossible to practically compare large sample numbers in the report.

I don't think that I can merge this module like this, so my suggestion is that we only generate these box plots when there are one or two samples. When there are more than that, we use a line graph - the same as the FastQC module. We miss out on the ranges, but that's ok. MultiQC was designed to only summarise outputs from other tools. Not replicate in full.

I'll see if I can put something together to do this now.

ewels:
🤦🏻‍♂️  moment - you already have the line graph 😅  Ok so it's just a case of skipping the box plot when there are lots of samples. Easy, will do that now.

ewels:
let's do this

PR #1831, opened 2022-12-27
fix (dragen_fastqc): clear accumulators between samples and mates

### LINKED ISSUE #1830: dragen_fastqc incorrectly computes N content
The dragen_fastqc module does not clear aggregated totals and N content across samples, causing the N content per-sample to be order-dependent and to approach a cross-sample homogenized average.  Problematic lines are [here](https://github.com/ewels/MultiQC/blob/8af61c1bde6ae0a38697e2f372866f8bc0e71a4a/multiqc/modules/dragen_fastqc/content_metrics.py#L43-L44):

https://github.com/ewels/MultiQC/blob/8af61c1bde6ae0a38697e2f372866f8bc0e71a4a/multiqc/modules/dragen_fastqc/content_metrics.py#L43-L44

This issue has been live in Illumina's for-profit Basespace multiqc module for at least a year. Unfortunately, I don't have any test data to show for this, as I originally found this problem in a prior job and the example data I was using didn't come with me; but the issue is very apparent once you know where to look.

cpalmer718:
The existing implementation of the N content plot for dragen_fastqc incorrectly maintains counts between samples and mates, causing the computed per-site N rate to be order-dependent (typically on alphabetical ordering of input sample IDs), and to approach a cross-sample/read/lane/etc. average. This isn't immediately apparent from output plots when the QCed libraries are performing well, but becomes easily visible when one of the lanes, for example, has a large spike in N content.

Closes #1830 

  - not including a change here, as dragen_fastqc itself is not released

### DIFF
--- a/multiqc/modules/dragen_fastqc/content_metrics.py
+++ b/multiqc/modules/dragen_fastqc/content_metrics.py
@@ -40,11 +40,11 @@ class DragenContentMetrics(BaseMultiqcModule):
     def n_content_plot(self):
         """Create the HTML for the per base N content plot"""
         data = dict()
-        totals = defaultdict(int)
-        non_n = defaultdict(int)
         GROUP = "POSITIONAL BASE CONTENT"
         for s_name in sorted(self.dragen_fastqc_data):
             for mate in sorted(self.dragen_fastqc_data[s_name]):
+                totals = defaultdict(int)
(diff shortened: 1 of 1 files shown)


ewels (approved):
Nice! Looks good to me, thanks 👍🏻

PR #1148, opened 2020-04-02
Pairtools module

sergpolly:
Hey @ewels !
this is a "Work in progress" for the [pairtools](https://github.com/mirnylab/pairtools) module:
 - `pairtools` is a set of command line framework for common operations on ["pairs"](https://github.com/4dn-dcic/pairix/blob/master/pairs_format_specification.md) generated by Hi-C and other "Chromatin Conformation Capture" ("C") techniques.
 - "Pairs" can be viewed as an [intermediate](https://pairtools.readthedocs.io/en/latest/) result of processing raw Hi-C (and other "C") data: raw paired-end "C"-data reads mapped to reference genome are parsed into "pairs", which can be further classified, sorted, flipped, deduplicated and filtered - these operations are the main domain of `pairtools`.
 - `Pairtools` is also responsible for collecting summary-statistics (as part of pipelines [[1]](https://data.4dnucleome.org/resources/data-analysis/hi_c-processing-pipeline) [[2]](https://github.com/mirnylab/distiller-nf), or [standalone](https://pairtools.readthedocs.io/en/latest/cli_tools.html#pairtools-stats)) and MultiQC is an amazing platform to aggregate, visualize and combine such pair-level stats with others (fastqc, adapter trimming, etc).

`Pairtools` is written by @golobor and influenced by many others from [4DN](http://dcic.4dnucleome.org/), [mirnylab](https://github.com/mirnylab/), etc

PS. I'm hoping that having this as a WIP PR would motivate/speed up the development of this module

Thank you so much!

## If this PR is for a new module
 - [x] There is example tool output for tools in the https://github.com/ewels/MultiQC_TestData repository
 - [x] Code is tested and works locally (including with `--lint` flag)
 - [x] `CHANGELOG.md` is updated
 - [x] `README.md` is updated
 - [x] `docs/README.md` is updated with link to below
 - [x] `docs/modulename.md` is created
 - [x] Everything that can be represented with a plot instead of a table is a plot
 - [x] Report sections have a description and help text (with `self.add_section`)
 - [x] There aren't any huge tables with > 6 columns (explain reasoning if so)
 - [x] Each table column has a different colour scale to its neighbour, which relates to the data (eg. if high numbers are bad, they're red)
 - [x] Module does not do any significant computational work

PR #1834, opened 2022-12-30
Pairtools module merge

ewels:
Continuation of #1148 as I don't have write access to push commits there.


ewels:
x-ref: https://github.com/ewels/MultiQC/pull/1148#issuecomment-1368105567

sergpolly:
Hi @ewels !
sorry for the delay ! I'm still alive... just slow
I've tried to do some cleanup, removed some unfinished code etc, but it looks like I'm unable to push to this branch - what should I do in this case ?


>  there's a lot of stuff happening with numpy (which is causing me headaches as the data structures are not easily exported into JSON)

I don't think I was able to address all of the numpy-concerns, but if you can point toward the most problematic structures - we would try to come with the solution on our side

ewels:
Hi @sergpolly!

You can continue to push to your previous branch. It's a bit convoluted but we should be able to go back and forth between the two 😅 Just check out this branch locally and merge it into the branch that you were doing before, then push. This branch contains the same git history so it should work fine.

Does that make sense? I can elaborate if needed.

Phil

PR #1836, opened 2023-01-01
Dockerfile - use python base

ewels:
Continuation of #1541

Rewrote the `Dockerfile` again, this time to use the `python-slim` base image instead of alpine.

Aim was to make the `Dockerfile` as succinct as possible, and the final image size as small as possible.


Dockerfile
```diff
@@ -1,19 +1,26 @@
...
-# Install MultiQC
-RUN python -m pip install .
+# - Install `ps` for Nextflow
+# - Install MultiQC through pip
+# - Delete unnecessary Python files
+# - Remove MultiQC source directory
+# - Add custom group and user
+RUN apt-get update && apt-get install -y procps && rm -rf /var/lib/apt/lists/* && \
+    pip install -v --no-cache-dir . && \
+    find /usr/local/lib/python3.11 \( -iname '*.c' -o -iname '*.pxd' -o -iname '*.pyd' -o -iname '__pycache__' \) | xargs rm -r && \
```
alexiswl, line 18: Is it possible to use a `find python3.11 (options) -exec rm -r {}\;` instead?

docs/installation.md
```diff
@@ -160,25 +160,25 @@ Note that support for using the base `multiqc` command was improved in MultiQC v
...
-You can also specify additional MultiQC parameters as normal:
+You can specify additional MultiQC parameters as normal:
 
 ```bash
-docker run -t -v `pwd`:`pwd` -w `pwd` ewels/multiqc . --title "My amazing report" -b "This was made with docker"
+docker run -t -v `pwd`:`pwd` -w `pwd` ewels/multiqc multiqc . --title "My amazing report" -b "This was made with docker"
 ```
 
+By default, docker will use the `:latest` tag. For MultiQC, this is set to be the most recent release.
+To use the most recent development code, use `ewels/multiqc::dev`.
```
alexiswl, line 179: Should this just be ewels/multiqc:dev

ewels: Yes, good spot

Dockerfile
ewels: I took this from a blog post which said this didn't work for the `__pycache__` dir for some reason. It was brief though and I didn't try it myself. I agree that it would be tidier so probably worth a go.

To be honest, this line only saves about 2MB so barely worth it 😅

docs/installation.md
ewels: Fixed in f15de216dcecb217879b66753adb87b85842b9ec

PR #1837, opened 2023-01-03
Add multi-arch docker builds

ewels:


ewels:
Live-testing on pushes 9ae6bb54a7e9dbc80ce044e0478cfe55c6153bd2 and bc4f724abf9e60deb0e335c57a6d76ef284a4845 too.

Looks like it's having difficulty with numpy :/

alexiswl:
Yep, I think pip needs to be upgraded before being built.

https://stackoverflow.com/a/70196702/6946787

Can't produce the numpy error on either arm/v7 (raspberry Pi v4) or amd64.  

Attempted fix in https://github.com/ewels/MultiQC/pull/1838

PR #1838, opened 2023-01-03
Upgrade pip before installing multiqc

alexiswl:
Patch to allow multi-platform build based on https://stackoverflow.com/a/70196702/6946787.  

Should resolve error found at https://github.com/ewels/MultiQC/actions/runs/3825879135/jobs/6509196703#step:7:7882 but can't test this locally myself on all platforms

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,7 @@ WORKDIR /usr/src/multiqc
 # - Remove MultiQC source directory
 # - Add custom group and user
 RUN apt-get update && apt-get install -y procps && rm -rf /var/lib/apt/lists/* && \
+    pip install --upgrade pip && \
     pip install -v --no-cache-dir . && \
     find /usr/local/lib/python3.11 \( -iname '*.c' -o -iname '*.pxd' -o -iname '*.pyd' -o -iname '__pycache__' \) | xargs rm -r && \
     rm -rf "/usr/src/multiqc/" && \


ewels (approved):
Awesome, let's give it a try.. Thanks!

alexiswl:
Damn, looks like we get the same error, I wonder if this is a 'buildx' issue rather than an specific arm numpy issue, let me try reproduce this on my amd server using docker buildx.

ewels:
Thanks for trying 👍🏻  Let me know how you get on. If in doubt I'll revert this, as I'd like to release ASAP and failing docker builds is a blocker..

alexiswl:
> Thanks for trying 👍🏻 Let me know how you get on. If in doubt I'll revert the multi-arch stuff, as I'd like to release ASAP and failing docker builds is a blocker..

Have managed to reproduce the error, I think it might be due to how a python package checks if it has a wheel for that arch.  

The `linux/arm64` compiles fine but the `linux/arm/v7` doesn't.  

I've added the build-essential package via apt and then removed it through apt as well after pip has installed. Takes a little longer to build though.  Should know in the next hour or so and will send through the image sizes.

ewels:
Amazing, thanks!

On that note, the architectures I picked in the build are a little arbitrary. Shout if you think we should be using a different list (I guess we could just do all of the ones available from the base Python image if we want..)

alexiswl:
No worries @ewels 

Had a couple of failures with zlib libraries not being found (I think for a pillow dependency), looks like installing zlib1g-dev might fix this (see https://stackoverflow.com/a/34631976/6946787) for more info.  

But I think we have two options here - 

1. Go down the compilation path for version specific arm
  * Takes a few hours to build, assume it improves compatibility for arm.
  * Not sure if a couple of hours of compilation is appropriate for a CI/CD pipeline.
2. Just do linux/amd64 and linux/arm64 as the platform types.
  * Builds in about five minutes.
  * Works fine on my raspberry pi!
  
I've pushed a version of the latter (with a few small changes to the dockerfile) to quay.io under the container `quay.io/alexiswl4/multiqc:dev` 

I think the main issue is trying to compile 32-bit binaries on a 64-bit system.  If you're happy having just the 64-bit images available I've put up a PR at https://github.com/ewels/MultiQC/pull/1840 for this.

ewels:
Nice, thanks for the detective work! I'm mostly thinking about AWS graviton instances to be honest, I guess that's the main place people will be interested in using this. I assume that's 64 bit, so that's fine by me..

PR #1840, opened 2023-01-06
Specify 64-bit images only for GH actions build

alexiswl:
Some image cleaning - 
* Added apt-clean to image after apt installation
* Use for-loop over xargs for deleting cache dirs.
* Moved groupadd / useradd onto separate lines

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -43,7 +43,8 @@ jobs:
         with:
           # All available with python:3.11-slim are:
           # platforms: linux/386,linux/amd64,linux/arm/v5,linux/arm/v7,linux/arm64/v8,linux/ppc64le,linux/s390x
-          platforms: linux/amd64,linux/arm/v7,linux/arm64/v8
+          # But 32-bit binaries likely require compilation from source so stick with linux/amd64 and linux/arm64 for now
+          platforms: linux/amd64,linux/arm64
           push: true
           tags: |
             ewels/multiqc:dev
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,9 +16,13 @@ WORKDIR /usr/src/multiqc
 RUN apt-get update && apt-get install -y procps && rm -rf /var/lib/apt/lists/* && \
     pip install --upgrade pip && \
     pip install -v --no-cache-dir . && \
-    find /usr/local/lib/python3.11 \( -iname '*.c' -o -iname '*.pxd' -o -iname '*.pyd' -o -iname '__pycache__' \) | xargs rm -r && \
+    apt-get clean -y && \
+    for cache_dir in $(find /usr/local/lib/python3.11 \( -iname '*.c' -o -iname '*.pxd' -o -iname '*.pyd' -o -iname '__pycache__' \)); do \
+      rm -rf "$cache_dir"; \
+    done; \
     rm -rf "/usr/src/multiqc/" && \
-    groupadd --gid 1000 multiqc && useradd -ms /bin/bash --create-home --gid multiqc --uid 1000 multiqc
+    groupadd --gid 1000 multiqc && \
+    useradd -ms /bin/bash --create-home --gid multiqc --uid 1000 multiqc
 
 # Set to be the new user
 USER multiqc


.github/workflows/docker.yml
```diff
@@ -43,7 +43,8 @@ jobs:
         with:
           # All available with python:3.11-slim are:
           # platforms: linux/386,linux/amd64,linux/arm/v5,linux/arm/v7,linux/arm64/v8,linux/ppc64le,linux/s390x
-          platforms: linux/amd64,linux/arm/v7,linux/arm64/v8
+          # But 32-bit binaries likely require compilation from source so stick with linux/amd64 and linux/arm64 for now
+          platforms: linux/amd64,linux/arm64
```
ewels, line 47: `arm64` and not `arm64/v8`?

Dockerfile
```diff
@@ -16,12 +16,16 @@ WORKDIR /usr/src/multiqc
 RUN apt-get update && apt-get install -y procps && rm -rf /var/lib/apt/lists/* && \
     pip install --upgrade pip && \
     pip install -v --no-cache-dir . && \
-    find /usr/local/lib/python3.11 \( -iname '*.c' -o -iname '*.pxd' -o -iname '*.pyd' -o -iname '__pycache__' \) | xargs rm -r && \
+    apt-get clean -y && \
```
ewels, line 19: Is there a reason to use this instead of `xargs`? Are they not doing the same thing?

ewels (approved):
Nice, thanks!

ewels:
Looks good! 🎉  https://github.com/ewels/MultiQC/actions/runs/3858788667/jobs/6577696099

ewels:
Looks good: https://github.com/ewels/MultiQC/pkgs/container/multiqc/62139002?tag=dev

https://hub.docker.com/layers/ewels/multiqc/dev/images/sha256-1d96ff55ddb150e93c0f92b811ebf8fbbe54c92706a8703f4a860596386e59a8?context=explore

.github/workflows/docker.yml
alexiswl: I wish this was clearer, but from what I can tell from a comment on this stackoverflow answer - https://stackoverflow.com/a/70819313/6946787 is that arm64 is the same as arm64/v8. arm/v6 and arm/v7 however are 32 bit images.

Dockerfile
alexiswl: Essentially yes, but the for loop was meant to be a bit more secure when it came to things like special characters in filenames, but I've realised I needed to quote it like `for cache_dir in "$(find ... )"; do`.

Dockerfile
```diff
@@ -16,12 +16,16 @@ WORKDIR /usr/src/multiqc
...
-    groupadd --gid 1000 multiqc && useradd -ms /bin/bash --create-home --gid multiqc --uid 1000 multiqc
+    groupadd --gid 1000 multiqc && \
+    useradd -ms /bin/bash --create-home --gid multiqc --uid 1000 multiqc
 
 # Set to be the new user
 USER multiqc
 
 # Set default workdir to user home
-WORKDIR /home/multiqc
+WORKDIR /home/multiqc
```
alexiswl, line 31: Ah looks like I've accidentally removed the last newline of the file as well

ewels:
haha, oops! We didn't update the GitHub Actions workflow for releases, only dev. However, it worked??

https://github.com/ewels/MultiQC/actions/runs/3869228401/jobs/6595193160

So we now have 3 architectures available for the v1.14 release..

- linux/amd64
- linux/arm/v7
- linux/arm64

https://hub.docker.com/r/ewels/multiqc/tags

Maybe we can add `linux/arm/v7` back to the `dev` version??

ewels:
Added in db5e3d79

Dockerfile
ewels: Added back in 41108ff0 😅

alexiswl:
Ah I don't think it did complete? 

https://github.com/ewels/MultiQC/actions/runs/3869392981/jobs/6595460306#step:7:7583

ewels:
Ah pants. Good spot! I wonder why it didn't crash the build job 🤔 Makes me a bit nervous.. I wonder if it makes sense to run a test command in the GitHub action after the build + push. Probably not, I guess it'll be pretty stable once we have it up and running.

Ok, but if the arm64 and amd builds did work and are functional then I don't think that this is a huge problem. We just revert to what you had put in place previously (for both dev and release this time 🙈 )

alexiswl:
If we really want the arm/v7, I've created another dockerfile that also installs the build-essentials, zlib1g-dev and libjpeg-dev (required for the pillow libraries), build times are quite long since they're installed from source.   

> I wonder why it didn't crash the build job

I think this might be an issue with pip, there are a few issues (like this one https://github.com/pypa/pip/issues/11463) highlighting that pip returns a zero exit code after failure. 

Maybe a line like `type multiqc` at the end would be a simple check to ensure pip has installed correctly?

ewels:
Ok, reverted to `linux/amd64,linux/arm64` in cc7f4e52 👍🏻 

> If we really want the arm/v7

Not super fussed to be honest, was more that I thought we got it for free so why not

> build times are quite long

How long is long? If measured in minutes and not hours could always add it to just master / release if you've done the work already. But to be honest, I suspect it'll be more hassle than it's worth so I'm happy to drop it if you're not bothered.

> Maybe a line like `type multiqc` at the end would be a simple check to ensure pip has installed correctly?

Yeah or `multiqc --help`. Where would that go though? I briefly wondered about about pulling the pushed image(s) at the end of the GitHub Actions workflow, but it seems hacky. And would have to do it for each arch somehow.

alexiswl:
The end of the main docker build chunk is 

```
#18 8933.6 Removing xz-utils (5.2.5-2.1~deb11u1) ...
#18 8934.0 Processing triggers for libc-bin (2.31-13+deb11u5) ...
#18 DONE 8964.4s
```

So 2.5 hours almost exactly!

I think this is feasible once GitHub actions properly supports ARM instances (only available at the moment via self-hosted runners).  

> Where would that go though?

Trying to look that up, maybe after the USER command one could try something like? 

```
RUN multiqc --help || (echo "Failed to run multiqc" && exit 1)
```

But to run full tests per image in the manifest without increasing the size of the container may be quite complex.

ewels:
Yup, let's leave that on the pile for "one day in the future" 👍🏻 

I like the idea of the Dockerfile `RUN` command. I was following similar lines but got distracted by reading blog posts about multi-stage builds being used to keep the container size slim and all sorts. I don't think that we need that level of testing, just checking that it installed properly is probably fine. And to be honest, I'm not convinced that even that is needed once we're happy that the build is working properly. Pip not exiting on failure is a bit of a random edge case..

alexiswl:
I'll put in a PR for the multiqc --help command

PR #1842, opened 2023-01-15
Run multiqc --help at the end of the dockerfile to check build is good

alexiswl:
Other additions:

* Added echo logs between each step of the run script, with 'Docker Build log' prefix to make easier to determine which step went wrong
* Added --quiet parameters to apt and pip upgrade
* Use xargs over for loop for removing cache with -printf to ensure each directory / file is quoted properly


alexiswl:
Justification of changes mentioned in discussion at: https://github.com/ewels/MultiQC/pull/1840

ewels (approved):
Thanks!

PR #1854, opened 2023-02-02
Fix dragen coverage metrics to show target bed and qc coverage metrics

### LINKED ISSUE #1844: MultiQC 1.14 no longer looks for .bed files for coverage metrics
### Description of bug

Somewhere between 1.13 and1.14, `search_patterns.yaml` was changed as follows: 
https://github.com/ewels/MultiQC/blob/v1.13/multiqc/utils/search_patterns.yaml#L262-L267
vs.
https://github.com/ewels/MultiQC/blob/v1.14/multiqc/utils/search_patterns.yaml#L284-L291

This change means that MultiQC no longer looks for `"*.target_bed_*.csv` when plotting coverage metrics. It seems like a fix would be similar to what was done in https://github.com/ewels/MultiQC/issues/1290, but I am not sure.

The following regexes also seem relevant, but I don't know how they are used:
https://github.com/ewels/MultiQC/blob/v1.14/multiqc/modules/dragen/coverage_metrics.py#L292
https://github.com/ewels/MultiQC/blob/v1.14/multiqc/modules/dragen/coverage_hist.py#L144
https://github.com/ewels/MultiQC/blob/v1.14/multiqc/modules/dragen/coverage_per_contig.py#L166

If someone can give advice, I am happy to write a PR for a fix.

_No response_

### MultiQC Error log

_No response_

alexiswl:
Refactored parse_wgs_coverage_metrics to generic parse_coverage_metrics.  

Fixes #1844

* Target bed, qc region and wgs qc then call parse_coverage_metrics each with different parameters.  
* Fixed search patterns to include target bed coverage metrics and qc region coverage metrics.

## Example outputs

QC metrics and target metrics are the same since I used the same bed file as input for both.

[wgs_with_qc_region.html.gz](https://github.com/ewels/MultiQC/files/10567515/wgs_with_qc_region.html.gz)  
[target_with_qc_region.html.gz](https://github.com/ewels/MultiQC/files/10567512/target_with_qc_region.html.gz)


ewels (approved):
Not my favourite MultiQC module, but tested locally and seemed to work and a read over of the code looks sensible.

Thanks!

Issue #1856, opened 2023-02-03
HUMID module is broken


Redmar-van-den-Berg:
### Description of bug

MultiQC (v1.14) always reports "No analysis results found" for the HUMID module.

Redmar-van-den-Berg:
Thanks!

PR #1857, opened 2023-02-03
Fix critical issue with HUMID module

Redmar-van-den-Berg:
See https://github.com/ewels/MultiQC/issues/1856 for details.


multiqc/modules/humid/humid.py
```diff
@@ -117,3 +117,5 @@ def process_stats(stats, s_name):
     except AssertionError:
         log.warning(f"HUMID stats looked wrong, skipping: {s_name}")
         return False
+    finally:
+        return True
```
ewels, line 121: Do you not need to return `stats`? Otherwise this function isn't doing anything..

```suggestion
    return stats
```

Maybe would be better still to include these functions inside the `MultiqcModule` (up one level of indentation) and handle everything there.

ewels: Oof, just noticed that you have two different functions both called `parse_stat_file` within one file - one inside the object scope and one outside. This needs refactoring... Sorry, must have missed this in the initial review.

ewels (changes requested):
Module needs a bit more refactoring to clean up functions outside object.

ewels:
I can try to have a go at the changes later if I get a moment (and if you don't beat me to it 👍🏻 )

Redmar-van-den-Berg:
I've refactored the code a bit more, hopefully it is clearer now.

ewels (approved):
Did my own minor refactoring, hope that's ok... Mostly a personal taste thing, but I find this a little less complex (no copying / passing dicts around) and easier to read / maintain.

Thanks for the fix! ✨

PR #1860, opened 2023-02-06
New Module: Xenome

MikeWLloyd:
This PR adds a new module for `xenome classify` parsing. Xenome is used for classifying DNA reads from xenograft sources to their respective sources.

## Example Log: 

```
Statistics
B       G       H       M       count   percent class
0       0       0       0       3006285 8.3833  "neither"
0       0       0       1       7393    0.0206161       "both"
0       0       1       0       3369033 9.39486 "definitely mouse"
0       0       1       1       1651249 4.60466 "probably mouse"
0       1       0       0       10152255        28.3105 "definitely human"
0       1       0       1       4392088 12.2477 "probably human"
0       1       1       0       51091   0.142472        "ambiguous"
0       1       1       1       57809   0.161206        "ambiguous"
1       0       0       0       29285   0.0816639       "both"
1       0       0       1       7016    0.0195648       "probably both"
1       0       1       0       135743  0.378532        "definitely mouse"
1       0       1       1       3250544 9.06444 "probably mouse"
1       1       0       0       540761  1.50796 "definitely human"
1       1       0       1       9007977 25.1196 "probably human"
1       1       1       0       15738   0.0438868       "ambiguous"
1       1       1       1       186135  0.519054        "ambiguous"

Summary
count   percent class
24093081        67.1858 human
8406569 23.4425 mouse
43694   0.121845        both
3006285 8.3833  neither
310773  0.866619        ambiguous
```


vladsavelyev (changes requested):
Thank you for contributing the module! Let a few comments.

Could you also please create a PR with some test data to https://github.com/ewels/MultiQC_TestData?

multiqc/modules/xenome/xenome.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def xenome_stats_plot(self):
+        """Make the HighCharts HTML to plot the classification rates"""
+
+        # Specify the order of the different possible categories
+        keys = OrderedDict()
+        keys["human_reads"] = {"color": "#000034", "name": "Human"}
+        keys["mouse_reads"] = {"color": "#328AE2", "name": "Mouse"}
+        keys["both_human_mouse_reads"] = {"color": "#7F9DA7", "name": "Both Species"}
+        keys["neither_human_mouse_reads"] = {"color": "#C1B8AA", "name": "Neither Species"}
```
vladsavelyev, line 102: I think the tool can be run on two arbitrary fastq files, which can belong to other species than Human and Mouse. Do you think it makes sense to generalise the columns, and call them Species 1 and Species 2? Or somehow parse the titles from the logs?

Also, see the [disambiguate](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/disambiguate/disambiguate.py) module. Not sure if it's a reference implementation - maybe it can improved as well - but it would be good to make the outputs for both modules look similar in terms of colors, scales, etc.

docs/modules/xenome.md
```diff
@@ -0,0 +1,10 @@
+---
+Name: Xenome
+URL: https://github.com/data61/gossamer/blob/master/docs/xenome.md
+Description: >
```
vladsavelyev, line 4: The CI is failing to validate this block.

That's because the keys need to be [lowercased](https://multiqc.info/docs/development/modules/#documentation).
```suggestion
name: Xenome
url: https://github.com/data61/gossamer/blob/master/docs/xenome.md
description: >
```

MikeWLloyd: Will correct this.

MikeWLloyd:
> Thank you for contributing the module! Let a few comments.
> 
> Could you also please create a PR with some test data to https://github.com/ewels/MultiQC_TestData?

So I am putting things in the correct place, the test data would go into `MultiQC_TestData/data/modules/xenome`?

vladsavelyev:
@MikeWLloyd - that's correct!

multiqc/modules/xenome/xenome.py
vladsavelyev: I think if it's available in the logs/outputs, it can be captured; otherwise, it can be by default "Species a" and "Species b" (like in disambiguate), and customised via a MultiQC config (similarly like there is a special [config section for qualimap](https://github.com/ewels/MultiQC/blob/8e471b39740ff8e1b31b499f9b7ba8ba42af6ed2/docs/modules/qualimap.md?plain=1#L41C1-L53))

MikeWLloyd: Regarding color matching among modules: when I make plots, I avoid using forms red/green in plots wherever possible. Diambiguate uses `#B1084C`, which I would prefer not to use. If you would like me to match though, I can implement that. The tool methods do differ in that disambiguate is based on mapped reads, whereas xenome uses a kmer approach to bin reads into either host or graft without mapping. 

Solving the flexibility issue is going to take me time. Being able to grep on a string was easy enough for me to code up. Matching strings based on physical position in the log file is a bit beyond me at the moment.  Hopefully I can find time next week

vladsavelyev: In terms of color matching, we can instead adjust Diambiguate, of the red color doesn't make sense here.

multiqc/modules/xenome/xenome.py
```diff
@@ -0,0 +1,123 @@
...
+                match = re.search(r, l)
+                if match:
+                    parsed_data[k] = float(match.group(1))
+
+        if len(parsed_data) > 0:
+            if s_name in self.xenome_data:
+                log.debug("Duplicate sample name found! Overwriting: {}".format(s_name))
+            self.add_data_source(f, s_name)
+            self.xenome_data[s_name] = parsed_data
+
```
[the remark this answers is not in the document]
vladsavelyev: Thank you @pontushojer :)

MikeWLloyd:
> Hey @MikeWLloyd - did you have a chance to share some test example? You can just attach it here if it's more convenient :)
> 
> Though I see you pasted a code block to the comment, there is a chance that potentially GitHub could strip/replace some whitespaces in the code block or when pasting it around. It happened before and caused elusive bugs, so we require the logs to be attached as original files.

I spent some time trying to work through the generalization, but didn't make substantial progress and then haven't had a a change to pick this back up. I don't have test examples yet.

tamuanand:
Hi @vladsavelyev 

Attaching test data for xenome
[Rep3.xenome_stats.txt](https://github.com/ewels/MultiQC/files/13231339/Rep3.xenome_stats.txt)

The key here are the last 7 lines - the lines starting after "Summary" - for MultiQC visualizations, the percentages from "percent class" are important 

```
Summary
count	percent	class
30220658	94.6246	human
892280	2.79384	mouse
32636	0.102187	both
73938	0.231509	neither
717908	2.24786	ambiguous
```

And `xenome` program has to be called this way (note - `xenome` works with both SE and PE data - I am just showing PE data here)

```
 xenome classify \
       -T 20 \
       --pairs \
       -i Rep3_1.trim.fastq \
       -i Rep3_2.trim.fastq \
       -P pdx \
       --graft-name human \
       --host-name mouse \
       --output-filename-prefix Rep3 > Rep3.xenome_stats.txt
```

Hope this helps.

vladsavelyev:
Thank you @tamuanand for the example. I added it in https://github.com/ewels/MultiQC_TestData/pull/300

I adjusted the module to support arbitrary host and graft species. Not sure how common that scenario in practice, but extra flexibility won't hurt.

Screenshot 2023-11-03 at 16 10 50

vladsavelyev (approved):
Looks good to me now!

MikeWLloyd:
> Looks good to me now!

Thanks for fixing the code! Sorry I didn't get it done. Glad to see this is now incorporated.

vladsavelyev:
@tamuanand good suggestions - I'll create another PR!

PR #1874, opened 2023-03-05
fix capitalization of frontmatter in some modules + docs

mashehu:


ewels (approved):
I had already fixed the modules, but the catch within the documentation body was a good one - thanks! 👍🏻

PR #1879, opened 2023-03-09
add admonition test page

mashehu:
Test page to see all the different kinds of admonitions in https://github.com/ewels/MultiQC_website/pull/33


ewels:
Nice, thanks! I think this will be genuinely useful going forward, so I made it into a proper docs page 😅

PR #1880, opened 2023-03-10
DRAGEN: Coverage metrics update

Just-Roma:
New version of the DRAGEN coverage metrics module.  
It is aimed at the issue [1781](https://github.com/ewels/MultiQC/issues/1781), which is currently closed , but was not solved.

- The new code can detect all CSV files, which correspond to the following file structure:
  \<output prefix>.\<coverage region prefix>_coverage_metrics\<arbitrary suffix>.csv
- Part of the code automatically detects metrics and assigns some header's configs when possible.
- A special codeblock is introduced to give users a tool to easily refine header's configs manually.
- An extra module is added, which handles the overall_mean_cov.csv files. These contain the source bed/wgs files,
  which are included in the HTML report. No separate PR is created for that module, because it is used only by
  the DRAGEN Coverage module and does not play any role beyond that.
[test_coverage_data.zip](https://github.com/ewels/MultiQC/files/10944125/test_coverage_data.zip)

PR #1905, opened 2023-04-14
DRAGEN Coverage Metrics Update: solve lint issues

Just-Roma:
The [DRAGEN data](https://github.com/ewels/MultiQC_TestData/tree/master/data/modules/dragen/pr_1880) uploaded for [PR1880](https://github.com/ewels/MultiQC/pull/1880) causes each new PR to fail lint test.
This little update for the current master-branch creates exclusive metric IDs to pass the lint test.


ewels:
Thanks for this! The same fix is in https://github.com/ewels/MultiQC/pull/1880 though right? I am planning to merge that as soon as I get a minute anyway, so would prefer that..

Just-Roma:
Sure. Yeah, it is practically the same fix.

PR #1912, opened 2023-04-25
DRAGEN: new module ROH Metrics

Just-Roma:

The lint tests will fail until [PR1880](https://github.com/ewels/MultiQC/pull/1880) is merged. Nevertheless you can use this module already if you'd like to.


your-highness:
For some python versions there seem to be some LINT errors which I can not really explain. Help wanted.

Issue #290, opened 2016-09-02
Parse software versions from logs


ewels:
Many (but not all) software logs contain information about what version of the tool was run. This would be nice to include in MultiQC report, for example:

> ### FastQC
> 
> All samples were run with v0.11.5 of FastQC

or for a mixture of version numbers:

> ### Kallisto
> 
> Samples were run with versions v0.42 and v0.42.5 `[ dropdown list ]`

As this will be a common feature for most modules, it could be nice to write a new common function to handle this somehow.

darogan:
+1
This will be a great feature (esp in conjunction with clusterflow)

ewels:
Yup, I was originally thinking of getting Cluster Flow to add specific syntax for this that MultiQC could pick up, but @Hammarn pointed out that a lot of the standard logs from tools have versions in. So hopefully for most I can add support for the tool irrespective of how it was run.

Later on I may add support for custom Cluster Flow version printing (and in our NextFlow pipelines), to pick up those tools which don't print their version numbers by default.

ewels:
I think this would work best if it is a new object in `multiqc.report` which modules can add to with versions if they want to. Much like how _General Statistics_ works. So we start by creating an empty placeholder variable [here](https://github.com/ewels/MultiQC/blob/master/multiqc/utils/report.py#L31), then write a new function in the [base module](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/base_module.py#L208) which adds information to that and is called from within modules. Finally, we update the [templates](https://github.com/ewels/MultiQC/blob/master/multiqc/templates/default/base.html#L43) to print this new object if it has data.

Phil

apeltzer:
Sounds like a great feature - also in terms of reproducibility which lots of journals e.g. require now, you could have a single report supporting versioning information of all tools (and most tools do write their versions in a logfile).

johandahlberg:
I was wondering if something has happened to this issue, and/or if there is any interest in a patch to add software versions to the MultiQC report? We are in the process of revamping our reports, and having the software versions in the report is something that has been requested.

apeltzer:
Most of the pipelines we use these days do have custom sections that generate a custom section using multiqc: https://github.com/nf-core/rnaseq/blob/e8376373da8c1dfb0cae5372947cf5b241056076/main.nf#L1041-L1065

johandahlberg:
@apeltzer Nice! Thank you. That is incredibly helpful, and pretty much exactly what we want do do.

[...]

Issue #1892, opened 2023-03-25
JSON Schema for MultiQC config


ewels:
It would be great to have the default config written as a JSON Schema (in YAML!) Likely the best way to do this is by making the config into a Pydantic object. From there we should be able to easily export a schema.

Could then use it for:

- Building reference docs on the website
- Type checking / validation
- Editor (VSCode) integration

Will need to be careful with validation at first, as the config handling is very flexible and quite permissive currently, so people may be doing all kinds of weird stuff (eg. it's used a lot with custom attributes).

We should be able to upload the schema to https://www.schemastore.org/json/ - then anyone editing a `multiqc_config.yml` file will automatically have validation / intellisense for it in VSCode.

ewels:
Should also look into generating automated documentation, to have feature set complete documentation of every config option and it's valid types etc.

PR #1927, opened 2023-05-23
Software versions 2.0

pontushojer:
This PR is inspired by https://github.com/ewels/MultiQC/pull/536 and is a first step to add software version information to the MultiQC report (see discussion in https://github.com/ewels/MultiQC/issues/290). 

So far I have only parsed version information from two tools (`samblaster` and `samtools stats`) but I figured it would be good to get some input on this before continuing. I have also added the ability to specify an auxiliary YAML file (default name `multiqc_versions.yaml`) with version information. 

**Examples**

Below is an example of how the version information is displayed in the report. Here the version are listed after the module section title. 

image

Here is another example with multiple version of the same software.

image

After all module sections a new section called `Software Versions` is added. See image below for example.

image

There will be quite some work to add version information for the other modules. I tired "grepping" the log files for version information and it seamed that at least 50% contained version info, so it will be quite a few. Possibly this could be done in separate PRs.

**Additional ideas**
- Some software that don't output version information to their statistics logs. For some of these you could probably get the version some other way, for example with bowtie2 you could call `bowtie2 --version > sample.version.txt` and parse this log for any discovered samples. Only downside is that MultiQC users would need to change their workflows. 
- List software version connected to each sample name. Something like the dropdown discussed in https://github.com/ewels/MultiQC/issues/290. 

**Summary of additions:**
- new `BaseMultiqcModule` method `add_software_version` to collect and report version extracted from logs
- new "special" module `software_versions` to for creating HTML for `Software Versions` section of the report. Also includes utility functions for loading version for a YAML file and matching softwares to executed multiqc modules.
- new config parameter `version_fn_name` that specifies the path to an auxiliary YAML with version information. 
- extract version info from `samtools stats` and `samblaster` modules. 
- updated docs

---

### DIFF
--- a/docs/core/development/modules.md
+++ b/docs/core/development/modules.md
@@ -639,7 +639,7 @@ The version number here (`0.1.22`) can be extracted using the regex pattern
 
 ```python
 for line in f.splitlines():
-    version = re.search(`r'Version\ (\d{1}.\d+.\d+)'`, line)
+    version = re.search(r'Version\ (\d{1}.\d+.\d+)', line)
     if version is not None:
         self.add_software_version(version.group(1))
 ```
--- a/docs/core/development/modules.md
+++ b/docs/core/development/modules.md
@@ -619,6 +619,40 @@ are as shown:
 self.add_data_source(f=None, s_name=None, source=None, module=None, section=None)
 ```
 
+### Saving version information
+
+Software version information may be present in the log files of some tools. The 
+version number can be included in the report by passing it to the method 
+`self.add_software_version`. Let's use this `samblaster` log below as an example.
+
+```bash
+samblaster: Version 0.1.22
+samblaster: Opening /home/vagrant/samblaster_standalone.sam for read.
+samblaster: Outputting to stdout
+samblaster: Loaded 86 header sequence entries.
+samblaster: Marked 0 of 117 (0.00%) read ids as duplicates using 2752k memory in 0.000S CPU seconds and 0S wall time.
+```
+
+The version number here (`0.1.22`) can be extracted using the regex pattern 
+`r'Version\ (\d{1}.\d+.\d+)'`. The we can use the input to call
+ `self.add_software_version`. 
+
+```python
+for line in f.splitlines():
+    version = re.search(`r'Version\ (\d{1}.\d+.\d+)'`, line)
+    if version is not None:
+        self.add_software_version(version.group(1))
+```
+
+The version number will now appear after the module header in the report as 
+well as in the section `Software Versions` in the end of the report. 
+
+:::tip
+For tools that don't output software versions in their logs these can instead 
+be provided in a separate YAML file. See [Customising Reports](https://multiqc.info/docs/reports/customisation/#listing-software-versions) 
+for details.
+:::
+
 ## Step 3 - Adding to the general statistics table
 
 Now that you have your parsed data, you can start inserting it into the
--- a/docs/core/reports/customisation.md
+++ b/docs/core/reports/customisation.md
@@ -84,6 +84,29 @@ Then this will be displayed at the top of reports:
 
 Note that you can also specify a path to a config file using `-c`.
 
+## Listing software versions
+
+For some modules the software version can be parsed from the log files. If version 
+information is not included in the log files you can instead add them by providing 
+a separate YAML file. By default MultiQC will look for a YAML file called 
+`multiqc_versions.yaml` in you current directory. You can also provide the path 
+the YAML file through the parameter `version_fn_name` in you configs, for example:
+
+```yaml
+version_fn_name: path/to/versions.yaml
+```
+
+Software versions are specified in the YAML as pairs of software names and version 
+strings. Multiple version of the same software are specified as lists. Example:
+
+```yaml
+samblaster: "0.1.24"
+samtools:
+  - "1.15"
+  - "1.10"
+some_other_tool: "2023-1"
+```
+
 ## Sample name replacement
 
 Occasionally, when you run MultiQC you may know that you want to change the resulting
--- a/multiqc/modules/base_module.py
+++ b/multiqc/modules/base_module.py
@@ -12,6 +12,7 @@ import textwrap
 from collections import OrderedDict
 
 import markdown
+from pkg_resources import packaging
 
 from multiqc.utils import config, report, util_functions
 
@@ -74,6 +75,9 @@ class BaseMultiqcModule(object):
             self.doi_link = '<em class="text-muted small" style="margin-left: 1rem;">DOI: {}.</em>'.format(
                 "; ".join(doi_links)
             )
+
+        # List of software version
+        self.versions = []
 
         if target is None:
             target = self.name
@@ -481,6 +485,27 @@ class BaseMultiqcModule(object):
             report.data_sources[module][section][s_name] = source
         except AttributeError:
             logger.warning("Tried to add data source for {}, but was missing fields data".format(self.name))
+
+    def add_software_version(self, version: str):
+        """Save software versions for module."""
+        # Check if version string is PEP 440 compliant to enable version normalization and proper ordering.
+        # Otherwise use raw string is used for version.
+        # - https://peps.python.org/pep-0440/
+        try:
+            version = packaging.version.parse(version)
+        except packaging.version.InvalidVersion:
+            logger.debug(f"Version '{version}' in module {self.name} does not conform to PEP 440 format")
+
+        if version in self.versions:
+            return
+
+        self.versions.append(version)
+
+        # Sort version in order newest --> oldest
+        self.versions.sort(reverse=True)
+
+        # Update version list for report section.
+        report.software_versions[self.name] = self.versions
 
     def write_data_file(self, data, fn, sort_cols=False, data_format=None):
         """Saves raw data to a dictionary for downstream use, then redirects
--- a/multiqc/modules/samblaster/samblaster.py
+++ b/multiqc/modules/samblaster/samblaster.py
@@ -12,6 +12,8 @@ from multiqc.plots import bargraph
 # Initialise the logger
 log = logging.getLogger(__name__)
 
+VERSION_REGEX = r"Version\ (\d{1}.\d+.\d+)"
+
 
 class MultiqcModule(BaseMultiqcModule):
     """Samblaster"""
@@ -86,6 +88,9 @@ class MultiqcModule(BaseMultiqcModule):
         s_name = None
         fh = f["f"]
         for l in fh:
+            match = re.search(VERSION_REGEX, l)
+            if match is not None:
+                self.add_software_version(match.group(1))
             # try to find name from RG-tag. If bwa mem is used upstream samblaster with pipes, then the bwa mem command
             # including the read group will be written in the log
             match = re.search(rgtag_name_regex, l)
--- a/multiqc/multiqc.py
+++ b/multiqc/multiqc.py
@@ -749,7 +749,7 @@ def run(
     # Parse software version from separate YAML file if it exists
     versions_from_file = software_versions.load_versions_from_yaml(config.version_fn_name)
     if versions_from_file:
-        name_to_module = {module.name: module for module in report.modules_output}
+        name_to_module = {module.anchor: module for module in report.modules_output}
         for software, versions in versions_from_file.items():
             # Try to find if the software is listed among the executed modules. Unlisted software are still
             # reported in the `Software Versions` section.
--- a/multiqc/multiqc.py
+++ b/multiqc/multiqc.py
@@ -28,7 +28,7 @@ import rich_click as click
 from rich.syntax import Syntax
 
 from .plots import table
-from .utils import config, lint_helpers, log, megaqc, plugin_hooks, report, util_functions
+from .utils import config, lint_helpers, log, megaqc, plugin_hooks, report, util_functions, software_versions
 
 # Set up logging
 start_execution_time = time.time()
@@ -746,10 +746,28 @@ def run(
         report.runtimes["mods"][run_module_names[mod_idx]] = time.time() - mod_starttime
     report.runtimes["total_mods"] = time.time() - total_mods_starttime
 
+    # Parse software version from separate YAML file if it exists
+    versions_from_file = software_versions.load_versions_from_yaml(config.version_fn_name)
+    if versions_from_file:
+        name_to_module = {module.name: module for module in report.modules_output}
+        for software, versions in versions_from_file.items():
+            # Try to find if the software is listed among the executed modules. Unlisted software are still
+            # reported in the `Software Versions` section.
+            module_name = software_versions.find_matching_module(software, name_to_module)
+            if module_name is None:
+                logger.debug(
+                    "No executed modules matches '{}' listed in '{}'.".format(software, config.version_fn_name)
+                )
+                versions = list(set(map(str, versions)))
+                versions.sort(reverse=True)
+                report.software_versions[software] = versions
+            else:
+                module = name_to_module[module_name]
+                for version in versions:
+                    module.add_software_version(str(version))
+
     # Add section for software versions if any are found
     if report.software_versions:
-        from multiqc.utils import software_versions
-
         report.modules_output.append(software_versions.MultiqcModule())
 
     # Special-case module if we want to profile the MultiQC running time
--- a/multiqc/multiqc.py
+++ b/multiqc/multiqc.py
@@ -746,6 +746,12 @@ def run(
         report.runtimes["mods"][run_module_names[mod_idx]] = time.time() - mod_starttime
     report.runtimes["total_mods"] = time.time() - total_mods_starttime
 
+    # Add section for software versions if any are found
+    if report.software_versions:
+        from multiqc.utils import software_versions
+
+        report.modules_output.append(software_versions.MultiqcModule())
+
     # Special-case module if we want to profile the MultiQC running time
     if config.profile_runtime:
         from multiqc.utils import profile_runtime
--- a/multiqc/templates/default/assets/css/default_multiqc.css
+++ b/multiqc/templates/default/assets/css/default_multiqc.css
@@ -190,6 +190,10 @@ kbd {
   border-left: 5px solid #8eb9dd;
   background-color: #e8f1f8;
 }
+.mqc-module-title{
+  display: inline;
+  margin-right: 5px;
+}
 #analysis_dirs_wrapper {
   max-height: 80px;
   overflow: auto;
--- a/multiqc/templates/default/content.html
+++ b/multiqc/templates/default/content.html
@@ -13,7 +13,10 @@ the output from each module and print it in sections.
     {% for s in m.sections %}
       {% if loop.first %}
       <div class="mqc-module-section-first">
-        <h2 id="{{ m.anchor }}">{{ m.name }}</h2>
+        <h2 class="mqc-module-title" id="{{ m.anchor }}">{{ m.name }}</h2>
+        {% if m.versions %}
+          <span><em>Version:</em> <code>{{m.versions|join("</code>, <code>")}}</code></span>
+        {% endif %}
         {{ m.intro if m.intro }}
         {% if m['comment'] %}<blockquote class="mqc-section-comment">{{ m['comment'] }}</blockquote>{% endif %}
       {% endif %}
--- a/multiqc/utils/report.py
+++ b/multiqc/utils/report.py
@@ -100,6 +100,10 @@ def init():
     global files
     files = dict()
 
+    # Map of software tools to a set of unique version strings
+    global software_versions
+    software_versions = defaultdict(list)
+
 
 def get_filelist(run_module_names):
     """
--- a/multiqc/utils/software_versions.py
+++ b/multiqc/utils/software_versions.py
@@ -4,9 +4,12 @@
 
 
 import logging
+import os
 
 from multiqc.modules.base_module import BaseMultiqcModule
 from multiqc.utils import report
+
+import yaml
 
 # Initialise the logger
 log = logging.getLogger(__name__)
@@ -37,3 +40,40 @@ class MultiqcModule(BaseMultiqcModule):
         content += "</dl>\n"
 
         self.add_section(name=None, content=content)
+
+
+def load_versions_from_yaml(file_name):
+    """Try to load software versions from any YAML file in current directory."""
+    if not os.path.isfile(file_name):
+        file_name = file_name.replace(".yaml", ".yml")
+
+    if not os.path.isfile(file_name):
+        return {}
+
+    with open(file_name) as f:
+        try:
+            log.debug("Reading software versions settings from: {}".format(file_name))
+            software_versions = yaml.safe_load(f)
+        except yaml.scanner.ScannerError as e:
+            log.error("Error parsing versions YAML: {}".format(e))
+            return {}
+
+    # Parse the versions in YAML and make sure that each software maps to a list of version strings
+    for software in list(software_versions):
+        versions = software_versions[software]
+        if type(versions) != list:
+            versions = [versions]
+        versions = [str(version) for version in versions]
+        software_versions[software] = versions
+
+    return software_versions
+
+
+def find_matching_module(software_name: str, modules):
+    if software_name in modules:
+        return software_name
+
+    if software_name.capitalize() in modules:
+        return software_name.capitalize()
+
+    return None
--- a/multiqc/utils/software_versions.py
+++ b/multiqc/utils/software_versions.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env python
+
+""" Super Special-Case MultiQC module to produce report section on software versions """
+
+
+import logging
+
+from multiqc.modules.base_module import BaseMultiqcModule
+from multiqc.utils import report
+
+# Initialise the logger
+log = logging.getLogger(__name__)
+
+
+class MultiqcModule(BaseMultiqcModule):
+    def __init__(self):
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__(
+            name="Software Versions",
+            anchor="multiqc_software_versions",
+            info="Versions of software tools extracted from file contents.",
+        )
+
+        # Don't repeat the Custom Content name in the subtext
+        if self.info or self.extra:
+            self.intro = "<p>{}</p>{}".format(self.info, self.extra)
+
+        self.report_software_versions()
+
+    def report_software_versions(self):
+        """Create section listing software versions."""
+        content = "<dl class=dl-horizontal>\n"
+        for tool_name in sorted(report.software_versions):
+            versions = [str(version) for version in report.software_versions[tool_name]]
+            versions_string = "</code>, <code>".join(versions)
+            content += f"  <dt>{tool_name}</dt><dd><code>{versions_string}</code></dd>\n"
+        content += "</dl>\n"
+
+        self.add_section(name=None, content=content)
--- a/multiqc/modules/samtools/stats.py
+++ b/multiqc/modules/samtools/stats.py
@@ -3,6 +3,7 @@
 """ MultiQC submodule to parse output from Samtools stats """
 
 import logging
+import re
 from collections import OrderedDict
 
 from multiqc import config
@@ -10,6 +11,9 @@ from multiqc.plots import bargraph, beeswarm
 
 # Initialise the logger
 log = logging.getLogger(__name__)
+
+# Regex to grab version number from samtools stats contents
+VERSION_REGEX = r"\((\d+.(\d+|\d+.\d+))\+"
 
 
 class StatsReportMixin:
@@ -22,6 +26,12 @@ class StatsReportMixin:
         for f in self.find_log_files("samtools/stats"):
             parsed_data = dict()
             for line in f["f"].splitlines():
+                # Get version number from file contents
+                if line.startswith("# This file was produced by samtools stats"):
+                    version_match = re.search(VERSION_REGEX, line)
+                    if version_match is not None:
+                        self.add_software_version(version_match.group(1))
+
                 if not line.startswith("SN"):
                     continue
                 sections = line.split("\t")
--- a/multiqc/utils/config_defaults.yaml
+++ b/multiqc/utils/config_defaults.yaml
@@ -38,6 +38,7 @@ base_count_desc: "millions"
 output_fn_name: "multiqc_report.html"
 data_dir_name: "multiqc_data"
 plots_dir_name: "multiqc_plots"
+versions_fn_name: "multiqc_versions.yaml"
 data_format: "tsv"
 module_tag: []
 force: false
--- a/multiqc/utils/config_defaults.yaml
+++ b/multiqc/utils/config_defaults.yaml
@@ -38,7 +38,7 @@ base_count_desc: "millions"
 output_fn_name: "multiqc_report.html"
 data_dir_name: "multiqc_data"
 plots_dir_name: "multiqc_plots"
-versions_fn_name: "multiqc_versions.yaml"
+version_fn_name: "multiqc_versions.yaml"
 data_format: "tsv"
 module_tag: []
 force: false


vladsavelyev:
This is really awesome! I'm eager to merge it, and implement version parsing for other modules whenever possible.

Just minor comments. I think it somehow slightly broke the padding between headers and subheaders. See the different padding under the `Samtools` header:

Before:

Screenshot 2023-08-28 at 23 21 15

<br /><br />
And after:

Screenshot 2023-08-28 at 23 21 31

<br />
<br />
<br />

Same tiny padding in the new section:

Screenshot 2023-08-28 at 23 22 48

<br />
<br />

The new section is great to have! The only thing I think I would only align the tool names in the table to the left - to keep the style of the whole report nicely left-aligned, like any of this:

240248068-515ed956-59b1-435b-bc2a-69f899f3cd23

240248068-515ed956-59b1-435b-bc2a-69f899f3cd23 copy2
Though I don't have a strong opinion regarding it.

vladsavelyev:
Really love that it's documented how to add versions both when developing a new module, and when customizing a report.

pontushojer:
Thanks for the review @vladsavelyev! The title padding (margins) should now be fixed as well as the left aligning the software names in the new section, se examples below.

image

image

Please try it out to see that it also looks good on your end 😄

multiqc/multiqc.py
```diff
@@ -746,6 +746,30 @@ def __rich_measure__(self, console: rich.console.Console, options: rich.console.
...
+                logger.debug(
+                    "No executed modules matches '{}' listed in '{}'.".format(software, config.version_fn_name)
+                )
+                versions = list(set(map(str, versions)))
+                versions.sort(reverse=True)
+                report.software_versions[software] = versions
+            else:
+                module = name_to_module[module_name]
+                for version in versions:
+                    module.add_software_version(str(version))
```
vladsavelyev, line 767: The method needs the sample name as a first argument, so it shouldn't work here?

pontushojer: You are absolutely right, this got broken in [df602d4](https://github.com/ewels/MultiQC/pull/1927/commits/df602d46f41b087554fb5f3d35432266a39032fa) but should now be fixed.

vladsavelyev: Excellent, thanks for quickly fixing it! I think it also makes sense to swap the version and sample name arguments and make the second one explicitly optional. I pushed the change now just for the sake of merging everything quickly, I hope you don't mind :)

vladsavelyev:
Thanks for addressing the review quickly. It looks really great now! I pushed a few changes myself:
* I think it makes sense to show all tools in the section in the bottom (both parsed automatically and pulled from the versions file)
* I think it's better to use `module.name` in that table as the software name, and sort case-insensitive It's a bit ugly that some tools are capitalised, and others are not; but I want it to have the same names and sorting order as the sections of the report itself.
* Nitpick - using singular or plural "Versions" depending if there is more than one.
* I'll also add a change log now, and good to merge.

vladsavelyev (approved):

vladsavelyev:
Attaching an example report for @ewels to take a look:
[multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/12464604/multiqc_report.html.zip)

I'll also add the `multiqc_version.yaml` into the test data repo.

multiqc/multiqc.py
```diff
@@ -746,6 +746,27 @@ def __rich_measure__(self, console: rich.console.Console, options: rich.console.
         report.runtimes["mods"][run_module_names[mod_idx]] = time.time() - mod_starttime
     report.runtimes["total_mods"] = time.time() - total_mods_starttime
 
+    # Parse software version from config if provided
+    versions_from_config = software_versions.load_versions_from_config(config)
+    if versions_from_config:
```
ewels, line 751: This function always returns a dict, so this isn't needed. If there was no config then the loop will just be skipped anyway, as it won't have any items.

multiqc/multiqc.py
```diff
@@ -746,6 +746,27 @@ def __rich_measure__(self, console: rich.console.Console, options: rich.console.
...
+        for software, versions in versions_from_config.items():
+            # Try to find if the software is listed among the executed modules.
+            # Unlisted software are still reported in the `Software Versions` section.
+            versions = list(set(map(str, versions)))
+            versions.sort(reverse=True)
```
ewels, line 756: I think this fits better in the `load_versions_from_config` function, along with the other cleanup we have there..

multiqc/multiqc.py
```diff
@@ -746,6 +746,27 @@ def __rich_measure__(self, console: rich.console.Console, options: rich.console.
...
+            module = software_versions.find_matching_module(software, report.modules_output)
+            if module is None:
+                logger.debug(f"No executed modules matches '{software}' listed in config.software_versions")
+            else:
```
ewels, line 760: Not sure we really need this..

```suggestion
            if module is not None:
```

multiqc/multiqc.py
```diff
@@ -746,6 +746,27 @@ def __rich_measure__(self, console: rich.console.Console, options: rich.console.
...
+                software = module.name
```
ewels, line 761:
```suggestion
```

multiqc/multiqc.py
```diff
@@ -746,6 +746,27 @@ def __rich_measure__(self, console: rich.console.Console, options: rich.console.
...
+                for version in versions:
+                    module.add_software_version(str(version))
+            report.software_versions[software] = versions
```
ewels, line 764:
```suggestion
            report.software_versions[module.name] = versions
```

multiqc/multiqc.py
```diff
@@ -746,6 +746,27 @@ def __rich_measure__(self, console: rich.console.Console, options: rich.console.
         report.runtimes["mods"][run_module_names[mod_idx]] = time.time() - mod_starttime
     report.runtimes["total_mods"] = time.time() - total_mods_starttime
 
+    # Parse software version from config if provided
```
ewels, line 749: I would prefer to move this whole block of code into a function in `software_versions.py` and just pass the required `report` objects to it.

I'm hoping to refactor `mutliqc.py` to split it up into more files so that it's easier to navigate.

'cc @vladsavelyev to check he agrees here. This might all need a bit of reorganising anyway when we refactor to use proper `report` classes in the future.

multiqc/templates/default/content.html
```diff
@@ -13,7 +13,12 @@
     {% for s in m.sections %}
       {% if loop.first %}
       <div class="mqc-module-section-first">
-        <h2 id="{{ m.anchor }}">{{ m.name }}</h2>
+        <div class="mqc-module-title-section">
+          <div><h2 class="mqc-module-title" id="{{ m.anchor }}">{{ m.name }}</h2></div>
+          {% if m.versions %}
+            <div><em>Version{% if m.versions|length > 1 %}s{% endif %}:</em> <code>{{m.versions|join("</code>, <code>")}}</code></div>
```
ewels, line 19: Prefer not to use `<em>` - use CSS instead (with `<span>` + class if needed).

multiqc/templates/default/content.html
```diff
@@ -13,7 +13,12 @@
     {% for s in m.sections %}
       {% if loop.first %}
       <div class="mqc-module-section-first">
-        <h2 id="{{ m.anchor }}">{{ m.name }}</h2>
+        <div class="mqc-module-title-section">
+          <div><h2 class="mqc-module-title" id="{{ m.anchor }}">{{ m.name }}</h2></div>
```
ewels, line 17: I don't think that we need a wrapper `div` here.
```suggestion
          <h2 class="mqc-module-title" id="{{ m.anchor }}">{{ m.name }}</h2>
```

multiqc/templates/default/content.html
ewels:
```suggestion
            <div class="mqc-module-versions"><span>Version{% if m.versions|length > 1 %}s{% endif %}:</span> <code>{{m.versions|join("</code>, <code>")}}</code></div>
```

multiqc/utils/report.py
```diff
@@ -100,6 +100,10 @@ def init():
     global files
     files = dict()
 
+    # Map of software tools to a set of unique version strings
+    global software_versions
```
ewels, line 104: Ugh, globals 😓 

Looking forward to refactoring how `report` works! 😅

multiqc/utils/software_versions.py
```diff
@@ -0,0 +1,64 @@
...
+        # Initialise the parent object
+        super(MultiqcModule, self).__init__(
+            name="Software Versions",
+            anchor="multiqc_software_versions",
+            info="Versions of software tools extracted from file contents.",
+        )
+
+        # Don't repeat the Custom Content name in the subtext
+        if self.info or self.extra:
+            self.intro = "<p>{}</p>{}".format(self.info, self.extra)
```
ewels, line 28: Not sure that this is needed? 🤔

multiqc/multiqc.py
```diff
@@ -746,6 +746,27 @@ def __rich_measure__(self, console: rich.console.Console, options: rich.console.
...
+            if module is None:
+                logger.debug(f"No executed modules matches '{software}' listed in config.software_versions")
+            else:
+                software = module.name
+                for version in versions:
+                    module.add_software_version(str(version))
+            report.software_versions[software] = versions
+
+    # Add section for software versions if any are found
+    if report.software_versions and len(report.modules_output) > 0:
```
ewels, line 767: We need a new config option to disable this, people may not always want it in reports.

ewels (changes requested):
Awesome work @pontushojer - thanks so much for pushing this! I was talking to @vladsavelyev about it and doing some archaeology on the issue, it's been in the works a long long time 😉 

I've left some fairly pedantic review comments, but it's all fairly minor stuff and hopefully not difficult to address.

One bigger thing - this will hopefully be used extensively in nf-core and I thought it good to compare to what we already have there.
Checking the [MultiQC report](https://nf-co.re/scrnaseq/2.4.0/results/scrnaseq/results-61d1919aa96a4d82517aeb6d04943262a42a055e/multiqc/?file=multiqc_report.html) first nf-core pipeline I clicked on, we see this:



Note that we have a first column there for `Process Name`. I remember this discussion when it happened. The logic with this was that a pipeline might include a software tool multiple times within multiple modules, and it's good to know which version number corresponds to which step in the pipeline.

Tools being included with multiple versions will hopefully be an edge case, and it might be that we decide we can't do much about this level of detail (and that's fine). But I thought it might be worth mentioning in case we can think of some elegant solution..

multiqc/utils/software_versions.py
```diff
@@ -0,0 +1,64 @@
...
+
+        self.report_software_versions()
+
+    def report_software_versions(self):
+        """Create section listing software versions."""
+        content = "<dl class=dl-horizontal>\n"
+        for tool_name in sorted(report.software_versions, key=lambda x: x.lower()):
```
ewels, line 35: _/ * waits for someone to request a custom sort order on this * /_

multiqc/templates/default/assets/css/default_multiqc.css
```diff
@@ -190,6 +190,16 @@ kbd {
   border-left: 5px solid #8eb9dd;
   background-color: #e8f1f8;
 }
+.mqc-module-title-section {
+  display: flex;
+  align-items: flex-end;
+  margin-top: 20px;
+  margin-bottom: 10px;
+  column-gap: 5px;
```
ewels, line 198:
```suggestion
  column-gap: 10px;
```

multiqc/templates/default/content.html
ewels: Actually, just use bootstrap classes to drop the text colour a bit, and I'm being hypocritical and using `<em>` already elsewhere, so meh - probably fine.. 😅 

```suggestion
            <div class="text-muted"><em>Version{% if m.versions|length > 1 %}s{% endif %}:</em> <code>{{m.versions|join("</code>, <code>")}}</code></div>
```

docs/core/reports/customisation.md
```diff
@@ -83,6 +83,22 @@ Then this will be displayed at the top of reports:
...
+Software versions are specified in the YAML as pairs of software names and version
+strings. Multiple version of the same software are specified as lists. Example:
+
+```yaml
+software_versions:
+  samblaster: "0.1.24"
+  samtools:
+    - "1.15"
+    - "1.10"
+  some_other_tool: "2023-1"
```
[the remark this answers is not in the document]
pontushojer: So on testing this samtools `1.10` was reported as `1.1` while `1.15` was actually fine. While it might be ok to leave out quotes in some cases from user standpoint I think it better to recommend passing strings here. At least until we have [proper config validation](https://github.com/ewels/MultiQC/issues/1892) inplace.

multiqc/modules/base_module.py
```diff
@@ -12,6 +12,7 @@
 from collections import OrderedDict
 
 import markdown
+from pkg_resources import packaging
```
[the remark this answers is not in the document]
pontushojer: Maybe this could be raised as a separate issue? `pkg_resources` is used in version parsing multiple places in the codebase besides here.

multiqc/modules/samblaster/samblaster.py
```diff
@@ -111,6 +117,9 @@ def parse_samblaster(self, f):
         if s_name is None:
             s_name = f["s_name"]
 
+        if version is not None:
+            self.add_software_version(version, s_name)
```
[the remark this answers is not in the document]
pontushojer: The sample name could be parsed from the samblaster log file (if not reading from stdin). Thus we don't have the sample name until after the loop.

multiqc/modules/base_module.py
```diff
@@ -75,6 +76,9 @@ def __init__(
                 "; ".join(doi_links)
             )
 
+        # List of software version
+        self.versions = []
```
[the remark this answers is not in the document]
pontushojer: Move this up a bit but it still listed it separate from `self.doi` and such as `self.versions` is not something that should be modified directly when creating a new module, just to make it clear to devs.

docs/core/reports/customisation.md
```diff
@@ -83,6 +83,22 @@ Then this will be displayed at the top of reports:
...
+```
+
```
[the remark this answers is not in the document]
ewels: hah, sorry 😅  Yes I think it's good to have both 👍🏻

docs/core/reports/customisation.md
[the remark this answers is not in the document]
ewels: Ah right, ok fair enough 👍🏻 

In that case, can we add a [warning admonition box](https://multiqc.info/docs/development/contributing/#warnings) pointing out this gotcha and saying to use quotes to be safe?

multiqc/modules/base_module.py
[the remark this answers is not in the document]
ewels: Yeah good call. Created https://github.com/ewels/MultiQC/issues/2012 👍🏻

multiqc/modules/base_module.py
```diff
@@ -41,7 +42,10 @@ def __init__(
         self.info = mod_cust_config.get("info", info)
         self.comment = mod_cust_config.get("comment", comment)
         self.extra = mod_cust_config.get("extra", extra)
-        self.doi = mod_cust_config.get("doi", doi)
+        self.doi = mod_cust_config.get("doi", (doi or []))
+
+        # List of software version(s) for module
+        self.versions = []
```
ewels, line 48:
```suggestion
        # List of software version(s) for module. Don't append directly, use add_software_version()
        self.versions = []
```

multiqc/modules/base_module.py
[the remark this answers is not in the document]
ewels: Sounds good 👍🏻  If we want to make it clear to devs then we should explicitly point this out with a code comment. Suggestion in https://github.com/ewels/MultiQC/pull/1927/files#r1309935462

ewels:
Ok, having slept on it I have a suggestion, up for discussion. In this PR we already support multiple software versions per tool:

```yaml
software_versions:
  tool_one:
    - 1.2
    - 1.3
  tool_two:
    - 4.5
  tool_two_related:
    - 12.3
```


I wonder if we use tuples / single key dicts in this optional array. For example:

```yaml
software_versions:
  tool_one:
    - tool_one: 1.2
    - tool_one: 1.3
  tool_two:
    - tool_two: 4.5
    - tool_two_related: 12.3
```

> [!NOTE]
> Array of dicts, because we may need duplicate keys. Could think of other ways to do the same thing here, feel free to suggest alternatives.

If the key matches the main tool name, it's ignored. If it differs then the name can be reported. So here, `tool_one` would render in exactly the same was as currently:



But tool two can now show associated versions (good for example with samtools, also showing htslib):



This would also allow us to render the extra tier of hierarchy used in the nf-core reports at the bottom, with `process_name` > `software` > `version` (whether as a table, or the current `dl`, don't mind - doesn't matter much for this discussion).

However, doing that for nf-core would mean that software versions _wouldn't_ be reported next to section titles. Except I guess if found in the log files in addition 🤔  In which case, nf-core pipelines could use a (as yet unwritten) config flag to not include the automatic version numbers in the bottom section or something.

Sorry, bit of a "thinking out loud" comment here. What do you guys think?

docs/core/reports/customisation.md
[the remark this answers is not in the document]
pontushojer: I was also thinking mostly on pipelines when writing this so that why I like the idea of a separate YAML file. 

Also one thing with providing version in the configs is that there are multiple scope-levels. If you, for whatever reason, add software version information on a system or user config level, it will appear in all reports.

multiqc/multiqc.py
pontushojer: Added a new option `--no-versions-section` for this.

docs/core/development/modules.md
```diff
@@ -626,6 +626,44 @@ are as shown:
...
+...
+```
+
+The version number here (`1.3`) can be extracted using the regex pattern
+`r"\((\d+.(\d+|\d+.\d+))\+"` that also allows for patches e.g. `1.3.1`. The we can use the input to call
+`self.add_software_version`. We also provide the sample name, `f["s_name"]` in this case, so that we don't add versions for samples that should be ignored.
+
+```python
+for line in f.splitlines():
+    version = re.search(r"\((\d+.(\d+|\d+.\d+))\+", line)
```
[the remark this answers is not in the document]
pontushojer: Yeah my regex was way more complicated. Changed to this pattern in docs and in the `samtools/stats` module.

docs/core/development/modules.md
```diff
@@ -626,6 +626,44 @@ are as shown:
...
+    if version is not None:
+        self.add_software_version(version.group(1), f["s_name"])
```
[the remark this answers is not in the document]
pontushojer: Added comment about this to the docs.

multiqc/modules/base_module.py
```diff
@@ -482,6 +486,32 @@ def add_data_source(self, f=None, s_name=None, source=None, module=None, section
...
+        # Otherwise use raw string is used for version.
+        # - https://peps.python.org/pep-0440/
+        # Use flag `is_compliant` so we don't log the same warning multiple times.
+        version, is_compliant = software_versions.parse_version(version)
+
+        if version in self.versions:
+            return
+
+        if is_compliant:
+            logger.debug(f"Version '{version}' in module {self.name} does not conform to PEP 440 format")
```
ewels, line 505: Honestly I'm not sure that we really need a debug log here at all. I think this will happen a lot and I wonder if it'll make people worry if they see it in the log.

```suggestion
            logger.debug(f"Version '{version}' in module {self.name} does not conform to PEP 440 format - cannot sort list of versions nicely")
```

multiqc/modules/samblaster/samblaster.py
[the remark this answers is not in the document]
ewels: Not sure I really understand. More verbose: Why can't we call the function as soon as `version` is assigned inside the loop (line 94), as we do in samtools?

multiqc/multiqc.py
```diff
@@ -245,6 +245,12 @@
     is_flag=True,
     help="Don't upload generated report to MegaQC, even if MegaQC options are found",
 )
+@click.option(
+    "--no-versions-section",
+    "no_versions_section",
+    is_flag=True,
+    help="Exclude section 'Software Versions' listing software versions from the report",
+)
```
ewels, line 253: I think this was added because of https://github.com/ewels/MultiQC/pull/1927#discussion_r1310019204 - I didn't really mean a new CLI flag. I'm not sure that this would be used enough to warrant another CLI flag? The list is already pretty long, I try to only add things when needed.

Additionally: when adding a new CLI flag, it should be configured for `rich.click` in the config above so that it gets grouped properly.

multiqc/multiqc.py
```diff
@@ -746,6 +753,28 @@ def __rich_measure__(self, console: rich.console.Console, options: rich.console.
...
 
+    # Parse software version from config if provided
+    versions_from_config = software_versions.load_versions_from_config(config)
+    for software, versions in versions_from_config.items():
+        # Try to find if the software is listed among the executed modules.
+        # Unlisted software are still reported in the `Software Versions` section.
+        module = software_versions.find_matching_module(software, report.modules_output)
+        if module is None:
+            logger.debug(f"No executed modules matches '{software}' listed in config")
+        else:
```
ewels, line 764: Repeat of https://github.com/ewels/MultiQC/pull/1927#discussion_r1309398815 as it became outdated.

I don't think that this debug log message is particularly helpful. I think it can be removed.

```suggestion
        if module is not None:
```

multiqc/multiqc.py
```diff
@@ -322,6 +328,7 @@ def run(
     lint=False,
     make_pdf=False,
     no_megaqc_upload=False,
+    no_versions_section=False,
```
ewels, line 331: When adding CLI flags for config options, the default needs to be merged with the user's config. See below after the comment `# Set up key variables (overwrite config vars from command line)`.

multiqc/multiqc.py
ewels: Ok cool - I didn't really mean CLI flag, I just meant something that can be set in a config file. I think I'd prefer to remove the new CLI flag if you don't mind. Sorry for not being clear.

multiqc/utils/software_versions.py
ewels: Semi serious about this. If we do write this so that people can for example list the processes in their pipeline with associated software versions, they will likely want to maintain the order.

Yet another config file option..? 🤔 As I think sorting makes sense most of the time..

ewels (commented):
Updating review comments..

pontushojer:
@ewels thanks for the review comments and suggestions! 

I have been thinking a bit about grouping of software versions and I think it makes sense to enable this. Especially as some tools report the version of multiple software, such as `samtools stats` you mentioned in a comment. I have now added a commit introducing this change. It follows the same structure as for nf-core with one process linking to one or multiple software with a list of versions for each. 

Here is an example of the updated `Software versions`  section, I basically copied the table from nf-core/rnaseq 
image

and here is the module section for `samtools` listing multiple `samtools` and `htslib` versions. 
image

I need to update the documentation but this would at least be something you can try out for now

>However, doing that for nf-core would mean that software versions wouldn't be reported next to section titles. Except I guess if found in the log files in addition 🤔 In which case, nf-core pipelines could use a (as yet unwritten) config flag to not include the automatic version numbers in the bottom section or something.

About this I think its reasonable to add a config (not CLI 😉 ) option to turn this of. In the pipeline application you would probably also have a process name that is distinct from the MultiQC module names in which case the version listed next to the module titles would not be affected.

multiqc/multiqc.py
pontushojer: Yes a CLI option is definitely unnecessary here, I think I got confussed in you previoud comment 😉 I will remove the `--no-versions-verions` flag.

multiqc/modules/samblaster/samblaster.py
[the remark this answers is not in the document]
pontushojer: This is two examples to samblaster logs
```
==> testdata/data/modules/samblaster/samblaster_standalone.log <==
samblaster: Version 0.1.22
samblaster: Opening /home/vagrant/samblaster_standalone.sam for read.
samblaster: Outputting to stdout
samblaster: Loaded 86 header sequence entries.
samblaster: Marked 0 of 117 (0.00%) read ids as duplicates using 2752k memory in 0.000S CPU seconds and 0S wall time.

==> testdata/data/modules/samblaster/stdinout.txt <==
samblaster: Version 0.1.24
samblaster: Inputting from stdin
samblaster: Outputting to stdout
samblaster: Loaded 86 header sequence entries.
samblaster: Marked 32 of 2761 (1.16%) read ids as duplicates using 1600k memory in 0.004S CPU seconds and 18S wall time.
```

In both cases the version is parsed from the first line. In the first example the sample name will be parsed from the log file, see second line: `/home/vagrant/samblaster_standalone.sam`. In the second example the input is stdin so the sample name is taken from the file name (`stdinout.txt` in this case). As the sample name is needed when calling `add_software_version()` to avoid ignored samples we need to wait until we have parsed at least the second log file line. In practice it is just easier to do this after parsing the full file. Does this make sense?

docs/core/development/modules.md
```diff
@@ -626,6 +626,44 @@ are as shown:
...
 ```
 
+### Saving version information
+
+Software version information may be present in the log files of some tools. The
+version number can be included in the report by passing it to the method
+`self.add_software_version`. Let's use this `samtools stats` log below as an example.
+
+```bash
+# This file was produced by samtools stats (1.3+htslib-1.3) and can be plotted using plot-bamstats
```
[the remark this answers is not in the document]
pontushojer: This is implemented in commit c49df0eba2b3b3a9f996f84d2c7e3966e68f9bbc

multiqc/modules/base_module.py
pontushojer: Removed in 29f82cf45f83e1bc5095fc9b6e932a872f81fea6

multiqc/multiqc.py
pontushojer: Fixed in 49647e1eb8816a8b2d245b04b3ccd0e9ca3b50fb

multiqc/multiqc.py
pontushojer: This debug message is now removed

docs/core/reports/customisation.md
```diff
@@ -83,6 +83,108 @@ Then this will be displayed at the top of reports:
...
+For modules where version information can be extracted from log files,
+or when not available in log files, you can alternatively add them in
+the configuration. Software versions can be specified in YAML format
+using two options.
+
+### Option 1: Grouping software sersions by process name
+
+For each process, software versions can be organized within a nested YAML
+dictionary with three levels: `process` -> `software` -> `version(s)`.
```
ewels, line 96: I think we should avoid the term `process` as that's a Nextflow thing, not used anywhere else in MultiQC.

Better to call it `group` maybe?

docs/core/reports/customisation.md
```diff
@@ -83,6 +83,108 @@ Then this will be displayed at the top of reports:
...
+
+An alternative is to specify pairs of software names and version
+strings directly. This option does not group software into processes.
+Instead the software name is used directly for the process name. Example:
+
+```yaml
+software_versions:
+  samblaster: "0.1.24"
+  quast:
+    - "5.2.0"
```
ewels, line 130: I think we probably have _too much_ flexibility with the format now. Feels like you can throw basically whatever you want at this, which will probably be confusing. I was thinking before to force people to use one of two formats: simple or complex.

Simple is just name + version:

```yaml
software_versions:
  samblaster: "0.1.24"
```

Complex is the full way in the other direction:

```yml
software_versions:
  quast:
   - quast: "5.2.0"
   - quast: "5.1.0"
```

NB: List of dicts, which is actually not one of the formats you have here. But I think that this syntax is clearer / more flexible. If we have a list of dicts then there's not really a need for list of strings etc.

What do you think?

ewels (commented):
Quick couple of comments before bed..

Getting there! Sorry for the intense review @pontushojer. It's a reflection of me thinking that this is a really awesome new feature and wanting to get it _just_ right before merge as a result. So take it as a complement 😅

docs/core/reports/customisation.md
pontushojer: I my opinion the list of dicts format is a bit repetitive. If you want to list multiple versions for a specific software, surely a list is cleaner. 

```yaml
software_versions:
  quast:
    quast: 
      - "5.2.0"
      - "5.1.0"
```

I also a bit confused about these two formats you suggest. 

In the *simple* format (that does not allow grouping) its software and version(s) is provided as a dict. Presumably this would also need to allow multiple versions so the format would be like this, not allowing single string inputs like above:

```yaml
software_versions:
  software:
    - version
    - other_version
```

In the *complex* format however, within each group you are instead passing a list of dicts specify versions.

```yaml
software_versions:
  group:
    - software: version
    - software: other_version
```

I think how you specify multiple versions should be the same if we choose to allow two formats. If we want to make the format a bit list flexible my suggestion would be these two formats. They would be similar to what we currently have but require a list of strings for each software. 

**1: Grouped input (complex)**
```yaml
software_versions:
  group:
    software: 
      - version
      - other_version
    other_software:
      - version
```

**2: Ungrouped input (simple)**
```yaml
software_versions:
  software: 
    - version
    - other_version
  other_software:
    - version
```

docs/core/reports/customisation.md
pontushojer: Yes, `process` might be mimicking Nextflow to much. Let's go for `group`.

multiqc/multiqc.py
pontushojer: Moved block in c70c0f30c4a7da9cf0b0801e08ff80ddd97fcd7c

docs/core/reports/customisation.md
ewels: Yup, you're right 👍🏻 

I can't remember the justification for the list of dicts now, which probably means it wasn't well thought through 😅  I think it was something along the lines of wanting to allow multiple identical keys - but there's no need for that if the key can have a list of versions. So yeah, I think I was just chasing my tail 😅 

Let's go for your two suggestions here 👍🏻

multiqc/utils/software_versions.py
```diff
@@ -0,0 +1,256 @@
...
+    software_versions_config = getattr(config, "software_versions", defaultdict(lambda: defaultdict(list)))
+    software_versions_config, is_valid = validate_software_versions(software_versions_config)
+
+    if not is_valid:
+        log.error("Software versions loaded config.software_versions is not in a valid format")
+
+    software_versions_file = defaultdict(lambda: defaultdict(list))
+    file_name = config.version_fn_name
+    if not os.path.isfile(file_name):
+        file_name = file_name.replace(".yaml", ".yml")
```
ewels, line 108: We cannot limit this to a specific file / filename, it must be a glob pattern. The idea of this is to allow many files to sit side by side - this means that we will very quickly have filename conflicts. This could be resolved by using subdirectories, but there's no need.

Better to use `*_mqc_version.yml` instead and allow people to have many of these files in one place.

I had envisaged using the standard MultiQC search mechanism to find these files, and then iterating over the resulting list in the same way as a module.

See original comment: https://github.com/ewels/MultiQC/pull/1927#discussion_r1309388032 (easy to get lost in the long review though 👍🏻 )

multiqc/modules/base_module.py
```diff
@@ -30,7 +31,7 @@ def __init__(
         extra=None,
         autoformat=True,
         autoformat_type="markdown",
-        doi=[],
+        doi=None,
```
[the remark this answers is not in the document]
ewels: Right! I knew this, I remember going through all of nf-core tools at one point switching out the default arguments. Totally forgot when seeing this 🙈 Ok, good call 👍🏻

multiqc/modules/samblaster/samblaster.py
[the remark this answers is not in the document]
ewels: Gotcha, absolutely 👍🏻  Thanks for the great explanation 😃

multiqc/utils/software_versions.py
ewels: Meh, can't think of a good way to handle this for now. Let's cross the bridge when someone asks for it.

multiqc/utils/software_versions.py
```diff
@@ -0,0 +1,256 @@
...
+    def _make_versions_html(versions):
+        """Generate a tabular HTML output of all versions."""
+        # source: https://github.com/nf-core/rnaseq/blob/3bec2331cac2b5ff88a1dc71a21fab6529b57a0f/modules/nf-core/custom/dumpsoftwareversions/templates/dumpsoftwareversions.py#L12
+        html = [
+            dedent(
+                """\
+                <table class="table mqc_versions_table">
+                    <thead>
+                        <tr>
+                            <th> Group </th>
```
ewels, line 47: Need a MultiQC config option to be able to customise this, as the group could be different things to different users.

multiqc/multiqc.py
ewels: Thank you, and sorry for not being clearer originally! 🌟

docs/core/reports/customisation.md
ewels: I think this is how it's now implemented, right? So good to resolve 👍🏻

ewels:
Ok, just gone through trying to resolve all of the review comments. I think we're very nearly there. Two things remaining as I see it:

- [ ] https://github.com/ewels/MultiQC/pull/1927#discussion_r1315185681 (glob of version files, not a single file)
- [ ] https://github.com/ewels/MultiQC/pull/1927#discussion_r1315193016 (new config option to customise table heading)

Only other "behaviour" comment / discussion I have left is about the _Software Versions_ report section. I wonder if this is overkill for most users, given that we have the information on the section headings already. How about we have some automatic behaviour for this, whereby it's only included if we have any groups that do not match the module names?

This could then be overwritten by having a three-state switch instead of the current `skip_versions_section`:

```yaml
show_versions_section: null   # Default - automatic behaviour
show_versions_section: true   # Always show table
show_versions_section: false  # Never show table
```

docs/core/reports/customisation.md
pontushojer: Great!  

So my suggestion above is mostly how I implemented it except one thing. We currently allow single string input, we could possibly drop this to make it a bit more straightforward and a little less flexible. 

i.e. this single string 
```yaml
software_versions:
  software: version
```
would have to become a list
```yaml
software_versions:
  software: 
    - version
```

It would also mean less mixing of styles which would make it easier to read and write.

i.e. comparing something like this

```yaml
software_versions:
  software: 
    - version
  other_software: version
  other software_with_multiple_versions:
    - version1
    - version2
```

to this

```yaml
software_versions:
  software: 
    - version
  other_software: 
    - version
  other software_with_multiple_versions:
    - version1
    - version2
```

I think I am leaning for removing this input option.

ewels:
Note that the globbing thing may not be trivial @pontushojer - so feel free to pull myself or @vladsavelyev in here to help with that if you need 🙏🏻

pontushojer:
I took a stab at adding the glob pattern thing. @vladsavelyev and @ewels please look through it and make any changes you see fit. I am not sure I will have much more time to work on this right now so I would be grateful for any help 🙏. I also added a PR to MultiQC_TestData with some YAML files I used to test this, see https://github.com/ewels/MultiQC_TestData/pull/274. 

> Only other "behaviour" comment / discussion I have left is about the _Software Versions_ report section. I wonder if this is overkill for most users, given that we have the information on the section headings already. How about we have some automatic behaviour for this, whereby it's only included if we have any groups that do not match the module names?

About this, I think that it is better to include this section as a default. The footprint of this section is quite small and it you make it easier to find all software versions connected to a certain report rather than looking at each separate section.

multiqc/utils/software_versions.py
```diff
@@ -0,0 +1,259 @@
...
+                            <th> Version  </th>
+                        </tr>
+                    </thead>
+                """
+            )
+        ]
+        for group, tmp_versions in sorted(versions.items()):
+            html.append("<tbody>")
+            for i, (tool, versions) in enumerate(sorted(tmp_versions.items())):
+                versions = list(map(str, versions))
```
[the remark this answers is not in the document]
vladsavelyev: Ah, my bad, they are already sorted, just in the descending order.

vladsavelyev:
Thank you so much @pontushojer - really great job!

> About this, I think that it is better to include this section as a default. The footprint of this section is quite small and it you make it easier to find all software versions connected to a certain report rather than looking at each separate section.

Agree

multiqc/utils/software_versions.py
```diff
@@ -0,0 +1,259 @@
...
+            elif isinstance(a[key], list) and isinstance(b[key], list):
+                a[key].extend(b[key])
+            else:
+                raise Exception("Conflict at " + ".".join(path + [str(key)]))
+        else:
+            a[key] = b[key]
+    return a
+
+
+def sort_versions(versions):
```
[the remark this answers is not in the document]
pontushojer: I am not sure I have done much thinking before selecting this 😉 Go for the natural sort order if you think its more readable.

ewels:
Let's use `table.plot()` but then I think we're probably good to go 👍🏻

pontushojer:
> Let's use `table.plot()` but then I think we're probably good to go 👍🏻

Using `table.plot()` for the software version table instead of the current custom html table? I am not sure is appropriate. Is there some functionally from there you would think is useful? Or is the idea just to not add too much custom code?

vladsavelyev:
@ewels, that's the best I can do with `table.plot()` without going all hacky:

Screenshot 2023-09-07 at 01 01 16

With `Group` as a first column it doesn't work well as it doesn't allow multiple empty entries in the first column.

```python
def report_software_versions(self):
    """Create section listing software versions."""
    headers = {
        "Version": {
            "title": "Versions", 
            "namespace": "Software versions", 
            "description": f"versions used in the {mqc_config.versions_table_group_header}"
        },
    }
    data = dict()
    for group, tmp_versions in sorted(mqc_report.software_versions.items()):
        if len(tmp_versions) == 1 and list(tmp_versions.keys())[0] == group:
            vs = list(tmp_versions.values())[0]
            versions = ", ".join(list(map(str, vs)))
        else:
            versions = ", ".join(f"<b>{s}</b>: {', '.join(list(map(str, vs)))}" for s, vs in tmp_versions.items())
        data[group] = {
            'Version': versions,
        }
    self.add_section(
        name=None,
        anchor="multiqc_software_versions",
        plot=table.plot(data, headers, pconfig={
            "id": "multiqc_software_versions", 
            "col1_header": mqc_config.versions_table_group_header
        }),
    )
```

pontushojer:
I also tried to make this work while keeping the style of the old table. 

```python
  def report_software_versions(self):
      """Create section listing software versions."""
      headers = {}
      headers["group"] = {
          "title": mqc_config.versions_table_group_header,
          "description": "Software group designation",
      }
      headers["software"] = {
          "title": "Software",
          "description": "Software name",
      }
      headers["version"] = {
          "title": "Version",
          "description": "Software version(s). Multiple versions are comma separated",
          "scale": False,
      }

      data = {}
      row_i = 0
      for group, tmp_versions in sorted(mqc_report.software_versions.items()):
          for software, versions in sorted(tmp_versions.items()):
              row_i += 1
              data[row_i] = {
                  "group": group,
                  "software": software,
                  "version": f"<code>{'</code>, <code>'.join(map(str, versions))}</code>"
              }

      pconfig = {
          "col1_header": "",
      }
      table_html = table.plot(data, headers=headers, pconfig=pconfig)
      
      self.add_section(name=None, content=table_html)
```
image

I don't really like the solution. Also comparing to @vladsavelyev solution I think I prefer the old table. It is cleaner and more readable. Also any solution with `table.plot()` would have to define something as a "sample". It would then be subject to the rename, hide and highlighting options in the toolbox, which is not really appropriate for this table. I think something bespoke might acctually be better suited in this case.

One thing I do like about these solutions is the `Copy table` button, which could be quite useful. But for this we could just add this button to the custom html solution. It would then look like this:
image

ewels:
ok, all valid points - you've convinced me 👍🏻  Let's add the copy button, then I'm happy.

multiqc/multiqc.py
```diff
@@ -746,6 +751,13 @@ def __rich_measure__(self, console: rich.console.Console, options: rich.console.
         report.runtimes["mods"][run_module_names[mod_idx]] = time.time() - mod_starttime
     report.runtimes["total_mods"] = time.time() - total_mods_starttime
 
+    # Update report with software versions provided in configs
+    software_versions.update_versions_from_config(config, report)
+
+    # Add section for software versions if any are found
+    if not config.skip_versions_section and report.software_versions and len(report.modules_output) > 0:
```
ewels, line 758: I'm not sure that we need to check that there is module output - why not let people generate a MultiQC report with _only_ software versions? (Presumably from `*_mqc_versions.yml` files?)

ewels: Changed in 23f55a03

ewels:
Ok, final points I'd like (first two are functionality from `table.plot()` - adding them myself now.

- [ ] We should save the resolved versions to `multiqc_data` as a file for downstream use
- [ ] Add copy button
- [ ] I would like to drop the Group column of the software table if it exactly matches Software

On the second point, my thinking is that many / most tools will only report one version. So the table will often end up looking like this for many users:
CleanShot 2023-09-14 at 11 19 58@2x

In which case, we can clean it up a bit:
CleanShot 2023-09-14 at 11 22 02@2x

Let me know ASAP if this is a bad idea.

docs/core/reports/customisation.md
ewels: Just saw this... I don't really mind at this point, but I do want to get this merged 😆 

Whilst hypocritical compared to some stuff I've said recently, I'm happy enough. Let's leave as-is and merge.

ewels (approved):
Let's do this!

Amazing work and patience @pontushojer and thanks for the support @vladsavelyev. This was a big one!

Very cool to get this in place after thinking about it for so many years!

pontushojer:
Awesome! Thanks for adding the final improvements @ewels, I think all of them are good.

Issue #1936, opened 2023-06-16
Request for support of Sentieon-Dedup


asp8200:
### Name of the tool

Sentieon Dedup

### Tool homepage

https://support.sentieon.com/manual/usages/general/#dedup-algorithm

### Tool description

The Dedup algorithm performs the marking/removing of duplicate reads.

### Tool output

[test.md.cram.metrics.zip](https://github.com/ewels/MultiQC/files/11771023/test.md.cram.metrics.zip)

### Data suitable for MultiQC plot(s)

I imagine that the metrics from Sentieon-Dedup should be displayed pretty much like the corresponding metrics from GATK's Markduplicates.

Here is an example of how the first part of the metrics-file from Dedup may look:

```
#SentieonCommandLine: /opt/sentieon/sentieon-genomics-202112.06/libexec/driver -t 2 -i test.paired_end.sorted.bam -r genome.fasta --algo Dedup --score_info test.md.score --metrics test.md.cram.metrics test.md.cram
LIBRARY	UNPAIRED_READS_EXAMINED	READ_PAIRS_EXAMINED	SECONDARY_OR_SUPPLEMENTARY_RDS	UNMAPPED_READS	UNPAIRED_READ_DUPLICATES	READ_PAIR_DUPLICATES	READ_PAIR_OPTICAL_DUPLICATES	PERCENT_DUPLICATION	ESTIMATED_LIBRARY_SIZE
testN	0	2820	2	2	0	828	0	0.293617	3807

#HISTOGRAM
BIN	VALUE
1.0	0.999986
2.0	1.476740
3.0	1.704038
...
```

And here is the complete set of metrics from GATK's markduplicates for the same sample:
```
## htsjdk.samtools.metrics.StringHeader
[...]

Issue #1180, opened 2020-05-07
Picard outputs from Sentieon not parsed


ewels:
_Pasted from an email from @chrispyatt:_

I am trying to pass some Picard metrics files to MultiQC (GCBias, AlignmentStat, InsertSize, MeanQualityByCycle, Duplication, and QualDistribution) but only the Duplication metrics are being found by MultiQC.

The config file I provide contains the section below, which matches the files I am trying to use, and the multiQC log confirms that these patterns are recognised.

```yaml
sp:
    fastqc/data:
        fn: '*.stats-fastqc.txt'
    picard/alignment_metrics:
        fn: '*.AlignmentStat_metrics.txt'
        shared: true
    picard/insertsize:
        fn: '*.InsertSize_metrics.txt'
        shared: true
    picard/markdups:
        fn: '*.Duplication_metrics.txt'
        shared: true
    picard/gcbias:
        fn: '*.GCBias_metrics.txt'
        shared: true
    picard/quality_by_cycle:
        fn: '*.MeanQualityByCycle_metrics.txt'
        shared: true
    picard/quality_score_distribution:
        fn: '*.QualDistribution_metrics.txt'
        shared: true
```

The files were generated by Sentieon, but apart from the Sentieon header they seem identical in format to the Picard specs. Also the Duplication module seems to work so I think the file contents are ok. For example the GCBias file begins with:

```
#SentieonCommandLine: /usr/local/sentieon-genomics-201911/libexec/driver -t 36 -i sorted_0.bam -i sorted_1.bam -r genome/hs37d5.fa --algo GCBias --summary gc_summary.txt gc_metric.txt
ACCUMULATION_LEVEL	READS_USED	GC	WINDOWS	READ_STARTS	MEAN_BASE_QUALITY	NORMALIZED_COVERAGE	ERROR_BAR_WIDTH	SAMPLE	LIBRARY	READ_GROUP	
All Reads	ALL	0	147859	1034	19	0.182258	0.005668
```

I presume the `find_log_files` function within the modules is not returning anything, but I cannot figure out why. Is there something else in the config file that I need to specify, or some other aspect that I have overlooked? Any insight into this issue would be most appreciated.

ewels:
Hi @chrispyatt,

[...]

PR #1943, opened 2023-06-25
New module: Sentieon Dedup Metrics

asp8200:
Adding support for Sentieon Dedup Metrics (corresponding to picard MarkDuplicates)

Issue #1936 

As far as I could tell, we can't just call the MarkDuplicates MultiQC-module for the Sentieon-Dedup (as is done in the 
 [biobambam2/bamsormadup MultiQC-module ](https://github.com/asp8200/MultiQC/blob/sentieon_dedup/multiqc/modules/biobambam2/biobambam2.py)), since, for instance, parsing of the Sentieon-command is different from the parsing of the MarkDuplicates command. 

image

This is how the Sentieon Dedup section looks if MultiQC findes to DedupMetrics-files:
image
 
The Sentieon Dedup Metrics are displayed just like the corresponding metrics from Picard MarkDuplicates, and the module code was copied and adjusted from the Markduplicates MultiQC-module.

- [x] There is example tool output for tools in the <https://github.com/ewels/MultiQC_TestData> repository : [data/modules/sentieon/issue_1180/P150_PE.st.metrics.DeDup.txt](https://github.com/asp8200/MultiQC_TestData/blob/master/data/modules/sentieon/issue_1180/P150_PE.st.metrics.DeDup.txt)
- [x] `docs/modules/sentieon.md` was already present but is now updated.

### DIFF
--- a/multiqc/utils/search_patterns.yaml
+++ b/multiqc/utils/search_patterns.yaml
@@ -730,6 +730,10 @@ sargasso:
 sentieon/alignment_metrics:
   contents: "--algo AlignmentStat"
   shared: true
+sentieon/dedup:
+  contents: "--algo Dedup"
+  num_lines: 3
+  shared: true
 sentieon/insertsize:
   contents: "--algo InsertSizeMetricAlgo"
   shared: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@
   - Added report section listing samples that had no adapters trimmed
 - **RSeQC**
   - Fix `ZeroDivisionError` error for `bam_stat` results when there are 0 reads ([#1735](https://github.com/ewels/MultiQC/issues/1735))
+- **Sentieon**
+  - Added support for metrics from Sentieon's DNAseq function Dedup (corresponding the Picard's Markduplicates). ([#1936](https://github.com/ewels/MultiQC/issues/1936))
 - **UMI-tools**
   - Fix bug that broke the module with paired-end data ([#1845](https://github.com/ewels/MultiQC/issues/1845))
 
--- a/docs/modules/sentieon.md
+++ b/docs/modules/sentieon.md
@@ -2,7 +2,7 @@
 name: Sentieon
 url: https://www.sentieon.com/products/
 description: >
-  Sentieon-dnaseq produces many outputs. This module deals with 3 Picard
+  Sentieon-dnaseq produces many outputs. This module deals with four Picard
   equivalents which do not transfer well to MultiQC. The code for each script
   is split into its own file and adds a section to the module output if
   logs are found.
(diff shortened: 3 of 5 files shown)


multiqc/utils/search_patterns.yaml
```diff
@@ -730,6 +730,10 @@ sargasso:
 sentieon/alignment_metrics:
   contents: "--algo AlignmentStat"
   shared: true
+sentieon/dedup:
+  contents: "--algo Dedup"
+  num_lines: 3
+  shared: true
```
ewels, line 736: If these logs can be shared, then setting `num_lines` doesn't really make sense - if the output was appended half way down an existing file then it won't be in the first 3 lines.

If Sentieon always creates these files then remove `shared` and keep `num_lines` (maybe also for the other blocks above and below?). If a likely use case is that the output from multiple samples / tools could be appended to a single file, remove `num_lines`

asp8200: Thanks for you prompt reply, @ewels. I probably didn't (don't) understand the purpose of `shared` and `num_lines` completely 😬 The new Sentieon-Dedup MultiQC-module only uses the first three lines from the sentieon-dedup-metrics files, that is why I set `num_lines` to three.  

Concerning the option `shared`, I think I looked at this:

https://github.com/ewels/MultiQC/blob/7129bd10420149c032e380bef57f9c9ca0044d33/docs/core/development/modules.md?plain=1#L381-L382

and I figured if some other MultiQC-module wanted to use the same log-file, then fine. Moreover, the other Sentieon-MultiQC-modules had `shared` set to `true` so I also just did that.

On re-reading your explanation and the docs, I get the impression that `shared` is actually more a way of having MultiQC handle the case where several pipeline-tools write to the same log/metrics/summary-file. If that is indeed the case, then I guess `shared` should be removed from all the Sentieon MultiQC-modules in the `search_patterns.yaml` 🤔

asp8200: @ewels : So I should just the field `num_lines` and `shared` for `sentieon`?

vladsavelyev:
Thank you a lot @asp8200 for raising the issue and for the contribution! We thought more about this as well as other PRs that request adding Sentieon modules (e.g. https://github.com/ewels/MultiQC/issues/1180), and decided to instead refactor the Picard module directly to support Sentieon: https://github.com/ewels/MultiQC/pull/2110

Now one can run MultiQC on Sentieon outputs (like the one you shared) and they will be interpreted as corresponding Picard outputs.

We restructured the test data as well, and whenever we have a test example for Sentieon tool corresponding to an existing Picard tool, we added that example into a subfolder, e.g. `sentieon_issue_1180` here: https://github.com/ewels/MultiQC_TestData/tree/master/data/modules/picard/MarkDuplicates 

Let me know if that works for you, and if you have other ideas on what Sentieon outputs to add and test 😌

PR #1946, opened 2023-07-04
new module: sincei

vivekbhr:
Added **sincei** module to summarise read and count-level single-cell QC metrics from [sincei](https://github.com/vivekbhr/sincei)


multiqc/modules/sincei/scFilterStats.py
```diff
@@ -0,0 +1,197 @@
...
+                )
+                return dict()
+
+            s_name = self.clean_s_name(cols[0], f)
+            if s_name in d:
+                log.debug("Replacing duplicate sample {}.".format(s_name))
+            d[s_name] = dict()
+
+            try:
+                d[s_name]["total"] = self._int(cols[1])
```
vladsavelyev, line 176: 1. It seems like the column only has integer values, so can just call `int()` here?

multiqc/modules/sincei/scFilterStats.py
```diff
@@ -0,0 +1,197 @@
...
+                d[s_name]["filtered"] = float(cols[2])
+                d[s_name]["blacklisted"] = self._int(cols[3])
```
vladsavelyev, line 178: This time all the values look like ratios, so rounding to an integer will lose all precision. So perhaps should just `float()` here?

multiqc/modules/sincei/scFilterStats.py
```diff
@@ -0,0 +1,197 @@
...
+            if len(cols) != 14:
+                # This is not really the output from scFilterStats!
+                log.warning(
+                    "{} was initially flagged as the tabular output from scFilterStats, but that seems to not be the case. Skipping...".format(
+                        f["fn"]
+                    )
+                )
+                return dict()
+
+            s_name = self.clean_s_name(cols[0], f)
```
vladsavelyev, line 170: Since we have the header, perhaps it would be more robust to use the header to index values - `cols['Cell_ID']` instead of `cols[0]`, and so on. Would be more robust in case if new columns are inserted into future versions

multiqc/modules/sincei/scFilterStats.py
```diff
@@ -0,0 +1,197 @@
...
+        self.sincei_scFilterStats = self.ignore_samples(self.sincei_scFilterStats)
+
+        if len(self.sincei_scFilterStats) > 0:
+            # Write data to file
+            self.write_data_file(self.sincei_scFilterStats, "sincei_read_filtering")
+
+            header = OrderedDict()
+            header["N Entries"] = {"title": "N entries", "description": "Number of entries sampled from the file"}
+            header["pct_Aligned"] = {
+                "title": "% Aligned",
```
vladsavelyev, line 35: For percentages, good to use `"suffix": "%"`

vladsavelyev (dismissed):

First, a minor comment, I'd try to refactor to not to assume the order for columns (i.e. indexing them with numbers like `cols[0]`, `cols[1]`). Since the header is available, I'd rather use `csv` to parse the file `reader = csv.DictReader(open(filename), delimiter=",")` and index with the column name (e.g. `cols['Cell_ID']`, `row["# Reads"]`, etc), to make it more robust to future changes of columns and their order. And probably use same column name in MultiQC table whenever reasonable.

A bigger concern is that I feel like using cells as samples is not what MultiQC was designed for. The generated beeswarm plot already hangs my browser even on the test data (I know the beeswarm is really inefficient at the first place and the problem will probably go away with a future new plotting library; but not until we add more data). MultiQC is designed to work with summarised per-sample metrics, and I don't think cells can be counted as samples, as there are way too many of them, and feels too raw to be fed into MultiQC, I think.

It would be good if more summarised per-samples stats can be produced here, e.g. with some different tool downstream.

ewels (dismissed):
Thanks for the PR @vivekbhr! As @vladsavelyev, I think that we need to rethink what data goes into the report here. The docs have a section called [Don’t add everything](https://multiqc.info/docs/development/modules/#dont-add-everything) which I think applies. MultiQC should only summarise, but here we're really just reporting on raw data.

The TSV files have a column called `sample` (eg. `sortChIC-BM-SL1-k4me1-1`) so maybe we could summarise stats for all cells under a given sample identifier?

vivekbhr:
Thanks a lot for the review and comments @vladsavelyev and @ewels . I understand what you say, I experienced the slow response on the output report myself when using bigger data, so summarizing per-sample stats, such as a sample-wise range and median value for each field could be a better way. 

Do you think Its OK to do these calculations from the output .tsv file in the multiqc module, or are you saying that this output should already be reported by sincei? I think calculating sample-wise range and median should not be a big overhead for the multiqc module.

vladsavelyev:
> Do you think Its OK to do these calculations from the output .tsv file in the multiqc module, or are you saying that this output should already be reported by sincei? I think calculating sample-wise range and median should not be a big overhead for the multiqc module.

I think it should be totally alright to summarise it in MultiQC!

vivekbhr:
@vladsavelyev I've finally found time to complete the changes as requested. The multiqc module now reports sample summaries (median values), instead of per-cell summary. Also, the column headers are used instead of integer positions. I hope the changes are satisfactory now.

Best,
Vivek

vivekbhr:
@vladsavelyev @ewels I have made all the requested changes, but the merging is still blocked with "changes requested". Please let me know if there are still unresolved issues. Thanks!

ewels:
Thanks @vivekbhr! We'll get back to this to review the updated code as soon as we can.

vivekbhr:
@ewels @vladsavelyev Did you find some time at the end to review the updates?

vivekbhr:
@ewels @vladsavelyev I made the fixes that we discussed but the module is still in review with "changes requested". Now the code in the master branch is restructured and the beeswarm plot (which I was using) seems to be gone. Is there still hope for my module to be merged in multiqc?

vladsavelyev:
@vivekbhr, the beeswarm is replaced with [the violin plot](https://multiqc.info/docs/development/plots/#violin-plots), the usage is the same. Other changes in the master branch shouldn't affect the module as well, apart from some formatting to make the 3rd party linting tools like `ruff` and `mypy` happy.

I'd be happy to review the code again and adjust it correspondingly, if you din't mind allowing commits into the PR branch from maintainers (there should be a [tickbox](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)).

vivekbhr:
@vladsavelyev Thanks. That box is already ticked!

vivekbhr:
@vladsavelyev anything we can do on our side to accelerate this merge?

ewels:
Hi @vivekbhr,

Apologies for the very very long delay on this PR. Keeping up with contributions is tough with MultiQC, but this one is a little extreme. I'm trying to get it tidied up and merged + released.

I'm nearly there now I think, but I was wondering why you wanted it as violin plots rather than regular tables? (which convert to violin plots automatically when you hit a certain number of samples). Note that all tables have a toggle button to switch to violin plot and back now anyway.

I think I will switch it to tables if that's ok, just to get it in and merged. If you have a strong preference for violin plots then let me know and we can always switch it back again.

Phil

ewels (approved):
Let's do it! Thanks!

PR #1958, opened 2023-07-24
Update rna quant metrics for dragen 4.2

alexiswl:
Dragen 4.2 uses suffix `.quant_metrics.csv` as opposed to previous versions `.quant.metrics.csv`.


ewels (approved):
Thanks!

This should probably be done outside of the module using sample name cleaning patterns, instead of a regex in the code to be honest. Then it would be configurable by the end user. But we can leave that for another day.

alexiswl:
Ah sorry @ewels looks like I didn't commit the search patterns file for this one. Adding it in now under a separate PR

PR #1959, opened 2023-07-25
new module Bakta

louperelo:
`Bakta` is a tool for rapid and standardized annotation of bacterial genomes, MAGs and plasmids. We are using `Bakta` as one of the annotation tool options in `nf-core/funcscan` and would like to include its output in the pipelines MultiQC report.

  - [x] There is example tool output for tools in the <https://github.com/ewels/MultiQC_TestData> repository: [#258](https://github.com/ewels/MultiQC_TestData/pull/258)


multiqc/modules/bakta/bakta.py
```diff
@@ -0,0 +1,151 @@
...
+            anchor="bakta",
+            href="https://github.com/oschwengers/bakta",
+            info="Rapid & standardized annotation of bacterial genomes, MAGs & plasmids",
+            doi="10.1099/mgen.0.000685",
+        )
+
+        # Parse logs (txt files)
+        self.bakta = dict()
+        for f in self.find_log_files("bakta"):  # , filehandles=True):
+            # self.parse_bakta(f)
```
vladsavelyev, line 29:
```suggestion
```

multiqc/modules/bakta/bakta.py
```diff
@@ -0,0 +1,151 @@
...
+            name="Bakta",
+            anchor="bakta",
+            href="https://github.com/oschwengers/bakta",
+            info="Rapid & standardized annotation of bacterial genomes, MAGs & plasmids",
+            doi="10.1099/mgen.0.000685",
+        )
+
+        # Parse logs (txt files)
+        self.bakta = dict()
+        for f in self.find_log_files("bakta"):  # , filehandles=True):
```
vladsavelyev, line 28:
```suggestion
        for f in self.find_log_files("bakta"):
```

louperelo:
Thank you for the review @vladsavelyev! Please check if I resolved the merge conflict for the CHANGELOG.md and README.md correctly.

multiqc/utils/search_patterns.yaml
```diff
@@ -13,6 +13,10 @@ afterqc:
 anglerfish:
   fn: "*.json"
   contents: "anglerfish_version"
+bakta:
+  fn: "*.txt"
+  contents: "Bakta:"
+  num_lines: 26
```
vladsavelyev, line 19:
```suggestion
```
Is there a chance that in some circumstances or in future Bakta versions there will be more metrics reported, and the "Bakta:" line will be pushed below the line 26? Perhaps makes sense to read more lines to be safe. Let's maybe remove it and just make it read the default 1000 lines

multiqc/modules/bakta/bakta.py
```diff
@@ -0,0 +1,150 @@
...
+        # Write parsed report data to a file
+        self.write_data_file(self.bakta, "multiqc_bakta")
+
+        # Add most important Bakta annotation stats to general table
+        headers = OrderedDict()
+        headers["Count"] = {
+            "title": "# contigs",
+            "description": "Number of contigs",
+            "min": 0,
+            "format": "{:i}%",
```
vladsavelyev, line 49: But it's not a percentage? Also, I don't think {:i} is a recognised format string. At least it doesn't seem to work, as numbers are displayed in the table as string (you can tell that by the lack of thousand groups separators!). I'll change them to `"{:,d}"` if that's okay.

multiqc/modules/bakta/bakta.py
```diff
@@ -0,0 +1,150 @@
...
+            raise UserWarning
+
+        log.info("Found {} logs".format(len(self.bakta)))
+
+        # Write parsed report data to a file
+        self.write_data_file(self.bakta, "multiqc_bakta")
+
+        # Add most important Bakta annotation stats to general table
+        headers = OrderedDict()
+        headers["Count"] = {
```
vladsavelyev, line 45: It's good to `"scale"` into the table columns config, so the columns have differrent colors ([see docs](https://multiqc.info/docs/development/modules/#step-3---adding-to-the-general-statistics-table))
https://multiqc.info/docs/development/modules/#step-3---adding-to-the-general-statistics-table

vladsavelyev (approved):
@louperelo - thanks for responding to the review quickly! Last time I didn't complete it, just left a few quick suggestions 😅  Now it's done. I mostly merged my suggestions directly into the branch, hope it's okay. Here is what I changed:
* fixed the format strings to `"format": "{:,d}"` 
* table [colors](https://multiqc.info/docs/development/modules/#colour-matters) - I tried to make them match similar metrics in the QUAST module, plus added a `shared_key: gene_count` for the CDS count to make the number scale comparable with QUAST
* got rid of spaces in the metric keys - it's not a problem for MultiQC, but theoretically might be some downstream tools that might use MultiQC output
* iterating over `f["contents_lines"]` to avoid extra splitting
* in search patterns, removed `num_lines: 26` to make it more future-proof

ewels (approved):
LGTM, thanks!

PR #1968, opened 2023-08-04
New submodule for GATK: AnalyzeSaturationMutagenesis

odcambc:
Adds support for a submodule for GATK: https://gatk.broadinstitute.org/hc/en-us/articles/360037594771-AnalyzeSaturationMutagenesis-BETA-

This submodule parses the sample.readCounts file, which enumerates how reads (and bases) are being handled by the tool. Two plots and one table are added. The plots provide summaries of reads and bases, while the table provides more granular information if columns are unhidden.

- [x] There is example tool output for tools in the <https://github.com/ewels/MultiQC_TestData> repository: [#261](https://github.com/ewels/MultiQC_TestData/pull/261)
- [x] `docs/modulename.md` is updated
   - The default is only 4 columns displayed
   - The hidden columns are important for troubleshooting bad samples, and contain important information


multiqc/utils/search_patterns.yaml
```diff
@@ -351,6 +351,8 @@ gatk/varianteval:
 gatk/base_recalibrator:
   contents: "#:GATKTable:Arguments:Recalibration"
   num_lines: 3
+gatk/analyze_saturation_mutagenesis:
+  fn: "*.readCounts"
```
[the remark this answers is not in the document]
vladsavelyev: Seems like that's the expected output file name: https://github.com/broadinstitute/gatk/blob/master/src/main/java/org/broadinstitute/hellbender/tools/AnalyzeSaturationMutagenesis.java#L541

Though it still feels pretty generic and there might be other tools using this suffix, so maybe we should check the file contents?

multiqc/modules/gatk/analyze_saturation_mutagenesis.py
```diff
@@ -0,0 +1,463 @@
...
+            "namespace": "GATK",
+        }
+        asm_headers["wt_reads_disjoint"] = {
+            "title": "WT (disjoint) (M)",
+            "description": "WT reads called from disjoint pairs (millions)",
+            "min": 0,
+            "scale": "Purples",
+            "shared_key": "disjoint_read_count",
+            "hidden": True,
+            "modify": lambda x: x / 1000000,
```
ewels, line 302: It would be nice to make all of these user-configurable as well, using a module-specific config.

ewels (approved):
Looking great, thanks very much! Just the minor addition of the configs that would be a nice-to-have. @vladsavelyev is looking into that now.

multiqc/modules/gatk/analyze_saturation_mutagenesis.py
vladsavelyev: I experimented with it a bit and realised it makes it difficult to read to the table when different prefixes are used for different read counts. E.g. you compare 42 to 42, but in one column it's 42 reads, and in the other one it's 42M reads. So I'd stick to `read_count_*` everywhere - it's already configurable.

ewels: Sounds good 👍🏻 

Nice compromise actually, as the different `shared_key` strings are there, so the background bar scales will still be grouped, even if the units are shared.

PR #1969, opened 2023-08-09
Add . or _ to search patterns for quant metrics csv

alexiswl:
Patch from #1958

### DIFF
--- a/multiqc/utils/search_patterns.yaml
+++ b/multiqc/utils/search_patterns.yaml
@@ -303,7 +303,7 @@ dragen/trimmer_metrics:
 dragen/time_metrics:
   fn: "*.time_metrics.csv"
 dragen/rna_quant_metrics:
-  fn: "*.quant.metrics.csv"
+  fn: "*.quant[._]metrics.csv"
 dragen/rna_transcript_cov:
   fn: "*.quant.transcript_coverage.txt"
 dragen/sc_rna_metrics:


ewels:
Hi @alexiswl - did you test this and it's definitely working? `fn` is a glob, not regex. I always have to google syntax but off the top of my head I thought it was squiggly brackets for this? But maybe square brackets work for glob syntax too..

vladsavelyev:
@ewels I tested it on my side and it worked well, but I agree that it would be nice to add test examples into `MultiQC_TestData` for `dragen/rna_quant_metrics` module. 

Also, it looks like the `parse_time_metrics_file` method was copied from `time_metrics.py`, I think it should be renamed, and the example data in the docstring should be updated. I'll do that now, but agree it would also be nice to refactor the Dragen modules to use cleaning patterns.

ewels:
Sounds good, thanks! Yeah a lot of the DRAGEN code is quite non-standard. Makes maintenance a bit of a headache..

alexiswl:
> Hi @alexiswl - did you test this and it's definitely working? fn is a glob, not regex. I always have to google syntax but off the top of my head I thought it was squiggly brackets for this? But maybe square brackets work for glob syntax too..

I just ran a python regex over it and also used the same syntax as https://github.com/alexiswl/MultiQC/blob/bc4f724abf9e60deb0e335c57a6d76ef284a4845/multiqc/utils/search_patterns.yaml#L303-L306

PR #1970, opened 2023-08-09
Replace 3 files (`__init__.py`, `multiqc.py`, `__main__.py`) with 2  (`__init__.py` and `__main__.py`).

vladsavelyev:
I noticed that there are 2 ways to run MultiQC: 

1. as a command line script (via the entry point in `__main__.py`),
2. and interactively from notebooks (i.e. import the `multiqc` module and call `run()`, defined in `multiqc/__init__.py`).

Looks like it makes sense to simplify it a bit and keep only `__init__.py` + `__main__.py`, and drop `multiqc.py`. It will also simplify the user namespace a bit. 

I'd even rename `__main__.py` to `main.py` to make it more clear to newcomers to find the entry point script).

Not sure of any caveats, let me know if I'm missing something. I'll test it more, that's why it's a draft PR.


ewels:
I think the dunder `__main__.py` filename is special, like `__init__.py`, so I'm not sure that we should rename that.

I've forgotten the the specifics now, but a speed google finds this: https://www.geeksforgeeks.org/usage-of-__main__-py-in-python/amp/ - can probably find more later.

ewels:
I _think_ it's three files because historically it started off as `multiqc.py` then `__main__.py` came later for Windows users (?) who wanted to run it as `python -m multiqc` 

...or something like this anyway.

Issue #1873, opened 2023-03-03
"Export Plots" functionality only downloads the first 10 plots.


robsyme:
When using the "Export" tab in the right-hand sidebar and more than ten plots are selected for export, only the first 10 plots are downloaded.

I have tried this on both Chrome and Safari, so I don't think that this is a browser limitation.

Example report: [5_multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/10886096/5_multiqc_report.html.zip)

ewels:
How has no-one ever reported this before? 🤯

vladsavelyev:
After playing with this a bit, it looks like it's a browser limitation after all. It seems to behave differently across browsers though. 

Here is a toy HTML page with one button that generates and downloads 20 files. In Chrome, for me, it would download only the first 10 files (`sample_1.txt through `sample_10.txt`). Safari though downloads only one last file (`sample_20.txt`).

```
<!DOCTYPE html>
<html>
<head>
  <title>Test downloads</title>
</head>
<body>
  <button id="downloadButton">Download Files</button>

  <script>
    document.getElementById("downloadButton").addEventListener("click", function() {
      for (let i = 1; i <= 20; i++) {
        const content = `sample_${i}`;
        const fileName = `sample_${i}.txt`;

        const blob = new Blob([content], { type: "text/plain" });
        const link = document.createElement("a");
        link.href = URL.createObjectURL(blob);
        link.download = fileName;
        link.click();
      }
    });
  </script>
</body>
</html>
```

Though after multiple attempts, Chrome suddenly learned to download all 20 at the same time, so it's not very reproducible.

One obvious solution would be to put all images into a zip archive and just download that archive instead. I tried to dig through the HighCharts codebase to find a more or less elegant solution for that with no success. I think since we will be looking for a HighCharts replacement, probably it's not worth it to hack it too much.

Another idea is to add a small timeout between each batch of 10 downloads. The problem is that it would require the user to confirm each batch in the browser, which is not super user-friendly, and also, quiet hacky.

So no solution for now, let's discuss options!

ewels:
I think zipping if there are > 8 plots or something could be good. Pretty sure I've written in-browser JS creation of zip files in the distant past at some point, feel like it should be possible..? 🤔 

I'm not sure that a new plotting library will solve this issue, as we will still want the ability for people to export multiple plots in one go.

Issue #1789, opened 2022-10-30
Replace plotting library


ewels:
HighCharts has served MultiQC very well, but it's time to move to something else. Key issues are:

- Limited numbers of plot types
- Requirement to write back-end plotting code in matplotlib manually to mimic appearance of front-end plots

Any new plotting library:

- Must have a permissive licence (eg. MIT)
- Should have a native Python implementation
- Should be able to generate static plots directly from the Python run time
- Should have a good degree of theme customisation options
- Should be extensible in the javascript front-end (eg. for highlight / hide / renaming tools)
- Should be compatible with MultiQC's efforts to compress data for small report sizes
- Work equally well in Jupyter notebooks with interactive use

The current top runner is [Plotly](https://plotly.com/), though [Vega-Altair](https://altair-viz.github.io/) could also be a good option (especially due to the open Vega grammar being good for reuse).

multimeric:
Agreed with the two contenders. Plotly has the advantage of offering a very large library of plots, including some bioinformatics specific things (dash-bio perhaps?), and having an enormous user base. On the other hand, Vega is an open standard. One advantage of this is that it's not dependent on the whims of Plotly, who have been a bit weird about the license for the library. This also means that plot config is standardised and machine readable, so I can imagine a scenario where we export the plot config and plot data from MultiQC, and provide it to a third party renderer so that it integrates into someone's existing dashboard, instead of needing to embed HTML. This also allows us to swap out our own rendering library from Altair to something else that supports Vega in the future, without breaking everything.

ewels:
> Plotly, who have been a bit weird about the license for the library.

Ah really? I wasn't aware of this. Could you elaborate?

multimeric:
Sorry I was misremembering. It wasn't the license (which was always MIT), but rather the way that they uploaded your plots to their online service by default (or something like that). It's briefly explained here:
https://medium.com/plotly/plotly-py-4-0-is-here-offline-only-express-first-displayable-anywhere-fc444e5659ee.

These days they're better about delineating the commercial from open source parts, however.

ewels:
If we do go for Plotly and also start using React then this looks like a pretty nice setup: https://plotly.com/javascript/react/

vladsavelyev:
Plotly is a great choice. Glad that there is no more confusion about online/offline and commercial/free versions as it was in the past. Also nice that plotly-express is bundled with the main library.

[...]

PR #1976, opened 2023-08-11
When exporting plots, make a more clear error message for unsupported FastQC dot plot

vladsavelyev:
Some refactoring while debugging https://github.com/ewels/MultiQC/issues/1873

Per-base content FastQC plot is not supported for export, and currently, there are alert messages shown to warn the user about that. I think one quick improvement here is to specify what type of plot exactly is failing, in case the user would export multiple plots at once.

Further improvements would be to perhaps replace the alert with a more friendly inline warning, or even to fix the export for this plot. But since we will likely be moving away from Highcharts (https://github.com/ewels/MultiQC/issues/1789), it's probably not worth refactoring it much for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Removed `simplejson` unused dependency ([#1973](https://github.com/ewels/MultiQC/pull/1973))
 - Give config `custom_plot_config` priority over column-specific settings set by modules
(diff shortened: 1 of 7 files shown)


ewels (approved):
LGTM, just a minor changelog entry please 👍🏻  then feel free to merge.

Issue #1883, opened 2023-03-15
multiqc errors when running viralrecon with samples with specific naming conventions


rrdavis77:
### Description of bug

Hi Phil,

I am having some issues with multiqc when running within viralrecon.  The issue seems to happen when multiqc is looking at files `*.variant_counts_mqc.tsv` which are generated from this script in viralrecon: [`ivar_variants_to_vcf.nf`](https://github.com/nf-core/viralrecon/blob/3ee1fe98fdf17a80922aa8cf4da4afaf483f3429/modules/local/ivar_variants_to_vcf.nf)

I don't think the files themselves are bad but it seems to be name of the samples that causes the error.
I tested the output with two samples named:

* `4056145068`
* `4056145117`

The pipeline completes without any errors. If I change the name of one of the samples to start with FS then the pipeline errors:

* `4056145068`
* `FS56145117`

Nothing changed except the name of the samples in the samplesheet.

Here is the original issue on viralrecon: https://github.com/nf-core/viralrecon/issues/345

### File that triggers the error

[variant_counts_mqc.zip](https://github.com/ewels/MultiQC/files/10984406/variant_counts_mqc.zip)

### MultiQC Error log

```console
NXF_VER=22.10.1 nextflow run nf-core/viralrecon --input samplesheet_all_samples_2.csv --platform illumina --genome 'NC_045512.2' --protocol amplicon --primer_bed /share/sequencing_effort/viralrecon/primer_files/NC_045512.2.v2.primers.bed --primer_fasta /share/sequencing_effort/viralrecon/primer_files/NC_045512.2.v2.primers.fna --skip_assembly -profile singularity --primer_left_suffix F --primer_right_suffix R --kraken2_db /share/sequencing_effort/viralrecon/krakendb/kraken2_human.tar.gz --ivar_trim_offset 5 -c /share/sequencing_effort/viralrecon/viralrecon_dev.config --outdir /share/Viralrecon_results/test_dev/ -r dev -resume --skip_asciigenome --skip_snpeff --skip_kraken2 --skip_pangolin --skip_nextclade


INFO:    Convert SIF file to sandbox...
                | v1.14
  /// MultiQC �
|           multiqc | Only using modules: custom_content, fastqc, fastp, kraken, bowtie2, samtools, mosdepth, bcftools, snpeff, quast, pangolin, cutadapt
|           multiqc | Search path : /share/Viralrecon_results/test_v2.5/work/49/2146dc958fadac4338a04b987fc007
[...]

PR #1977, opened 2023-08-11
Fix sample names convert to float in custom content

vladsavelyev:
Fix https://github.com/ewels/MultiQC/issues/1883 by keeping sample names as strings in TSV custom content parsing.

Long term, refactoring to use a more type-safe approach would help to avoid such problems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 
 ### Module updates
 
+- **Custom content**
+  - Don't convert sample IDs to floats ([#1883](https://github.com/ewels/MultiQC/issues/1883))
 - **Kraken**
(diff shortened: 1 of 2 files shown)


ewels (approved):
Code change LGTM, thanks 👍🏻  Just needs a changelog entry, then can merge.

ewels:
> Long term, refactoring to use a more type-safe approach would help to avoid such problems.

Agreed - this error has come up a bunch of times in the past with other modules as well. It'd be nice to have a better global solution to this. But can tackle that in later refactoring.

ewels:
Added changelog entry and pushed example data to the testdata repo - merging. Thanks @vladsavelyev !

PR #1981, opened 2023-08-15
WhatsHap: ensure that TSV is only split on tab character. Allows sample names with spaces

caspargross:
This pull requests ensures that the whatshap module splits the TSV datafile only on 'tab' characters. This allows to run with samplenames containing white spaces, which lead to errors previously.


vladsavelyev:
Thank you @caspargross!

Just a note, when contributing a bug fix, try to also add a test example that reproduced this bug into MultiQC_TestData, like here: https://github.com/ewels/MultiQC_TestData/pull/266

Thank you for adding a clear line to the change log. Try also to name the PRs similarly to the change log line you used, this way the commit history looks cleaner :)

vladsavelyev (approved):
Thank you!

Issue #1919, opened 2023-05-04
Remove pre-Python 3.6 style code


ewels:
The MultiQC code base has a bunch of stuff written for Python 2 that we no longer need. We should modernise this for a cleaner, more readable, better code strucutre.

Examples, but not limited to:

- [x] Using `OrderedDict`s everywhere when we don't need to
- [x] `__future__` statements
- [x] `encode` and other pre-unicode hacks
- [x]  Format strings (run `flynt`)
- ...

vladsavelyev:
I think those creeping back shouldn't happen often, and we can handle that manually when reviewing the old PRs. If it becomes a problem, I'd add a check.

PR #1982, opened 2023-08-15
Add mosdepth XY coverage plot and contig filtering options

### LINKED ISSUE #1978: Reproduce samtools idxstats XY counts plot for Mosdepth module.
### Description of feature

image

Hi,

I was wondering if it is possible to reproduce the XY counts plot used in the samtools idxstats multiqc module as an output of the mosdepth module as another means of sex QC. I currently use the `Average coverage per contig` mosdepth plot to do so in the `sarek`, however, it is very noisy and `sarek` does not run `samtools idxstats`:

image

If there was another output of the mosdepth module that showed the average converage per contig for just X and Y chromosomes as a horizontal bar plot that would be pretty useful.

Here is the original issue where this feature was added to the samtools idxstats module:
https://github.com/ewels/MultiQC/issues/251#issuecomment-242232484

vladsavelyev:
1. Adds an X/Y coverage plot, similar to the one in `samtools-idxstats`:

Screenshot 2023-08-15 at 19 12 25

2. Add chromosome filtering options similar to `samtools-idxstats` to avoid clutter. I didn't set a default threshold though to avoid breaking compatibility. Happy to discuss if we want to sync that between samtools or not.

Fixes #1978

### DIFF
--- a/multiqc/modules/mosdepth/mosdepth.py
+++ b/multiqc/modules/mosdepth/mosdepth.py
@@ -15,6 +15,49 @@ from multiqc.plots import bargraph, linegraph
 log = logging.getLogger(__name__)
 
 
+def read_config():
+    cfg = getattr(config, "mosdepth_config", dict())
+    if type(cfg) != dict:
+        return {}
+
+    cfg["include_contigs"] = cfg.get("include_contigs", [])
+    if type(cfg["include_contigs"]) != list:
+        cfg["include_contigs"] = []
+
+    cfg["exclude_contigs"] = cfg.get("exclude_contigs", [])
+    if type(cfg["exclude_contigs"]) != list:
+        cfg["exclude_contigs"] = []
+
+    cfg["xchr"] = cfg.get("xchr", None)
+    if type(cfg["xchr"]) != str:
+        cfg["xchr"] = None
+
+    cfg["ychr"] = cfg.get("ychr", None)
+    if type(cfg["ychr"]) != str:
+        cfg["ychr"] = None
+
+    if cfg["include_contigs"]:
+        log.info("Trying to include these contigs in mosdepth: {}".format(", ".join(cfg["include_contigs"])))
+    if cfg["exclude_contigs"]:
+        log.info("Excluding these contigs from mosdepth: {}".format(", ".join(cfg["exclude_contigs"])))
+    if cfg["xchr"]:
+        log.info('Using "{}" as X chromosome name'.format(cfg["xchr"]))
+    if cfg["ychr"]:
+        log.info('Using "{}" as Y chromosome name'.format(cfg["ychr"]))
+
+    # The coverage cutoff to display chromosomes to avoid clutter
+    cutoff = cfg.get("perchrom_fraction_cutoff", 0.001)
+    try:
+        cutoff = float(cutoff)
+    except ValueError:
+        cutoff = 0.001
+    if cutoff != 0.001:
+        log.info(f"Setting mosdepth per-chrom coverage cutoff to display contigs to: " f"{cutoff * 100.0}%")
+    cfg["perchrom_fraction_cutoff"] = cutoff
+
+    return cfg
+
+
 class MultiqcModule(BaseMultiqcModule):
     """
     Mosdepth can generate multiple outputs with a common prefix and different endings.
@@ -78,8 +121,9 @@ class MultiqcModule(BaseMultiqcModule):
             doi="10.1093/bioinformatics/btx699",
         )
 
+        self.cfg = read_config()
         genstats_headers = defaultdict(OrderedDict)
-        genstats, cumcov_dist_data, cov_dist_data, xmax, perchrom_avg_data = self.parse_cov_dist()
+        genstats, cumcov_dist_data, cov_dist_data, xmax, perchrom_avg_data, xy_cov = self.parse_cov_dist()
 
         # Filter out any samples from --ignore-samples
         if genstats:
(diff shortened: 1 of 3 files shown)


multiqc/modules/mosdepth/mosdepth.py
```diff
@@ -15,6 +15,49 @@
...
+    if type(cfg["exclude_contigs"]) != list:
+        cfg["exclude_contigs"] = []
+
+    cfg["xchr"] = cfg.get("xchr", None)
+    if type(cfg["xchr"]) != str:
+        cfg["xchr"] = None
+
+    cfg["ychr"] = cfg.get("ychr", None)
+    if type(cfg["ychr"]) != str:
+        cfg["ychr"] = None
```
ewels, line 37: Looking forward to using Pydantic for some of this stuff...!

multiqc/modules/mosdepth/mosdepth.py
```diff
@@ -15,6 +15,49 @@
...
+
+    if cfg["include_contigs"]:
+        log.info("Trying to include these contigs in mosdepth: {}".format(", ".join(cfg["include_contigs"])))
+    if cfg["exclude_contigs"]:
+        log.info("Excluding these contigs from mosdepth: {}".format(", ".join(cfg["exclude_contigs"])))
+    if cfg["xchr"]:
+        log.info('Using "{}" as X chromosome name'.format(cfg["xchr"]))
+    if cfg["ychr"]:
+        log.info('Using "{}" as Y chromosome name'.format(cfg["ychr"]))
```
ewels, line 46: I think that `log.debug` is probably sufficient here. Or nothing - the log already contains some stuff about loaded config file changes, which I think should give a rough log of this already.

multiqc/modules/mosdepth/mosdepth.py
```diff
@@ -209,31 +273,18 @@ def __init__(self):
         self.general_stats_addcols(genstats, genstats_headers)
 
     def parse_cov_dist(self):
-        genstats = defaultdict(OrderedDict)  # mean coverage
-        cumcov_dist_data = defaultdict(OrderedDict)  # cumulative distribution
-        cov_dist_data = defaultdict(OrderedDict)  # absolute (non-cumulative) coverage
+        genstats = defaultdict(dict)  # mean coverage
+        cumcov_dist_data = defaultdict(dict)  # cumulative distribution
+        cov_dist_data = defaultdict(dict)  # absolute (non-cumulative) coverage
```
ewels, line 278: There's a lot of this.. I've been thinking of stripping out `OrderedDict` throughout the entire repo, but it's a bit scary in case it breaks stuff. One for a separate PR anyway..

multiqc/modules/mosdepth/mosdepth.py
```diff
@@ -286,6 +337,65 @@ def parse_cov_dist(self):
...
                     self.add_data_source(f, s_name=s_name, section="genome_results")
 
+        # Applying the contig coverage cutoff. First, count the total coverage for
+        # every contig.
+        total_cov_per_contig = defaultdict(lambda: 0)
+        total_cov = 0
+        for s_name, perchrom in perchrom_avg_data.items():
+            for contig, cov in perchrom.items():
+                total_cov_per_contig[contig] += cov
+                total_cov += cov
```
ewels, line 347: Can we collect this info during parsing, to avoid needing to loop through it again here?

multiqc/modules/mosdepth/mosdepth.py
```diff
@@ -286,6 +337,65 @@ def parse_cov_dist(self):
...
+        for s_name, perchrom in perchrom_avg_data.items():
+            for contig, cov in perchrom.items():
+                if float(total_cov_per_contig[contig]) > req_cov:
+                    if contig not in passing_contigs:
+                        passing_contigs.add(contig)
+
+        rejected_contigs = set()
+        filtered_perchrom_avg_data = defaultdict(dict)
+        for s_name, perchrom in perchrom_avg_data.items():
+            for contig, cov in perchrom.items():
```
ewels, line 361: Feels a little ugly looping over this so often - I assume it's quick though? (Thinking of that other issue where the user was running 12k mosdepth reports or something)

ewels (approved):
Couple of minor comments, but general code read-through looks good. Haven't tested locally, but assuming you have then happy to merge once the comments are addressed 👍🏻

multiqc/modules/mosdepth/mosdepth.py
vladsavelyev: Oh for sure!

multiqc/modules/mosdepth/mosdepth.py
vladsavelyev: True, I'll switch to `log.debug`

Here just tried to replicate the `samtools idxstats` behavior: https://github.com/ewels/MultiQC/blob/master/multiqc/modules/samtools/idxstats.py#L47-L66 But we can refactor that separately.

multiqc/modules/mosdepth/mosdepth.py
vladsavelyev: Ran profiling to make sure. Those lines take a negligent amount of time compared to other places: for 100 copies of mosdepth test data, these loop run in ~1000ms each, whereas the line `if any(fnmatch.fnmatch(contig, str(pattern)) for pattern in self.cfg["exclude_contigs"])` takes 1104668ms :) Because it's run for each line in mosdepth output (`for line in f["f"].split("\n")`), and repeated millions of times because of that.

multiqc/modules/mosdepth/mosdepth.py
vladsavelyev: Yeah we can. (though as I mentioned below, doing another sequential loop is _really_ cheap, so as long as it improves readability, I prefer doing that. Though perhaps doesn't matter in this case!)

multiqc/modules/mosdepth/mosdepth.py
ewels: Yup 👍🏻  I already have an issue for this here: https://github.com/ewels/MultiQC/issues/1919

multiqc/modules/mosdepth/mosdepth.py
ewels: That `fnmatch` sounds like a smoking gun! Are you addressing that somewhere?

PR #1986, opened 2023-08-16
Mosdepth: separate section for per-region coverage plots

vladsavelyev:
Fix an effective bug happening when both `region` and `global` coverage histograms for a sample are available (i.e. when mosdepth was run with `--by`, see [mosdepth docs](https://github.com/brentp/mosdepth#usage)). In this case, data was effectively merged. Instead, summarise data separately for each scope, and add a separate report section for the region-based coverage data.


ewels:
Do we have any test data generated in this way? It'd be nice to have an example of it in the testdata repo..

vladsavelyev:
There is an example in test data repo already :)

The issue is that only the visual look changed, so the test has been passing before and it will be passing now. Would be nice to support some more interesting tests though in the moduels, like automated checking if some plot is in the HTML, or some general stats value is within range, or some value is present on the plot 🤔

PR #1990, opened 2023-08-17
Adding Element Biosciences AVITI Bases2fastq support to multiqc

blajoie:
- [X] Adding support for Element Biosciences Bases2Fastq (Fastq generation & Demultiplexing) to MultiQC
- [X] https://github.com/ewels/MultiQC_TestData/pull/263
- [ ] Code is tested and works locally (including with `--lint` flag) - **linting in progress.**

cc @yuhecheng62 @ajaltomare


blajoie:
Example MultiQC reports cc @YuheCheng62 

Cloudbreak-DVT-ecoli-wgs-2x150
12 runs, 1152 samples total.  Each run is a 96plex of Ecoli WGS libraries, sequenced on AVITI.
https://element-public-data.s3.us-west-2.amazonaws.com/multiqc/Cloudbreak-DVT-ecoli-wgs-2x150.html

Cloudbreak-DVT-human-wgs-2x150
10 runs, 20 samples total.  Each run is a 2plex of Human WGS libraries, sequenced on AVITI.
https://element-public-data.s3.us-west-2.amazonaws.com/multiqc/Cloudbreak-DVT-human-wgs-2x150.html

Cloudbreak-DVT-human-rnaseq-2x75
10 runs, 160 samples total.  Each run is a 16plex of Human UHHR Rna-Seq libraries, sequenced on AVITI.
https://element-public-data.s3.us-west-2.amazonaws.com/multiqc/Cloudbreak-DVT-human-rnaseq-2x75.html

blajoie:
PR into MultiQC_TestData for supprting test data. cc @YuheCheng62 
https://github.com/ewels/MultiQC_TestData/pull/263

docs/modules/base2fastq.md
```diff
@@ -0,0 +1,16 @@
+---
+name: bases2fastq
+url: https://docs.elembio.io/docs/bases2fastq/
+description: >
+  bases2fastq convert raw data from Element AVITI system to fastq and output sequencing run statistics
+---
+
+Note that the default maximum file size is 50Mb. Some Runstats.json file will exceed that size if you have many samples or longer reads per run. If promoted, please check the size of all json files in the folder and change the maximum file size accordingly.
```
ewels, line 8: I upped this from 10MB only the other day 😅

ewels:
```suggestion
Note that the default maximum file size is 50Mb. Some Runstats.json file will exceed that size if you have many samples or longer reads per run. If samples are missing, please check the size of all json files in the folder and change the maximum file size accordingly.
```

docs/modules/base2fastq.md
```diff
@@ -0,0 +1,16 @@
...
+
+You can configure the threshold and parse your files by changing the
+`log_filesize_limit` config option. For example, to parse files up to 100MB in
+size, add the following to your MultiQC config file:
+
+```yaml
+log_filesize_limit: 100000000
+```
```
ewels, line 16: Better to link to the docs, in case this changes in the future.

````suggestion
You can configure the threshold and parse your files by changing the
`log_filesize_limit` config option.
See the [documentation](https://multiqc.info/docs/usage/troubleshooting/#big-log-files) for more information.
````

File: multiqc/utils/search_patterns.yaml [hunk 3]
```diff
@@ -17,6 +17,14 @@ bamtools/stats:
   contents: "Stats for BAM file(s):"
   shared: true
   num_lines: 10
+bases2fastq/persample:
+  fn: "*_stats.json"
+bases2fastq/run:
+  fn: "RunStats.json"
+bases2fastq/project:
+  fn: "*_RunStats.json"
```
ewels, line 25: These filename patterns are not very specific, especially `*_stats.json`. Is there anything else that we can also filter on? More specific is good, as it'll return a smaller list of files to the module (faster) and also because when a file is found it's "consumed", meaning if it's for another tool then it won't be passed to that module.

Can we guarantee that these files always start with the string `{"AnalysisVersion":`? Could then combine both `fn` and `content` to make a more specific (`AND`) filter.

See Bcl2fast for a similar example:
https://github.com/ewels/MultiQC/blob/41c1c85813c27a6cec2158d664594afacf178417/multiqc/utils/search_patterns.yaml#L89-L92

setup.py
```diff
@@ -63,6 +63,7 @@
         "rich>=10",
         "rich-click",
         "spectra>=0.0.10",
+        "seaborn",
```
ewels, line 66: It looks like this new dependency is used to generate colour palettes.

MultiQC already has a bunch of functionality built in to do this, see https://github.com/ewels/MultiQC/blob/master/multiqc/utils/mqc_colour.py

Please use that instead and remove seaborn.

multiqc/modules/bases2fastq/bases2fastq.py
```diff
@@ -0,0 +1,278 @@
...
+            # get run + analysis
+            run_name = data_dict.get("RunName", None)
+            analysis_id = data_dict.get("AnalysisID", None)[0:4]
+
+            if not run_name or not analysis_id:
+                log.error("Error with RunStats.json.  Either RunName or AnalysisID is absent.")
+                log.error("Please visit Elembio docs for more information - https://docs.elembio.io/docs/bases2fastq/")
+                raise UserWarning
+
+            run_analysis_name = "__".join([run_name, analysis_id])
```
ewels, line 54: Please use `self.clean_s_name()` on any custom method to generate sample names - see docs: https://multiqc.info/docs/development/modules/#custom-sample-names

This is needed for user-configurable sample name cleanup, as well as functionality such as `--dirs` (prepending directory names) and other bits and pieces.

ewels: You also need to remove ignored samples. See https://multiqc.info/docs/development/modules/#filtering-by-parsed-sample-names

This could be a good place to do that with something like:

```python
if self.is_ignore_sample(run_analysis_name):
    continue
```

multiqc/modules/bases2fastq/bases2fastq.py
```diff
@@ -0,0 +1,278 @@
...
+            # get run + analysis
+            run_name = data_dict.get("RunName", None)
+            analysis_id = data_dict.get("AnalysisID", None)[0:4]
+
+            if not run_name or not analysis_id:
+                log.error(f"Error with {f['root']}.  Either RunName or AnalysisID is absent.")
+                log.error("Please visit Elembio docs for more information - https://docs.elembio.io/docs/bases2fastq/")
+                raise UserWarning
+
+            run_analysis_name = "__".join([run_name, analysis_id])
```
ewels, line 111: As above (sample name cleaning + ignoring).

ewels (changes requested):
Very fast first-pass review to check some of the common gotchas that come up in PRs. A few tweaks to change here, I'll come back for a more thorough review soon. (Haven't even tried running it yet).

multiqc/modules/bases2fastq/bases2fastq.py
```diff
@@ -0,0 +1,278 @@
...
+        run_prefix_len = None
+        for f in self.find_log_files("bases2fastq/run"):
+            data_dict = json.loads(f["f"])
+
+            # get run + analysis
+            run_name = data_dict.get("RunName", None)
+            analysis_id = data_dict.get("AnalysisID", None)[0:4]
+
+            if not run_name or not analysis_id:
+                log.error("Error with RunStats.json.  Either RunName or AnalysisID is absent.")
```
vladsavelyev, line 50:
```suggestion
                log.error("Error with RunStats.json. Either RunName or AnalysisID is absent.")
```

multiqc/modules/bases2fastq/plot_runs.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+def tabulate_run_stats(runData, colorDict):
+    """
+    Tabulate general information and statistics of each run
+    """
+    plotContent = dict()
+    for s_name in runData.keys():
+        runStats = dict()
+        runStats.update({"#Polonies": runData[s_name]["NumPolonies"]})
```
vladsavelyev, line 62: `#Polonies` should be rounded to [read or bases counts](https://github.com/ewels/MultiQC/blob/41c1c85813c27a6cec2158d664594afacf178417/multiqc/utils/config_defaults.yaml#L29-L37C28) in the general stats table. See how [it's done](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/bcl2fastq/bcl2fastq.py#L394-L401
) for the bcl2fastq module as a reference.

vladsavelyev: Same for other metrics. Please take a look if similar metrics that are already reported in other modules, and try to use the same naming style, value formatting, and color scheme. E.g. [bcl2fastq has](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/bcl2fastq/bcl2fastq.py#L462C1-L469) Q30, yield, mean base quality. FastQC can be a reference for other metrics and plots.

multiqc/modules/bases2fastq/plot_samples.py
```diff
@@ -0,0 +1,312 @@
...
+    provides enrichement of a number of different K-mers at the 5' end of the reads.
+    Whilst this is a true technical bias, it isn't something which can be corrected
+    by trimming and in most cases doesn't seem to adversely affect the downstream
+    analysis._
+    """
+    plotContent = data
+    return html, plotName, anchor, description, helptext, plotContent
+
+
+def plot_per_cycle_N_content(sampleData, groupLookupDict, colorDict):
```
vladsavelyev, line 170: Can the [FastQC code for that plot](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/fastqc/fastqc.py#L693) be adapted here?

vladsavelyev:
I tried to open the first example that you shared (
https://element-public-data.s3.us-west-2.amazonaws.com/multiqc/Cloudbreak-DVT-ecoli-wgs-2x150.html) and it froze my browser :( I didn't investigate much but I suspect it might be the `Per Cycle Sequence Composition` plot?

Overall, the module is a great addition, but I would ask to adjust the code style (particularly, Python's standard is to use lower_case_underscore naming for variables and functions), and to look at similar modules like [fastqc](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/fastqc/fastqc.py#L693) [bclf2fastq](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/bcl2fastq/bcl2fastq.py) to reuse and adapt the existing code for similar plots and metrics.

vladsavelyev (changes requested):
Also, a CI test is [failing](https://github.com/ewels/MultiQC/actions/runs/5896544120/job/16010305973?pr=1990), would be great to resolve it.

multiqc/modules/bases2fastq/plot_samples.py
```diff
@@ -0,0 +1,312 @@
...
+    config = {
+        "description": "GC",
+        "xlab": "GC content",
+        "ylab": "Percentage",
+        "colors": sampleColor,
+        "id": "gc_hist",
+        "title": "bases2fastq: per sample GC content histogram",
+        "ylab": "Percentage",
+    }
+    plotName = "per sample GC histogram"
```
vladsavelyev, line 254:
```suggestion
    plotName = "Per-sample GC histogram"
```

blajoie:
Thank you for the review/revisions @ewels & @vladsavelyev  !  We will push the fixes very shortly.

blajoie:
> I tried to open the first example that you shared ( https://element-public-data.s3.us-west-2.amazonaws.com/multiqc/Cloudbreak-DVT-ecoli-wgs-2x150.html) and it froze my browser :( I think the culprits are the sample-level plots, but can't understand what exactly is the bottleneck in rendering. Would be good to look more because even the test sample data makes the rendering veeery slow. We might want to revert to static plots here. But something to get back to when we replace the plotting library.
> 
> Overall, the module is a great addition, but I would ask to adjust the code style (particularly, Python's standard is to use lower_case_underscore naming for variables and functions), and to look at similar modules like [fastqc](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/fastqc/fastqc.py#L693) [bclf2fastq](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/bcl2fastq/bcl2fastq.py) to reuse and adapt the existing code for similar plots and metrics.

Yes, the first plot was meant to be an upper-limit test, 10 runs + 1052 samples total.  It does open on my end AOK, but agreed it is quite large.  We will apply a linter and make the style changes as well as look more into fastqc/bcl2fastq for opportunities to reuse existing code.

multiqc/utils/search_patterns.yaml
blajoie: Our jsons can be quite large - we try to give the customer an abundance of metrics at the run, sample, and cycle level.
Recently we've moved to flatten our json (e.g. jq -c) to reduce file size, which then causes the jsons to only have a single (flattened) line :)

For our sample specific jsons, the `fn: "*_stats.json"` will only ever match to only file.  The * is there to capture the <sample_id> that we use to prefix these file names.  It may be better to not search for a contents across N lines given the above?  Thoughts?

setup.py
blajoie: we will use mpc colors instead.

multiqc/modules/bases2fastq/bases2fastq.py
blajoie: this is now addressed.

multiqc/modules/bases2fastq/bases2fastq.py
blajoie: fixed.

blajoie:
> * Got errors about making Matplotlib figures. Haven't looked into why yet.
This is now resolved

> * Colours in tables are all defaults. [Colour matters](https://multiqc.info/docs/development/modules/#colour-matters).
We will take this back and discuss internally.  We only have a spec for %Q30.  For all others the metrics are often specific to the experiment/run/etc..    But we can perhaps had some more useful coloring.

> * Please add a column or two to general stats for comparison against results from other tools. Yield maybe?
yes will look into adding

> * Run name - set `Scale` to `None` for text cells, it'll fix text wrapping.
great - thanks

> * Please write more help text. Explain the why of the plot, what to look for, implications, how to spot bad samples etc. Current help text is more suitable as the description field.
Will work with marketing to find appropriate language to use :)


More middle of next week - thank you !

multiqc/utils/search_patterns.yaml
ewels: > Will this cause issues on the multiqc side given that now you must slurp in the entire json to then search for our `content` pattern?

It'll slow it down for sure, though the `fn` pattern is there at least so it hopefully wouldn't be so many files.

This means that the only remaining reason for doing this really is in case this search pattern is preventing files from being found by other tools. I think it's still worth adding it in, if that's ok.

@vladsavelyev another search optimisation that we could do is to read in chunks of bytes to check for `contents` rather than just counting lines, to avoid this situation of having to read a 100MB file with a single line, just to match the first 10 characters. Pretty rare edge case though, so maybe not worth the effort.

setup.py
ewels: Thank you!

ewels:
Thanks for the fast responses @blajoie!

> We will take this back and discuss internally. We only have a spec for %Q30. For all others the metrics are often specific to the experiment/run/etc.. But we can perhaps had some more useful coloring.

When it comes to colours, it's less about exact thresholds and more about what the colours suggest. For example, with %Q30 - a low value is bad and a high value is good. So an appropriate colour scheme would be `RdYlGn` (see available colour schemes [in the docs](https://multiqc.info/docs/development/plots/#table-colour-scales)).

Also, I'm a bit of a stickler for not having two columns with the same colour scheme side by side. This is so that people repeatedly viewing the same types of reports can recognise the columns quickly at a glance, without having to read the text in the table header. It's also easier to follow long tables.

setup.py
```diff
@@ -62,7 +62,7 @@
         "requests",
         "rich>=10",
         "rich-click",
-        "spectra>=0.0.10",
+        "spectra>=0.0.10"
```
ewels, line 65: The Black code linter will complain with out this..

```suggestion
        "spectra>=0.0.10",
```

I recommend setting up [pre-commit](https://pre-commit.com/) to run these tests automatically before pushing.

```bash
pip install pre-commit
# Set up MultiQC repo specifically, run from repo root dir
pre-commit install
# After that, pre-commit will run tests every time you do:
git commit -m "My changes"
# If anything fails (eg. a comma breaking the linting) it'll hopefully be fixed automatically, but it'll cancel the commit. So you need to then do:
git add .
git commit -m "My changes"
#.. then the pre-commit tests should pass.
```

multiqc/modules/bases2fastq/plot_samples.py
blajoie: this is now fixed

multiqc/utils/search_patterns.yaml
blajoie: ok - marking this as resolved.  We can look into not completely flattening the entire json to help improve this.

docs/README.md
```diff
@@ -1,3 +1,157 @@
...
+Custom Content:
+  Introduction: custom_content.md
+Coding with MultiQC:
+  Writing new modules: modules.md
+  Plotting Functions: plots.md
+  MultiQC Plugins: plugins.md
+  Writing new templates: templates.md
+  Updating for compatibility: compatibility.md
+---
+
```
vladsavelyev, line 154: This block was recently removed in the master branch, I think it's left here as a merge artefact :)

```suggestion
```

docs/modules/bases2fastq.md
```diff
@@ -0,0 +1,12 @@
+---
+name: bases2fastq
+url: https://docs.elembio.io/docs/bases2fastq/
+description: >
+  bases2fastq convert raw data from Element AVITI system to fastq and output sequencing run statistics
```
vladsavelyev, line 5:
```suggestion
name: Bases2Fastq
url: https://docs.elembio.io/docs/bases2fastq/
description: >
  Bases2Fastq converts raw data from Element AVITI system to FastQ and outputs sequencing run statistics.
```

vladsavelyev (changes requested):
A few change suggestions, otherwise good with me.

vladsavelyev (approved):
Awesome stuff!

multiqc/modules/bases2fastq/plot_runs.py
```diff
@@ -0,0 +1,326 @@
...
+        plot_content.update({s_name: run_stats})
+
+    headers = OrderedDict()
+
+    headers["num_polonies_run"] = {
+        "title": "Number of Polonies",
+        "format": "{d}",
+        "description": "The (total) number of polonies calculated for the run",
+        "min": 0,
+        "scale": "RdYlGn",
```
vladsavelyev, line 110: Just repeating some code comments that might be lost now :)

`#Polonies` should be rounded to [read or bases counts](https://github.com/ewels/MultiQC/blob/41c1c85813c27a6cec2158d664594afacf178417/multiqc/utils/config_defaults.yaml#L29-L37C28) in the general stats table. See how [it's done](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/bcl2fastq/bcl2fastq.py#L394-L401
) for the bcl2fastq module as a reference.

vladsavelyev: Same for other metrics. Could you take a look if similar metrics that are already reported in other modules, and try to use the same naming style, value formatting, and color scheme. E.g. [bcl2fastq has](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/bcl2fastq/bcl2fastq.py#L462C1-L469) Q30, yield, mean base quality. FastQC can be a reference for other metrics and plots.

multiqc/modules/bases2fastq/plot_samples.py
```diff
@@ -0,0 +1,338 @@
...
+    provides enrichement of a number of different K-mers at the 5' end of the reads.
+    Whilst this is a true technical bias, it isn't something which can be corrected
+    by trimming and in most cases doesn't seem to adversely affect the downstream
+    analysis._
+    """
+    plot_content = data
+    return html, plot_name, anchor, description, helptext, plot_content
+
+
+def plot_per_cycle_N_content(sample_data, group_lookup_dict, color_dict):
```
vladsavelyev, line 196: Would be great to adapt the [FastQC code for that plot](https://github.com/ewels/MultiQC/blob/master/multiqc/modules/fastqc/fastqc.py#L703)

vladsavelyev (changes requested):
Minor changes requested - see the code comments above!

ewels:
ok, the remaining changes are very minor. I just tried to fix the merge conflict but couldn't push it back - because the fork is under Elembio (and not a user account), GitHub is a bit more strict about allowing us to push into the PR directly.

To keep the momentum, I've created a new branch that we can work in and an associated PR: https://github.com/ewels/MultiQC/pull/2044

Hopefully we can do these final minor changes and get this merged soon 😄

multiqc/utils/search_patterns.yaml
```diff
@@ -17,6 +17,20 @@ bamtools/stats:
...
+bases2fastq/run:
+  fn: "RunStats.json"
+  contents_re: ".*NumPolonies.*"
+  num_lines: 100
+bases2fastq/project:
+  fn: "*_RunStats.json"
+  contents_re: ".*NumPolonies.*"
+  num_lines: 100
+bases2fastq/group:
+  fn: "*_b2fgroup.csv"
```
vladsavelyev, line 33: We don't seem to have any test data for the group file, do we? 
https://github.com/ewels/MultiQC_TestData/tree/master/data/modules/bases2fastq

Would it be possible to add some? @blajoie

multiqc/modules/bases2fastq/bases2fastq.py
```diff
@@ -0,0 +1,321 @@
...
+        self.palette = self.palette[len(self.group_dict) :]
+
+        # Read custom group info
+        self.group_info_exist = False
+        for f in self.find_log_files("bases2fastq/group"):
+            if self.group_info_exist:
+                log.warning(
+                    "More than one group assignment files are found. Please only keep one assignment file in the analysis folder. Bases2fastq stats will not be plotted"
+                )
+            group_info = pd.read_csv(StringIO(f["f"]))
```
vladsavelyev, line 243: Pandas is not imported anywhere.

Note that we also don't install pandas with MultiQC, so we can't really use it. I think it's pretty straightforward to get around without pandas here?

vladsavelyev: Modified it in the new :)

ewels:
Locking the conversation here so that we remember to move over to https://github.com/ewels/MultiQC/pull/2044

ewels:
Actually I think I will just close this PR for clarity.

PR #1991, opened 2023-08-18
Update installation.md

antass:
Add `multiqc` function call at the end of the bash alias definition. Without it, to run multiqc one would have to call `multiqc multiqc .` instead of `multiqc .`


ewels (approved):
Good spot, thanks!

PR #576, opened 2017-09-05
new group_samples() function

ewels:
New base module function `group_samples()` for bundling samples into groups, such as read pairs. A default config added and FastQC module updated to make use of the read pair grouping.

See #542 for discussion.

PR #1992, opened 2023-08-18
feat: new 'endedness' submodule for ngsderive

a-frantz:

### DIFF
--- a/multiqc/modules/ngsderive/ngsderive.py
+++ b/multiqc/modules/ngsderive/ngsderive.py
@@ -33,6 +33,7 @@ class MultiqcModule(BaseMultiqcModule):
         self.readlen = {}
         self.encoding = {}
         self.junctions = {}
+        self.endedness = {}
 
         # parse ngsderive summary file
         expected_header_count_strandedness = 5
@@ -40,6 +41,8 @@ class MultiqcModule(BaseMultiqcModule):
         expected_header_count_readlen = 4
         expected_header_count_encoding = 3
         expected_header_count_junctions = 9
+        expected_header_count_endedness_wo_rpt = 6
+        expected_header_count_endedness_w_rpt = 7
 
         for f in self.find_log_files("ngsderive/strandedness"):
             self.parse(
@@ -81,18 +84,38 @@ class MultiqcModule(BaseMultiqcModule):
                 expected_header_count_junctions,
             )
 
+        for f in self.find_log_files("ngsderive/endedness"):
+            self.parse(
+                self.endedness,
+                f,
+                "endedness",
+                expected_header_count_endedness_wo_rpt,
+            )
+            self.parse(self.endedness, f, "endedness", expected_header_count_endedness_w_rpt)
+
         self.strandedness = self.ignore_samples(self.strandedness)
         self.instrument = self.ignore_samples(self.instrument)
         self.readlen = self.ignore_samples(self.readlen)
         self.encoding = self.ignore_samples(self.encoding)
         self.junctions = self.ignore_samples(self.junctions)
+        self.endedness = self.ignore_samples(self.endedness)
 
         num_results_found = max(
-            [len(d) for d in [self.strandedness, self.instrument, self.readlen, self.encoding, self.junctions]]
+            [
+                len(d)
+                for d in [
+                    self.strandedness,
+                    self.instrument,
+                    self.readlen,
+                    self.encoding,
+                    self.junctions,
+                    self.endedness,
+                ]
+            ]
         )
         if num_results_found == 0:
             raise UserWarning
-        log.info("Found {} reports".format(num_results_found))
+        log.info(f"Found {num_results_found} reports")
 
         if self.strandedness:
             self.add_strandedness_data()
@@ -108,6 +131,9 @@ class MultiqcModule(BaseMultiqcModule):
 
         if self.junctions:
             self.add_junctions_data()
+
+        if self.endedness:
+            self.add_endedness_data()
 
     def probe_file_for_dictreader_kwargs(self, f, expected_header_count):
         """In short, this function was created to figure out which
@@ -454,4 +480,80 @@ class MultiqcModule(BaseMultiqcModule):
             description="""Junction annotations provided by ngsderive. For more information, please see
             [the documentation](https://stjudecloud.github.io/ngsderive/subcommands/junction_annotation/).""",
             plot=bargraph.plot([bardata, bardata], cats, pconfig),
+        )
+
+    def add_endedness_data(self):
+        # Write data to file
+        self.write_data_file(self.endedness, "ngsderive_endedness")
+
+        general_data = {}
+        rpt_present = False
+        for sample, endedness_data in self.endedness.items():
+            general_data[sample] = {
+                "endedness": endedness_data.get("Endedness"),
+            }
+            if "Reads per template" in endedness_data:
+                rpt_present = True
+
+        general_headers = OrderedDict()
+        general_headers["endedness"] = {
+            "title": "Predicted Endedness",
+            "description": "Predicted library endedness from ngsderive",
+        }
+        self.general_stats_addcols(general_data, general_headers)
+
+        headers = OrderedDict()
+        headers["f+l-"] = {
+            "title": "f+l- count",
+            "description": "Number of reads with the 'first in template' FLAG set and the 'last in template' FLAG unset",
+            "format": "{:,.d}",
+        }
+        headers["f-l+"] = {
+            "title": "f-l+ count",
+            "description": "Number of reads with the 'first in template' FLAG unset and the 'last in template' FLAG set",
+            "format": "{:,.d}",
+        }
+        headers["f-l-"] = {
+            "title": "f-l- count",
+            "description": "Number of reads with the 'first in template' FLAG unset and the 'last in template' FLAG unset",
+            "format": "{:,.d}",
+        }
+        headers["f+l+"] = {
+            "title": "f+l+ count",
+            "description": "Number of reads with the 'first in template' FLAG set and the 'last in template' FLAG set",
+            "format": "{:,.d}",
+        }
+        if rpt_present:
+            headers["RPT"] = {
+                "title": "Reads per template",
+                "description": "Average number of reads per QNAME template",
+                "format": "{:,.4f}",
+            }
+        headers["endedness"] = {
+            "title": "Predicted Endedness",
+            "description": "Predicted library endedness from ngsderive",
+        }
+
+        table_data = {}
+        for sample, endedness_data in self.endedness.items():
+            table_data[sample] = {}
+            table_data[sample]["f+l-"] = endedness_data.get("f+l-")
+            table_data[sample]["f-l+"] = endedness_data.get("f-l+")
+            table_data[sample]["f-l-"] = endedness_data.get("f-l-")
+            table_data[sample]["f+l+"] = endedness_data.get("f+l+")
+            table_data[sample]["RPT"] = endedness_data.get("Reads per template")
+            table_data[sample]["endedness"] = endedness_data.get("Endedness")
+
+        # Config for the plot
+        config = {
+            "id": "ngsderive_endedness_plot",
+            "title": "ngsderive: Endedness",
+        }
+
+        self.add_section(
+            name="Endedness",
+            anchor="ngsderive-endedness",
+            description="""Predicted library endedness provided by ngsderive. For more information, please see
+            [the documentation](https://stjudecloud.github.io/ngsderive/subcommands/endedness/).""",
+            plot=table.plot(table_data, headers, config),
         )
--- a/multiqc/utils/search_patterns.yaml
+++ b/multiqc/utils/search_patterns.yaml
@@ -511,6 +511,9 @@ ngsderive/encoding:
 ngsderive/junction_annotation:
   contents: "File	total_junctions	total_splice_events	known_junctions	partial_novel_junctions	complete_novel_junctions	known_spliced_reads	partial_novel_spliced_reads	complete_novel_spliced_reads"
   num_lines: 1
+ngsderive/endedness:
+  contents: "File	f+l-	f-l+	f-l-	f+l+	"
+  num_lines: 1
 optitype:
   contents: "	A1	A2	B1	B2	C1	C2	Reads	Objective"
   num_lines: 1


multiqc/modules/ngsderive/ngsderive.py
```diff
@@ -455,3 +481,79 @@ def add_junctions_data(self):
...
+        }
+
+        table_data = {}
+        for sample, endedness_data in self.endedness.items():
+            table_data[sample] = {}
+            table_data[sample]["FyLn"] = endedness_data.get("f+l-")
+            table_data[sample]["FnLy"] = endedness_data.get("f-l+")
+            table_data[sample]["FnLn"] = endedness_data.get("f-l-")
+            table_data[sample]["FyLy"] = endedness_data.get("f+l+")
+            table_data[sample]["RPT"] = endedness_data.get("Reads per template")
```
vladsavelyev, line 544: Unfortunately formatting in the tables doesn't work properly when numbers are passed as strings:

```suggestion
            table_data[sample]["FyLn"] = int(endedness_data.get("f+l-"))
            table_data[sample]["FnLy"] = int(endedness_data.get("f-l+"))
            table_data[sample]["FnLn"] = int(endedness_data.get("f-l-"))
            table_data[sample]["FyLy"] = int(endedness_data.get("f+l+"))
            table_data[sample]["RPT"] = float(endedness_data.get("Reads per template"))
```

Would be good to do that for other subcommands as well, including the general stats table

vladsavelyev (changes requested):
Thanks for adding another feature! Looks great, just some comments mostly relevant to formatting.

multiqc/modules/ngsderive/ngsderive.py
```diff
@@ -455,3 +481,79 @@ def add_junctions_data(self):
...
+            "title": "Predicted Endedness",
+            "description": "Predicted library endedness from ngsderive",
+        }
+        self.general_stats_addcols(general_data, general_headers)
+
+        headers = OrderedDict()
+        headers["FyLn"] = {
+            "title": "f+l- count",
+            "description": "Number of reads with the 'first in template' FLAG set and the 'last in template' FLAG unset",
+            "format": "{:,.d}",
```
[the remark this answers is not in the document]
a-frantz: > For the metrics showing read counts it's nice to use the `config.read_count_prefix`, `config.read_count_desc`, and `config.read_count_multiplier`, like here: https://github.com/ewels/MultiQC/blob/master/multiqc/modules/verifybamid/verifybamid.py#L178-L186

I can't tell from that code snippet what the variable `config` is. From context, I'm gathering this is a way to toggle units( i.e. "millions of reads")? I think for this table we want to prevent that. The sanity check this is performing (for Paired-End data) is that the _exact_ same number of "first" reads and "last" reads exist in the file. Hence we don't want to lose any precision. We want the full number of reads displayed, not an approximation.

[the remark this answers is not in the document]
vladsavelyev: Got it, that makes sense! Agree with that. 
I'm mostly concerned with the readability of large numbers, but fixing the `"{:,d}"` format will add spaces between decimal groups, so it will be readable anyway.

vladsavelyev (approved):

vladsavelyev:
Thanks for addressing the suggestions quickly @a-frantz, looks good to me! 
Will leave it to @ewels to merge.

a-frantz:
@ewels
The new `endedness` command (and the old `strandedness` command) support a parameter called `--split-by-rg`. This does what it sounds like it does; instead of one entry per sample (BAM), it will have X+1 entries per sample, where X is the number of read groups in the file (plus 1 for an `overall` entry).
When first writing the `strandedness` integration a few years ago, I was under the impression that there wasn't a good way in MultiQC to present multiple data points per sample (in the simplest form, think of a table where multiple rows correspond to the same sample). Because of that belief, I simply decided MultiQC should ignore any files with a `ReadGroup` column header and called it a day.
I'm rethinking that decision, and wanted to check with you if what I want is already present in MultiQC. We've found a decent number of cases where splitting these analyses by read-group has elucidated something interesting. It would be ideal if we can get this slight change of data format through the MultiQC parsing.

ewels:
@a-frantz not really I'm afraid, at least - not yet. I've been wanting to add some core functionality to handle groups of samples / results for years. I actually had a stab at it back in 2017 (time flies 🤯 ) in #576 👀 

It will happen at some point. But it's very complex and might be one of the things that ends up in the bigger v2 milestone that we're starting to build. Feel free to follow https://github.com/ewels/MultiQC/issues/542 for progress.

In the mean time, could you just treat each as a separate sample - just adding the read group as a suffix to the sample name? It's a little crude, but would get the data into the report at least. I think that Picard MarkDuplicates does this already: 

https://github.com/ewels/MultiQC/blob/a5a31290ecc540c377a5569d5912d1758fab389c/docs/modules/picard.md?plain=1#L134-L135

Also DRAGEN, VerifyBamId and some other modules do stuff with read groups too it seems, based on a quick project-wide search.

multiqc/modules/ngsderive/ngsderive.py
```diff
@@ -33,66 +33,51 @@ def __init__(self):
...
-                self.junctions,
-                f,
-                "junctions",
-                expected_header_count_junctions,
-            )
+            self.parse(self.junctions, f, "junctions", expected_header_count=9)
+
+        for f in self.find_log_files("ngsderive/endedness"):
+            self.parse(self.endedness, f, "endedness", expected_header_count=6)
+            self.parse(self.endedness, f, "endedness", expected_header_count=7)
```
ewels, line 56: I take it that this is because the file could contain either 6 or 7 columns?

The problem with this is that we now get a warning log message for every sample:
```
|         ngsderive | Could not easily detect delimiter for file. Trying csv.Sniffer()
```
I think that we can drop this to debug at the best of times, but even so - I feel like a range could be a better approach?

Note that we only have a single test data file to work with here, so I can't be sure / test. It would be nice to have a few more example data files, covering different styles of output, if possible.

multiqc/modules/ngsderive/ngsderive.py
```diff
@@ -455,3 +447,80 @@ def add_junctions_data(self):
...
+            "title": "Predicted Endedness",
+            "description": "Predicted library endedness from ngsderive",
+        }
+        self.general_stats_addcols(general_data, general_headers)
+
+        headers = OrderedDict()
+        headers["FyLn"] = {
+            "title": "f+l- count",
+            "description": "Number of reads with the 'first in template' FLAG set and the 'last in template' FLAG unset",
+            "format": "{:,d}",
```
ewels, line 475: Should use `"shared_key": "read_count"` and the appropriate number base modifiers. See [docs](https://multiqc.info/docs/reports/customisation/#number-base-multiplier).

Hmm, docs are a bit thin on this. Example here:

https://github.com/ewels/MultiQC/blob/57f3588c5a524f855b0706022e6db44f423588d1/multiqc/modules/ivar/ivar.py#L110-L116

multiqc/modules/ngsderive/ngsderive.py
```diff
@@ -455,3 +447,80 @@ def add_junctions_data(self):
...
+            "title": "ngsderive: Endedness",
+            "namespace": "ngsderive",
+        }
+
+        self.add_section(
+            name="Endedness",
+            anchor="ngsderive-endedness",
+            description="""Predicted library endedness provided by ngsderive. For more information, please see
+            [the documentation](https://stjudecloud.github.io/ngsderive/subcommands/endedness/).""",
+            plot=table.plot(table_data, headers, config),
```
ewels, line 525: Is it possible to represent these counts as a stacked bar plot, instead of a table? They're all subsets of a whole, right? Or could they sum to > 100%?

a-frantz: They do sum to 100%, so a stacked bar plot is possible. I'm just afraid it might be less informative than the raw numbers. Which is admittedly counter-intuitive. Let me explain how these data will be used.

These data are more or less meant as sanity checks. The patterns you expect to see are:
Paired-End sample: `f+l-` == `f-l+` && `f+l+`==`0` && `f-l-`==`0`
Single-End sample: `f+l+` > `0` && `everything else == 0`
So in my mind, the easiest way to check those assumptions is with a table, where I can quickly glance and say "Great, these numbers are the same number" or "O no, something is greater than zero". The degree by which a category is off from it's assumed value doesn't really matter that match.

In practice, I'm running this on our data and seeing some off-by-one cases. I'm afraid that presenting this as a barplot would mask an off-by-one case, because in a plot the difference between 3million and 3million+1 would be impossible to see. But in a table, super easy to see a value is off by one.

I could see using the conditional formatting available to tables to color-code some of these cases. Would that be possible? First instinct is "Give each unique numerical value a distinct color" (there will be at most 4). Equal cells will have the same color. Unequal cells would have some contrast. That should facilitate the "at-a-glance" assumption checking.

So far I've ignored the `Reads Per Template` value. That is supposed to be exactly `2.0` (Paired-End) or exactly `1.0`(Single-End). Deviation from those exact values is what's concerning. Any ideas for how to present that more intuitively?

multiqc/modules/ngsderive/ngsderive.py
a-frantz: Would this automatically scale the values? As I shared in another comment, these data are being checked for exact equality. Any scaling would mask problems users should know about. The value of the "one's" place is important, as off-by-one errors are the most common class of error.

a-frantz:
> @a-frantz not really I'm afraid, at least - not yet. I've been wanting to add some core functionality to handle groups of samples / results for years. I actually had a stab at it back in 2017 (time flies 🤯 ) in #576 👀
> 
> It will happen at some point. But it's very complex and might be one of the things that ends up in the bigger v2 milestone that we're starting to build. Feel free to follow #542 for progress.
> 
> In the mean time, could you just treat each as a separate sample - just adding the read group as a suffix to the sample name? It's a little crude, but would get the data into the report at least. I think that Picard MarkDuplicates does this already:
> 
> https://github.com/ewels/MultiQC/blob/a5a31290ecc540c377a5569d5912d1758fab389c/docs/modules/picard.md?plain=1#L134-L135
> 
> Also DRAGEN, VerifyBamId and some other modules do stuff with read groups too it seems, based on a quick project-wide search.

This seems like a decent enough solution. Not ideal, but I think we can make it work. My big concern with taking this approach is that I'm not sure if there's a way to "hide" these suffixed "samples" from other sections, mainly the General Stats table. I think having 7 entries with the same sample prefix in one block of the table, and then 6 of them being blank for every other section of the table would be pretty ugly.

What I'm envisioning: `ngsderive` inserts an `overall` entry when `--split-by-rg` is True. This entry would be parsed according to standard sample cleaning rules and would appear as an entry in the General Stats Table. Every other ngsderive result with a `ReadGroup` creates a new "sample" by suffixing the `overall` sample name with the RG ID. These "new samples" don't appear anywhere outside of `ngdserive`'s block of the MultiQC report. No entries go into General Stats.

Is this possible?

multiqc/modules/ngsderive/ngsderive.py
a-frantz: This has been refactored to use a `set` of Ints.

And another example with a different style output uploaded: https://github.com/ewels/MultiQC_TestData/pull/269

multiqc/modules/ngsderive/ngsderive.py
vladsavelyev: I think that for readability, decimal separators do the job, so there is no need to use read count multipliers that hide the precision that's important here. Unless we add another helpful metric like the difference between f+l- and f-l+. Though I'm not sure if MultiQC should be creative in this case.

multiqc/modules/ngsderive/ngsderive.py
vladsavelyev: I agree that we should keep the raw numbers instead of rounding or plotting them. But perhaps, indeed, these checks can be automated, so conditional formatting to highlight samples that break the condition is a good direction to look into.

Another approach would be to just add one more derived metric into the general stats that would simply show if the sample passes those conditions or not.

multiqc/modules/ngsderive/ngsderive.py
vladsavelyev: Looks like it actually fails with the new test, so it's great that we added it :D

vladsavelyev: Pushed a fix

multiqc/modules/ngsderive/ngsderive.py
vladsavelyev: It would be also helpful to add a "bad" example into the test data, where a sample doesn't pass the endedness criterias.

a-frantz: Let me complicate things for us: There is a parameter to this command which enables only processing a subset of the BAM file. When this mode is active, there is obviously some amount of deviation from the _exact_ conditions I've outlined previously, just due to the randomness inherent in only processing a portion of the available reads. When run in this mode, the user supplies a tolerance range for determining the conditions. e.g. we can still call the sample `Paired-End` so long as `0.45 < (f+l-)/((f+l-)+(f-l+)) < 0.55`. This is in contrast to the typical case, when every read is processed, that requires `(f+l-)==(f-l+)`. Similarly RPT gets rounded to the nearest `int` during checks when run in "subset mode".

Generally we would discourage users from using this mode as it will mask errors. But we felt the need to enable it for consistency with the rest of the `ngsderive` commands. As you can imagine, checking out every read is the file takes a long time. A user has the option to get a decent enough guess in a few seconds if that's what their use-case calls for.

> Another approach would be to just add one more derived metric into the general stats, that would simply show if the sample passes those conditions or not. Similarly formatted to the existing "predicted instrument" and "predicted endedness" columns.

Because of the inconsistency outlined above, MultiQC can't get into the business of checking these conditions itself. I would not want to imply "conditions met" or "conditions unmet" with colors such as "green" and "red". The conditions checking is already done and presented by the value of `Endedness`. That will be `Unknown` if no conditions are met (or `Paired-End` or `Single-End` were _those_ conditions met). I wouldn't be opposed to coloring `Unknown` red, but I'm not sure it really adds a whole lot. Same for coloring `Paired-End` and `Single-End` green. I could take it or leave it.

I think my first instinct:
>  First instinct is "Give each unique numerical value a distinct color" (there will be at most 4). Equal cells will have the same color. Unequal cells would have some contrast. That should facilitate the "at-a-glance" assumption checking.

is still our best bet. I haven't come up with anything better. It wouldn't be particularly useful for the niche case that only a subset of the BAM was processed, but I think that's fine. We can't make everyone happy, so let's focus on the predominant use-case.

I wouldn't want to color the RPT cell because it is allowed to deviate some. I suppose `>=3` would be universally "a problem" today, but in the future we might find a valid use case for that (circular templates?). So to future-proof, I wouldn't want to call any RPT value "bad".

a-frantz: > It would be also helpful to add a "bad" example into the test data, where a sample doesn't pass the endedness criteria.

https://github.com/ewels/MultiQC_TestData/blob/master/data/modules/ngsderive/test.endedness.without_rpt.txt

This represents the common "off by one" error case

multiqc/modules/ngsderive/ngsderive.py
ewels: ok 👍🏻 

Worth considering a custom shared key anyway, to preserve bar scaling across columns? I _think_ that `read_count` might do some magic with modifiers (?) but a custom string shouldn't.

multiqc/modules/ngsderive/ngsderive.py
ewels: Wow, this is a fairly atypicaly module 😆  It would be great to write all of this in the description / help text above the table, so that users can read how about to interpret the results.

I can't think of any other modules doing this, but it should be possible to get what you want by setting `scale` to `False` and `bgcols` to a dynamically generated dict of discovered read counts + assigned colours. Then you can effectively colour code the table cells according to their numerical status.

If the actual number isn't very interesting, then I think this makes sense for me. As the normal cell background bars don't really tell you anything.

ewels:
> Is this possible?

Yup 👍🏻  The module code dictates which samples + data go into the General Stats table / report section separately, so you should have complete control over what goes where.

Plenty of precedent on this one. For example, FastQC appends suffixes to sample names for the adapter trimming plot.

a-frantz:
@ewels I'm working on a [v4](https://github.com/stjudecloud/ngsderive/pull/139) release of ngsderive. It involves some breaking changes that will need to be accommodated for in MultiQC*.

Would you rather me knock this current PR out first and open another PR to address the v4 changes, or do it all in one PR? What's easiest for your workflow?

*the changes relevant to MultiQC are:
* reporting percentages instead of proportions in the range of 0-1
    * we'll have to detect what the scale is in MultiQC (does this value need to be multiplied by `100` before going into the MultiQC report?)
* Headers have been made to consistently use PascalCase, instead of the current mishmash of cases
    * MultiQC will need to be able to parse the data of either

vladsavelyev:
@a-frantz, it's really up to you. If you already know what will the new ngsderive output look like, perhaps it makes sense to update the code in this PR, and also update the MultiQC_TestData examples.

multiqc/modules/ngsderive/ngsderive.py
```diff
@@ -124,37 +112,61 @@ def probe_file_for_dictreader_kwargs(self, f, expected_header_count):
...
         for row in csv.DictReader(io.StringIO(found_file["f"]), **kwargs):
             if not row.get("File"):
                 continue
             sample_name = self.clean_s_name(row.get("File"), found_file)
+            read_group = row.get("ReadGroup")
+            if read_group and read_group != "overall":
+                sample_name += "-" + read_group
+                row["is_read_group"] = True
+            else:
+                row["is_read_group"] = False
```
a-frantz, line 134: @vladsavelyev At this stage of development, the only subcommands which support read group level data are `strandedness` and `endedness`. That data is not guaranteed to be present, as a user must opt into that level of granularity.

I would like to support two different modes for the `strandedness` barplot. The default would be the current implementation (only one "overall" entry per-sample). The user could then toggle the "Read Group" mode which shows samples + their read group level entries. I'll need to play around with the sorting here to see what would be most useful. I'm not sure ATM.

I don't think we need any toggles for the `endedness` table. I think it's suitable to display all found information by default.

Is there an example elsewhere in the code base I could follow for what I'm describing for `strandedness`? I'm not quite sure the best way to approach the problem.

multiqc/modules/ngsderive/ngsderive.py
```diff
@@ -124,37 +112,61 @@ def probe_file_for_dictreader_kwargs(self, f, expected_header_count):
...
 
         data = {}
         for sample, strandedness in self.strandedness.items():
-            data[sample] = {
-                "predicted": strandedness.get("Predicted"),
-                "forward": round(float(strandedness.get("ForwardPct")) * 100.0, 2),
-                "reverse": round(float(strandedness.get("ReversePct")) * 100.0, 2),
-            }
+            if float(strandedness.get("ForwardPct")) == 0 and float(strandedness.get("ReversePct")) == 0:
+                # TODO how to present this case? No reads found for this Read Group.
```
a-frantz, line 149: @vladsavelyev This is a case I'm stumped over. In v4 of ngsderive, I added a feature where the read group definitions in the BAM header are validated against what appears in the sequences. This now means it's possible for an entry to be NULL, if the read group is defined in the header but lacks any sequences associated with it. There's no data to present here, so it can't go into the barplot. This is most likely an error in the BAM that should be communicated to the user. However I'm not sure how to go about that. I suppose we could `log.warning("Read Group info messed up! Yada yada")` and just ignore it for the output report... I would like to see it somewhere in the report, but I have no idea about an appropriate way to achieve that.

Issue #264, opened 2016-07-12
Point to example config file in docs


ewels:
The documentation talks about `multiqc_config_example.yaml`, however this file doesn't exist if the package is installed through `pip` or `conda`.

Add to the docs a link to view this file, and make it more obvious about where to find it.

On a side note, it could be nice to move the hardcoded defaults into a yaml file for consistency?

PR #1993, opened 2023-08-21
New module: mapDamage

TCLamnidis:
This PR adds a module to add aNDA damage plots and statistics to a MultiQC report. It creates:
 - 4 general stats table columns with the deamination damage % in 1st & 2nd bp of the 5' and 3' ends of reads.
 - 2 linegraphs of the deamination % at each bp on the 5' and 3' ends of reads respectively.
 - 2 linegraphs with the length distribution of reads mapping to the forward and reverse strands respectively.
The module output is as close as possible to that of DamageProfiler, and the module code is largely based on that of the `damageprofiler` module, except the input data is read from individual files instead of a nice JSON. 


TCLamnidis:
Test Data PR here: [MultiQC/TestData #264](https://github.com/ewels/MultiQC_TestData/pull/264)

multiqc/modules/mapdamage/mapdamage.py
```diff
@@ -0,0 +1,282 @@
...
+
+        # Init empty dictionaries
+        self.threepGtoAfreq_data = dict()
+        self.fivepCtoTfreq_data = dict()
+        self.lgdist_fw_data = dict()
+        self.lgdist_rv_data = dict()
+
+        # Find and load log files
+        for f in self.find_log_files("mapdamage", filehandles=True):
+            self.parseInput(f)
```
vladsavelyev, line 36: I would stick to the conventional pythonic naming style, which is lowercase_underscore. And for MultiQC (somewhat loose) [convention](https://multiqc.info/docs/development/modules/#step-2---parse-data-from-the-input-files), would be good to call the method `parse_logs()`:

```suggestion
            self.parse_logs(f)
```

multiqc/modules/mapdamage/mapdamage.py
```diff
@@ -0,0 +1,282 @@
...
+
+                    length_distribution_dict["lendist_" + strand][line[1]] = int(line[2])
+            return length_distribution_dict
+        except Exception as e:
+            print(e)
+            log.warning("Could not parse mapDamage length distribution file: '{}'".format(f["fn"]))
+            return None
+
+    # Parse input files
+    def parseInput(self, f):
```
vladsavelyev, line 134:
```suggestion
    def parse_logs(self, f):
```

multiqc/modules/mapdamage/mapdamage.py
```diff
@@ -0,0 +1,282 @@
...
+                name="Reverse read length distribution",
+                description="Read length distribution for reverse strand (-) reads.",
+                helptext="""
+                This plot shows the read length distribution of the reverse reads in the investigated sample. Reads below lengths of 30bp are typically filtered, so the plot doesn't show these in many cases. A shifted distribution of read lengths towards smaller read lengths (e.g around 30-50bp) is also an indicator of ancient DNA.
+                """,
+                plot=self.lgdistplot(self.lgdist_rv_data, "Reverse"),
+            )
+
+    ## Parse misincorporation file into a dict with 1 key ('dmg_5p' or 'dmg_3p') and 1 value (list of 2nd column values)
+    def parseMisincorporation(self, f):
```
vladsavelyev, line 88:
```suggestion
    def parse_misincorporation(self, f):
```

Please rename other methods declarations and calls similarly :)

multiqc/modules/mapdamage/mapdamage.py
```diff
@@ -0,0 +1,282 @@
...
+                name="Forward read length distribution",
+                description="Read length distribution for forward strand (+) reads.",
+                helptext="""
+                This plot shows the read length distribution of the forward reads in the investigated sample. Reads below lengths of 30bp are typically filtered, so the plot doesn't show these in many cases. A shifted distribution of read lengths towards smaller read lengths (e.g around 30-50bp) is also an indicator of ancient DNA.
+                """,
+                plot=self.lgdistplot(self.lgdist_fw_data, "Forward"),
+            )
+            self.add_section(
+                name="Reverse read length distribution",
+                description="Read length distribution for reverse strand (-) reads.",
```
vladsavelyev, line 80: Minor detail - using a minus sign Unicode character instead of a dash (so it's the same size as the plus sign and looks neater):
```suggestion
                description="Read length distribution for reverse strand (−) reads.",
```

vladsavelyev (approved):
Great stuff! Thanks for putting work into this module, adding proper formatting to columns and detailed description texts for all plots and metrics. 

Minor comments on code style and some suggestions, otherwise it looks good to me.

multiqc/modules/mapdamage/mapdamage.py
```diff
@@ -0,0 +1,282 @@
...
+                        values.append(float(line[1]))
+            misincorporation_dict["dmg_" + readend] = values
+            return misincorporation_dict
+        except Exception as e:
+            print(e)
+            log.warning("Could not parse mapDamage misincorporation file: '{}'".format(f["fn"]))
+            return None
+
+    ## Parse length distribution file into a dict with 2 keys ('lendist_fw' and 'lendist_rv') and 1 value each (dict of {length : count})
+    def parseLengthDistribution(self, f):
```
TCLamnidis, line 110:
```suggestion
    def parse_length_distribution(self, f):
```

multiqc/modules/mapdamage/mapdamage.py
```diff
@@ -0,0 +1,282 @@
...
+            # Add 3' G to A data
+            self.threepGtoAfreq_data[s_name] = self.parseMisincorporation(f)["dmg_3p"]
+
+        elif f["fn"].endswith("_freq.txt") and f["fn"].startswith("5p"):
+            # Add 5' C to T data
+            self.fivepCtoTfreq_data[s_name] = self.parseMisincorporation(f)["dmg_5p"]
+
+        elif f["fn"] == "lgdistribution.txt":
+            # Add lendist forward & reverse
+            lgdist_data = self.parseLengthDistribution(f)
```
TCLamnidis, line 152:
```suggestion
        if f["fn"].endswith("_freq.txt") and f["fn"].startswith("3p"):
            # Add 3' G to A data
            self.threepGtoAfreq_data[s_name] = self.parse_misincorporation(f)["dmg_3p"]

        elif f["fn"].endswith("_freq.txt") and f["fn"].startswith("5p"):
            # Add 5' C to T data
            self.fivepCtoTfreq_data[s_name] = self.parse_misincorporation(f)["dmg_5p"]

        elif f["fn"] == "lgdistribution.txt":
            # Add lendist forward & reverse
            lgdist_data = self.parse_length_distribution(f)
```

TCLamnidis:
The current header of the general stats fields is `mapDamage_mqc-generalstats-mapdamage-mapdamage_3_Prime1` which seems rather superfluous, but I'm not sure where to limit the mapdamages 😅 
I guess something like `mapDamage_mqc-generalstats-mapdamage-3_Prime1` might be enough to differentiate the columns from those of DamageProfiler though

ewels:
Looks like the CI tests are failing because there's no `self.ignore_samples()` call (see [docs](https://multiqc.info/docs/development/modules/#filtering-by-parsed-sample-names)).

ewels:
Ah sorry, just looked at the code. It's not the absence of ignoring samples, it's because you're not raising a `UserWarning` when no samples are found. See the [docs](https://multiqc.info/docs/development/modules/#no-files-found).

multiqc/modules/mapdamage/mapdamage.py
```diff
@@ -0,0 +1,291 @@
...
+        self.lgdist_rv_data = self.ignore_samples(self.lgdist_rv_data)
+
+        ## Stop computations if there is no data after ignoring samples
+        if (
+            len(self.threepGtoAfreq_data) == 0
+            or len(self.fivepCtoTfreq_data) == 0
+            or len(self.lgdist_fw_data) == 0
+            or len(self.lgdist_rv_data) == 0
+        ):
+            raise UserWarning
```
ewels, line 51: @TCLamnidis - is this meant to be `or`? Currently this will not produce a report if any of these four are empty. 

Typically I'd be used to the opposite - make a report if we found any data at all..

Probably doesn't make any difference, but I thought I'd ask.

ewels:
```suggestion
        ## Stop computations if there is no data after ignoring samples
        if (
            len(self.threepGtoAfreq_data) == 0
            and len(self.fivepCtoTfreq_data) == 0
            and len(self.lgdist_fw_data) == 0
            and len(self.lgdist_rv_data) == 0
        ):
            raise UserWarning
```

ewels (approved):

Issue #1865, opened 2023-02-09
DRAGEN changing name is not working


NicoleGruenheit:
### Description of bug

I'm trying to change the name of the default file pattern for `dragen/wgs_contig_mean_cov` using a config file:

```yml
sp:
  dragen/wgs_contig_mean_cov:
    fn: 'contig_mean_test.csv'
```

The error message is:
```
The last file found was: ./contig_mean_test.csv                                          │
                                                                                                                                        
 Traceback (most recent call last):                                                                                                     
   File "/usr/local/lib/python3.11/site-packages/multiqc/multiqc.py", line 654, in run                                                  
     output = mod()                                                                                                                     
   File "/usr/local/lib/python3.11/site-packages/multiqc/modules/dragen/dragen.py", line 89, in __init__                                
     samples_found |= self.add_coverage_per_contig()                                                                                    
   File "/usr/local/lib/python3.11/site-packages/multiqc/modules/dragen/coverage_per_contig.py", line 17, in add_coverage_per_contig    
     s_name, perchrom_data_by_phenotype = parse_wgs_contig_mean_cov(f)                                                                  
   File "/usr/local/lib/python3.11/site-packages/multiqc/modules/dragen/coverage_per_contig.py", line 167, in parse_wgs_contig_mean_cov 
     sample, phenotype = m.group(1), m.group(2)                                                                                         
 AttributeError: 'NoneType' object has no attribute 'group'
```

I think the problem is that in line 167 the pattern of the file is hard coded?

```py
m = re.search(r"(.*).wgs_contig_mean_cov_?(tumor|normal)?.csv", f["fn"])
```

The file works if the filename ends with `.wgs_contig_mean_cov.csv`

Cheers,
Nicole

### File that triggers the error

[contig_mean_test.csv](https://github.com/ewels/MultiQC/files/10697648/contig_mean_test.csv)

### MultiQC Error log

```console
docker run -u $UID_USER:$GID_BFX --rm -it -v `pwd`:`pwd` -w `pwd` ewels/multiqc:dev multiqc . -i S000021_S1163Nr32_ignore -f --config config1.txt

[2023-02-09 13:05:30,806] multiqc                                            [DEBUG  ]  This is MultiQC v1.15.dev0
[2023-02-09 13:05:30,810] multiqc                                            [DEBUG  ]  Loading config settings from: config1.txt
[...]

PR #1994, opened 2023-08-21
DRAGEN: use sample name cleaning patterns instead of hardcoded ones

vladsavelyev:
Add `fn_clean_exts` for DRAGEN sample name cleaning, instead of hardcoding patterns into the code.

Fixes https://github.com/ewels/MultiQC/issues/1865


ewels:
Suggestion is to have a top level config attribute called `dragen_config` with the search patterns under there, rather than changing the search pattern config structure.

So something like:

```yaml
dragen_config:
  filename_suffixes:
    - _tumor
```

vladsavelyev:
Actually, there is no need to have `_tumor` and `_normal` as the cleaning patterns at all, since `s_name = s_name.split(ext["pattern"], 1)[0]` strips a suffix even if it's in the middle of a sample name (that feels even more dangerous, though!).

So removing merging the list of DRAGEN cleaning patterns with the main least, and removing `_tumor` and `_normal` from it.

vladsavelyev:
@ewels, good to take another look now.

multiqc/utils/config_defaults.yaml
```diff
@@ -256,7 +256,26 @@ fn_clean_exts:
...
+  - ".wgs_contig_mean_cov"
+  - ".overall_mean_cov"
+  - "_coverage"
+  - ".wgs_coverage"
+  - ".wgs_fine_hist"
+  - ".fragment_length_hist"
+  - ".mapping"
+  - ".gc"
+  - ".trimmer"
+  - ".time"
```
ewels, line 273: This one feels maybe a little risky - would it work if we moved it to `fn_clean_trim` so that it only gets removed if it's at the very end of a sample name? Just in case people have sample names like `mysample.timepoint1` for example 🤔

ewels (approved):
Awesome, thanks 👌🏻 It was a good discussion earlier, I'm still a little on the fence! Let's keep it in the back of our minds when we do some of the heavier refactoring in v2. Happy to reconsider the best approach.

I think there are a couple which are maybe a bit risky, in that I could imagine them legitimately coming up in some sample names (`.time`, `.gc`, `.vc`) - would it work to shift them to the other config where they're only removed from the end of sample names and not the middle onwards? I've done that to compromise with other similarly risky ones in the past.

multiqc/utils/config_defaults.yaml
ewels: I _think_ this should work as the suffixes are removed in order. So from `*.time_metrics.csv` the `_metrics.csv` would be removed from the `_metrics` above here, leaving `.time` to be removed afterwards by `fn_clean_trim`...

ewels: Alternatively we could be less conservative and list the combinations here. eg. `vc_metrics`, `gc_metrics`, `time_metrics` etc..

vladsavelyev: Okay, now it looks better

vladsavelyev: The problem with `quant_metrics` is that we also have to add `quant.metrics.csv` due to the older DRAGEN format, and it will get a bit bloated. But maybe why not

ewels: Great 👍🏻

PR #2000, opened 2023-08-24
CI: only run full workflow on Linux 3.11. Add helpful messages

vladsavelyev:
1. Slightly refactor `multiqc_linux.yml` and `multiqc_windows.yml` to avoid code duplication: combine them into one file, only run the test variations on Linux. Additionally, set up to only run test variations with the latest Python version.
2. Add messages showing the reason why no output was produced when no output is expected.


.github/workflows/multiqc.yml
```diff
@@ -0,0 +1,177 @@
...
+      - name: All modules / Log filename as s_name, no cleaning
+        if: matrix.python-version == env.latest_python
+        run: multiqc ${{ needs.changes.outputs.single_module }} --lint test_data/data/modules/ --fullnames --fn_as_s_name
+
+      - name: Filter out all filenames (confirm no report)
+        if: matrix.python-version == env.latest_python
+        run: |
+          multiqc ${{ needs.changes.outputs.single_module }} test_data/data/modules/ --filename all_ignored.html --ignore-samples '*'
+          if [[ -f all_ignored.html ]]; then
+            echo "Either self.ignore_samples() was not called in a module, or UserWarning was not raised by the module when no matching files are found" >&2
```
ewels, line 101:
```suggestion
            echo "No report should have been generated! Likely due to a module missing self.ignore_samples() or UserWarning when no matching files are found" >&2
```

ewels (approved):
Code LGTM, thanks 👍🏻 

One day maybe we'll have a solid unit test suite and we can reduce the number of runs we do here 😅

ewels:
Total CI run time is now 19 minutes, down from a total of 51 minutes before this change. Considering that this is running every time a commit is pushed to a PR / master, that's quite a significant amount of compute time saved! (Would have been even more prior to the search speedup 😅 ).

Now:

CleanShot 2023-08-24 at 21 11 20@2x

Before:

CleanShot 2023-08-24 at 21 11 47@2x

CleanShot 2023-08-24 at 21 11 56@2x

PR #2002, opened 2023-08-24
Log the absolute path to report in verbose mode

vladsavelyev:
When debugging MultiQC in PyCharm, run configurations are often used, and their working directory might be different from the working directory of the terminal. So it's helpful to log the absolute path in the end in order to quickly copy and open a report in the browser.

Since I'm not sure if there are other use cases for this, not insisting on it - feel free to just close.


ewels (approved):
Yup, like it 👍🏻

PR #2004, opened 2023-08-25
fix parsing html string in json custom content

jchorl:
Parsing json custom content with html fails

I just copied the logic from yaml cc parsing to json.

I suspect this isn't caught in testing because the html cc tests seem to be yaml: https://github.com/ewels/MultiQC_TestData/tree/master/data/custom_content/embedded_config

I'm not super caught up on mqc testing though.

thanks as always!

- [x] `CHANGELOG.md` has been update

If you have good ideas for testing, I might be able to give it a stab. I'm a bit slammed right now.


vladsavelyev (approved):
Thank you a lot @jchorl for the fix!

I added minor refactoring and moved the repetitive check outside the if clauses, and also dropped the `OrderedDict` (at some point we will go through the entire code to get rid of `OrderedDict`s for good 😓 )

PR #2015, opened 2023-08-31
fix(conpair): don't re search entire lines

ewels:
Fix the search slow down.

TODO: Check if it still matches all the files it needs to match 👀


vladsavelyev (approved):
I tested - works correctly for the old and new test example.

PR #2017, opened 2023-08-31
refactor(core) add color scale `Highcharts`, load it instead of hardcoding colors for plots

vladsavelyev:


multiqc/plots/bargraph.py
```diff
@@ -472,8 +460,8 @@ def matplotlib_bargraph(plotdata, plotsamples, pconfig=None):
                         prevdata[i] += prev_values[i]
                 # Default colour index
                 cidx = idx
-                while cidx >= len(default_colors):
-                    cidx -= len(default_colors)
+                while cidx >= len(scale.colours):
+                    cidx -= len(scale.colours)
```
ewels, line 464: I wonder if we can move this logic into `get_colour()` (allowing idx > array length, using remainder).

multiqc/utils/mqc_colour.py
```diff
@@ -382,6 +384,18 @@ def get_colours(self, name="GnBu"):
                 "#fddaec",
                 "#f2f2f2",
             ],
+            "Highcharts": [
```
ewels, line 387: Lowercase?
```suggestion
            "highcharts": [
```

ewels: Or maybe nicer to drop the name "Highcharts" completely - they have updated their default colour palette since this, plus I guess we will continue to use these colours in Plotly..

```suggestion
            # Originally from Highcharts
            "plot_defaults": [
```

ewels (commented):
Looking great!

multiqc/plots/bargraph.py
vladsavelyev: This logic is already there, actually, so can just remove this block.

multiqc/modules/preseq/preseq.py
```diff
@@ -348,25 +349,14 @@ def _prep_real_counts(real_cnts_all, real_cnts_unq, is_basepairs, counts_in_1x,
...
-    ]
+    scale = mqc_colour.mqc_colour_scale("Highcharts")
+
     series = []
     for si, sn in enumerate(sorted(data.keys())):
         if sn in xs_by_sample:
             x = float(xs_by_sample[sn])
             point = {
-                "color": default_colors[si % len(default_colors)],
+                "color": scale.get_colour(si),
```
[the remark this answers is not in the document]
vladsavelyev: It's there already!

multiqc/plots/bargraph.py
```diff
@@ -183,6 +183,14 @@ def plot(data, cats=None, pconfig=None):
         logger.warning(f"Tried to make bar plot, but had no data: {pconfig.get('id')}")
         return '<p class="text-danger">Error - was not able to plot data.</p>'
 
+    # Add colors to the categories if not set. Since the "plot_defaults" scale is
+    # identical to default scale of the Highcharts JS library, this is not strictly
```
ewels, line 187: NB: It is needed even before updating / changing highcharts, as the static plots use matplotlib and not highcharts. So that code needs these colours.

PR #2018, opened 2023-08-31
Spelling standardization

vladsavelyev:
Work on spelling americanization: 

- [x] color -> colour


ewels:
😭 🇬🇧 

I'm thinking that this might be worth pushing to v2, together with any other spelling standardi- _z_ -ations. As it's technically a breaking change, it might affect plugins / custom modules that call these functions.

Issue #1987, opened 2023-08-16
Better changelog management


ewels:
A classic problem with managing pull-requests in this project is merge conflicts arising from changelog entries.

This isn't a new problem, and there are tools that have been written to address this. For example:

- [Scriv](https://scriv.readthedocs.io/)
- [Towncrier](https://towncrier.readthedocs.io/)
- ..probably others.

It'd be good to have a play with these and see if we like them / if they're worth the effort of setting up and using.

I suspect that many contributors won't want to use these tools locally. So ideally it would be paried with a GitHub action where a changelog entry can be added to a PR via a special comment on the PR or something. That'd be nice.

vladsavelyev:
The way I see it is that when we are preparing a release, we run a script that takes descriptions of all merged pull requests, and simply populates a change log. 

So care should be taken into writing the pull request titles. And they should be prefixed with labels like `New module: Samtools` and `Samtools: fix parsing of input files` (or `[New module] Samtools`/`[Samtools]: fix parsing of input files`)

ewels:
Let's use standard semantic commit message structure: https://www.conventionalcommits.org/

Can then use stuff like https://github.com/marketplace/actions/semantic-pull-request

Need to think about our own categories, for example to distinguish between adding a new module and adding a new feature to an existing module.

vladsavelyev:
I tried https://www.conventionalcommits.org for a few PRs, and it doesn't look very aesthetic to me:

<img width="859" alt="Screenshot 2023-09-02 at 23 46 27" src="https://github.com/ewels/MultiQC/assets/1575412/07ef3341-54cf-424d-abe7-f183055b45a5">

----

I think their model implies that there is a limited number of possible "scopes" that you can list in the CI config: like `ui`, `api`, `backend`, etc. For us, we have modules that are added all the time, and using a module name as a scope in parentheses looks ugly'ish. Also, constructions like `feat(core)` look a bit unnecessary geeky and could be confusing to contributors who only start programming.

Also, it's not clear how to name a "new module" PR in their model - and that's the most frequent type of PR in MultiQC.

Maybe we could stick to the following - pretty much the existing but not strictly followed - guideline:
* `New module: Bakta` (must start with the words "New module: " or similar, will be listed under the corresponding section in the change log).
* `Kraken: Support new format` (if a PR name starts with a name of an existing module, it will go under "Module updates" in the change log).
* Everything else will go under "MultiQC updates" in the changelog.

[...]

PR #2025, opened 2023-09-04
Add action to populate the change log from PR titles triggered by `@multiqc-bot changelog`

vladsavelyev:
Following the [discussion](https://github.com/ewels/MultiQC/issues/1987).

Add a CI action triggered when a PR is approved. 

A new line will be added either under "MultiQC updates", "New Modules", or "Module updates", so it pretty much keeps the existing structure, in it's simplest form (i.e. not separating between classes of core updates). That new line will be committed into the same branch.

I decided not to use any of the existing actions on the marketplace, as configuring them gets too hacky and/or puts too much restrictions.

Updated on 24 Sep 2023:

Triggering a workflow on review approval is not working for forks. Unfortunately such event is triggered on the PR merge branch, so the secrets are not available to the workflow for safety. So working around this by triggering the workflow on a comment. The idea is to support a `@multiqc-bot changelog` command. The command can take an optional argument. It will trigger a workflow that:

- [x] Determines the section where the entry should go.
   - [x] If the PR title starts with "New module:" and a single module was added, the change will go into "New modules"
   - [x] If a single module was altered and the PR title starts with this module's name, it will go into "Module updates" 
   - [x] Otherwise, it will go into general "MultiQC updates".
- [x] Checks if an entry is already added in the changelog (a comment always contains the link to the PR), and adds or replaces an entry based on this.
- [x] If additional section was specified (`@multiqc-bot changelog An amazing change`), it will take it as a description; otherwise,  takes a changelog entry from the PR title.


.github/workflows/changelog.py
```diff
@@ -0,0 +1,202 @@
...
+if pr_title.endswith("(chore)") or pr_title.endswith("(docs)"):
+    print("Skipping logging this change as it's a chore or docs update")
+    sys.exit(0)
+
+
+def find_module_info(module_name):
+    """
+    Helper function to load module meta info. With current setup, can't really just import
+    the module and call `mod.info`, as the module does the heavy work on initialization.
+    But that's good - we avoid installing and importing MultiQC here, and the action runs faster.
```
ewels, line 48: okay-seinfeld-gif

😆

.github/workflows/changelog.py
```diff
@@ -0,0 +1,202 @@
...
+    with py_path.open("r") as f:
+        contents = f.read()
+    if not (m := re.search(r'name="([^"]+)"', contents)):
+        return None
+    name = m.group(1)
+    if not (m := re.search(r'href="([^"]+)"', contents)):
+        return None
+    url = m.group(1)
+    if not (m := re.search(r'info="([^"]+)"', contents)):
+        if not (m := re.search(r'info="""([^"]+)"""', contents)):
```
ewels, line 73: Will this match strings including newlines?

.github/workflows/changelog.py
```diff
@@ -0,0 +1,202 @@
...
+    if mod is not None:
+        section = "### Module updates"
+        pr_title = pr_title.split(":")[1].strip().capitalize()
+
+# Now that we determined the PR type, preparing the change log entry.
+pr_link = f"([#{pr_number}]({REPO_URL}/pull/{pr_number}))"
+if section == "### New modules":
+    new_lines = [
+        f"- [**{mod['name']}**]({mod['url']})\n",
+        f"  - {mod['name']} {mod['info']}\n",
```
ewels, line 110: I have a feeling that GitHub markdown rendering needs 4 spaces for list indentation. Maybe check, but I think that's the case.

ewels (commented):
Amazing stuff 🤩 

Bonus points in the workflow YAML:

* Don't run if `CHANGELOG.md` has been modified within the pull request (?)
* Add a comment on the PR with the stderr output from the script, saying what happened
    * This will alert us to errors, as otherwise we'd need to go in and check the actions logs manually I think
    * If we include the git commit hash of the pushed commit, it will make it easy to quickly view the diff of the automated update, for sanity checking

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,43 @@
...
+        description: PR number
+        required: true
+env:
+  PR_TITLE: ${{ github.event.pull_request.title }} # this is returning empty
+  PR_NUMBER: ${{ github.event.number }}
+
+jobs:
+  update_changelog:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.merged == true
```
ewels, line 22: Should also check the repository name here, so that it doesn't accidentally run on people's forks.

.github/workflows/changelog.py
vladsavelyev: Yep! Tested :)

.github/workflows/changelog.py
vladsavelyev: Our changelog file uses 2 spaces there: https://raw.githubusercontent.com/ewels/MultiQC/master/CHANGELOG.md

As long as they are not tabs!

vladsavelyev:
Alright, it works now. Tested here https://github.com/vladsavelyev/MultiQC/pull/21 :)

ewels (changes requested):
Looks like a couple of lines in the Changelog got messed with in this PR

ewels (commented):
It would be good to document how this works for the developer in https://multiqc.info/docs/development/contributing/

ewels (approved):
LGTM once we have some docs 👍🏻

.github/workflows/changelog.py
```diff
@@ -0,0 +1,313 @@
...
+    info = m.group(1)
+
+    # Reduce consecutive spaces and newlines.
+    info = re.sub(r"\s+", " ", info)
+    return {"name": name, "anchor": anchor, "url": url, "info": info}
+
+
+def _files_altered_by_pr(pr_number, types=None) -> set[Path]:
+    """
+    Returns a list of files added by the PR.
```
vladsavelyev, line 79:
```suggestion
    Returns a list of files added or modified by the PR (depending on `types`, 
    which can be a subset of `{'added', 'modified'}`
```

PR #2030, opened 2023-09-06
File search: fix off-by-1 error in num_lines

vladsavelyev:
When a search pattern has `num_lines: 1`, it actually makes it check 2 lines

Normally it's alright as the `num_lines` option was originally thought to optimise the file search.

However, to tell Kraken output from Bracken, we actually have to rely on this.

Tested on all existing modules, doesn't break anything.


ewels (approved):
LGTM, thanks 👍🏻

Issue #1051, opened 2019-11-11
Refactor code for use within interactive environments


ewels:
Now that #1042 is merged, I think that we should be able to import MultiQC into a Jupyter Notebook and use sort of interactively. It would be nice to refactor the code so that it can be used for specific plots rather than _only_ generating a big report, though I don't know how easy this will be.

- [x] Make a new repository for an example notebook
- [x] Get a basic notebook that imports MultiQC and just runs it on some data to generate a report
- [x] Look into whether we can access results in the `multiqc.report` object
- [x] Figure out whether we can print individual plots and result tables
- [x] Refactor the code to make the above more intuitive
- [x] Interactively add and remove sections using custom data (https://github.com/MultiQC/MultiQC/issues/2452)
- [x] Write documentation

This could be quite a bit task with all of that, but would be good to get at least a basic run working.

ewels:
This needs fixing for good Jupyter usage: https://github.com/ewels/MultiQC/issues/1141

ewels:
Tightly linked to the refactoring described in https://github.com/ewels/MultiQC/issues/1790

ewels:
Maybe related, just stumbled across https://github.com/olkuka/multiqc_jupyterlab/ - may be useful.

PR #2036, opened 2023-09-11
Core: Refactor code for use within interactive environments

vladsavelyev:
Moving towards the [interactive use](https://github.com/ewels/MultiQC/issues/1051):

* Move the module report building logic out of `__init__()`
* Make `Report` a normal class instead of a module singleton


ewels:
Wondering if we can add in logic here to check if the `build()` def exists within a module. If so then run it the new way. However, if it's _not_ there, then run the old way with `__init__()`.

This way, the change is not a breaking change and we can push the refactoring before v2.0, essentially as progressive enhancement. We can also use it as an opportunity to log depreciation warnings so that folks know that the new syntax is ready and should be switched to.

Things like lint tests can abandon the old syntax completely, as all _core_ MultiQC modules should be using the new syntax immediately.

ewels:
Outdated, replacement now merged 🚀

Issue #2032, opened 2023-09-08
[Question]: How to customize "group column name" for "Bcftools Stats"


tamuanand:
Hi

Given below is a snippet from my multiqc_config.yaml  

When I run MultiQC (1.15), the general stats table shows SAMTools (after Recalibrate) in the Group column, but does not show the same for BCFTools (pre-Normalization) or for BCFTools (post-Normalization) - screenshot attached. 

Do I have to do anything different for bcftools? Thanks in advance.
```yaml
- samtools:
    name: 'SAMTools (after Recalibrate)'
    info: 'This section of the report shows SAMTools results after Recalibrate'
    path_filters:
        - '*samtools_recalibrated*'
- bcftools:
    name: 'BCFTools (pre-Normalization)'
    info: 'This section of the report shows BCFTools results before normalization'
    path_filters:
        - '*bcftools_raw*'
- bcftools:
    name: 'BCFTools (post-Normalization)'
    info: 'This section of the report shows BCFTools after normalization'
    path_filters:
        - '*bcftools_normalized*'
 ```

![image](https://github.com/ewels/MultiQC/assets/8467214/b998a810-23e8-4259-a8c8-7a650ebd02e7)

vladsavelyev:
While we are working on a bore robust solution, I added a quick fix that works for bcftools: https://github.com/ewels/MultiQC/pull/2043

tamuanand:
Thanks a lot @vladsavelyev

PR #2037, opened 2023-09-13
Append `namespace` in general stats to the module name

vladsavelyev:
More robust fix to https://github.com/ewels/MultiQC/issues/2032

I think we overriding the `namespace` parameter in the code for the general stats is not really good, and it's much clearer when the actual module name is displayed there (or whatever the user sets explicitly in their custom config, e.g. when using groups).

The only real scenario is the module like DRAGEN where there are so many submodules involved, that you might want to distinguish them somehow in the table customisation view. But such bloated modules are not ideal anyway, I would even look into breaking DRAGEN into separate modules. Or we can add some native support for submodules into MultiQC.

UPD: actually, `namespace` can be also set in in column configurations, so there is another way to override it in the module, which also breaks the groups functionality. Removed that as well. The only place where it's used is to distinguish the CellRanger's "Count" and "VDJ" submodules in the general stats; but that can be done in by adding prefixes to the column titles and descriptions.

UPD: as the best of both worlds, keeping the `namespace` parameter, but appending it to the module name (or whatever it is is overridden with in the grouping config).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 - Fixed parsing of `plot_type: "html"` `data` in json custom content
 - Add software version information to the MultiQC report (thanks [pontushojer](https://https://github.com/pontushojer) for the [contribution](https://github.com/ewels/MultiQC/pull/1927)). Modules can parse versions from logs, and users can pass versions with `config.software_versions`.
 - Replace deprecated `pkg_resources`
-- [Fix](https://github.com/ewels/MultiQC/issues/2032) module groups configuration where namespace is set explicitly in `general_stats_addcols`. Namespace is now always appended to the module name in general stats ([2037](https://github.com/ewels/MultiQC/pull/2037))
+- [Fix](https://github.com/ewels/MultiQC/issues/2032) the module groups configuration for modules where the namespace is passed explicitly to `general_stats_addcols`. Namespace is now always appended to the module name in the general stats ([2037](https://github.com/ewels/MultiQC/pull/2037)).
 
 ### New Modules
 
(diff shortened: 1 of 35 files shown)


ewels:
What is the namespace used for? Will a lot of people have this in their config files and have them suddenly break? 🤔  Need to go back and refresh my memory about the attribute....

vladsavelyev:
> Will a lot of people have this in their config files and have them suddenly break?

I don't think it's configurable via the config? It can be set in the modules code, and people use it to label columns in their custom tables - which appears before the colon in the column pop-up text, as well as the first text column in the table configuration:

Screenshot 2023-09-14 at 10 13 44



Some modules also override it for the columns in general stats, which I think is undesirable, and the module name should go there (unless the module name is overridden with a `path_filters ` config, of course).

ewels:
What about when the module name is not informative? eg. `custom_content`. Maybe we can make that one a special case 🤔 

What I'm wondering is whether we leave the feature in tact for now, but just remove the usage from the modules.. Or whether it's best to completely kill it as you have here.

ewels:
Using it to identify parts of massive modules, eg. which Picard tool generated a column in general stats, sounds familiar too.

What if we made `namespace` a suffix instead? Then module renaming would work (https://github.com/ewels/MultiQC/issues/2032) but we could keep the benefits of clarity in big modules / custom content..

vladsavelyev:
Yeah, we can look into using it as a suffix. I'll take a look into that! 

To unblock the original issue, I just removed the namespace parameter where it's truly redundant to resolve the original issue: https://github.com/ewels/MultiQC/pull/2043

vladsavelyev:
Alright, it works nice if appended as a suffix. For general stats, we want to show both the module name and the namespace if provided. For other tables objects, we just show the namespace, as the module name is in the section title already.

Screenshot 2023-09-14 at 14 50 09

Screenshot 2023-09-14 at 14 50 29

ewels (approved):
Really nice stuff, thanks 👍🏻

PR #2043, opened 2023-09-14
Remove redundant namespace in `general_stats_addcols` usages

vladsavelyev:
Remove the namespace parameter for `general_stats_addcols` in a few cases where it's redundant (as it matches the default value equivalent to the module name). Fixes https://github.com/ewels/MultiQC/issues/2032

A more robust solution to namespaces might take a bit longer to discuss and implement, so leaving https://github.com/ewels/MultiQC/pull/2037 for that.

Not adding a changelog entry as I expect the more robust solution to be a part of the next release as well.


ewels (approved):
Thanks 👍🏻

PR #2044, opened 2023-09-14
Elembio bases2fastq

ewels:
We don't have permission to push directly to the source fork in https://github.com/ewels/MultiQC/pull/1990, so creating this branch + PR to continue work so that we can push minor changes directly to be able to finish the merge.

Please see https://github.com/ewels/MultiQC/pull/1990 for discussion and context.


vladsavelyev:
What I don't like is the way is the group names are broken up:

Screenshot 2023-09-15 at 10 44 57

It would be nice to have an option to make the column take the width of the widest cell, and not the header width, similar to the first column.

ewels:
Set `scale: False` and everything will work as you'd expect.

vladsavelyev:
On more issue is that when run on all test inputs, it reports that no samples are found:

```
[2023-09-15 11:33:31] multiqc.modules.bases2fastq.bases2fastq            [ERROR  ]  More than one read length configurations are found in the dataset:150+150,75+75
[2023-09-15 11:33:31] multiqc.modules.bases2fastq.bases2fastq            [ERROR  ]  These runs have 150+150 read length configuration:DVT-0008__2d47,DVT-0013__04fd,DVT-0015__1fa3
[2023-09-15 11:33:31] multiqc.modules.bases2fastq.bases2fastq            [ERROR  ]  These runs have 75+75 read length configuration:DVT-0023__38a4,DVT-0019__85ac,DVT-0018__e1b7,DVT-0020__f923,DVT-0031__0e11,DVT-0030__8a89,DVT-0025__2b41,DVT-0024__17cf,DVT-0021__3f2c,DVT-0022__08c4
[2023-09-15 11:33:31] multiqc                                            [DEBUG  ]  No samples found: bases2fastq
```

Probably in this case it should raise an exception to be more explicit about the errors.

vladsavelyev (approved):
Looks great!

ewels:
@vladsavelyev - getting the same thing here as we just fixed in bclconvert, where the module exits with no results if there are mixtures of runs:

```console
❯ multiqc -f .

  /// MultiQC 🔍 | v1.16.dev0 (de0264c)

|           multiqc | Search path : /Users/ewels/GitHub/MultiQC/TestData/data/modules/bases2fastq
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 303/303
|       bases2fastq | Found 13 total RunStats.json
|       bases2fastq | More than one read length configurations are found in the dataset:150+150,75+75
|       bases2fastq | These runs have 150+150 read length configuration:DVT-0008__2d47,DVT-0013__04fd,DVT-0015__1fa3
|       bases2fastq | These runs have 75+75 read length configuration:DVT-0023__38a4,DVT-0019__85ac,DVT-0018__e1b7,DVT-0020__f923,DVT-0031__0e11,DVT-0030__8a89,DVT-0025__2b41,DVT-0024__17cf,DVT-0021__3f2c,DVT-0022__08c4
|           multiqc | No analysis results found. Cleaning up..
```

Do you think we can refactor to be able to still generate a report successfully in this case?

multiqc/utils/search_patterns.yaml
```diff
@@ -20,6 +20,20 @@ bamtools/stats:
...
+bases2fastq/run:
+  fn: "RunStats.json"
+  contents_re: ".*NumPolonies.*"
+  num_lines: 100
+bases2fastq/project:
+  fn: "*_RunStats.json"
+  contents_re: ".*NumPolonies.*"
+  num_lines: 100
+bases2fastq/group:
+  fn: "*_b2fgroup.csv"
```
vladsavelyev, line 36: We don't seem to have any test data for the group file, do we? 
https://github.com/ewels/MultiQC_TestData/tree/master/data/modules/bases2fastq

Would it be possible to add some? @blajoie

vladsavelyev (changes requested):
Did one last pass through and added a few changes:

* Worked around the use of pandas which we don't install along with MultiQC,
* Got rid of `UserWarning`. We want to have only one `UserWarning` per module, which indicates that no data is found. Otherwise, a module should ideally do its best to generate a report for whatever data it found. It doesn't seem like it depends on the read lengths to be identical across all samples, so we can get rid of the `UserWarning` as well.

What we are missing is some test data for the following:

```yaml
bases2fastq/project:
  fn: "*_RunStats.json"
  contents_re: ".*NumPolonies.*"
  num_lines: 100
bases2fastq/group:
  fn: "*_b2fgroup.csv"
```

It would be nice to add it, otherwise a fair share of the module code is not tests at all.

ewels:
Still not able to generate a report from the test data:

```
  /// MultiQC 🔍 | v1.16.dev0 (2e8e5dc)

|           multiqc | Search path : /Users/ewels/GitHub/MultiQC/TestData/data/modules/bclconvert
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 30/30
|        bclconvert | Detected different read lengths across the RunXml files. We cannot handle this - read length across all input directories must be the same.
|           multiqc | No analysis results found. Cleaning up..
```

ewels:
@blajoie / @YuheCheng62 - we are close to getting the module merged now on the code side. However, the size of the test data is presenting some fairly significant problems.

For starters, it's bloated the MultiQC test data a lot, making all the CI checks run quite a lot slower. It's 141MB, which is 30% of the size of the entire test data repo - for one module out of 130 😓 

Also the module runs very slowly. Probably the most comparable MultiQC module is bclconvert, and that runs in `0.34` seconds. But bases2fastq takes `20.56` seconds. That is really too slow, when one of the golden rules of MultiQC modules is the _"Module should not not do any significant computational work"_. This is bad for simple stuff like CI tests, but also bad for people running MultiQC at scale.

@blajoie - we spoke on Slack about potentially being able to generate a different flavour of outputs, which are not huge files containing _everything_ in one place. But instead having a lightweight `*.bases2fastq-metrics.json` file which would be appropriate for MutliQC. This would be:

* Faster to parse
* Smaller for the CI
* Not require user-config overrides to push up the maximum log filesize

Is this a possibility in the near-ish future? I think that the report filesize may be a blocker for merging into MultiQC as it stands..

blajoie:
Sorry all - had to pivot to some other high priority work.  Hoping to come back to this next week to finish this off! 

Appreciate your patience on this ! :)

ewels:
Quick update after a catch up with @blajoie and @rosibaj. The example file used for this PR was kind of extreme, so knowing this means that most of the blockers we had with it are not actually blockers..

* A typical JSON file is more like 1-2MB, instead of 120MB. Parsing speed shouldn't be a blocker
* 98% of typical user JSON files should come in under the 50MB default limit, so no config override needed for most users
* We now have new test data for CI which is a fraction of the size

So now we have the new test data, I think that we can bring this PR up to speed and get it merged. Once the conflicts are resolved and it's been updated with new syntax and functionality that's been added to MultiQC in the past year or so, we can have a final check on functionality and output.

vladsavelyev:
I merged with main, but can't get it to work - the new shared example files RunStats_MultiQC_WES.json RunStats_MultiQC_WGS.json do not match the search patterns. Should they be updated?

vladsavelyev:
Thanks!

Even after renaming `RunStats_MultiQC_WGS.json`/`RunStats_MultiQC_WES.json` to `RunStats.json`, the module won't work as it needs sample- and -project level jsons as well. Wonder if you have those around?

blajoie:
Will take a look and get back to you asap @vladsavelyev   !

ewels:
Closing in favour of https://github.com/MultiQC/MultiQC/pull/3123

PR #2048, opened 2023-09-17
Fix circular imports in `software_versions`

vladsavelyev:
Move the modules `software_versions` and `profile_runtime` into the `modules` module.

Motivation: 

Really by chance, Python doesn't throw a circular import exception with the current setup:

.multiqc _imports_ .utils.software_versions
.utils.software_versions _imports_ .modules.base_module
.modules.base_module _imports_ .utils.software_versions

The reason the circular import error doesn't actually happen seems to be that the MultiqcModule in software_versions doesn't need other utility functions defined in the module, so they are really independent and can be imported separately.

When I started refactoring to add custom exceptions, this circular import became apparent.

I suggest moving the code that defines a `MultiqcModule` into the `modules` folder, keeping the utility functions in `utils`. So `profile_runtime` should also be moved.


ewels (approved):
LGTM 👍🏻

Issue #2047, opened 2023-09-15
Custom exception types


ewels:
Start creating a set of custom exception types to properly handle certain behaviour.

* `ModuleNoSamplesFound` - replace current `UserWarning` usage (add progressively, depreciate `UserWarning` and remove in v2)
* `ModuleFatalError` - a module decides that something is _so_ badly wrong that the entire MultiQC run needs to stop (exit code `1`)

I'm sure that we can think of many more. If we have a proper code structure for them, we can add more in as we go through refactoring.

PR #2049, opened 2023-09-17
Use custom exception type instead of `UserWarning` when no samples are found.

vladsavelyev:
Fixes https://github.com/ewels/MultiQC/issues/2047

Adding `ModuleNoSamplesFound` to replace `UserWarning`.

Also, in this and in a [follow-up PR](https://github.com/ewels/MultiQC/pull/2057), making sure that exception is only raised when there are actually no proper inputs found, following the convention that modules should always either work, or skip samples. Refactor `flash` module to avoid self-catching `UserWarning`s.

Also, following that convention, I'm not adding `ModuleFatalError`.


multiqc/modules/base_module.py
```diff
@@ -18,6 +18,25 @@
...
+
+
+class ModuleBadInputError(Exception):
+    """Something with the input files, so we cannot proceed running the module"""
+
+    def __init__(self):
+        # Overriding the default init to explicitly forbid passing messages.
+        # We want users to print their own messages with log.error(), so it
+        # is gets automatically labelled with the module name in the log.
+        super().__init__()
```
ewels, line 32: Would be nice if we could automate the logging. Would need to update the logger to be able to accept an optional namespace value, and use this over module name if supplied.

Low priority, so if not immediately easy to do we stick with this ☝🏻

multiqc/multiqc.py
```diff
@@ -689,8 +690,10 @@ def run(
                 except AttributeError:
                     pass
 
-        except UserWarning:
```
ewels, line 692: We should still capture this for now, though maybe show a depreciation log message.

multiqc/multiqc.py
```diff
@@ -689,8 +690,10 @@ def run(
...
+        except ModuleNoSamplesFound:
             logger.debug("No samples found: {}".format(list(mod_dict.keys())[0]))
+        except ModuleBadInputError:
+            logger.debug("Could not parse inputs: {}".format(list(mod_dict.keys())[0]))
```
ewels, line 696:
```suggestion
            logger.error("Could not parse inputs: {}".format(list(mod_dict.keys())[0]))
```

ewels: Should maybe raise an exit code 1

multiqc/modules/base_module.py
vladsavelyev: I'd actually stick to a for conventional use of the exception by passing a message:

```py
raise ModuleBadInputError("Mixed 'TOTAL_READS' and 'TOTAL_BASES' reports")
```

And then print the module name in `multiqc.py `:

```py
        except ModuleBadInputError as e:
            logger.error(f"{this_module}: could not run module: {e}")
```

multiqc/multiqc.py
vladsavelyev: Done!

ewels (commented):
Need more documentation - this part describes `UserWarning` so needs updating as a minimum: https://multiqc.info/docs/development/modules/#no-files-found

multiqc/multiqc.py
```diff
@@ -689,8 +690,8 @@ def run(
                 except AttributeError:
                     pass
 
-        except UserWarning:
-            logger.debug("No samples found: {}".format(list(mod_dict.keys())[0]))
+        except (ModuleNoSamplesFound, UserWarning):  # UserWarning deprecated from 1.16
+            logger.debug(f"No samples found: {this_module}")
```
ewels, line 694: Let's throw a debug log message here about the depreciation, maybe use `warnings.DepreciationWarning` ?
```suggestion
        except ModuleNoSamplesFound:
            logger.debug(f"No samples found: {this_module}")
        except UserWarning:  # UserWarning deprecated from 1.16
            logger.debug(f"DEPRECIATED: Please use ModuleNoSamplesFound instead of UserWarning")
            logger.debug(f"No samples found: {this_module}")
```

Actually could also use use `config.lint` here so that we throw an error when running with `--lint`. That would be good to catch this in any new modules etc.

vladsavelyev: Tried with `DeprecationWarning`, it's a bit ugly, and running python with `-W error` also fails on warnings from the library code which is not what we want.

But we already have the mechanism for handling this, as you said: `config.lint` and `report.linting_errors`. Implemented it this way instead.

ewels: Ok, sounds good! 👍🏻

multiqc/multiqc.py
```diff
@@ -689,8 +690,16 @@ def run(
                 except AttributeError:
                     pass
 
-        except UserWarning:
-            logger.debug("No samples found: {}".format(list(mod_dict.keys())[0]))
+        except ModuleNoSamplesFound:
+            logger.debug(f"No samples found: {this_module}")
+        except UserWarning:  # UserWarning deprecated from 1.16
+            msg = f"Please raise ModuleNoSamplesFound instead of UserWarning (in module: {this_module})"
```
ewels, line 696:
```suggestion
            msg = f"DEPRECIATED: UserWarning. Please raise ModuleNoSamplesFound instead in module: {this_module}"
```

ewels (commented):
Needs docs. I'll write something quickly now to get this merged.

ewels (approved):

ewels:
@multiqc-bot changelog

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### MultiQC updates
 
 - Add CI action [changelog.yml](.github%2Fworkflows%2Fchangelog.yml) to populate the changelog from PR titles, triggered by a comment `@multiqc-bot changelog` ([#2025](https://github.com/ewels/MultiQC/pull/2025))
+- Use custom exception type instead of `UserWarning` when no samples are found. ([#2049](https://github.com/ewels/MultiQC/pull/2049))
```
ewels, line 8: ❤️

Issue #2042, opened 2023-09-14
Software Versions: Add support to all modules


ewels:
Now that https://github.com/ewels/MultiQC/pull/1927 is merged and stable, we should go through all modules adding support for reporting version numbers wherever we can!

pontushojer:
I started to work a bit on this in PR #2051

I am using a Google Sheet to make it easier to keep track of which modules has version info, if its been added and any other relevant information. That way it should be easier to collaborate since its soooo many modules to go through. Here is a link: https://docs.google.com/spreadsheets/d/1amPRcK_Lewudyli5ql9We73tC1Txm8eNvjKfQVXW0u0/edit?usp=sharing

Two suggestions
1. Should we add version information if is in a file that is not currently parsed but part of the software output? For example, for  `fastp` we only parse the JSON ATM while version info is only found in the HTML report. I think that we should.
2. As an extension of (1) should we parse the output of something like `tool --version &> sample_name.module_version.log`. This is primarily for tools that have no version info in their normal output. For example `mosdepth` does not output version info but calling `mosdepth --version` produces output from which the version could be parsed. We could allow users to direct this output to a file like this `my_sample.mosdepth_version.log` from which we could get the sample (`my_sample`) and the module (`mosdepth`) require to include sample version info in the report. This would probably be quite a big addition, maybe for its better a separate issue.

ewels:
Just chatting to @vladsavelyev - it's too much work to tick them all off manually here. We should add a lint test to automatically highlight any that are missing.

We could adapt the software versions function to be able to explicitly report `False` or `None` when it's impossible to scrape the version. Then we can hopefully check if the function was called, and fail `--lint` runs if it's not. Then any new modules added will fail CI if the author hasn't added this. Also it provides a way for us to keep track of what's done as we move towards checking them off.

ewels:
Completed, even with the linting to check that every module attempts this ✅ 

Thanks @pontushojer! 🎉

PR #2051, opened 2023-09-18
Software Versions: Add support to all modules

pontushojer:
This PR works towards solving https://github.com/ewels/MultiQC/issues/2042

Version parsing added for modules:

- adaptorremoval
- anglerfish
- bakta
- bbduk
- bcftools
- bismark
- cellranger
- clipandmerge
- cutadapt
- damageprofilier
- dedup
- diamond
- eigenstratdatabasetools
- fastq_screen
- fastqc
- flash
- flexbar
- gffcompare
- gopeaks
- hifiasm
- interop
- longranger
- macs2
- malt
- mirtop
- mtnucratio
- multivcfanalyzer
- pangolin
- purple
- pycoqc
- qc3C
- salmon
- sambamba

---


vladsavelyev:
Wow, that's really amazing!! Thanks so much for doing this, @pontushojer

multiqc/modules/cutadapt/cutadapt.py
```diff
@@ -100,6 +100,7 @@ def parse_cutadapt_logs(self, f):
         s_name = None
         end = "default"
         cutadapt_version = None
+        python_version = None
```
vladsavelyev, line 103: I think we should only report versions of the scientific software, so python version probably is not important

vladsavelyev (approved):
It looks great, approved!

Some late thoughts after looking at the table again, should have been on the original PR, but 😅 
* I'm still not convinced that reporting `htslib` versions for `bcftools` and `samtools` is a useful information, as almost always they going to be identical to the actual tools versions. I'd remove it, or at least add some logic to only show it if it differs 🤔  And similarly, no need to report the Python version for cutadapt.
* Also, another late thought I don't think we should lowercase module names in the table. It's nicer when the names match the sections names in the report (e.g. "Adapter Removal" instead of "adapter removal"). And no need to wrap module names in the code tags.

ewels:
@pontushojer - you're a machine! 🤯  Amazing work! 👏🏻 

@vladsavelyev - agree on both points. I realise that I pushed for the htslib thing 👀 but I also like clean output 😅

pontushojer:
Thanks guys! 

> * I'm still not convinced that reporting `htslib` versions for `bcftools` and `samtools` is a useful information, as almost always they going to be identical to the actual tools versions. I'd remove it, or at least add some logic to only show it if it differs 🤔  And similarly, no need to report the Python version for cutadapt.

From my perspective I would assume its useful information if the tool author(s) decides to include it in their log output. Thus I would say its probably better to include it than not. Probably samtools can be compiled with different versions of htslib thus it is valuable to report this.

> * Also, another late thought I don't think we should lowercase module names in the table. It's nicer when the names match the sections names in the report (e.g. "Adapter Removal" instead of "adapter removal"). And no need to wrap module names in the code tags.

This I agree would be a nice change. I will open a separate PR with this.

multiqc/modules/bcftools/stats.py
```diff
@@ -41,6 +45,16 @@ def parse_bcftools_stats(self):
...
             for line in f["f"].splitlines():
+                # Get version number from file contents
+                if line.startswith("# This file was produced by bcftools stats"):
+                    version_match = re.search(VERSION_REGEX, line)
+                    if version_match is not None:
+                        self.add_software_version(version_match.group(1), f["s_name"])
+
+                    htslib_version_match = re.search(HTSLIB_REGEX, line)
+                    if htslib_version_match is not None:
+                        self.add_software_version(htslib_version_match.group(1), f["s_name"], "htslib")
```
vladsavelyev, line 56:
```suggestion
                        if htslib_version_match.group(1) != version_match.group(1):
                            self.add_software_version(htslib_version_match.group(1), f["s_name"], "htslib")
```

ewels:
> Probably samtools can be compiled with different versions of htslib thus it is valuable to report this.

How about we add some code in the samtools module which checks if samtools == htslib. If they're different, report both as done currently. If they're the same, just report samtools (can always do a debug log message if we're feeling paranoid).

Then we get clean output in the 99.99% of cases, but in the edge cases we still show the important information.

multiqc/modules/bcftools/stats.py
vladsavelyev: I'd suggest to only report it if it's different. What do you think?

pontushojer:
@ewels, @vladsavelyev I changed both samtools and bcftools to only report htslib version if its different. 

What do you think about the `cutadapt` case? This can definitely be run using different version of python so I think its best to include. 

Also I now added `pagolin` ([c595cad](https://github.com/ewels/MultiQC/pull/2051/commits/c595cade13ef903dd4b6a47a91ee99fd241f5128)) which reports a whole array of software version for tools it depends on. Not sure if all are relevant but maybe that is for someone more familiar with the tool to decide whats should be included. For now I added most of them.

ewels:
I think Python version is too deep. It's not directly related / _part of_ the tool, but rather it's basically a dependency. Let's keep it to just the cutadapt version.

Pangolin - haven't checked what the output looks like, but here I'm more tempted report everything. This tool is made of several subcomponents and stuff like the database version etc is very important. But I can't remember what is already reported and the details of what you've put here, so I'll check properly later. Maybe we can ask on the @nf-core viralrecon slack channel if we're not sure.

ewels:
Yeah, Pangolin is kind of scrappy but I think that it's probably fairly unique. That information is important, so let's keep it like that for now. Can always revise according to community feedback.

ewels (approved):
Amazing work @pontushojer ! 👏🏻 

Going to merge now as we are building towards a release. But keep 'em coming! 😆

pontushojer:
> Yeah, Pangolin is kind of scrappy but I think that it's probably fairly unique. That information is important, so let's keep it like that for now. Can always revise according to community feedback.

Sound good to me! 

I was also looking into the testdata we have for Pangolin in `MultiQC_TestData` . It seems like that columns are left-shifted for some sample which makes version end up in the wrong columns: see https://github.com/ewels/MultiQC_TestData/blob/master/data/modules/pangolin/v4.0/pangolin_4-0.csv

Posted an issue here: https://github.com/cov-lineages/pangolin/issues/531. Let's see if it is just a misstake or might occur in the actual tool output.

PR #2055, opened 2023-09-19
Don't sys.exit from run() function

ewels:
Noticed that we have a stray `sys.exit()` call in the `run()` function. This should be removed for usage in interactive environments.

Not checked thoroughly yet, so needs sanity checking / testing before merge.

### DIFF
--- a/multiqc/multiqc.py
+++ b/multiqc/multiqc.py
@@ -586,7 +586,7 @@ def run(
         run_modules = [m for m in run_modules if list(m.keys())[0] not in config.exclude_modules]
     if len(run_modules) == 0:
         logger.critical("No analysis modules specified!")
-        sys.exit(1)
+        return {"report": report, "config": config, "sys_exit_code": 1}
     run_module_names = [list(m.keys())[0] for m in run_modules]
     logger.debug("Analysing modules: {}".format(", ".join(run_module_names)))
 
@@ -981,7 +981,7 @@ def run(
                     logger.error("Output directory {} already exists.".format(config.plots_dir))
                     logger.info("Use -f or --force to overwrite existing reports")
                     shutil.rmtree(tmp_dir)
-                    sys.exit(1)
+                    return {"report": report, "config": config, "sys_exit_code": 1}
             logger.info(
                 "Plots       : {}{}".format(
                     os.path.relpath(config.plots_dir),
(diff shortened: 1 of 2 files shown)


ewels:
Need to double check that I haven't missed any other `sys.exit()` calls here.

ewels:
Scooped up a couple of others, some remain but I think that we can tackle those during wider refactoring. Hopefully this PR replaces the ones within `multiqc.run()` which is the priority here.

Still needs testing, running each of the edge cases where these trigger, to make sure I haven't screwed up.

- [ ] No modules specified
- [ ] Module run error
- [ ] Report filename already exists

multiqc/multiqc.py
```diff
@@ -586,7 +586,7 @@ def run(
         run_modules = [m for m in run_modules if list(m.keys())[0] not in config.exclude_modules]
     if len(run_modules) == 0:
         logger.critical("No analysis modules specified!")
-        sys.exit(1)
+        return {"report": report, "config": config, "sys_exit_code": 1}
```
vladsavelyev, line 589: Is there a scenario when it's triggered?

multiqc/multiqc.py
```diff
@@ -981,7 +981,7 @@ def __rich_measure__(self, console: rich.console.Console, options: rich.console.
                     logger.error("Output directory {} already exists.".format(config.plots_dir))
                     logger.info("Use -f or --force to overwrite existing reports")
                     shutil.rmtree(tmp_dir)
-                    sys.exit(1)
+                    return {"report": report, "config": config, "sys_exit_code": 1}
```
vladsavelyev, line 984: I guess because of these lines https://github.com/ewels/MultiQC/blob/master/multiqc/multiqc.py#L933-L934 it also should never happen, unless a folder was created by another process. Simulated that, works fine

vladsavelyev (approved):
Approving, works as expected. Will do more notebook environment testing when the refactoring is complete

multiqc/multiqc.py
ewels:
```console
❯ multiqc -m fastqc -e fastqc .

  /// MultiQC 🔍 | v1.16.dev0 (f7ad10c)

|           multiqc | Only using modules: fastqc
|           multiqc | Excluding modules 'fastqc'
|           multiqc | No analysis modules specified!
```

multiqc/multiqc.py
ewels: Yes, you might be right there. I suspect that this code predates the rotating suffix thing being added. In which case, we can presumably just delete this chunk of code.

ewels:
Feel free to merge when you're happy @vladsavelyev 👍🏻

Issue #2052, opened 2023-09-18
bcftools: TypeError


ewels:
### Description of bug

_From [Slack](https://multiqc.slack.com/archives/C04QMP84K5L/p1695033854534519):_

---

Hi, I am new to MultiQC. I am running testfull profile on nf-core Sarek pipeline v3.2.3. The whole pipeline (including MultiQC v1.14) worked well before, but I am recently encountering an issue.  Is there any possible way to fix this? Sincerely thank you for your support!

### File that triggers the error

[HCC1395N.tiddit.bcftools_stats.txt](https://github.com/ewels/MultiQC/files/12652690/HCC1395N.tiddit.bcftools_stats.txt)

### MultiQC Error log

```console
Oops! The 'bcftools' MultiQC module broke... ────────────────╮
  │ Please copy this log and report it at                                        │
  │ https://github.com/ewels/MultiQC/issues                                      │
  │ Please attach a file that triggers the error. The last file found was:       │
  │ ./37/HCC1395N.tiddit.bcftools_stats.txt                                      │
  │                                                                              │
  │ Traceback (most recent call last):                                           │
  │   File "/usr/local/lib/python3.11/site-packages/multiqc/multiqc.py", line 65 │
  │     output = mod()                                                           │
  │              ^^^^^                                                           │
  │   File "/usr/local/lib/python3.11/site-packages/multiqc/modules/bcftools/bcf │
  │     n["stats"] = self.parse_bcftools_stats()                                 │
  │                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^                                 │
  │   File "/usr/local/lib/python3.11/site-packages/multiqc/modules/bcftools/sta │
  │     self.add_data_source(f, s_name, section="stats")                         │
  │   File "/usr/local/lib/python3.11/site-packages/multiqc/modules/base_module. │
  │     source = os.path.abspath(os.path.join(f["root"], f["fn"]))               │
  │                                           ~^^^^^^^^                          │
  │ TypeError: string indices must be integers, not 'str'
```

### Before submitting

- [X] I have read the [troubleshooting documentation](https://multiqc.info/docs/#troubleshooting-1).
- [X] I am using the latest release of MultiQC.
- [X] I have included a full MultiQC log, not truncated.
- [X] I have attached an input file (**.zip** if necessary) that triggers the error.

vladsavelyev:
[...]

PR #2056, opened 2023-09-19
Bcftools: fix parsing of multi-sample stats outputs

vladsavelyev:
Fixes https://github.com/ewels/MultiQC/issues/2052

That's why linters tend to flag single-letter variables 😅


ewels (approved):
Good catch :+1:

PR #2057, opened 2023-09-20
Modules should handle any input. Fix `Preseq` and `BCL Convert` on mixed input

vladsavelyev:
There are currently a few modules that need to collate information across all input logs, and bail if there are mixed inputs:

* Preseq cannot run on a mixture of logs based on reads and on basepairs, because it needs to understand what are the plot labels are,
* BCL Convert similarly can only run when all logs are based on the same number of cycles.

Following the de-facto convention that modules should either work on an inputs, or skip a sample, or raise a `UserWarning`/`ModuleNoSamplesFound` exception, we should fix those modules to properly work around those limitations. It's not great that these two modules fail on our input test data, so not get tested fully.

* Preseq will build 2 plots, one with the number of reads as a base, and another with the basepairs as a base.
* BCL Convert matches samples with run IDs, records the cluster length for each run, and properly re-calculate some metrics based on the cluster length.

Along those lines, fix a few bugs in BCL Convert, in particular
* when a mixture of post-3.9.3 runs and pre-3.9.3 runs are passed, and the quality metrics files aren't available for some runs
* a potential usage of a variable defined at another scope
* rename "reads" to "clusters" to better reflect that those are read _pairs_ or molecules

### DIFF
--- a/multiqc/utils/search_patterns.yaml
+++ b/multiqc/utils/search_patterns.yaml
@@ -102,7 +102,7 @@ bclconvert/quality_metrics:
 bclconvert/adaptermetrics:
   fn: "Adapter_Metrics.csv"
 bclconvert/unknown_barcodes:
-  fn: "Top_Unknown_Barcodes.csv"
+  fn: "Top_Unk"
 biobambam2/bamsormadup:
   contents: "# bamsormadup"
   num_lines: 2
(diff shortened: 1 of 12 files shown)


multiqc/utils/search_patterns.yaml
```diff
@@ -102,7 +102,7 @@ bclconvert/quality_metrics:
 bclconvert/adaptermetrics:
   fn: "Adapter_Metrics.csv"
 bclconvert/unknown_barcodes:
-  fn: "Top_Unknown_Barcodes.csv"
+  fn: "Top_Unk"
```
vladsavelyev, line 105:
```suggestion
  fn: "Top_Unknown_Barcodes.csv"
```

ewels (approved):
LGTM from a code skim, thanks! 👍🏻 

Approving on the basis that you've tested locally and it looks good.

nvnieuwk:
ping @matthdsm

PR #2059, opened 2023-09-20
Export plots in a zip file

vladsavelyev:
Address the browser limitations on the number of downloaded files at the same time. Fixes https://github.com/ewels/MultiQC/issues/1873

TODO:
[] Avoid adding a JS dependency.

### DIFF
--- a/multiqc/templates/default/assets/js/multiqc_toolbox.js
+++ b/multiqc/templates/default/assets/js/multiqc_toolbox.js
@@ -344,6 +344,7 @@ $(function () {
         var f_scale = parseInt($("#mqc_export_scaling").val());
         var f_width = parseInt($("#mqc_exp_width").val()) / f_scale;
         var f_height = parseInt($("#mqc_exp_height").val()) / f_scale;
+        var zip = new JSZip();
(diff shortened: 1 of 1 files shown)


ewels:
Same thing needs to be added for downloading data CSVs as well..

multiqc/templates/default/assets/js/multiqc_toolbox.js
```diff
@@ -4,6 +4,7 @@
 
 var mqc_colours_idx = 0;
 var mqc_colours = ["#e41a1c", "#377eb8", "#4daf4a", "#984ea3", "#ff7f00", "#ffff33", "#a65628", "#f781bf", "#999999"];
+var zip_threshold = 1;
```
ewels, line 7: Let's only use zip files when we really have to..

```suggestion
var zip_threshold = 8;
```

ewels (approved):
Let's still go ahead, even though it's SVG without export config. Something fixed is better than nothing, even if it's missing some of the config.

We can make a new issue for scooping up the config properly for the zip files to do later, but it'll probably be overtaken by replacing the plotting library anyway.

PR #2062, opened 2023-09-21
NanoStat: module clean-up by @schorlton 

vladsavelyev:
As we are not going with renaming the NanoStat module, I'm pulling the clean-up from @schorlton's pull request https://github.com/ewels/MultiQC/pull/1627, as we still want that in.


ewels (approved):
Nice, thanks!

Issue #1869, opened 2023-02-16
Table column color scaling not working when there are negative numbers


zxl124:
### Description of bug

I am trying to use a diverging color scale such as "RdYlGn" and "bars_zero_centrepoint: True" to plot my tables, in order to show the magnitude of change with bars, but direction of change, i.e., whether the value is positive or negative, with colors. However, while the bar size function works as expected, the color scaling is not working. I would have negative and positive numbers with the same color, or numbers with signs but different values with the same color. See a screenshot below.
![image](https://user-images.githubusercontent.com/37191025/219507958-ee53eef0-374a-4110-aa7a-ced74ec98351.png)
Col_1 and col_2 have the same numbers, except that sample_a has a negative number. You can see that col_2 behaving as expected, because the mid point of the scale would be at 0.95, and 0.4 should have a orange color. However, col_1 is not displaying the right colors at all. I would expect sample_a to have red, sample_b to have a darker shade of green, sample_c to have a lighter shade of green. Instead, they are all the same color.

I've tried adding `min: -1.9 max: 1.9` to the headers setting, it made col_2 look the same as col_1.

### File that triggers the error

[test_mqc.zip](https://github.com/ewels/MultiQC/files/10761753/test_mqc.zip)

### MultiQC Error log

```console
/// MultiQC 🔍 | v1.14

|           multiqc | Search path : /data
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 3/3
|    custom_content | test: Found 4 samples (table)
|           multiqc | Compressing plot data
|           multiqc | Deleting    : multiqc_report.html   (-f was specified)
|           multiqc | Deleting    : multiqc_data   (-f was specified)
|           multiqc | Report      : multiqc_report.html
|           multiqc | Data        : multiqc_data
|           multiqc | MultiQC complete
```

### Before submitting

- [X] I have read the [troubleshooting documentation](https://multiqc.info/docs/#troubleshooting-1).
- [X] I am using the latest release of MultiQC.
- [X] I have included a full MultiQC log, not truncated.
- [X] I have attached an input file (**.zip** if necessary) that triggers the error.

vladsavelyev:
It looks like the code trims the leading `-` sign while determining the min and max value for the color scale:

[...]

PR #2064, opened 2023-09-21
Fix table color scaling with negative numbers

vladsavelyev:
Escape the dash in the regular expressions to avoid trimming the leading minus in the negative numbers.

Fixes https://github.com/ewels/MultiQC/issues/1869


ewels (approved):
Nice, thanks!

PR #2065, opened 2023-09-22
Remove print statments 

pontushojer:
@ewels Seems some print statements were include when https://github.com/ewels/MultiQC/pull/2051 go merged.


ewels (approved):
ooops 👀  Good spot, thanks for the fast fix!

PR #2069, opened 2023-09-22
Fix BamQC GC color scheme, show warning about categorical scale for sequential values

vladsavelyev:
* Use a sequential color scale for the average GC color in QualiMap BamQC, instead of a qualitative Set1.
* Show a warning if it happens.
* Tiny refactoring - move the colors static dictionary as a constant class field to be accessible by all functions.

Perhaps we should add some linting that checks if a qualitative scale is specified for the values that are clearly sequential (like in `"suffix": "%"` or `min` and `max` set are good signs of that).


multiqc/utils/mqc_colour.py
```diff
@@ -51,6 +339,14 @@ def get_colour(self, val, colformat="hex", lighten=0.3):
         rgb_converter = lambda x: max(0, min(1, 1 + ((x - 1) * lighten)))
 
         try:
+            if self.name in mqc_colour_scale.qualitative_scales and isinstance(val, float):
```
ewels, line 342: I think this is all that's needed to hide this behind linting:
```suggestion
            if self.name in mqc_colour_scale.qualitative_scales and isinstance(val, float) and config.lint:
```

multiqc/utils/mqc_colour.py
```diff
@@ -51,6 +339,14 @@ def get_colour(self, val, colformat="hex", lighten=0.3):
...
+                sequential_scales = [
+                    s for s in mqc_colour_scale.COLORBREWER_SCALES if s not in mqc_colour_scale.qualitative_scales
+                ]
+                logger.warning(
+                    f"A qualitative scale is used for float values. Consider using "
```
ewels, line 361:
```suggestion
                    f"A qualitative scale ('{self.name}') is used for float values. Consider using "
```

ewels (approved):
Great! 🙌🏻

PR #2074, opened 2023-09-24
Just run CI on the oldest + newest supported Python versions

ewels:
Small measure to reduce the amount of CI jobs that we run.

I think it's vanishingly rare that anything fails only in a _middle_ version of Python - it's basically always one end of the range. So maybe it makes sense to only run tests on the extremes.


vladsavelyev (approved):

vladsavelyev:
@multiqc-bot changelog

vladsavelyev:
@multiqc-bot changelog

PR #2079, opened 2023-09-27
Plotly backend for graphs. Adds `plotly` template

ewels:
Very early stage WIP code where I was playing with using Plotly when creating MultiQC charts.

Beginning of efforts to work on https://github.com/ewels/MultiQC/issues/1789

This PR may be too basic to be worth using, but if nothing else I wanted to push the code just so that it's visible for reference.

[Vlad] additional TODO list after first review:

- [x] Table: make the "Plot" button work
- [x] Table: refactor as a new Plot class subtype
- [x] Table: add violin/default switch to all tables
- [x] Export plot: apply scale correctly
- [x] Remove hover delay
- [x] Violin: add main title
- [x] Violin: fix flat version
- [x] Resize flat versions: up to max width
- [x] Flat plots: maintain aspect ratio when resizing
- [x] Violin: perhaps show only values on hover tooltips, and sample name show separately
- [x] Violin: label contrastive colors
- [x] Export plot button: open the toolbar instead of downloading a PNG directly
- [x] Export plot button: use the original button, disable the camera icon
- [x] Violin: debug outliers and border in edge cases in test_data
- [x] Violin/table: disable switch when > config.max_table_rows and add information tooltips
- [x] Violin: with large datasets, pass plot only and not the data
- [x] Violin: fix HTML in hover labels
- [x] Bug: if a plot is resized vertically, horizontal resizing doesn't work any more
- [x] Violin: debug setting range for xaxis0
- [x] Violin: pre-parse numerical values
- [x] Violin plot: respect `shared_key`
- [x] "Samtools alignment metrics" suffixes
- [x] Add "Created with MultiQC"
- [x] Heatmaps don't have "Export plot" button above plot area
- [x] Icon for violin plot - would be nice to be different to scatter. Maybe glyphicon glyphicon-align-left ?
Could perhaps also swtich the scatter plot icon as it's clear bars - maybe glyphicon glyphicon-equalizer?
- [x] "Created with MultiQC" into flat plots
- [x] Make "Created with MultiQC" smaller
- [x] > In the previous template, samples unmatched in an active highlight are shown with grey in tables "sample id" column. This seems to have been lost, are now showing with regular black text
- [x] Address vertical labels overflowing plot handle
- [x] Export to TSV/CSV of other plot types
- [ ] "Created with MultiQC" into SVG
- [ ]  Add a tiny transparent logo as well?
- [ ] Line plots `cpswitch` should affect Y axis 
- [x] Export plot: Plot scaling should be set to 2 by default
- [x] Highlighting samples seems to completely hide un-highlighted rows in the FastQC status heatmap
- [x] Outliers are not shown for massive violin plots now, but the hover help text says that they are


vladsavelyev:
Starting a lot of non-obvious things to implement with Plotly:

- [ ] Export data as CSV/TSV

ewels:
Things we can skip doing:

* FastQ Screen: Custom plot type when only a few samples (instead, always do the stacked bar)
* If it's too much work, can skip / defer to a later stage the "Custom plots on click":
    * RSeQC Junction Saturation
    * FastQC per-base sequence content

vladsavelyev:
Actionable points before the release:

- [ ] Table: make "Plot" button work
- [ ] Table: refactor as a new Plot class subtype
- [ ] Table: add violin/default switch to all tables
- [ ] Export plot: apply scale correctly
- [ ] Remove hover delay
- [ ] Violin: add main title
- [ ] Violin: fix flat version
- [ ] Violin: perhaps show only values on hover tooltips, and sample name show separately
- [ ] Violin: debug outliers and border in edge cases
- [ ] Export plot button: open the toolbar instead of downloading a PNG directly
- [ ] Export plot button: use the original button, disable the camera icon
- [ ] Add "Created with MultiQC"
- [ ] Flat plots: maintain aspect ratio when resizing
- [ ] Line plot: convert into an interval line + a violin of means
- [ ] Violin: fix HTML in hover labels
- [ ] Violin: label contrastive colors

ewels:
Remaining work:

* Default order of series in scatter plot
* MultiQC footer text 🙈 

Otherwise ready for more testing and edge-case-finding...

ewels:
Getting to the really small stuff now..

* Heatmaps don't have "Export plot" button above plot area
* If you can change the axis label text colour for dotplots, does this mean that you could do the same for barplot sample names when using the toolbox highlighter tool? (Could never do this in HighCharts, hence the bar outline)
* In the previous template, samples unmatched in an active highlight are shown with grey in tables "sample id" column. This seems to have been lost, are now showing with regular black text
    * If we can do barplot labels as above, then would be great to make unmatched the same grey here too.
* Icon for violin plot - would be nice to be different to scatter. Maybe `glyphicon glyphicon-align-left` ?
    * Could perhaps also swtich the scatter plot icon as it's clear bars - maybe `glyphicon glyphicon-equalizer`?
    * 
* As discussed on Slack - having the "Created with MultiQC" text in exported plots (if not super difficult)

vladsavelyev:
> If you can change the axis label text colour for dotplots, does this mean that you could do the same for barplot sample names when using the toolbox highlighter tool? (Could never do this in HighCharts, hence the bar outline)

Unfortunately, it's not straightforward: for violins, the setup is more complex as we build a separate subplot for each violin, with a unique layout, and then compose them together. But I can try doing the same for bars. Hopefully it won't affect performance too much.

ewels:
Ok, I think I'm about at the limit of things I can find here. A handful of very small things:

* Export plot: Plot scaling should be set to `2` by default
* Highlighting samples seems to completely hide un-highlighted rows in the FastQC status heatmap
* Outliers are not shown for massive violin plots now, but the hover help text says that they are

I can think of a few other things, but it's at the point now where they'd probably do better as separate small feature request issues and PRs, rather than slowing this PR from being merged.
We'll also need to update the docs, but I think that's probably also better in a separate PR..

Otherwise, I think it's a ✅ from me to merge, once you're happy / done with your own list @vladsavelyev 😄 

I can't review this PR as I started it, so feel free to merge when you want.

ewels:
oh, and if we're going to make this the default right away now, we should rename `plotly` to `default` and make `default` &rarr; `highcharts`.

vladsavelyev (approved):

Issue #2063, opened 2023-09-21
Picard MarkDuplicates 1 out many thousands of runs broke


a-frantz:
### Description of bug

Very odd. I've run this version of Picard and this version of MultiQC literally thousands of times without an issue. I took a look at the offending MarkDups file, and it seems well-formed to me. Also looked over the stderr logs from my `picard MarkDuplicates` call, and nothing seems off (although with how verbose Picard logs are, I might've missed something).

I'm going to clear the `miniwdl` cache for this run and see if it happens again. Either way, thought you'd want to see this. Based on the traceback looks like it should be an easy fix.

### File that triggers the error

[SJMB027_D.WholeGenome.MarkDuplicates.metrics.txt](https://github.com/ewels/MultiQC/files/12689618/SJMB027_D.WholeGenome.MarkDuplicates.metrics.txt)

### MultiQC Error log

```console
$         multiqc -v \
            --no-ansi \
            -c /mnt/miniwdl_task_container/work/_miniwdl_inputs/0/multiqc_config_hg38.yaml \
            --file-list file_list.txt \
            -o SJMB027_D.WholeGenome.multiqc

  /// MultiQC 🔍 | v1.14

[2023-09-21 12:02:42,107] multiqc                                            [DEBUG  ]  This is MultiQC v1.14
[2023-09-21 12:02:42,116] multiqc                                            [DEBUG  ]  Loading config settings from: /mnt/miniwdl_task_container/work/_miniwdl_inputs/0/multiqc_config_hg38.yaml
[...]

PR #2094, opened 2023-10-01
Picard: MarkDuplicates: account for missing trailing `0`

vladsavelyev:
Fixes https://github.com/ewels/MultiQC/issues/2063

Some MarkDuplicates outputs have a missing `0` value for the last column for libraries with all zeroes, so need to parse an empty string properly as 0.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,11 +17,8 @@
 
 ### Module updates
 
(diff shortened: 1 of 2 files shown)


vladsavelyev:
@multiqc-bot changelog

ewels (approved):
LGTM, looks cleaner logic now. Prefer having it at the end like that. Thanks!

Issue #2093, opened 2023-09-29
whatshap: truncated stats file gives `KeyError`


splaisan:
### Description of bug

multiqc (1.14) of truncated whatshap results (head command retaining only the 23 first lines and removing all unplaced contig lines that came after that.

The inputs are attached for reproducibility

### File that triggers the error

[input.zip](https://github.com/ewels/MultiQC/files/12766475/input.zip)

### MultiQC Error log

```console
$ multiqc .

  /// MultiQC 🔍 | v1.14

|           multiqc | MultiQC Version v1.16 now available!
|           multiqc | Search path : /...../main_chr
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 10/10  
╭────────────────────────────────────────────────── Oops! The 'whatshap' MultiQC module broke... ──────────────────────────────────────────────────╮
│ Please copy this log and report it at https://github.com/ewels/MultiQC/issues                                                                    │
│ Please attach a file that triggers the error. The last file found was: ./887604734_main_stats.tsv                             │
│                                                                                                                                                  │
│ Traceback (most recent call last):                                                                                                               │
│   File "/opt/miniconda3/envs/multiqc/lib/python3.11/site-packages/multiqc/multiqc.py", line 654, in run                                          │
│     output = mod()                                                                                                                               │
│              ^^^^^                                                                                                                               │
│   File "/opt/miniconda3/envs/multiqc/lib/python3.11/site-packages/multiqc/modules/whatshap/whatshap.py", line 53, in __init__                    │
│     self.whatshap_add_general_stats()                                                                                                            │
│   File "/opt/miniconda3/envs/multiqc/lib/python3.11/site-packages/multiqc/modules/whatshap/whatshap.py", line 208, in whatshap_add_general_stats │
│     general_stats = {field: summary_data[field] for field in general_stats_headers}                                                              │
│                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                              │
│   File "/opt/miniconda3/envs/multiqc/lib/python3.11/site-packages/multiqc/modules/whatshap/whatshap.py", line 208, in <dictcomp>                 │
│     general_stats = {field: summary_data[field] for field in general_stats_headers}                                                              │
│                             ~~~~~~~~~~~~^^^^^^^                                                                                                  │
[...]

PR #2095, opened 2023-10-02
Whatshap: process truncated input with no ALL chromosome

vladsavelyev:
MultiQC can pretty much summarise the ALL "chromosome" given available information: all reported fields are either sums, or averages across blocks, so it's a relatively easy addition. Fixes https://github.com/ewels/MultiQC/issues/2093

### DIFF
--- a/multiqc/modules/whatshap/whatshap.py
+++ b/multiqc/modules/whatshap/whatshap.py
@@ -127,7 +127,7 @@ class MultiqcModule(BaseMultiqcModule):
         # as a fictional 'ALL' chromosome.
         if ALL_CHROM not in results and len(results) > 1:
             log.warning(
-                f"Could not find the 'ALL' chromosome for {sample}. Make sure that the input is not truncated. Will summarize information for all available chromosomes instead."
+                f"Could not find the 'ALL' chromosome for {sample}. Make sure that the input is not truncated. MultiQC will summarize information for all available chromosomes instead."
             )
             sum_fields = {
                 "variants",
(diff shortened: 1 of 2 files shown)


vladsavelyev:
@multiqc-bot changelog

multiqc/modules/whatshap/whatshap.py
```diff
@@ -91,6 +97,9 @@ def process_data(data):
         header = next(file_content).strip().split("\t")
         header[0] = header[0].lstrip("#")
 
+        if sample == "GM24385_0_1x_full_opt":
+            pass
```
ewels, line 101: Debugging?

vladsavelyev: Fixed!

multiqc/modules/whatshap/whatshap.py
```diff
@@ -103,25 +109,63 @@ def process_data(data):
...
             results = dict()
-            results["ALL"] = defaultdict(int)
+            results[ALL_CHROM] = defaultdict(int)
+
+        # If ALL chromosome is not present, and there are multiple contigs available,
+        # e.g. in case of a truncated input file, we summarize available contigs
+        # as a fictional 'ALL' chromosome.
+        if ALL_CHROM not in results and len(results) > 1:
+            log.warning(
+                f"Could not find the 'ALL' chromosome for {sample}. Make sure that the input is not truncated. Will summarize information for all available chromosomes instead."
```
ewels, line 130:
```suggestion
                f"Could not find the 'ALL' chromosome for {sample}. Make sure that the input is not truncated. MultiQC will summarize information for all available chromosomes instead."
```

ewels (approved):
Ok, nice! This goes beyond what I had envisaged - I was thinking that we would just catch the exception and throw an error saying "file is truncated, skipping". You're being very kind to the end user 😅 

Great stuff! ✅

Issue #2099, opened 2023-10-03
mosdepth module not prioritizing region over global information


jamigo:
### Description of bug

The [documentation](https://multiqc.info/modules/mosdepth/) of the mosdepth module on multiqc v1.16 states this:

> Using "region" if exists, otherwise "global". Plotting 3 figures:

But this is not the case, and both global and region information are plotted. Either the selection of the "region" plots should be prioritized, hiding global information when region information is available (I would personally go for that), or either the documentation needs to be updated stating that both global and region information are always shown if available.

Also, the Mean Coverage information displayed on the main General Statistics table has changed from region to global too. This [commit](https://github.com/ewels/MultiQC/commit/1bf611b177669dce381d75e6ccb0ae8a46b6776f) has changed obtaining that information from the global information line only of each *.mosdepth.summary.txt file, rather than being overwritten by the region information if available as it has always been until now.

This change is located in current multiqc v1.16 mosdepth.py file at line 131, that went from
`if line.startswith("total"):`
to
`if line.startswith("total\t"):`
In order to prioritize region information over global (which, again, I would personally go for that), that "\t" character should be removed.

vladsavelyev:
@jamigo, thank you for reporting this issue, it's a good catch. I agree that the global information should be hidden when the region information is available.

And good catch on checking "total" vs. "total_region" - I'll add a fix for that as well.

Really appreciate a detailed report and suggestions to fix!

PR #2106, opened 2023-10-09
Mosdepth: fix prioritizing region over global information 

vladsavelyev:
* Fix bug parsing sample summary: when parsing summary, prioritize `total_region` over `total` for region-based reports.
* When both "global" and "region" reports found for the same sample, pick only the "region" report, and discard global information for plots. Add a warning about it into the plot descriptions.

Fixes https://github.com/ewels/MultiQC/issues/2099


ewels (approved):
Nice 👍🏻

Issue #2031, opened 2023-09-07
Fastp: Sample name incorrectly parsed when there are spaces in it


dpryan79:
### Description of bug

The fastp module parses sample names from a json file. It does so by just pulling the first string after the -i option. That's fine until that name contains a space like "Campaign 3 sample 1.fastq.gz". I unfortunately don't have a very elegant solution to this or I'd make a pull request to go along with this.

You can reproduce this with something like:

```bash
fastp -c -g -y -i 'Campaign 3 sample 1_1.fastq.gz' -o 'Campaign 3 sample 1_1.trimmed.fastq.gz' -I 'Campaign 3 sample 1_2.fastq.gz' -O 'Campaign 3 sample 1_2.trimmed.fastq.gz' -j 'Campaign 3 sample 1_fastp.json' -w 4
```

This ends up in the json being something like:

```json
"command": "fastp -c -g -y -i Campaign 3 sample 1_1.fastq.gz -o Campaign 3 sample 1_1.trimmed.fastq.gz -I Campaign 3 sample 1_2.fastq.gz -O Campaign 3 sample 1_2.trimmed.fastq.gz -j Campaign 3 sample 1_fastp.json -w 4 "
```

That's understandably not exactly fun to parse. Using sed to quote these also doesn't help since the python being used is this: 

https://github.com/ewels/MultiQC/blob/d51c895a186191ccccbfe209a54bec7e37bf2b0f/multiqc/modules/fastp/fastp.py#L161-L163

I've taken to just replacing spaces within filenames with underscores in this file using sed, which kind of works around the issue. If any one else has some elegant parsing solution that'd be nicer long-term :)

PR #2108, opened 2023-10-11
Fastp: parse unescaped sample names with white spaces

vladsavelyev:
Fixes https://github.com/ewels/MultiQC/issues/2031


ewels (approved):
Managed to find an edge case where this doesn't work:

```bash
fastp -c -g -y -i "Campaign 3 -sample 1_1.fastq.gz" -o
```

Which ends up in the log as this:

```bash
fastp -c -g -y -i Campaign 3 -sample 1_1.fastq.gz -o
```

Which gets truncated. But that is _very_ awkward to cater for and is hopefully enough of an edge case that no-one will hit it.

LGTM to merge until someone else complains 😆  👍🏻

Issue #2039, opened 2023-09-14
Drop official support for Python 3.7


ewels:
Python 3.7 hit EOL (end of life) in June: https://endoflife.date/python

I'm not aware of any 3.7-specific language constructs that we're specifically maintaining, but it might be nice to drop official support (CI) anyway to try to encourage users to update.

ewels:
This needs updating in the CI, the docs, and also in the version check:
https://github.com/ewels/MultiQC/blob/f4878540c0b1e77fa419b589ec33f3a5039d17da/multiqc/multiqc.py#L477-L484

vladsavelyev:
If we drop 3.7 support, it would simplify https://github.com/ewels/MultiQC/pull/2113 (we would be able to use `shutil.copytree(..., dirs_exist_ok=True)` introduced in 3.8).

However, we are still supporting even 3.6. I'd go with dropping both 3.6 and 3.7 in 1.17.

ewels:
Done in https://github.com/ewels/MultiQC/pull/2121 but looks like we missed that `multiqc.py` mention..

PR #2113, opened 2023-10-12
Fix for python 3.12: replace removed `distutils`

vladsavelyev:
`distutils` was fully removed in Python 3.12: https://peps.python.org/pep-0632/#migration-advice

1. Replace `distutils.version.StrictVersion` with `packaging.version`
2. Replace `distutils.util.strtobool` with re-implementation
3. Replace `distutils.dir_util.copy_tree` with `shutil.copytree` calls


vladsavelyev:
@multiqc-bot changelog

multiqc/modules/picard/picard.py
```diff
@@ -157,3 +157,25 @@ def multiply_hundred(self, val):
         except ValueError:
             pass
         return val
+
+
+def strtobool(val):
```
ewels, line 162: Better to put this in the MultiQC utils, rather than having it in a module?

ewels (approved):
Nice 👍🏻

ewels:
@multiqc-bot fix linting

multiqc/modules/picard/picard.py
vladsavelyev: It think it makes sense to move it if/when it's going to be used by other modules

vladsavelyev: Though it will actually fix the test, so doing it now 😅

vladsavelyev:
(Please merge https://github.com/ewels/MultiQC/pull/2115 first)

vladsavelyev:
@multiqc-bot changelog

vladsavelyev:
Sigh, `dirs_exist_ok` to `shutil.copytree()` was only added in Py 3.8 :( https://docs.python.org/3/whatsnew/3.8.html#shutil

Should probably add an if-else to import distutils for 3.6 and 3.7 instead of calling shutils.

ewels:
If we do, maybe we add a note to https://github.com/ewels/MultiQC/issues/2039

It would be really nice to know what Python versions people are using! No idea how many people are still on 3.7..

vladsavelyev:
Good point. I thought we are dropping only 3.6; but if we drop 3.7 as well, it should be nice.

ewels (approved):
Great!

Issue #2112, opened 2023-10-12
Python 3.12 installation problems


ewels:
Originally reported [in the community forum](https://community.seqera.io/t/multiqc-modulenotfounderror-no-module-named-imp/147/2)

---

Looks like MultiQC is having some problems running in Conda with Python 3.12, due to a dependency error of some kind:

```console
❯ conda create --name py3.12 python=3.12 multiqc --yes --quiet
Collecting package metadata (current_repodata.json): ...working... done
Solving environment: ...working... done

## Package Plan ##

  environment location: /Users/ewels/.miniconda3/miniconda3/envs/py3.12

  added / updated specs:
    - multiqc
    - python=3.12


The following packages will be downloaded:
```

<details>

```
    package                    |            build
    ---------------------------|-----------------
    brotli-1.1.0               |       hb547adb_1          19 KB  conda-forge
    brotli-bin-1.1.0           |       hb547adb_1          17 KB  conda-forge
    brotli-python-1.1.0        |  py312h9f69965_1         335 KB  conda-forge
    ca-certificates-2023.7.22  |       hf0a4a13_0         146 KB  conda-forge
    certifi-2023.7.22          |     pyhd8ed1ab_0         150 KB  conda-forge
    charset-normalizer-3.3.0   |     pyhd8ed1ab_0          45 KB  conda-forge
    click-8.1.7                |unix_pyh707e725_0          82 KB  conda-forge
    contourpy-1.1.1            |  py312h389731b_1         222 KB  conda-forge
    cycler-0.12.1              |     pyhd8ed1ab_0          13 KB  conda-forge
    fonttools-4.43.1           |  py312he37b823_0         2.5 MB  conda-forge
    freetype-2.12.1            |       hadb7bae_2         582 KB  conda-forge
    humanfriendly-10.0         |     pyhd8ed1ab_6          72 KB  conda-forge
    importlib-metadata-6.8.0   |     pyha770c72_0          25 KB  conda-forge
    kiwisolver-1.4.5           |  py312h389731b_1          60 KB  conda-forge
    lcms2-2.15                 |       hf2736f0_3         210 KB  conda-forge
    libblas-3.9.0              |18_osxarm64_openblas          15 KB  conda-forge
    libbrotlicommon-1.1.0      |       hb547adb_1          67 KB  conda-forge
    libbrotlidec-1.1.0         |       hb547adb_1          28 KB  conda-forge
[...]

PR #2119, opened 2023-10-13
Bundle lzstring to fix Python 3.12

vladsavelyev:
Addresses https://github.com/ewels/MultiQC/issues/2112

Bundling the 450 lines of code must be easier if we aim in supporting Py 3.12 quickly.

Stacked on https://github.com/ewels/MultiQC/pull/2113


ewels (approved):
Great! 👍🏻

Issue #2114, opened 2023-10-12
htseq-count output files skipped when more than two columns present


gl-eb:
### Description of bug

If htseq-count output files contain more than the two standard columns `feature` and `count`, they are ignored. The htseq-count user has the option to add further columns for e.g. feature name or chromosome on which the feature is located.
This behaviour is present in both MultiQC 1.15 and 1.16. Versions 1.14 and 1.13 work as expected.

The current regex pattern `^(feature\tcount|\w+\t\d+)$` does not allow for additional columns. Adding the following pattern to my system-wide config file fixed the issue: `^(feature\tcount|\w+.*\t\d+)$`

### File that triggers the error

[WTCHG_536273_229157.tsv.zip](https://github.com/ewels/MultiQC/files/12886168/WTCHG_536273_229157.tsv.zip)

### MultiQC Error log

```console
multiqc -v --force --interactive --filename multiqc-test --outdir <outdir> <srcdir>

  /// MultiQC 🔍 | v1.16

[2023-10-12 20:53:29] multiqc                                            [DEBUG  ]  This is MultiQC v1.16
[2023-10-12 20:53:29] multiqc                                            [DEBUG  ]  Command used: /Users/gleb/micromamba/envs/multiqc-new/bin/multiqc -v --force --interactive --filename multiqc-test --outdir <outdir> <srcdir>
[2023-10-12 20:53:30] multiqc                                            [DEBUG  ]  Latest MultiQC version is v1.16
[2023-10-12 20:53:30] multiqc                                            [DEBUG  ]  Working dir : /Users/gleb
[2023-10-12 20:53:30] multiqc                                            [DEBUG  ]  Template    : default
[2023-10-12 20:53:30] multiqc                                            [DEBUG  ]  Running Python 3.11.6 | packaged by conda-forge | (main, Oct  3 2023, 10:37:07) [Clang 15.0.7 ]
[...]

PR #2129, opened 2023-10-16
Htseq: allow counts files with more than 2 columns

vladsavelyev:
Fixes https://github.com/ewels/MultiQC/issues/2114

Thanks a lot @gl-eb for the bug report and for the fix suggestion!


ewels (approved):
LGTM 👍🏻

Issue #2130, opened 2023-10-17
custom_content module adds "base" to the namespace of the column of general stats table


zxl124:
### Description of bug

When using custom_content module to add a column to the general stats table, for example, via a JSON file, the module changes the mouse-over text of the added column by the following behavior:
1. When `namespace` is not defined, the module uses "base" as the namespace. The mouse-over text appears as "base: {title/description}". 
![image](https://github.com/ewels/MultiQC/assets/37191025/b3930f69-e85e-43f4-aab2-c02103500b42)
2. When `namespace` is defined in the input, the module prepends "base " to the name space. The mouse-over text appears as "base {namespace}: {title/description}".
![image](https://github.com/ewels/MultiQC/assets/37191025/66494e27-7647-4221-b968-044a7ad7d956)

This bug occurs in version 1.16. I've tested 1.14. This bug is not there. In version 1.14, the mouse-over text would just be "{namespace}: {title/description}". This is caused by the new addition in #2037 to address a different bug. The relevant lines are https://github.com/ewels/MultiQC/blob/c10cdcb963909b9dddd76558e88cd3bc87005f65/multiqc/modules/base_module.py#L468-L470
This is well designed, except for that the custom_content module does not recognize anything in the input to modify `BaseMultiqcModule.name` in any way, it just has a default value "base". Therefore, the user has no way to configure the mouse-over text correctly. It will always has "base" in it. One add code to the custom_content module to modify `BaseMultiqcModule.name` based on a value in the input, or modify the code above to have a different behavior when `self.name == "base"`. 

### File that triggers the error

[gs_perc_passvariant_mqc.json](https://github.com/ewels/MultiQC/files/12922905/gs_perc_passvariant_mqc.json)

### MultiQC Error log

```console
# multiqc -f .

  /// MultiQC 🔍 | v1.16

|           multiqc | Search path : /data
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 5/5
|    custom_content | vcftools: Found 2 General Statistics columns
|           multiqc | Report      : multiqc_report.html   (overwritten)
|           multiqc | Data        : multiqc_data   (overwritten)
|           multiqc | MultiQC complete
```

### Before submitting

- [X] I have read the [troubleshooting documentation](https://multiqc.info/docs/#troubleshooting-1).
- [X] I am using the latest release of MultiQC.
- [X] I have included a full MultiQC log, not truncated.
[...]

PR #2131, opened 2023-10-17
Set name and anchor for the custom content "module"

vladsavelyev:
Fixes https://github.com/ewels/MultiQC/issues/2130

In reports built from custom content, a superfluous "base" is appearing, coming from the default `name` field of the `BaseMultiQCModule` class:

Setting the name and the anchor to the BaseMultiQCModule class instantiated for custom content to make it appear less meaningless:

Screenshot 2023-10-17 at 12 35 50


ewels (approved):
Great 👍🏻

PR #2136, opened 2023-10-17
Allow unquoted versions in versions YAML

vladsavelyev:
Currently, MultiQC will break if a version value is parsed by YAML into a non-string (float or int):

```yaml
$ cat multiqc my_mqc_versions.yaml
myprocess:
  mytool: 1.2
  mytool2: 1.2.3
```

```sh
$ multiqc my_mqc_versions.yaml ...
[2023-10-17 18:05:36] multiqc.utils.software_versions                    [ERROR  ]  Software versions loaded from my_mqc_versions.yaml is not in a valid format
```

This is perhaps overprotective to require users to quote strings in YAML, so we need to handle unquoted versions as well.

UPD: as @pontushojer noted, allowing PyYAML to parse numbers would turn `1.10` into `1.1`. A workaround here is to pass `Loader=yaml.BaseLoader` into `yaml.load`, which will force PyYAML to treat all scalar values as strings, keeping `1.10` as `"1.10"`. Updated the PR.

cc @adamrtalbot


maxulysse (commented):
looking forward to it

pontushojer:
We considered this in https://github.com/ewels/MultiQC/pull/1927, but the issue was versions like `1.10` being converted to `1.1` when this is parsed as a float rather than a string. Is it not better to be protective in this case?

vladsavelyev:
@pontushojer, that's a really great point, shame on me for forgetting that discussion :( I agree it makes sense to keep to the requirement to quote the versions.

vladsavelyev:
Revisiting this PR. I think there is a good workaround that doesn't require quoting versions in the `*_mqc_versions.yaml` files. If we parse them with `yaml.load(f, Loader=yaml.BaseLoader)` it will force PyYAML to treat all scalar values as strings, keeping `1.10` as `"1.10"`. 

It seems to be safe to do as `*_mqc_versions.yaml` are not expected to contain anything apart from versions in nested structures. For the general config section `config.software_versions`, we still assert that all versions are quoted strings.

I updated to the PR, it's good to review again now.

maxulysse:
So it looks nice to me, but I don't think I'm the person you want to review this

maxulysse (approved):

multiqc/utils/software_versions.py
```diff
@@ -61,22 +62,29 @@ def load_versions_from_config(config):
...
             try:
-                log.debug("Reading software versions settings from: {}".format(file_name))
-                software_versions_file_tmp = yaml.safe_load(f)
+                log.debug(f"Reading software versions settings from: {file_name}")
+                software_versions_file_tmp = yaml.load(
+                    f,
+                    # We need to be cautious when loading unquoted version strings from a YAML file.
+                    # For instance, the version `1.10` will be parsed as a float by default, this converted
+                    # into `1.1`. Passing yaml.BaseLoader explicitly makes YAML treat all scalar values
+                    # as strings, so `1.10` will turn into a string `"1.10"` as we want.
```
ewels, line 78:
```suggestion
                    # as strings, so `1.10` will turn into a string `"1.10"` as we want.
                    # From https://pyyaml.org/wiki/PyYAMLDocumentation
                    #      BaseLoader(stream) does not resolve or support any tags
                    #      and constructs only basic Python objects: lists, dictionaries and Unicode strings.
```

ewels (approved):
LGTM, thanks!

Issue #2138, opened 2023-10-19
MultiQC 1.17 fails to recognize fastp files


tamuanand:
### Description of bug

Hi @vladsavelyev and @ewels 

MultQC 1.17 fails to detect fastp files - the same works with MultiQC 1.16

I will attach the verbose output from MultiQC 1.16 as a comment so that you can see the difference(s).

### File that triggers the error

[HG001.fastp.json](https://github.com/ewels/MultiQC/files/13038975/HG001.fastp.json)
[HG002_son.fastp.json](https://github.com/ewels/MultiQC/files/13038976/HG002_son.fastp.json)
[HG003_father.fastp.json](https://github.com/ewels/MultiQC/files/13038977/HG003_father.fastp.json)
[HG004_mother.fastp.json](https://github.com/ewels/MultiQC/files/13038978/HG004_mother.fastp.json)

### MultiQC Error log

```console
/ MultiQC 🔍 | v1.17

[2023-10-19 02:47:16] multiqc                                            [DEBUG  ]  This is MultiQC v1.17
[2023-10-19 02:47:16] multiqc                                            [DEBUG  ]  Command used: /usr/local/bin/multiqc /path_to/temp_dir/2023_Oct_19 --force --verbose --interactive
[2023-10-19 02:47:16] multiqc                                            [DEBUG  ]  Could not connect to multiqc.info for version check: module 'packaging.version' has no attribute 'StrictVersion'
[2023-10-19 02:47:16] multiqc                                            [DEBUG  ]  Working dir : /path_to/temp_dir/2023_Oct_19
[2023-10-19 02:47:16] multiqc                                            [DEBUG  ]  Template    : default
[2023-10-19 02:47:16] multiqc                                            [DEBUG  ]  Running Python 3.12.0 | packaged by conda-forge | (main, Oct  3 2023, 08:43:22) [GCC 12.3.0]
[...]

PR #2139, opened 2023-10-19
Fastp: correctly parse sample name from --in1/--in2 command. Fallback to file name

vladsavelyev:
Fixes https://github.com/ewels/MultiQC/issues/2138

The sample name parsing logic:

* Parse the `command` field found in JSON. E.g. `"fastp -c -g -y --in1 Sample 1 1.fastq.gz --in2 Sample 1_2.fastq.gz --out1 ..."` will parse everything between `--in1 ` and the following ` -`, plus do normal file name trimming, ending up with `Sample 1 1`.
* If the step above failed, fall back to the file name (e.g. `fastp.json`) to extract the sample name (e.g. `fastp`).
* Add `config.fastp.s_name_filenames` option to force using the file names for sample names.

Additionally, fix the `self.ignore_samples()` logic for the module (it previously affected only the main `self.fastp_data`, but not the derived dictionaries).

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Install packages
         run: |
           python -m pip install --upgrade pip
-          pip install yaml
+          pip install pyyaml
 
       - name: Update CHANGELOG.md from the PR title
         env:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
       - id: check-hooks-apply
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: "v4.4.0" # Use the ref you want to point at
+    rev: "v4.5.0" # Use the ref you want to point at
     hooks:
       - id: check-added-large-files
       - id: check-merge-conflict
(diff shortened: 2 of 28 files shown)


ewels:
Let's switch the priority, as discussed here: https://github.com/ewels/MultiQC/issues/2138#issuecomment-1780763764

ewels (approved):

tamuanand:
Hi @ewels and @vladsavelyev 

with your proposed switch, how will this below be handled

```
fastp --in1 Sample_A_1.fastq.gz --in2 Sample_A_2.fastq.gz --out SampleA.fastp.json
```

Will my general stats table show this as `Sample_A_1`?

vladsavelyev:
Hi Anand - yes, exactly. Unless you specify `extra_fn_clean_trim: "_1"` in the config.

tamuanand:
Hi @vladsavelyev and @ewels - I do not agree with this  - https://github.com/ewels/MultiQC/issues/2138#issuecomment-1780763764 - please correct me if I am wrong

```
My preference is to use the FastQ in the command line as the top priority - this is the same behaviour as we have in most other modules, such as Trimmomatic, Cutadapt, FastQC etc
```

- with FastQC - both "Sample_A_1" and "Sample_A_2" make it to the general stats table with the Read count data, %GC data etc. 

Will `fastp` too have "Sample_A_1"  and  "Sample_A_2" etc reported?
- if yes, then I agree with @ewels comment above
- if no, and if only "Sample_A_1" data gets shown in the general stats table where in fact it is referring to consolidated stats from both _1 and _2 then it is going to convey something incorrect and unwanted.

That's why I asked here - https://github.com/ewels/MultiQC/issues/2138#issuecomment-1770718289
- `why fix/change something that has been working for years now`

I request you to revisit this.

vladsavelyev:
@tamuanand - 

We want to generalise the MultiQC behaviour as much as possible, and for that reason we want `fastp` to work similarly to other modules that take pairs of FASTQ files as input: `cutadapt`, `trimmomatic`. For that reason, the best bet would be to take the first FASTQ file name as a sample name by default, but allow users to override that behaviour with either module-specific file name trimming, or by forcing MultiQC to use the log file names as sample names:

```yaml
fastp:
  s_name_filenames: true
```

That option can be conveniently specified in the command line as well:

```sh
multiqc -m fastp test_data/data/modules/fastp -f -v --cl-config "fastp: {s_name_filenames: true}"
```

Hope that helps.

ewels:
> why fix/change something that has been working for years now

It's been working for you, but equally for other users it's been broken for years now. It depends on your analysis setup. That's what we try to avoid - the previous behaviour required you to organise your data in a particular way. The newer behaviour (consistent with the rest of MultiQC) does not.

There is a standard way to achieve the behaviour that you want via a config file, mentioned by Vlad. I think this is fine.

> if only "Sample_A_1" data gets shown in the general stats table where in fact it is referring to consolidated stats from both _1 and _2 then it is going to convey something incorrect and unwanted.

I agree that this isn't ideal. Again it's not specific to this module, and MultiQC is riddled with imperfections such as this. The data that we have to obtain sample names is by nature imperfect.

I've created a new issue to discuss a potential new general functionality to better clean sample names when we have a pair of filenames available: https://github.com/ewels/MultiQC/issues/2162

PR #2142, opened 2023-10-19
Changelog bot: detect module updates when only `search_patterns.yaml` changed

vladsavelyev:
E.g. will detect this https://github.com/ewels/MultiQC/pull/2129/files as a "HtSeq Count" module change, and put the entry correctly under the "Module Updates" section.


vladsavelyev:
@multiqc-bot changelog

ewels (approved):
Can we also not match for square brackets when looking for `skip changelog` in the PR title?

I did `(skip changelog)` instead of `[skip changelog]` the other day and it didn't work.

PR #2147, opened 2023-10-22
Docs + warning - supports 3.8+ not 3.6+ [skip changelog]

ewels:
Final bits to close https://github.com/ewels/MultiQC/issues/2039

PR #2148, opened 2023-10-23
Fix Python and MultiQC version checks

vladsavelyev:
Few more bits for https://github.com/ewels/MultiQC/pull/2147

* Put the oldest supported version into a variable
* Return when version doesn't match
* Make the check as early as possible
* Also, fix MultiQC remote version check: replace `version.StrictVersion` with `version.parse`

### DIFF
--- a/multiqc/multiqc.py
+++ b/multiqc/multiqc.py
@@ -359,6 +359,16 @@ def run(
 
     Author: Phil Ewels (http://phil.ewels.co.uk)
     """
+    # Throw an error if we are using an unsupported version of Python
+    OLDEST_SUPPORTED_PY = "3.8.0"
+    if sys.version_info[:3] < tuple(map(int, OLDEST_SUPPORTED_PY.split("."))):
+        current_version = ".".join(str(component) for component in sys.version_info[:3])
+        logger.critical(
+            "You are running MultiQC with Python {}. "
+            "Please upgrade Python! MultiQC does not support Python < {}, "
+            "things will break.".format(current_version, OLDEST_SUPPORTED_PY)
+        )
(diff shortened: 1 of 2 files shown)


multiqc/multiqc.py
```diff
@@ -359,6 +359,16 @@ def run(
...
+    # Throw an error if we are using an unsupported version of Python
+    OLDEST_SUPPORTED_PY = "3.8.0"
+    if sys.version_info[:3] < tuple(map(int, OLDEST_SUPPORTED_PY.split("."))):
+        current_version = ".".join(str(component) for component in sys.version_info[:3])
+        logger.critical(
+            "You are running MultiQC with Python {}. "
+            "Please upgrade Python! MultiQC does not support Python < {}, "
+            "things will break.".format(current_version, OLDEST_SUPPORTED_PY)
+        )
+        return {"report": None, "config": None, "sys_exit_code": 1}
```
ewels, line 371: You think it's worth completely ditching execution? Previously it's been a warning, as whether it crashed or not depended on which modules you ran etc.

If it is certain that it'll crash with 3.7 then a critical + return is best though.

vladsavelyev: It can lead to unexpected behaviour or to confusing errors. I think we either should stay silently (but fully) supporting 3.7; or give ourselves full freedom to use 3.8 features, and bail early. 

Maybe it makes sense to stick to silently supporting 3.7 for some time, just showing a warning.

vladsavelyev: It's actually not going to work at all with 3.7, so we need to fail early: https://github.com/ewels/MultiQC/actions/runs/6651649342/job/18074099768?pr=2147

PR #2151, opened 2023-10-25
New module: Seqera Platform CLI

vladsavelyev:
Parses `runs_*.tar.gz` dumps containing logs and stats from a Seqera Platform run. To be triggered in the https://github.com/seqeralabs/nf-aggregate workflow, or can parse the output of `tw runs dump ... --output=runs_SmUkr43Nul49G.tar.gz` directly.

Note that binary files by default are ignored by MultiQC, so you need to set `ignore_images: false` in the MultiQC config to make it parse the input tar-gz file. This can be done via the CLI:

```sh
multiqc -m seqera_cli test_data/data/modules/seqera_cli -f -v --cl-config "ignore_images: false"
```

- [x] Think what general stats cols to hide?
- [x] Figure out how to better present start date/end date
- [x] Test with many runs
- [x] Add a bar plot with a category per status of workflows/processs?

Screenshot 2023-10-25 at 22 45 10


ewels:
Given that the tar step is from the workflow and not from the Seqera CLI itself, I think we should also cater for uncompressed files. Then can handle the custom stuff in the pipeline but it stays as general use as possible.

drpatelh:
The Seqera CLI natively dumps tar files. I was going to uncompress them in the same process that dumps the run info but if we don't have to then that would be a bonus. Supporting both would be great. The JSON files all have standard names so will be easy to scrape if we stage each run in its individual directory / archive.

vladsavelyev:
Added support for both compressed and uncompressed dumps. Also added a barplot and parsing of the platform version.

ewels:
> Figure out how to better present start date/end date

Maybe show a duration as a column, keep start + end dates but hide by default.

ewels:
👉🏻  Should specify the task status order + colours in the bar plot to match Tower:

* Pending: `#8f4199`
* Submitted: `#e68642`
* Running: `#4256e7`
* Cached: `#939598`
* Succeeded: `#28ae61`
* Failed: `#e7363e`

docs/core/development/modules.md
```diff
@@ -823,9 +824,14 @@ headers['name'] = {
...
     show "percentage aligned" on page load but hide "number of reads aligned".
 - `placement`
-  - If you feel that the results from your module should appear at the left side
+  - If you feel that the results from your module should appear on the left side
     of the table set this value less than 1000. Or to move the column right, set
     it greater than 1000. This value can be any float.
+- `to_float`
+  - Lambda function to convert value to a number, to support the quantitative
+    color code and table sorting. By default, MultiQC attempt to convert each
+    value to float(), and if it fails, it will leave the value as a string without
```
ewels, line 833:
```suggestion
    value to `float()`, and if it fails, it will leave the value as a string without
```

docs/core/development/modules.md
```diff
@@ -823,9 +824,14 @@ headers['name'] = {
     be useful when data could be sometimes useful. For example, some modules
     show "percentage aligned" on page load but hide "number of reads aligned".
 - `placement`
-  - If you feel that the results from your module should appear at the left side
+  - If you feel that the results from your module should appear on the left side
     of the table set this value less than 1000. Or to move the column right, set
     it greater than 1000. This value can be any float.
+- `to_float`
+  - Lambda function to convert value to a number, to support the quantitative
+    color code and table sorting. By default, MultiQC attempt to convert each
```
ewels, line 832:
```suggestion
    color code. By default, MultiQC attempt to convert each
```

docs/core/development/modules.md
```diff
@@ -823,9 +824,14 @@ headers['name'] = {
...
+    value to float(), and if it fails, it will leave the value as a string without
+    color code and sorting.
```
ewels, line 834:
```suggestion
    color code.
```

ewels (approved):
Minor comments, otherwise LGTM!

ewels:
LGTM to me to merge, waiting on review from @drpatelh before hitting the button.

multiqc/modules/seqera_cli/seqera_cli.py
```diff
@@ -0,0 +1,218 @@
...
+    that is reading workflow.json and workflow-load.json files directly.
+    """
+
+    def __init__(self):
+        super(MultiqcModule, self).__init__(
+            name="Seqera Platform CLI",
+            anchor="seqera_cli",
+            href="https://github.com/seqeralabs/tower-cli",
+            info="reports statistics generated by the Seqera Platform CLI.",
+            doi="10.1016/j.ajhg.2017.01.017",
```
ewels, line 76: This is showing the citation for Peddy 😅 

I wonder if we can have some kind of CI check that looks for duplicated DOIs. Might be some valid reasons, but I bet that most would be accidental copy+paste artefacts.
```suggestion
            # doi="",  # No DOI for this tool
```

ewels:
Suggestion: Move CPU time and Cost columns up, alongside pipeline duration. These 3 columns are likely typically the most interesting.

multiqc/modules/seqera_cli/seqera_cli.py
```diff
@@ -0,0 +1,218 @@
...
+            "duration": {
+                "title": "Duration",
+                "description": "Duration of the workflow",
+                "scale": "BuPu",
+                "to_float": lambda x: x.total_seconds(),
+            },
+            "cpuEfficiency": {
+                "title": "CPU Efficiency",
+                "description": "Percentage of CPU time used by the workflow",
+                "format": "{:,.2f}",
```
robsyme, line 160:
```suggestion
                "format": "{:,.2f}",
                "suffix": "%",
                "max": 100,
```

multiqc/modules/seqera_cli/seqera_cli.py
```diff
@@ -0,0 +1,218 @@
...
+                "scale": "RdYlGn",
+            },
+            "memoryEfficiency": {
+                "title": "Memory Efficiency",
+                "description": "Percentage of memory used by the workflow",
+                "format": "{:,.2f}",
```
robsyme, line 166:
```suggestion
                "format": "{:,.2f}",
                "suffix": "%",
                "max": 100, 
```

robsyme (changes requested):
We know that cpuEfficiency and memoryEfficiency will always be percentages, so should format as such.

drpatelh:
Initial comments based on report @ewels sent to me:


In General stats:

### Change names in header:

- [ ]  `Sample Name` -> `Run ID`
- [ ] `Duration` -> `Wall Time`
- [ ] `Cost` -> `Estimated Cost ($)`
- [ ] [NEW] `Pipeline version`

### Default column order in (after renaming based on above):

- [ ] `Repository`
- [ ] `Pipeline version`
- [ ] `Run ID`
- [ ] `Wall Time`
- [ ] `CPU Time`
- [ ] `Estimated Cost ($)`
- [ ] `Read GB`
- [ ] `Write GB`
- [ ] `CPU Efficiency`
- [ ] `Memory Efficiency`

### Default table sorting

- [ ] First by `Repository` and then by `CPU hours`

### Check Wall time calculation

- [ ] For Run ID: `1h5PlYK4uXvNyE` there is a small discrepancy in Duration (9:09:11; MultiQC) vs Wall time (9:09:31; [Seqera Platform](https://tower.nf/orgs/seqeralabs/workspaces/benchmarks/watch/1h5PlYK4uXvNyE)). Are we calculating this correctly? There appears to be a consistent discrepancy for other runs too so we will need to double-check this.

### Process status plot

image

- [ ] Remove the  "Versions: 2.6.0, 3.3.2, 3.12.0" next to the `Seqera Platform CLI` header. Doesn't add make any sense having that there and better to include pipeline versions in the General stats table
- [ ] Change x-axis labels to repository name and instead show Run id on hover.
- [ ] The DOI in the description under the header points to something random and needs to be removed.

### Additional Plots

It would be great to at least have bar plots for the more pertinent metrics so they can be dumped directly into presentations if required rather than having to generate them via another tool:

- [ ] `CPU Time`
- [ ] `Wall Time`
- [ ] `Estimated cost ($)`

###  Enhancements

- [ ] Can we include a link to the actual run in Seqera Platform. Can't find a run URL in the JSON files so might need to add that in as an enhancement later.
- [ ] Include run labels in the report for filtering. I will take a look at this now.
- [ ] Include Nextflow and Seqera Platform version in stats

Let me know when you have made these changes and regenerated the report and I'll take another look. Thank you!! Looking great already.

ewels:
>  `Sample Name` -> `Run ID`

Can't do this in the General Stats table, but we could move all of these out into a module-specific table where it is possible. If there are no other modules running it'll look essentially the same.

> Default table sorting

MultiQC does not have this concept, it'll always be sorted by the primary identifier - the run ID in this case.

>  Remove the "Versions: 2.6.0, 3.3.2, 3.12.0" next to the Seqera Platform CLI header. Doesn't make any sense having that there and better to include pipeline versions in the General stats table

Disagree on this point - that's the version of the CLI that was used. It's a core MultiQC feature now, not sure that it hurts to report this?

> Change x-axis labels to repository name and instead show Run id on hover.

Not sure if this is possible, I have a feeling that these need to be unique during report generation (though can be renamed interactively).

drpatelh:
>> Sample Name -> Run ID

> Can't do this in the General Stats table, but we could move all of these out into a module-specific table where it is possible. If there are no other modules running it'll look essentially the same.

Ok, happy to see what the alternative looks like. Sample Name is misleading in this context.

>> Default table sorting

> MultiQC does not have this concept, it'll always be sorted by the primary identifier - the run ID in this case.

That's a shame. But it is what it is!

>> Remove the "Versions: 2.6.0, 3.3.2, 3.12.0" next to the Seqera Platform CLI header. Doesn't make any sense having that there and better to include pipeline versions in the General stats table

> Disagree on this point - that's the version of the CLI that was used. It's a core MultiQC feature now, not sure that it hurts to report this?

You sure? Those look like pipeline versions to me? There should only be a single version for the CLI which I agree would be good to include.

>> Change x-axis labels to repository name and instead show Run id on hover.

> Not sure if this is possible, I have a feeling that these need to be unique during report generation (though can be renamed interactively).

Given that those run ids are essentially random strings, they mean nothing to anyone without additional metadata if we just look at that plot. Be great if we can find some sort of workaround to make that more intuitive by default.

ewels:
> You sure? Those look like pipeline versions to me?

That's a bug in the module if so, 'cc @vladsavelyev 

> Be great if we can find some sort of workaround to make that more intuitive by default.

Best I can think of is the run name, instead of ID? It has to be unique to each run so it doesn't leave us with much. But as mentioned in https://github.com/ewels/MultiQC_TestData/pull/298#issuecomment-1788858338:

> Bit of a shame that there's no other label for the runs beyond the somewhat obscure run ID. But could easily switch these out with something custom using `--sample-names`: https://multiqc.info/docs/reports/customisation/#sample-name-replacement

drpatelh:
I don't think the version of the Seqera Platform CLI is actually dumped anywhere in those JSON files. I didn't follow how you are now including versions for tools in MultiQC? For the moment, I am doing this manually via the pipeline at the end of a Nextflow process like this:

```
cat <<-END_VERSIONS > versions.yml
    "${task.process}":
        seqera-cli: \$(echo \$(tw --version 2>&1) | sed 's/^.*Tower CLI version //; s/ *\$//')
    END_VERSIONS
```

The versions we can get from the JSON files that we can include in the General Stats table are:
- Nextflow Version (`workflow.json['nextflow']['version']`)
- Seqera Platform Version (`service-info.json['version']`)

vladsavelyev:
Updated version:

Screenshot 2023-11-01 at 19 03 56

One thing I'm not sure about is putting all the 3 versions into the table as columns. It is already too wide for a laptop screen, and as mentioned before, we have a separate mechanism in MultiQC for the tool/software version reporting, so the platform and the nextflow versions (and the CLI version if provided) should perhaps be reported there (see the screenshot), and we can just add one column for the pipeline versions as it will more likely be different between runs.

We can also add the nextflow/platform versions as hidden columns if it's useful for table exporting 🤔

vladsavelyev:
Regarding the run names, it would be nice to append the repository IDs to the run IDs and make them clickable, instead of having a separate column. This way, the IDs would be more meaningful in the plots as well. That would need some adjustments in the core MultiQC.

ewels:
I don't think that we should touch the run IDs. I think we should suggest to the user to use a second separate manually created CSV to use with `--sample-names` or `--replace-names`, as a meaningful identifier will be custom for every run.

The pipeline doing these runs could potentially generate such a file, if there is logic in the workflow to do so.

ewels:
> We can also add the nextflow/platform versions as hidden columns if it's useful for table exporting 🤔

I think this is good, if nothing else so that if there is more than one value you can see which run used which.

drpatelh:
> I think this is good, if nothing else so that if there is more than one value you can see which run used which.

Think it's fine to scroll of the screen and not hiding? Having versions per run will be quite useful as a sanity check too ensure exactly which versions were used for each run as they could be different across a run set.

Can we label the y-axis in the plots by repository name somehow and then show the run id on hover? Or is that a no-go?

I will look into the labels aspect when I'm on a laptop as I think that would be incredibly useful to help with filtering and visualisation.

drpatelh:
> Think it's fine to scroll off the screen and not hiding? Having versions per run will be quite useful as a sanity check to ensure exactly which versions were used for each run as they could be different across a run set.

As a compromise, maybe we can add the `Nextflow Version` to the General Stats and leave the Seqera Platform in the versions section at the bottom of the report as you have done already @vladsavelyev . It's quite unlikely that the latter will be different across pipeline runs.

vladsavelyev:
Sharing an updated version:

Screenshot 2023-11-02 at 14 22 17

Figured it would be nice to color the platform/nextflow versions with colored badges, as otherwise it's hard to tell apart outliers in the list :)

vladsavelyev:
Done! (updated the comment above)

drpatelh:
Beautiful! Last main thing to add would be the Plots I outlined in https://github.com/ewels/MultiQC/pull/2151#issuecomment-1788912050

I have created issues on the Seqera CLI for things that would be good to add in the future. For now, I can try to find a workaround for using run names in the plots via a mapping file.

Copied again below:

## Additional Plots

It would be great to at least have bar plots for the more pertinent metrics so they can be dumped directly into presentations if required rather than having to generate them via another tool:

- [ ] CPU Time
- [ ] Wall Time
- [ ] Estimated cost ($)

vladsavelyev:
> It would be great to at least have bar plots for the more pertinent metrics so they can be dumped directly into presentations if required rather than having to generate them via another tool:

I'm not sure I understand though what would be the categories in such bar plots. In the dump, there is only one value per run, which fits a table column - which translates into a simple bar plot by the means of the background color bar:

Screenshot 2023-11-02 at 15 44 40

But that doesn't fit a bar plot, where you want multiple non-overlapping categories that can sum up.

If we had per-process numbers, that might work - we could use the N heaviest processes as categories, plus "the remaining" as the last tail category.

drpatelh:
We would have a separate plot for each metric so which should be relatively simple?

vladsavelyev:
Such a plot for a single-value metric would just duplicate the corresponding table column.

Screenshot 2023-11-02 at 15 55 37

We do have this plot in the table already:

Screenshot 2023-11-02 at 15 57 20

drpatelh:
Yes, exactly! One for each of the metrics I listed. The problem with having that in the table is you can't export it as a plot. Also, it will be nice to bring a larger focus on the core metrics if we have individual plots.

vladsavelyev:
Alright, let's add those. It would be nice to add per-process data into the run dump though, to make the bars more informative.

[multiqc_report.html.zip](https://github.com/ewels/MultiQC/files/13242750/multiqc_report.html.zip)

Screenshot 2023-11-02 at 18 26 24

multiqc/modules/seqera_cli/seqera_cli.py
```diff
@@ -0,0 +1,364 @@
...
+            anchor="seqera_cli_run_stats_table",
+            plot=table.plot(data_by_run, headers, {"col1_header": "Run ID"}),
+        )
+
+    def _plots(self, data_by_run):
+        """
+        Add bar plots
+        """
+        self.add_section(
+            name="Processes statuses",
```
drpatelh, line 278:
```suggestion
            name="Workflow status",
```

multiqc/modules/seqera_cli/seqera_cli.py
```diff
@@ -0,0 +1,364 @@
...
+                    "pending": {"name": "Pending", "color": "#8f4199"},
+                    "submitted": {"name": "Submitted", "color": "#e68642"},
+                    "running": {"name": "Running", "color": "#4256e7"},
+                    "cached": {"name": "Cached", "color": "#939598"},
+                    "succeeded": {"name": "Succeeded", "color": "#28ae61"},
+                    "failed": {"name": "Failed", "color": "#e7363e"},
+                },
+                pconfig={
+                    "id": "seqera_cli_process_statuses_plot",
+                    "title": "Seqera platform CLI: processes statuses",
```
drpatelh, line 292:
```suggestion
                    "title": "Seqera platform CLI: Workflow status",
```

multiqc/modules/seqera_cli/seqera_cli.py
```diff
@@ -0,0 +1,364 @@
...
+                    "ylab": "Number of processes",
```
drpatelh, line 293:
```suggestion
                    "ylab": "Number of tasks",
```

multiqc/modules/seqera_cli/seqera_cli.py
```diff
@@ -0,0 +1,364 @@
...
+
+        self.add_section(
+            name="Wall Time",
+            anchor="seqera_cli_wall_time_section",
+            plot=bargraph.plot(
+                plot_data,
+                {"wallTime": {"name": "Wall time"}},
+                {
+                    "id": "seqera_cli_wall_time_plot",
+                    "title": "Seqera platform CLI: wall time",
```
drpatelh, line 316:
```suggestion
                    "title": "Seqera platform CLI: Wall time",
```

multiqc/modules/seqera_cli/seqera_cli.py
```diff
@@ -0,0 +1,364 @@
...
+            plot_data[sn] = dict()
+            if "wallTime" in data:
+                plot_data[sn]["wallTime"] = data["wallTime"] / 60 / 60  # hours
+            if "cpuTime" in data is not None:
+                plot_data[sn]["cpuTime"] = data["cpuTime"] // 1000 / 60 / 60  # hours
+            if "cost" in data is not None:
+                plot_data[sn]["cost"] = data["cost"]
+
+        self.add_section(
+            name="Wall Time",
```
drpatelh, line 309:
```suggestion
            name="Wall time",
```

multiqc/modules/seqera_cli/seqera_cli.py
```diff
@@ -0,0 +1,364 @@
...
+                    "tt_suffix": "&nbsp;h",
+                    "tt_percentages": False,
+                    "cpswitch": False,
+                    "use_legend": False,
+                    "hide_zero_cats": False,
+                },
+            ),
+        )
+        self.add_section(
+            name="CPU Time",
```
drpatelh, line 328:
```suggestion
            name="CPU time",
```

multiqc/modules/seqera_cli/seqera_cli.py
```diff
@@ -0,0 +1,364 @@
...
+            anchor="seqera_cli_cpu_time_section",
+            plot=bargraph.plot(
+                plot_data,
+                {"cpuTime": {"name": "CPU time"}},
+                {
+                    "id": "seqera_cli_cpu_time_plot",
+                    "title": "Seqera platform CLI: CPU time",
+                    "ylab": "hours",
```
drpatelh, line 336:
```suggestion
                    "ylab": "CPU hours",
```

multiqc/modules/seqera_cli/seqera_cli.py
```diff
@@ -0,0 +1,364 @@
...
+                    "tt_suffix": "&nbsp;h",
+                    "tt_percentages": False,
+                    "cpswitch": False,
+                    "use_legend": False,
+                    "hide_zero_cats": False,
+                },
+            ),
+        )
+        self.add_section(
+            name="Estimated Cost",
```
drpatelh, line 347:
```suggestion
            name="Estimated cost",
```

multiqc/modules/seqera_cli/seqera_cli.py
```diff
@@ -0,0 +1,364 @@
...
+            anchor="seqera_cli_cost_section",
+            plot=bargraph.plot(
+                plot_data,
+                {"cost": {"name": "Estimated Cost"}},
```
drpatelh, line 351:
```suggestion
                {"cost": {"name": "Estimated cost"}},
```

multiqc/modules/seqera_cli/seqera_cli.py
```diff
@@ -0,0 +1,364 @@
...
+                {
+                    "id": "seqera_cli_cost_plot",
+                    "title": "Seqera platform CLI: estimated cost",
```
drpatelh, line 354:
```suggestion
                    "title": "Seqera platform CLI: Estimated cost",
```

drpatelh (commented):
Looks awesome @vladsavelyev ! 🤩 Thank you! Some minor comments for consistency but otherwise LGTM! Be great to add in some other things in the future but this is definitely a good start.

drpatelh (approved):

ewels:
Looks great! 👏🏻

PR #2157, opened 2023-10-31
Pangolin: update for v4: add QC Note , update tool versions columns

schorlton:

Hi @ewels and team! This PR:
* Adds an additional column "QC Note" present in the latest pangolin outputs: https://cov-lineages.org/resources/pangolin/output.html
* Report the versions of `constellations` and `scorpio` (used by Pangolin since v4) instead of `PangoLEARN` and `pango` (used by older versions).

I did run the MultiQC test data through and it all passed and looked good.
image

### DIFF
--- a/multiqc/modules/pangolin/pangolin.py
+++ b/multiqc/modules/pangolin/pangolin.py
@@ -6,6 +6,7 @@
 import csv
 import logging
 from collections import OrderedDict
+from typing import Optional
 
 from multiqc.modules.base_module import BaseMultiqcModule, ModuleNoSamplesFound
 from multiqc.plots import table
@@ -103,6 +104,9 @@ class MultiqcModule(BaseMultiqcModule):
                 # Avoid generic header ID that clashes with other modules
                 if "qc_status" not in row:
                     row["qc_status"] = row.pop("status")
+                if "qc_notes" in row:
+                    row["qc_notes"] = _format_qc_notes(row["qc_notes"])
+
                 self.pangolin_data[s_name] = row
                 # Just save the lineage key for now - we will sort out the colours later
                 self.lineage_colours[row["lineage"]] = None
@@ -116,7 +120,7 @@ class MultiqcModule(BaseMultiqcModule):
                 add_version_not_none(row.get("pango_version"), s_name, "Pango")
                 add_version_not_none(row.get("pangoLEARN_version"), s_name, "PangoLEARN")
                 add_version_not_none(row.get("scorpio_version"), s_name, "Scorpio")
-                # constellation_version is someimes "TRUE" or "FALSE" - ignore these
+                # constellation_version is sometimes "TRUE" or "FALSE" - ignore these
                 constellation_version = row.get("constellation_version")
                 if constellation_version not in {None, "TRUE", "FALSE"}:
                     self.add_software_version(constellation_version, s_name, "Constellations")
@@ -225,7 +229,6 @@ class MultiqcModule(BaseMultiqcModule):
             "title": "QC Note",
             "description": "Notes specific to the QC checks run on the sequences.",
             "scale": False,
-            "modify": _format_qc_notes,
         }
 
         headers["note"] = {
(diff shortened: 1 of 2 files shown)


multiqc/modules/pangolin/pangolin.py
```diff
@@ -103,6 +104,9 @@ def add_version_not_none(version, sample, name):
                 # Avoid generic header ID that clashes with other modules
                 if "qc_status" not in row:
                     row["qc_status"] = row.pop("status")
+                if "qc_notes" in row:
+                    row["qc_notes"] = _format_qc_notes(row["qc_notes"])
+
```
vladsavelyev, line 109: I moved parsing the QC notes here, so the parsing errors would more clearly be marked and cought as happening in this module, rather than in the code plotting module.

multiqc/modules/pangolin/pangolin.py
```diff
@@ -235,3 +245,23 @@ def pangolin_table(self):
...
+    Parses QC notes, they appear to come from:
+    https://github.com/cov-lineages/pangolin/blob/361f49cbffbf26eb28bed2f4a4c0e7f3d5a054cc/pangolin/utils/preprocessing.py#L91-L97
+    https://github.com/cov-lineages/pangolin/blob/361f49cbffbf26eb28bed2f4a4c0e7f3d5a054cc/pangolin/utils/preprocessing.py#L179
+    """
+    if raw.startswith("Ambiguous_content"):
+        # e.g. Ambiguous_content:0.03
+        split = raw.split(":")
+        if len(split) != 2:
+            logging.warning(f"Expected label of format 'Ambiguous_content:0.01', found: '{raw}'")
+            return None
```
vladsavelyev, line 261: I replaced the assertion with a warning, so the module wouldn't fail completely if one input file is off.

multiqc/modules/pangolin/pangolin.py
```diff
@@ -200,16 +204,16 @@ def pangolin_table(self):
             "hidden": True,
         }
 
-        headers["pangoLEARN_version"] = {
-            "title": "PangoLEARN version",
-            "description": "The dated version of the pangoLEARN model installed.",
+        headers["scorpio_version"] = {
```
vladsavelyev, line 207: I'm not familiar with Pangolin, but could you please comment on changing the tools which versions are reported here? To make sure it's not accidental.

multiqc/modules/pangolin/pangolin.py
```diff
@@ -200,16 +204,16 @@ def pangolin_table(self):
...
+            "title": "Scorpio version",
+            "description": "The version of the scorpio software installed.",
             "scale": False,
             "hidden": True,
         }
 
-        headers["pango_version"] = {
-            "title": "Pango version",
-            "description": "The version of pango-designation lineages that this assignment is based on.",
+        headers["constellation_version"] = {
```
vladsavelyev, line 214: Same here

vladsavelyev (changes requested):
The QC notes addition looks good, but would be good to add some comment justifying the version reporting

multiqc/modules/pangolin/pangolin.py
schorlton: Thanks!

multiqc/modules/pangolin/pangolin.py
schorlton: Sure - in the code or just here? Pangolin is for assigning lineages to SARS-CoV-2. Over time, the method of performing this has changed, which also impacts various aspects of performance including speed and accuracy. In v4 (the latest), the method changed from [PangoLEARN](https://cov-lineages.org/resources/pangolin/pangolearn.html) to UShER. There is also another tool (scorpio) and database ([constellations](https://cov-lineages.org/constellations.html)), built in to Pangolin. The latest output of Pangolin runs these tools and reports their results. Therefore the version is important. Here are the [full docs](https://cov-lineages.org/resources/pangolin/output.html) for each field in the ouput; you'll notice that `pangoLEARN_version` is no longer reported as it is no longer run by default.

vladsavelyev: Thanks a lot for the explanation! Was curious about removing the previous `pangoLEARN_version` and `pango_version` columns, instead of keeping them along for users of older Pangolin versions. But perhaps we don't need to worry about that, the columns were hidden anyway so unlikely were very popular.

vladsavelyev: I slightly updated the PR title and the description to reflect this change, if you don't mind (this is what I meant about leaving the comment 😅 )

vladsavelyev (approved):

vladsavelyev:
@multiqc-bot changelog

PR #2159, opened 2023-11-03
fastp: add version parsing

### LINKED ISSUE #2085: Fastp: Parse version information
### Description of feature

From fastp v0.22 (https://github.com/OpenGene/fastp/commit/8c8ddb6a79844ee137835d65adf60773708def80) the JSON includes version information. There is however no such file in https://github.com/ewels/MultiQC_TestData. 

This version information has however been included in the HTML report prior to v0.22. This HTML file is currently unparsed. So if we want to include version parsing for `fastp` < 0.22 we need to parse this from the HTML report.

pontushojer:
fix #2085

The fastp testdata included in #2138 contains version info which we now can parse and test for. 

image

- [x ] This comment contains a description of changes (with reason)


vladsavelyev (approved):
Thank you!!

Issue #2158, opened 2023-11-01
Module request:  bamdst 


hongtir:
### Name of the tool

bamdst

### Tool homepage

https://github.com/shiquan/bamdst

### Tool description

Bamdst is a lightweight tool to stat the depth coverage of target regions of bam file(s).

### Tool output
```
## The file was created by bamdst
## Version : 1.0.9
## Files : ST0217_Lg.bam 
                               [Total] Raw Reads (All reads)	766026326
                                       [Total] QC Fail reads	0
                                        [Total] Raw Data(Mb)	113517.44
                                        [Total] Paired Reads	766026326
                                        [Total] Mapped Reads	750267810
                            [Total] Fraction of Mapped Reads	97.94%
                                     [Total] Mapped Data(Mb)	111466.09
                         [Total] Fraction of Mapped Data(Mb)	98.19%
                                     [Total] Properly paired	735084416
                         [Total] Fraction of Properly paired	95.96%
                                [Total] Read and mate paired	744300282
                    [Total] Fraction of Read and mate paired	97.16%
                                          [Total] Singletons	5967528
                       [Total] Read and mate map to diff chr	5429012
                                               [Total] Read1	383013163
                                               [Total] Read2	383013163
                                        [Total] Read1(rmdup)	359498611
                                        [Total] Read2(rmdup)	359195220
                                [Total] forward strand reads	375142297
                               [Total] backward strand reads	375125513
                                 [Total] PCR duplicate reads	31573979
                     [Total] Fraction of PCR duplicate reads	4.21%
                            [Total] Map quality cutoff value	20
                       [Total] MapQuality above cutoff reads	694167039
                 [Total] Fraction of MapQ reads in all reads	90.62%
              [Total] Fraction of MapQ reads in mapped reads	92.52%
                                       [Target] Target Reads	343938951
              [Target] Fraction of Target Reads in all reads	44.90%
           [Target] Fraction of Target Reads in mapped reads	45.84%
                                    [Target] Target Data(Mb)	50324.92
                              [Target] Target Data Rmdup(Mb)	47435.18
                [Target] Fraction of Target Data in all data	44.33%
             [Target] Fraction of Target Data in mapped data	45.15%
                                      [Target] Len of region	1442021955
                                      [Target] Average depth	34.90
                               [Target] Average depth(rmdup)	32.89
                                     [Target] Coverage (>0x)	93.55%
                                    [Target] Coverage (>=4x)	93.52%
[...]

PR #2161, opened 2023-11-06
New module: Bamdst

vladsavelyev:
Module for Bamdst - resolves https://github.com/ewels/MultiQC/issues/2158

Parses `coverage.report` to add to general stats and a separate table with all metrics included.

Also parse `chromosomes.report` to generate 2 plots:

Screenshot 2023-12-13 at 15 20 27

To avoid clutter, chromosome filtering patterns `include_contigs`/`exclude_contigs` are enabled, and can be overridden with the config:

```yaml
bamdst:
  exclude_contigs:
    - "*_alt"
    - "*_decoy"
    - "*_random"
    - "*_fix"
    - "HLA*"
    - "chrUn*"
    - "chrM"
    - "chrEBV"
```

Additionally, added an option to exclude contigs by total fraction, like it's implemented in mosdepth:

```yaml
bamdst:
  # Should be a fraction, e.g. 0.001 (exclude contigs with 0.1% coverage of sum of
  # coverages across all contigs)
  perchrom_fraction_cutoff: 0.001
```

Similarly to mosdepth, if all samples come with a single contig, the module will build a bar graph:

Screenshot 2023-12-13 at 15 28 06


hongtir:
Thanks a lot!

tamuanand:
Hi @vladsavelyev  and @hongtir 

Please check out this comment before merging - https://github.com/ewels/MultiQC/issues/2158#issuecomment-1801675805

vladsavelyev:
Added a separate table with all metrics included.

Now also parsing `chromosomes.report` to generate 2 plots:

Screenshot 2023-12-13 at 15 20 27

To avoid clutter, chromosome filtering patterns `include_contigs`/`exclude_contigs` are enabled, and can be overridden with the config:

```yaml
bamdst:
  exclude_contigs:
    - "*_alt"
    - "*_decoy"
    - "*_random"
    - "*_fix"
    - "HLA*"
    - "chrUn*"
    - "chrM"
    - "chrEBV"
```

Additionally, added an option to exclude contigs by total fraction, like it's implemented in mosdepth:

```yaml
bamdst:
  # Should be a fraction, e.g. 0.001 (exclude contigs with 0.1% coverage of sum of
  # coverages across all contigs)
  perchrom_fraction_cutoff: 0.001
```

Similarly to mosdepth, if all samples come with a single contig, the module will build a bar graph:

Screenshot 2023-12-13 at 15 28 06

multiqc/modules/bamdst/bamdst.py
```diff
@@ -0,0 +1,419 @@
...
+                "description": f"Target coverage ({short})",
+                "min": 0,
+                "suffix": "%",
+                "format": "{:.2f}",
+                "scale": "RdYlGn",
+                "modify": lambda x: x * 100.0,
+                "hidden": short != ">0x",
+            }
+
+        self.general_stats_addcols(data_by_sample, headers)
```
ewels, line 253: As with the other module, I don't love just sticking all columns into General Stats. If we do this a lot, or if there are many samples, it'll bloat the report size. It also makes the "Configure Columns" dialogue massive and difficult to use if there are hundreds of columns to choose from..

multiqc/utils/search_patterns.yaml
```diff
@@ -16,6 +16,9 @@ anglerfish:
 bakta:
   fn: "*.txt"
   contents: "Bakta:"
+bamdst/coverage:
+  contents: "## The file was created by bamdst"
+  num_lines: 1
```
ewels, line 21: In case of unexpected crap from elsewhere
```suggestion
  num_lines: 5
```

multiqc/utils/config_defaults.yaml
```diff
@@ -394,6 +394,10 @@ module_order:
       module_tag:
         - DNA
         - RNA
+  - bamdst:
+      module_tag:
+        - DNA
+        - RNA
```
ewels, line 400: We should really ditch these, they're pointless.

multiqc/modules/bamdst/bamdst.py
```diff
@@ -0,0 +1,419 @@
...
+
+
+import logging
+from collections import defaultdict
+from typing import Dict, Union
+import csv
+import os
+import fnmatch
+
+import math
```
ewels, line 11: Weird, why has Ruff moved this here, as if it's not a core Python library?

multiqc/modules/bamdst/bamdst.py
```diff
@@ -0,0 +1,419 @@
...
+
+        self.cfg = _read_config()
+        data_by_sample = dict()
+        data_by_chromosome_by_sample = dict()
+        for f in self.find_log_files("bamdst/coverage"):
+            if data := self._parse_coverage_report(f):
+                data_by_sample[f["s_name"]] = data
+
+                # Parse "chromosomes.report" sitting next to the "coverage.report" file:
+                if os.path.exists(chroms_path := os.path.join(f["root"], "chromosomes.report")):
```
ewels, line 87: Refactoring idea for MultiQC v2 - generally switch from using `os.path` to the `Path` library. `f["root"]` could then be a `Path` object, meaning that this kind of syntax could be simplified.

Instead of:

```python
os.path.join(f["root"], "chromosomes.report")
```

We could simply have:
```python
f["root"] / "chromosomes.report"
```

multiqc/modules/bamdst/bamdst.py
```diff
@@ -0,0 +1,419 @@
...
+        """
+        data = dict()
+        s_name = None
+        version = None
+        for line in f["f"].split("\n"):
+            line = line.strip()
+            if line.startswith("## Files : "):
+                # Split a line like "## Files : example/test1.bam example/test2.bam "
+                # Cannot split by spaces because the file names can contain spaces as well
+                path_line = line.split(":")[1].strip().replace(".bam", "<EXT>").replace(".cram", "<EXT>")
```
ewels, line 115: `.sam`?

ewels: Any idea what shows if the tool is used in "pipeline mode"?

From the readme:

> Pipeline mode:
>
> ```bash
> samtools view in1.bam -u | bamdst -p x.bed -o ./ -
> ```

multiqc/modules/bamdst/bamdst.py
```diff
@@ -0,0 +1,419 @@
...
+                names = [self.clean_s_name(sn) for sn in path_line.split("<EXT>") if sn]
+                f["s_name"] = "-".join(names)  # concatenating file names
```
ewels, line 117: Maybe tweak the code to not overwrite `f["s_name"]` if no valid filenames are found, so that it falls back to using the filename.

```suggestion
                if names:
                    f["s_name"] = "-".join(names)  # concatenating file names
```

ewels (changes requested):
Is there a reason that the report I get only has `ST0217_Lg` present in the plots?
I can see other `coverage.report` files, so I was expecting more samples to be plotted..

Other than that, a few minor comments but mostly good to go 👍🏻

multiqc/modules/bamdst/bamdst.py
vladsavelyev: Good point! Added a test case for this. In this case, the log file name will be used for the sample name i.e. (`coverage.report` -> `coverage`), so the user will have to run the tool with `--dirs`. I updated the docs for that

vladsavelyev:
> Is there a reason that the report I get only has ST0217_Lg present in the plots? I can see other coverage.report files, so I was expecting more samples to be plotted..

There are default rules to exclude alt contigs (similar to `mosdepth`) and keep only chr1 ... chrY. Other samples have only data for chrM, which is excluded. 

To make it more transparent, I now added tabs to the plots so users can switch between showing all contigs, and showing only the main ones.

ewels (approved):
Looks good, thanks!

Issue #2173, opened 2023-11-13
Logging problem with full Test Data suite.


ewels:
Something weird is happening with the loggers, when running on the full size test dataset.

It looks like the secondary logger, which typically writes to a file in `multiqc_data`, is starting to spit out logs to the CLI. Or a secondary logger from _somewhere_, anyway. The formatting is wrong, and it is printing debug messages. Info-level messages are duplicated.

You can see this in the CI [here](https://github.com/ewels/MultiQC/actions/runs/6852015507/job/18629583075#step:7:19), for example:

```
/// MultiQC| v1.18.dev0

|    strict_helpers | Checking docs readmes in '/home/runner/work/MultiQC/MultiQC/docs/modules' as --strict specified
|           multiqc | --strict specified. Being strict with validation.
|           multiqc | Search path : /home/runner/work/MultiQC/MultiQC/test_data/data/modules
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 2504/2504  
|               ccs | Found 2 reports
|         ngsderive | Found 2 reports
WARNING:root:ngsderive_instruments_plot: Error getting colour: could not convert string to float: 'e'
WARNING:root:ngsderive_instruments_plot: Error getting colour: could not convert string to float: 'e'
DEBUG:multiqc.modules.purple.purple:Duplicate PURPLE output prefix found! Overwriting: FFPE_sample
DEBUG:multiqc.modules.purple.purple:Duplicate PURPLE output prefix found! Overwriting: good_quality_sample
DEBUG:multiqc.modules.purple.purple:Duplicate PURPLE output prefix found! Overwriting: SEQCII_50pc_purity_cell_line
|            purple | Found 3 reports
INFO:multiqc.modules.purple.purple:Found 3 reports
DEBUG:multiqc.modules.conpair.conpair:Duplicate sample name found! Overwriting: NA[12](https://github.com/ewels/MultiQC/actions/runs/6852015507/job/18629583075#step:7:13)878_tumor80x--NA12878_normal40x
|           conpair | Found 3 reports
INFO:multiqc.modules.conpair.conpair:Found 3 reports
|              lima | Found 2 reports
INFO:multiqc.modules.lima.lima:Found 2 reports
|              lima | Found [27](https://github.com/ewels/MultiQC/actions/runs/6852015507/job/18629583075#step:7:28) samples
INFO:multiqc.modules.lima.lima:Found 27 samples
[...]

PR #2174, opened 2023-11-14
Fix logging spillover

vladsavelyev:
Fixes https://github.com/ewels/MultiQC/issues/2173

There were several sources of logging outside the main multiqc logger:

* Warnings about unknown to Python escape sequences (`\d`, `\W`, etc.) in regular expressions. To make it intentional and remove the warning, escape them with a raw specifier (`r""`).
* Using the `logging` module directly vs the `logger` object (in one place).
* A `print` leftover in the VEP module.


ewels (approved):
Great!

ewels:
ngsderive is throwing a warning with colours, which is why that statement was being triggered:

```
|         ngsderive | Found 2 reports
|        mqc_colour | ngsderive_instruments_plot: Error getting colour: could not convert string to float: 'e'
|        mqc_colour | ngsderive_instruments_plot: Error getting colour: could not convert string to float: 'e'
```

Issue #1899, opened 2023-04-08
Use Click envvars for CLI options / env var templating in config


ewels:
Should use `auto_envvar_prefix` from Click in the command line inferface ([docs](https://click.palletsprojects.com/en/8.1.x/options/#values-from-environment-variables)).

This would make all CLI options settable via exporting environment variables.

For example, instead of:

```bash
multiqc . --title example
```

You could do:

```bash
export MULTIQC_TITLE="example"
multiqc .
```

Could be nice for people running in automated environments / pipelines.

jchorl:
This is a great start. It certainly allows dynamism. I think it would be well paired with string-templating. I'm envisioning:
```yaml
report_header_info:
  - Contact E-mail: "${MULTIQC_CONTACT_EMAIL}"
 ```

Of course, this adds a bunch of complexity. I suppose this would be an easier start:
```
MULTIQC_REPORT_HEADER_INFO='
  - Contact E-mail: "foo@example.com"
'
```

But then you're mixing config between cli/yaml/env-vars, instead of _only_ using env-vars for the template values and leaving the config to yaml.

**Comment by ewels** [comment]
Ok cool, thanks for the example - I understand now what you meant about string-templating 👍🏻 

I'm _slightly_ nervous about blindly replacing things that look like env vars across all config parsing, but it could be done.

What about using a [syntax like mkdocs](https://www.mkdocs.org/user-guide/configuration/#environment-variables)?

```yaml
report_header_info:
  - Contact E-mail: !ENV MULTIQC_CONTACT_EMAIL
```

or with a default:

```yaml
report_header_info:
  - Contact E-mail: !ENV [MULTIQC_CONTACT_EMAIL, "default@email.com"]
```

Same concept, but the `!ENV` prefix is a little more explicit and less likely to have off-target instances.

ewels:
Ahh, I love Python. Someone has already made a library to handle this syntax: https://github.com/mkaranasou/pyaml_env

Also gives a nicer syntax for defaults, and the ability to handle string interpolation:

```yaml
report_header_info:
  - Contact E-mail: !ENV "${NAME:info}@${DOMAIN:example.com}"
```

Issue #2160, opened 2023-11-05
Kaiju: division by zero


kayobianco:
### Description of bug

Hi all,
I have the following problem.

```console
(base) kayobianco@arqueas:~$ multiqc -m kaiju --outdir /home/kayobianco/Documentos/MiSeq_Arqueas_RBD/kaiju --filename kiaju_teste /home/kayobianco/Documentos/MiSeq_Arqueas_RBD/kaiju

  /// MultiQC 🔍 | v1.17

|           multiqc | Only using modules: kaiju
|           multiqc | Search path : /home/kayobianco/Documentos/MiSeq_Arqueas_RBD/kaiju
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 184/184  
|             kaiju | Found 46 reports
╭───────────────── Oops! The 'kaiju' MultiQC module broke... ──────────────────╮
│ Please copy this log and report it at                                        │
│ https://github.com/ewels/MultiQC/issues                                      │
│ Please attach a file that triggers the error. The last file found was:       │
│ /home/kayobianco/Documentos/MiSeq_Arqueas_RBD/kaiju/13_S13_L001_R1_001_kaiju │
│ .out.tsv                                                                     │
│                                                                              │
│ Traceback (most recent call last):                                           │
│   File "/home/kayobianco/miniconda3/lib/python3.8/site-packages/multiqc/mult │
│     output = mod()                                                           │
│   File "/home/kayobianco/miniconda3/lib/python3.8/site-packages/multiqc/modu │
│     self.kaiju_stats_table()                                                 │
│   File "/home/kayobianco/miniconda3/lib/python3.8/site-packages/multiqc/modu │
│     self.kaiju_sample_unclassified[s_name] * 100 / self.kaiju_sample_total_r │
│ ZeroDivisionError: division by zero                                          │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
|           multiqc | No analysis results found. Cleaning up..
|           multiqc | MultiQC complete
```

### File that triggers the error

[kaiju_teste.zip](https://github.com/ewels/MultiQC/files/13261190/kaiju_teste.zip)

### Before submitting

- [X] I have read the [troubleshooting documentation](https://multiqc.info/docs/#troubleshooting-1).
- [X] I am using the latest release of MultiQC.
- [X] I have included a full MultiQC log, not truncated.
[...]

Issue #1790, opened 2022-10-30
Refactor core code to use Pydantic objects


ewels:
Much of the core MultiQC code was written early on when (a) I didn't know very much Python and (b) everything had to work on Python 2.7. I also wrote a lot of the code to be super forgiving. Things like needing a list but allowing a string (and coercing it into a list) if the user wanted. Also tonnes of defaults for everything, even when it doesn't always make sense. This sounds like a nice thing to do, but actually tends to confuse things because stuff doesn't always fail when it probably should do.

Now that MultiQC supports only Python 3 we can use static typing, and specifically the [Pydantic](https://pydantic-docs.helpmanual.io/) library. I think that this it's time to do a major code clean up in the core library to sort out a lot of the mess that has crept in. We can use this opportunity to be much less forgiving with the code and make it clear about what is expected (type hints will help developers a lot). By using Pydantic data models we can get a tonne of nice validation for free, without using much in the way of additional code.  We can also extend the validation methods for custom classes easily.

`pydantic-settings` specifically ([docs](https://docs.pydantic.dev/latest/concepts/pydantic_settings/)) has a lot of functionality that we currently do ourselves, or would like to have.

Pydantic should also be able to help with https://github.com/ewels/MultiQC/issues/1256 by providing access to a wide range of nice output formats. Of particular interest is [Apache Parquet](https://parquet.apache.org/) and JSON Schema, which I think we should be well positioned for.

```[tasklist]
### Progress
- [ ] https://github.com/MultiQC/MultiQC/pull/2442
- [ ] https://github.com/MultiQC/MultiQC/pull/2534
- [ ] https://github.com/MultiQC/MultiQC/pull/2573
```

ewels:
Moving https://github.com/MultiQC/MultiQC/issues/2535 out as it's low priority, so it can sit on the backlog.

PR #2178, opened 2023-11-15
Config with environment variables, and config strict typing

vladsavelyev:
Address https://github.com/ewels/MultiQC/issues/1899

* Allow `MULTIQC_`-prefixed environment variables to override config entries. The load order is after `MULTIQC_CONFIG_PATH`. E.g.

```sh
$ export MULTIQC_TITLE="My report"
$ multiqc -v .
[DEBUG  ]  New config: {'title': 'My report'}
[INFO   ]  Report title: My report
Report      : My-report_multiqc_report.html
```

* Load user configs with [pyaml_env](https://github.com/mkaranasou/pyaml_env) to support interpolation with environment variables. E.g. 

```sh
$ cat multiqc_config.yaml                                                                                                                                                                                                                                                                                                                                                     
report_header_info:
  - Contact E-mail: !ENV "${NAME:info}@${DOMAIN:example.com}"
$ export NAME=me
$ multiqc -v .
[DEBUG  ]  New config: {'report_header_info': [{'Contact E-mail': 'me@example.com'}]}
```

* Add typing to config, and make sure only config is used after the CLI arguments are fully processed in `multiqc.py`. This fixes the bug when `config.title` wasn't used for the report file name.


ewels:
Super cool! 🎉 

@jchorl - does this meet your requirements?

vladsavelyev:
The CI is failing because I added a test example for the fix in https://github.com/ewels/MultiQC/issues/2160 - @ewels please take a look at 2160, it's small

multiqc/multiqc.py
```diff
@@ -499,9 +502,43 @@ def run(
         config.custom_css_files.extend(custom_css_files)
     config.kwargs = kwargs  # Plugin command line options
 
-    # Clean up analysis_dir if a string (interactive environment only)
-    if isinstance(config.analysis_dir, str):
-        config.analysis_dir = [config.analysis_dir]
+    # Discarding the CLI variables to make sure we use only config down below.
```
ewels, line 505: I feel like we could probably avoid this with some more refactoring of the "mega do everything function" approach that is currently in `multiqc.py`.

eg. have dedicated functions to parse config, parse cli, then move around a config object cleanly instead.

Just a thought off the top of my head, can return to this another day in another PR..

ewels (changes requested):
Just gone through the code - wasn't exactly the implementation that I was expecting (I had imagined using Click support for env vars), but I think this is better as it's not limited to just the CLI options.

I really don't like a lot of the code that we have around configs, I'm hoping that we can remove a lot of this at some point in the near-ish future by using Pydantic (see https://github.com/ewels/MultiQC/issues/1790). eg. take a look at `pydantic-settings` here: https://docs.pydantic.dev/latest/concepts/pydantic_settings/

But that shouldn't stop us from moving forward here with this improvement. I think that the functionality here is great and having it in place means that when we do move to Pydantic we have a solid feature base to move from. For example, I'm not 100% sure how we should handle variable interpolation with Pydantic, so having it here gives a good base to use in the new system.

Missing a bit on the docs front:

* Need a dedicated section about environment vars I think, explaining how to find what they should be called (eg. all config options, see the defaults yaml, case doesn't matter etc) with a few examples.
* Need a dedicated section about string interpolation, how it works with some examples.

multiqc/utils/config.py
```diff
@@ -39,12 +41,116 @@
...
+fn_ignore_files: List
+top_modules: List
+module_order: List
+
+# Populating the variables above from the default MultiQC config
+config_defaults_path = os.path.join(MULTIQC_DIR, "utils", "config_defaults.yaml")
+with open(config_defaults_path) as f:
+    _default_config = yaml.safe_load(f)
+for c, v in _default_config.items():
+    globals()[c] = v
```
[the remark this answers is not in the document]
vladsavelyev: Haha. That will be refactored with pydantic etc, but for now I really like how just declaring the variables with type hints enabled static typing with a very minimal invasiveness. Nice incremental improvement

multiqc/multiqc.py
vladsavelyev: There was an old Carmack article where he argued in favour of not breaking up long functions. I think this is related: http://number-none.com/blow/blog/programming/2014/09/26/carmack-on-inlined-code.html
He suggests only refactoring code into a function if it's actually used multiple times, otherwise inline.

But we're going refactor this anyway to support interactive usage, plus to support config with Pydantic, so let's see how it looks.

vladsavelyev:
Done with the docs! @ewels, take a look :)

docs/core/getting_started/config.md
```diff
@@ -13,14 +13,16 @@ it collects the configuration settings from the following places in this order
...
    - Manual installations only, not `pip` or `conda`
-3. User config in `~/.multiqc_config.yaml`
-4. File path set in environment variable `MULTIQC_CONFIG_PATH`
+3. User config in `$XDG_CONFIG_HOME/multiqc_config.yaml` (or `~/.config/multiqc_config.yaml` if `$XDG_CONFIG_HOME` is not set)
+4. User config in `~/.multiqc_config.yaml`
+5. File path set in environment variable `$MULTIQC_CONFIG_PATH`
    - For example, define this in your `~/.bashrc` file and keep the file anywhere you like
-5. Config file in the current working directory: `multiqc_config.yaml`
-6. Config file paths specified in the command with `--config` / `-c`
+6. Environment variables prefixed with `MULTIQC_` (e.g. `$MULTIQC_TITLE`, `$MULTIQC_TEMPLATE`, etc.)
```
ewels, line 20:
```suggestion
6. Environment variables prefixed with `MULTIQC_` (e.g. `$MULTIQC_TITLE`, `$MULTIQC_TEMPLATE`, etc. - see ["Config with environment variables"](#config-with-environment-variables) below)
```

docs/core/getting_started/config.md
```diff
@@ -441,6 +443,63 @@ Qualimap module: _(as [described in the docs](https://multiqc.info/modules/quali
...
+filesearch_lines_limit: 10
+```
+
+:::tip
+Some variables such as `title` can be also directly set through the command line
+options: `--title "My report"`. For a list of all parameters, run `multiqc --help`.
+:::
+
+Note that it is _not_ possible to set nested config parameters through environment
+variables, such as those that expect lists or dicts as values (e.g. `fn_clean_exts`).
```
ewels, line 469:
```suggestion
:::note
It is _not_ possible to set nested config parameters through environment
variables, such as those that expect lists or dicts as values (e.g. `fn_clean_exts`).
:::
```

ewels (approved):
LGTM!

Issue #2162, opened 2023-11-07
Discussion: Sample name cleaning with pairs of input filenames


ewels:
Putting here for discussion, as a thorny issue that could be dangerous if implemented incorrectly. Feedback welcome!

We generally take sample names from input filenames where possible. If a tool takes more than one input (eg. a pair of FastQ files), we typically ignore the second and take the first. This issue is about a suggested method to use both to try to create a "cleaner" resulting sample identifier (typically without a `_1` suffix).

We could look into a generalised function that we could use for every module that has the possibility to find a _pair_ of input FastQ filenames. To try to resolve that pair of names into a single identifier. This could be done by doing a diff of the two filenames (we can't make assumptions about syntax like `_1` because this could break many valid sample identifiers).

For example, given:
* `sample_1_R1_L1`
* `sample_1_R2_L1`

We could remove the `_R1`/`_R2` diff from the two strings to get `sample_1_L1` as a sample identifier.

We would need to be very careful not to remove other data here, so maybe we _only_ do this if the diff is `_R1`/`_R2` or `_1`/`_2` or something. As we don't want to do this:

* `sample_one_R1_Lane1`
* `sample_1_2_L1`

Going to `sample` (or similar). In this case we should just leave the behaviour as current, that is - use the FastQ input 1 (`sample_one_R1_Lane1`).

tamuanand:
Thanks @ewels for starting this discussion.

My thoughts here:

- `fastp` - if you run fastp in PE mode and output general stats as is done currently in 1.17, it mistakenly conveys the impression that the analyses began wit a paired end dataset but then fastp somehow collapsed it all to be interleaved and/or have only `_R1`  datasets coming from the `--in1` argument of fastp. I have had detailed discussions with @vladsavelyev on this and he can fill in on what potental problems this could cause. 
- FWIW, `fastp` also gives separate _R1 and _R2 stats 

```
"read1_before_filtering": {
                "total_reads": 114077968,
                "total_bases": 17111695200,
                "q20_bases": 16805928223,
                "q30_bases": 16233110450,
                "total_cycles": 150,
                "quality_curves": {
SNIPPED
        "read2_before_filtering": {
                "total_reads": 114077968,
                "total_bases": 17111695200,
                "q20_bases": 16355237292,
                "q30_bases": 15526339378,
                "total_cycles": 150,
                "quality_curves": 
----
----
----
        "read1_after_filtering": {
                "total_reads": 110452640,
[...]

PR #2181, opened 2023-11-16
Sample name cleaning with pairs of input filenames

vladsavelyev:
Address https://github.com/ewels/MultiQC/issues/2162

Make `self.clean_s_name` support a lists of paths. Each name will be cleaned separately first, and then a common prefix will be found. If the common prefix is empty, the cleaned names will be dash-concatenated. Since the use case if primarily for pairs of FASTQ files, common FASTQ suffixes will be trimmed.

First check for Illumina conventional [naming](https://support.illumina.com/help/BaseSpace_Sequence_Hub_OLH_009008_2/Source/Informatics/BS/NamingConvention_FASTQ-files-swBS.htm):

```py
>>> self.clean_s_name(["EDM_TAGGCA_L001_R1_001", "EDM_TAGGCA_L001_R2_001"])
"EDM_TAGGCA_L001"
```

Other cases when `_R1`/`_R2` is anywhere:

```py
>>> self.clean_s_name(["sample_1_R1_trimmed", "sample_1_R2_trimmed"])
"sample_1_trimmed"
>>> self.clean_s_name(["TE18-018_12S_R1", "TE18-018_12S_R2"])
"TE18-018_12S"
```

We don't want to remove anything less conventional from the middle, so we only check for other variations in the end (.1, _1, r1, etc):

```py
>>> self.clean_s_name(["fastq_r1", "fastq_r2"])
"fastq"
>>> self.clean_s_name(["Sample_1_1", "Sample_1_2"])
"Sample_1"
>>> self.clean_s_name(["SAMPLE1_PE.1", "SAMPLE1_PE.2")
"SAMPLE1_PE"
```

Adjusting read processing modules to make use of this feature and pass found FASTQ pairs to `self.clean_s_name`:

- [x] trimmomatic
- [x] cutadapt
- [x] fastp

-----

Additionally, fixes a bug in `cutadapt` when an option parameter was taken as a sample name:

```sh
head -n2 issue_1328/SRR4238351_subsamp_m_9999999999.cutadapt.log
This is cutadapt 3.4 with Python 3.9.2
Command line parameters: -a GGGGGGGGGGATATGCGATCGC...GTGTCTCAACGGTATCCGCAACT -o SRR4238351_subsamp.trimed.fastq.gz SRR4238351_subsamp.fastq.gz -m 999999999999999
```
Screenshot 2023-11-16 at 15 10 18


multiqc/modules/base_module.py
```diff
@@ -281,14 +281,70 @@ def add_section(
...
+
+        if len(clean_names) == 2:
+            # Possibly a FASTQ pair - try trimming FASTQ suffixes.
+            r1, r2 = sorted(clean_names)
+
+            # Try trimming the conventional illumina suffix with a tail 001 ending. Refs:
+            # https://support.illumina.com/help/BaseSpace_Sequence_Hub_OLH_009008_2/Source/Informatics/BS/NamingConvention_FASTQ-files-swBS.htm
+            # https://support.10xgenomics.com/spatial-gene-expression/software/pipelines/latest/using/fastq-input#:~:text=10x%20pipelines%20need%20files%20named,individual%20who%20demultiplexed%20your%20flowcell.
+            cleaned_r1 = re.sub(r"_R1_\d\d\d$", "", r1)
+            cleaned_r2 = re.sub(r"_R2_\d\d\d$", "", r2)
```
ewels, line 300:
```suggestion
            cleaned_r1 = re.sub(r"_R1_\d{3}$", "", r1)
            cleaned_r2 = re.sub(r"_R2_\d{3}$", "", r2)
```

multiqc/modules/base_module.py
```diff
@@ -281,14 +281,70 @@ def add_section(
...
+            if cleaned_r1 == cleaned_r2:  # trimmed successfully
+                return cleaned_r1
```
ewels, line 302: If we don't return here, we continue with partially-cleaned sample names, right? Do we want to revert to the full filenames before continuing on to the next attempt?

multiqc/modules/base_module.py
```diff
@@ -281,14 +281,70 @@ def add_section(
...
+
+            # Try removing _R1 and _R2 from the middle.
+            cleaned_r1 = re.sub(r"_R1_", "_", r1)
+            cleaned_r2 = re.sub(r"_R2_", "_", r2)
+            if cleaned_r1 == cleaned_r2:  # trimmed successfully
+                return cleaned_r1
+
+            # Try trimming other variations from the end (-R1, _r1, _1, .1, etc).
+            cleaned_r1 = re.sub(r"([_.-][rR]?1)?$", "", r1)
+            cleaned_r2 = re.sub(r"([_.-][rR]?2)?$", "", r2)
```
ewels, line 312:
```suggestion
            cleaned_r1 = re.sub(r"([_\.-][rR]?1)?$", "", r1)
            cleaned_r2 = re.sub(r"([_\.-][rR]?2)?$", "", r2)
```

multiqc/modules/base_module.py
```diff
@@ -281,14 +281,70 @@ def add_section(
...
+            if cleaned_r1 == cleaned_r2:  # trimmed successfully
+                return cleaned_r1
+
+        # More than 2 names, so can't assume a FASTQ pair. Just trying to keep a common prefix.
```
ewels, line 316: Not necessarily more than 2 names, could be 2 but weren't able to trim.

```suggestion
        # Couldn't clean, or more than 2 names. Just trying to keep a common prefix.
```

multiqc/modules/base_module.py
```diff
@@ -281,14 +281,70 @@ def add_section(
...
+            if len(s_name) == 0:
+                raise ValueError("Empty list of sample names passed to clean_s_name()")
+            if len(s_name) >= 2:
+                # Extract a sample name from a list of file names (for example, FASTQ pairs).
+                # Each name is cleaned separately first.
+                clean_names = [
+                    self.clean_s_name(sn, f=f, root=root, filename=filename, seach_pattern_key=seach_pattern_key)
+                    for sn in s_name
+                ]
+                return self._clean_fastq_name_pair(clean_names)
```
ewels, line 344: I think that we don't want to return yet, no?

eg. if two identical filenames `sample.fastq.gz` and `sample.fastq.gz` are given, it'll return the full `sample.fastq.gz` instead of cleaning to `sample`.

Also, returning at this stage means we miss directory prefixes etc.

multiqc/modules/base_module.py
```diff
@@ -281,14 +281,70 @@ def add_section(
...
+                return cleaned_r1
+
+        # More than 2 names, so can't assume a FASTQ pair. Just trying to keep a common prefix.
+        prefix = os.path.commonprefix(clean_names)
+        prefix = prefix.rstrip("_.- ")
+        # If the prefix is empty, join the names with a dash.
+        if prefix == "":
+            s_name = "_".join(clean_names)
+        else:
+            s_name = prefix
```
ewels, line 323: Not sure if this is too generic 🤔  Need to think about some edge cases to test it with.

ewels (commented):
Nice, couple of questions / comments but basically good to go I think.

multiqc/modules/base_module.py
vladsavelyev: I personally prefer to duplicate, but that's alright :)

multiqc/modules/base_module.py
vladsavelyev: Not needed inside square brackets

multiqc/modules/base_module.py
vladsavelyev: > eg. if two identical filenames `sample.fastq.gz` and `sample.fastq.gz` are given, it'll return the full `sample.fastq.gz` instead of cleaning to `sample`.

It will be cleaned to two identical [`sample`, `sample`] in the list comprehension above. That's why the variable is called `clean_names` :)

multiqc/modules/base_module.py
vladsavelyev: We clean the full names `r1`/`r2` on every attempt, and `cleaned_r1`/`cleaned_r2` are discarded

ewels: Ah you're right, was misreading sorry 👍🏻

multiqc/modules/base_module.py
ewels: huh, nice!

multiqc/modules/base_module.py
ewels: Ooh I didn't notice that - calling the function on itself, nice!

multiqc/modules/base_module.py
ewels: So am I right in thinking that the following would be true here?

`sample_1` + `sequence_2` goes to `s`

vladsavelyev: Yeah, that's right. Not sure how common that usecase is. Perhaps we should always "_"-merge?

vladsavelyev: Done!

ewels (approved):
Code LGTM 👍🏻

Issue #2182, opened 2023-11-17
GATK: Rescale Emperical Quality scatter plot


sainsachiko:
### Description of feature

In this scatter plot, scales are approximately equal, but for some reason, two axes were not plotted with the same scale.
Please consider adding `square: true `to pconfig in file `multiqc/modules/gatk/base_recalibrator.py` and maybe add a dotted diagonal line too.
Thanks!
![image](https://github.com/ewels/MultiQC/assets/134130358/0d68017d-e3a7-4d3b-b88d-25bb6a4f14a5)

ewels:
Square config is for this plot config:

https://github.com/ewels/MultiQC/blob/5eb990d3db406c7bcf1800e630c05492890fc191/multiqc/modules/gatk/base_recalibrator.py#L181-L188

Additional task is to ensure that `ymax = xmax`. We could consider making this doable via plot config (eg. `"match_axis_limits" = True` so that it's easy to apply in a more generalisable way.

vladsavelyev:
I added `"square": True` to the plot config here. I also thought it's a good idea to ensure `xmax` == `ymax` in the general plotting code, so the user code only sets `"square"` in the config. And passing `endOfTick` to HighCharts to make sure it doesn't break the square limits by adding inconsistent padding. 

https://github.com/ewels/MultiQC/pull/2189

Also, _scatter_ plot doesn't seem to support drawing a diagonal line :-( Something to come back to with Plotly.

PR #2189, opened 2023-11-17
Fix `"square": True` for scatter plot. Make the GATK BaseRecalibrator plot square

vladsavelyev:
Fix https://github.com/ewels/MultiQC/issues/2182

- Pass `"square"` to the plot config
- Set `xmin=0` and `ymax=0` .
- Assure `xmax` == `xmin` in the python plotting code.
- Fix the scatter plot plotting code to re-calculate the total max value, pass `endOfTick` and calculate padding manually to make sure HighCharts doesn't get creative with padding


ewels (approved):
LGTM 👍🏻

Issue #2190, opened 2023-11-17
Dropping quast namespace broke `table_columns_visible`


jchorl:
### Description of bug

We used to have a config like:
```yaml
table_columns_visible:
  QUAST:
    "# misassemblies": false
```

This commit removed the quast namespace: https://github.com/ewels/MultiQC/commit/056d21d3006561f8ce31fdea53c06a8336744ad8#diff-6d91e36cef5a41720305bb1bd9f0ab612e3830d0cce62efc0933a1d982b5296dL252

We now need to specify this config as:
```yaml
table_columns_visible:
  "":
    "# misassemblies": false
```

Is this expected/desirable?

Thanks!

### Before submitting

- [X] I have read the [troubleshooting documentation](https://multiqc.info/docs/usage/troubleshooting/).
- [X] I am using the latest release of MultiQC.
- [X] I have included a full MultiQC log, not truncated.
- [X] I have attached an input file (**.zip** if necessary) that triggers the error.

vladsavelyev:
Hi @jchorl,

That worked before really by accident, as we don't normally pass the `namespace` to `table.plot()` config in modules unless we want to tell apart between different submodules/subtools (like in DRAGEN and CellRanger), and QUAST was an exception.

In the future refactoring, we'll make it more flexible, but the way it is supported now is exactly as you modified it:

```py
table_columns_visible:
  "":
    "# misassemblies": false
```

I think I'll simplify the usage to:

```py
table_columns_visible:
  "# misassemblies": false
```

vladsavelyev:
Done (see https://github.com/ewels/MultiQC/pull/2191) - now you can also use a flat config like this:

```yaml
table_columns_visible:
  "# misassemblies": false
```

The column IDs should be unique anyway :)

PR #2191, opened 2023-11-20
`table_columns_visible` and `table_columns_name`: support flat config and `table_id` as a group

vladsavelyev:
Fix https://github.com/ewels/MultiQC/issues/2190

Using "Group" in `table_columns_visible` and `table_columns_name` only works for the general stats table:

```yaml
table_columns_visible:
  QUAST:
    "# misassemblies": false
```

As we don't normally pass the namespace to `table.plot()` (QUAST was an exception). Refactoring the plotting code and calling the functions through `self.table_plot()` would allow access to the module name in the plotting code, and make it work everywhere. But for now, we cannot rely on the namespace here.

Instead, we can support passing the table ID:

```yaml
table_columns_visible:
  quast_table:
    "# misassemblies": false
```

Or just making it flat  - the IDs should be unique anyway:

```yaml
table_columns_visible:
  "# misassemblies": false
```

Same is done for `table_columns_name`.

I also allowed column titles as keys:

```yaml
table_columns_visible:
  "Misassemblies": false
```

Finally, re-added namespace for QUAST for back-compatibility.

### DIFF
--- a/multiqc/plots/table_object.py
+++ b/multiqc/plots/table_object.py
@@ -158,39 +158,43 @@ class DataTable:
 
                 # Overwrite "name" if set in user config
                 # Key can be a column ID, a table ID, or a namespace in the general stats table.
-                for key in config.table_columns_name.keys():
+                for key, val in config.table_columns_name.items():
+                    key = key.lower()
                     # Case-insensitive check if the outer key is a table ID or a namespace.
-                    if key.lower() == pconfig["id"].lower() or key.lower() == headers[idx][k]["namespace"].lower():
+                    if key in [pconfig["id"].lower(), headers[idx][k]["namespace"].lower()] and isinstance(val, dict):
                         # Assume a dict of specific column IDs
-                        headers[idx][k]["title"] = config.table_columns_name[key][k]
+                        for key2, new_title in val.items():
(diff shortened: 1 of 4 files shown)


multiqc/plots/table_object.py
```diff
@@ -156,33 +156,43 @@ def __init__(self, data, headers=None, pconfig=None):
...
+                for key in config.table_columns_name.keys():
+                    # Case-insensitive check if the outer key is a table ID or a namespace.
+                    if key.lower() == pconfig["id"].lower() or key.lower() == headers[idx][k]["namespace"].lower():
+                        # Assume a dict of specific column IDs
+                        headers[idx][k]["title"] = config.table_columns_name[key][k]
+
+                    # Case-insensitive check if the outer key is a column ID
+                    elif key.lower() == k.lower() or key.lower() == headers[idx][k]["title"].lower():
+                        if isinstance(config.table_columns_visible[key], bool):
+                            headers[idx][k]["title"] = config.table_columns_visible[key]
```
ewels, line 170: Should this be `config.table_columns_name`, not `config.table_columns_visible`?

ewels (approved):
LGTM! 👍🏻

PR #2202, opened 2023-11-29
Re-add `run()` into the `multiqc` namespace

vladsavelyev:
Ruff [suggested](https://github.com/ewels/MultiQC/commit/64f9c63a789aa24cfecb9c826effb5fb6c95ed2c#diff-272bc7573ee7d640f5ae69b4bdf83c95f2d754fe4a615a1787b37f634f8ff1b1L16) removing the `from .multiqc import run` from `multiqc/__init__.py` as an "unused import". 

Adding `__all__ = ["run", "config", "__version__"]` to properly indicate which imports we want to add into this namespace.


ewels (approved):
Nice catch. Thanks!

Issue #1228, opened 2020-06-17
Kraken module contains 'uncustomisable' dynamically generated column names


jfy133:
**Description of bug:**

Originally reported on gitter: https://gitter.im/ewels/MultiQC?at=5ee0edd4035dfa1261112e3f

The Kraken module will report a '%' of the top taxonomic hit in the general stats table. The key is then 'dynamically generated', which means if you have different samples/data, you cannot use a common key to place the column in the desired position in the table with a `config.yml`, as this column will be named differently in each dataset.

Example
![image](https://user-images.githubusercontent.com/17950287/84878185-c6c48f00-b089-11ea-8a4d-39b2767c2eea.png)

**MultiQC run details (please complete the following):**
* Command used to run MultiQC: 
```bash
cd MultiQC_TestData/data/modules/kraken/
multiqc .
## dynamic ID will be: % Peptoclostridium difficile 
```
* MultiQC Version: 1.9
* Operating System: Ubuntu 18.04
* Method of MultiQC installation: pip

vladsavelyev:
Fixed by https://github.com/ewels/MultiQC/pull/2205 - thank a lot @gsmashd!

PR #2205, opened 2023-12-01
kraken: fixed column keys in genstats

gsmashd:
Change header keys in kraken module as described in Issue #1228. Column headers should not be set dynamically since this makes it impossible to predictably place or disable columns. 

Keys for the two other columns from kraken in genstats are also changed to be more in line with naming conventions for other modules. 

The results for this branch on MultiQC_Testdata/data/modules/kraken yields identical results when compared to ewels/MultiQC/master.


vladsavelyev (approved):
Thank you @gsmashd for the contribution! It definitely makes a lot of sense, merging.

Issue #2203, opened 2023-11-30
Data missing from JSON, when module run multiple times.


fgvieira:
### Description of bug

I am trying to plot the `FastQC` module multiple times (like [here](https://multiqc.info/docs/reports/customisation/#running-modules-multiple-times)) but, even though all looks good in the `html` report, only the latest data is stored in the `json`. That is, in the example above, only the `FastQC (trimmed)` data is stored in the `json`; is this intended?

Would it be possible to save both? Maybe by renaming `fastqc_sequence_length_distribution_plot` to `fastqc_trimmed-sequence_length_distribution_plot` and `fastqc_raw-sequence_length_distribution_plot` (like in the rest of the report)?

### MultiQC Error log

```console
No error reported.
```

### Before submitting

- [X] I have read the [troubleshooting documentation](https://multiqc.info/docs/usage/troubleshooting/).
- [X] I am using the latest release of MultiQC.
- [X] I have included a full MultiQC log, not truncated.
- [X] I have attached an input file (**.zip** if necessary) that triggers the error.

vladsavelyev:
Thank you @fgvieira for reporting an issue - we definitely should save both, sounds like a bug. I'll take a look now

fgvieira:
Which version are you using?
Because it is not on the JSON file [multiqc_data.json](https://github.com/ewels/MultiQC/files/13532051/multiqc_data.json)

vladsavelyev:
I tried both `1.18 (af792aa)` (the latest stable release installed with `pip install multiqc==1.18`) as well as master (`1.19.dev0 (af792aa)`.

I wonder if it has to do with the input data? I can see e.g. both `fastqc_per_base_sequence_quality_plot` and `fastqc_per_base_sequence_quality_plot-1` in your JSON. And `sequence_length_distribution-1` might be missing because the `sequence_length_distribution` data was missing in the FastQC output for the `FastQC (raw)` samples?

Wonder if you run MultiQC in verbose mode with `-v`, it will print `sequence_length_distribution not found in FastQC reports` into the stderr?

If you could attach the FastQC reports are you running on, along with the `multiqc_config.yaml`, that would be really helpful for me to reproduce the issue!

fgvieira:
I did not see anything out of the ordinary, but here is some example data:

[fastqc_raw.zip](https://github.com/ewels/MultiQC/files/13532292/fastqc_raw.zip)
[...]

Issue #2076, opened 2023-09-26
label output plots using anchor


sstrong99:
### Description of feature

When a module is used more than one time in the multiqc config, I can control the names of the output data files using the `anchor` key in the config, but this does not affect some of the ouput files, e.g. plots. For example, with the samtools module, one of the output data files is named `multiqc_samtools_idxstats_<anchor>.txt`, but i also get a bunch of files named `mqc_samtools-idxstats-mapped-reads-plot-<number>_*.txt`. It would be useful if the plot files (and any other output data files) also used the anchor instead of the index number.

ewels:
This is likely not tied into the plot data export code, but rather the plot generation code - as I think that the data filenames are based on the plot HTML IDs.

Need to go into that code and get the IDs to use the module anchor.

ewels:
Oof, ok. Does it make sense to push this back to the bigger module refactoring that we're planning, where we will make heavier use of classes and this kind of info will be easier to get at?

It would be nice to update all plots to supply as simple an `id` as possible and automatically prefix, however we don't want to break a tonne of filenames that people may be relying on. So that bigger refactoring across all modules might be a better v2-ish target.. 🤔

vladsavelyev:
I guess so. Can't get around without changing some IDs and breaking something for someone, unfortunately. But hopefully it will be a rare case, and only for some possible power users who will know how fix this quickly on their site.

PR #2206, opened 2023-12-01
Use `self.anchor` to prepend plot IDs to make sure custom anchor is applied

vladsavelyev:
Use `self.anchor` to prepend plot IDs to make sure custom anchor is applied
Helpful when module is run twice with a `module_order` [config](https://multiqc.info/docs/reports/customisation/#running-modules-multiple-times):

```yaml
module_order:
  - fastqc:
      name: "FastQC (trimmed)"
      anchor: "fastqc_trimmed"
      info: "This section of the report shows FastQC results after adapter trimming."
      target: ""
      path_filters:
        - "*_1_trimmed_fastqc.zip"
  - cutadapt
  - fastqc:
      name: "FastQC (raw)"
      anchor: "fastqc_raw"
      path_filters:
        - "*_1_fastqc.zip"
```

In the result, instead of numerical suffixes in the JSON outputs:

```json
"fastqc_sequence_length_distribution_plot": {
...
"fastqc_per_base_sequence_quality_plot": {
...
"fastqc_per_base_sequence_quality_plot-1": {
...
```

It becomes more clear which data belongs to which section:

```json
"fastqc_trimmed_sequence_length_distribution_plot": {
...
"fastqc_trimmed_per_base_sequence_quality_plot": {
...
"fastqc_raw_per_base_sequence_quality_plot": {

```

Which is especially relevant for the sequence_length_distribution plot that would only be built for trimmed reads, and would always be skipped for raw reads due to identical sequence length.

Only fastqc module for now - WIP. 

Fixes https://github.com/ewels/MultiQC/issues/2076
Fixes https://github.com/ewels/MultiQC/issues/2203

UPD: Should probably make it automatic, which requires https://github.com/ewels/MultiQC/pull/2036 to be completed first

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Create CODE_OF_CONDUCT.md ([#2195](https://github.com/ewels/MultiQC/pull/2195))
 - BCLConvert: fix mean quality, fix count-per-lane barplot ([#2197](https://github.com/ewels/MultiQC/pull/2197))
 - Re-add `run()` into the `multiqc` namespace ([#2202](https://github.com/ewels/MultiQC/pull/2202))
+- Use `self.anchor` to prepend plot IDs to make sure custom anchor is applied ([#2206](https://github.com/ewels/MultiQC/pull/2206))
 
 ### New Modules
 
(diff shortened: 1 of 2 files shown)


ewels:
Maybe we can do this prepending of the anchor to the ID in the table / plot code instead, so that people don't have to manually remember to do it.

Means we have to clean up the duplicated prefixes across all modules 😱 and where possible avoid changes in the resulting IDs, but hopefully not too bad..

ewels:
Maybe fits nicely into the refactoring that we're doing for https://github.com/ewels/MultiQC/issues/1051

Issue #2204, opened 2023-11-30
Version 1.18 ignores ordering of an OrderedDict


rich7409:
### Description of bug

We have an internal plugin that creates a horizontal bar plot. The order of the rows is defined by how they are read from a file and we need that order preserved. Thus we've used OrderedDict for the plot data.

This has been working correctly for all versions up to and including 1.17 but the latest 1.18 disregards the order of the object and we end up with a plot ordered by the alphabetic keys.

I suspect a copy into another dictionary has been introduced between the two versions with natural ordering of the keys. I've not poked around in the source code but it appears the object we create is not the one that is used to render the chart's data.

If this is intended behaviour we will need to change our code (or stay on 1.17) but I'm raising this as it might be an unintended change.

### MultiQC Error log

```console
python3.9 -m multiqc -n zz_SINAA9.html --clarity https://limsdev.cruk.cam.ac.uk --force --no-data-dir --barcode-file testdata/clarityBarcodes.json -e bcl2fastq -e bclconvert --title 'SINA testing' --config /home/bowers01/work/pipelines/nf-sequencing/resources/multiqc_config.yaml testdata/231120_A00489_2045_BHNV7LDRX3

  /// MultiQC 🔍 | v1.18

|           multiqc | Report title: SINA testing
|           multiqc | Excluding modules 'bcl2fastq', 'bclconvert'
|           multiqc | Search path : /home/bowers01/work/multiqcplugins/multiqc_barcodebalance_plugin/testdata/231120_A00489_2045_BHNV7LDRX3
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 3/3  
|    barcodebalance | Using https://genomicsequencing.cruk.cam.ac.uk as the Clarity server.
|    barcodebalance | Found 2 demultiplexing stats files
|           multiqc | Report      : zz_SINAA9.html   (overwritten)
|           multiqc | Data        : None
|           multiqc | MultiQC complete
```

### Before submitting

- [X] I have read the [troubleshooting documentation](https://multiqc.info/docs/usage/troubleshooting/).
- [X] I am using the latest release of MultiQC.
- [X] I have included a full MultiQC log, not truncated.
- [X] I have attached an input file (**.zip** if necessary) that triggers the error.

vladsavelyev:
Thank you @rich7409 for creating an issue. It sounds like an unintended behaviour, so sorry about it - I'll take a look into this asap.

[...]

PR #2154, opened 2023-10-27
Add Ruff linter. Fix formatting

vladsavelyev:
* Apply [ruff linter](https://github.com/ewels/MultiQC/issues/2149) to the entire code base: 
    * `ruff format .`
    * `ruff --fix .`
    * Manually addressing issues reported by `ruff .`
* Remove `OrderedDict`s (built in `dict`s are de facto sorted since Python 3.6, and guaranteed to be sorted since Python 3.7)
* Remove Python 2 code

Fixes https://github.com/ewels/MultiQC/issues/2149

PR #2210, opened 2023-12-04
Bar graphs: add `sort_samples: false` config option, also do not sort `OrderedDict` data

vladsavelyev:
Fix an artefact of removing `OrderedDict`s https://github.com/ewels/MultiQC/pull/2154

The idea of the original code seems this: before Python 3.6, regular dict were randomly shuffled, so MultiQC needed to post-sort the sample list to get some reasonable order. But if the module wrapped data in an `OrderedDict`, it meant that the user explicitly provided the order, and it's not needed to be sorted further: 

```py
if isinstance(d, OrderedDict):
    hc_samples = list(d.keys())
else:
    hc_samples = sorted(list(d.keys()))
```

After Python 3.6, regular dicts keep order, so the need for that sorting has disappeared. So instead, module devs started to assume that `OrderedDict` is a hacky way to tell MultiQC to disable auto-sorting. Instead, we should always keep the order of passed dictionaries, and treat `OrderedDict`s and `dict`s equally (we wouldn't use `OrderedDict` anywhere internally anyway).

Fixes https://github.com/ewels/MultiQC/issues/2204


ewels:
Happy only if:

* The order of samples is maintained between runs
* The default of order samples is alphabetical for the majority of modules where sample order is not important

It would be nice to see an example where this PR changes the output behaviour to think about this.

May need a new config option to disable alphabetical sorting? I agree that the previous behaviour with OrderedDicts was a bit hacky.

If basically nothing is changing, then happy to merge 👍🏻

vladsavelyev:
Thinking about it more, decided to keep the order of `OrderedDict` inputs for legacy reasons, and also added an non-hacky way to disable sorting with the option `pconfig.sort_samples` which is `True` by default.

ewels (approved):
Perfect, thanks!

Issue #2198, opened 2023-11-24
Error when running with `--no-report`


lacek:
### Description of bug

Example command that triggers the error:

```bash
multiqc --no-report stats
```

When `--no-report` is given, `config.output_fn` is None (https://github.com/ewels/MultiQC/blob/v1.18/multiqc/multiqc.py#L676) and therefore the debug statement (https://github.com/ewels/MultiQC/blob/v1.18/multiqc/multiqc.py#L1064) will throw error.

### File that triggers the error

Just arbitrary valid input, e.g. a bcftools stats output txt.

### MultiQC Error log

```console
/// MultiQC 🔍 | v1.18

|           multiqc | Search path : /runtime/stats
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 1/1
|          bcftools | Found 1 stats reports
|           multiqc | Report      : None
|           multiqc | Data        : multiqc_data
Traceback (most recent call last):
  File "/usr/local/bin/multiqc", line 8, in <module>
    sys.exit(run_multiqc())
             ^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/multiqc/__main__.py", line 23, in run_multiqc
    multiqc.run_cli(prog_name="multiqc")
  File "/usr/local/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/rich_click/rich_command.py", line 126, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/multiqc/multiqc.py", line 300, in run_cli
    multiqc_run = run(**kwargs)
                  ^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/multiqc/multiqc.py", line 1064, in run
[...]

PR #2212, opened 2023-12-04
Fix running `--no-report`

vladsavelyev:
Fixes the bug introduced with adding a "full path to the HTML" debug message.

Fixes https://github.com/ewels/MultiQC/issues/2198


ewels (approved):
Good catch!

Issue #2172, opened 2023-11-13
No analysis results found for fastp JSON report


timz0605:
### Description of bug

Hello, I am running multiqc in the same directory as where I store my JSON reports for data processed by pastp. However, I have been getting this error message:

### MultiQC Error log

```console
/// MultiQC 🔍 | v1.17

|           multiqc | Search path : /mnt/c/Baker Lab/ARMS_data/raw_fastq_files
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 514/514
|           multiqc | No analysis results found. Cleaning up..
|           multiqc | MultiQC complete
```

### Before submitting

- [X] I have read the [troubleshooting documentation](https://multiqc.info/docs/#troubleshooting-1).
- [X] I am using the latest release of MultiQC.
- [X] I have included a full MultiQC log, not truncated.
- [X] I have attached an input file (**.zip** if necessary) that triggers the error.

ewels:
Please attach some example reports, there's not much that we can do without being able to replicate the error. It works fine with our test data :)

ailtonpcf:
I'm having the same issue with multiqc v.1.18. Below follows one fastp report example. Thank you in advance.
[example_fastp.zip](https://github.com/ewels/MultiQC/files/13448003/example_fastp.zip)

vladsavelyev:
@ailtonpcf, thanks so much for providing the reproducible example!

MultiQC needs the fastp log files to end with `fastp.json`: https://multiqc.info/modules/fastp/

So `SRR8487000.json` doesn't work, but `SRR8487000_fastp.json` would.

Otherwise, it's hard to tell if a file comes to fastp. Checking the contents for `fastp_version` line looks promising, but the older versions don't have this line as it was added after v0.22.0: https://github.com/OpenGene/fastp/commit/8c8ddb6a79844ee137835d65adf60773708def80

I might look into allowing _either_ `fastp_version` to be found in contents, or the file name ending with `fastp.json`. But for now, renaming the file would help!

PR #2213, opened 2023-12-04
Fastp: search pattern: look at content instead of file name

vladsavelyev:
Fixes https://github.com/ewels/MultiQC/issues/2172

MultiQC requires fastp inputs to end with `*fastp.json` (https://github.com/ewels/MultiQC/issues/2172#issuecomment-1839141197), but we actually have enough data to tell them apart if we require any "*.json" and check for one or more lines that we always expect there to be.

It would be nicer to be able to check for `fastp_version`, but it was only added 2 years ago: https://github.com/OpenGene/fastp/commit/8c8ddb6a79844ee137835d65adf60773708def80


ewels (approved):
As long as the change doesn't significantly impact run time, then I'm happy 👍🏻

Issue #1944, opened 2023-06-28
nonpareil module


fgvieira:
### Name of the tool

nonpareil

### Tool homepage

https://github.com/lmrodriguezr/nonpareil

### Tool description

Estimate metagenomic coverage and sequence diversity

### Tool output

This is a second attempt to issue #474.

Tool outputs a TSV that needs to be post-processed through some `R` functions to estimate the model.

I've recently created a [snakemake wrapper](https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/nonpareil/plot.html) to plot the results from `nonpareil`, but not sure how to get them into `MultiQC`.

The wrapper outputs the main `R` object as a `.RData` file (example [here](https://github.com/ewels/MultiQC/files/11894162/sample_data.zip) for both single- and multi-sample analysis), but not sure how to import it into `MultiQC`. Maybe a `json`?

### Data suitable for MultiQC plot(s)

In a plot, I think it would be useful to see the same as in [here](http://enve-omics.ce.gatech.edu/nonpareil/results?jid=example1).

Mainly, the observed curve, the model curve (if estimated), and desired threshold (red dotted line).

### Most interesting data for the General Stats table

In the `R` object there are several values that could be interesting to show on the General Stats table:
```
          Redundancy Avg. coverage Seq. effort Model correlation
sample_10       0.60          0.63       12.48                 1
sample_11       0.75          0.77       16.83                 1
sample_12       0.44          0.47        6.45                 1
sample_13       0.50          0.53        9.02                 1
sample_1        0.29          0.32        8.23                 1
sample_2        0.52          0.55       14.48                 1
sample_3        0.38          0.42       12.20                 1
sample_4        0.31          0.35       10.29                 1
sample_5        0.36          0.39        9.92                 1
sample_6        0.35          0.38       10.83                 1
sample_7        0.67          0.69       13.87                 1
sample_8        0.50          0.53        6.63                 1
sample_9        0.55          0.58        8.88                 1
          Required seq. effort Diversity
sample_10               170.47     22.14
sample_11                62.80     21.58
sample_12               171.31     22.53
sample_13               198.09     22.51
sample_1                464.34     23.72
sample_2                182.51     22.98
sample_3                372.83     23.54
sample_4                439.33     23.77
sample_5                370.82     23.48
sample_6                445.06     23.63
sample_7                 82.80     21.97
sample_8                126.29     22.17
sample_9                112.49     22.19
```

### Before submitting

[...]

Issue #474, opened 2017-06-08
Module request: nonpareil


ekopylova:
Nonpareil: Nonpareil uses the redundancy of the reads in a metagenomic dataset to estimate the average coverage and predict the ammount of sequences that will be required to achieve "nearly complete coverage", defined as ≥95% or ≥99% average coverage. [Link to website](http://enve-omics.ce.gatech.edu/nonpareil/).

Examples of coverage curve, log and output files on tool [website](http://enve-omics.ce.gatech.edu/nonpareil/results?jid=example1).

Coverage, actual effort and required effort all important values.

Thanks!

ekopylova:
I added output files in [PR #38](https://github.com/ewels/MultiQC_TestData/pull/38).

iimog:
I looked into the four provided log files. It appears that they do not contain the numbers you are interested in. Instead those have to be calculated frorm the `.npo` file using the `Nonpareil.curve()` R function (see the [documentation](http://nonpareil.readthedocs.io/en/latest/curves.html)). As this is a non-standard function and there is no standardized way how the calculated values are stored alongside the log files those can not be easily integrated into MultiQC reports.

The actual effort can be read from the `.npo` file directly but neither the required effort nor the coverage (those require the model to be fitted). I assume that the actual effort alone is not useful? Are there any other numbers that can be read from the log files directly?

ewels:
Hi @ekopylova,

I'm going to close this issue for now as I can't see how we're going to be able to make a standard module with useful summary statistics for this tool as it stands. Please do come back with comments if you're able to and we can reopen the issue.

Phil

fgvieira:
I've recently created a [snakemake wrapper](https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/nonpareil/plot.html) to plot the results from `nonpareil`, but it would be nicer to have the plots on MultiQC.

Since we need to run `Nonpareil.curve()` to estimate the model, I could add an output to the wrapper with the model data for input to MultiQC.
Would it make sense?

ewels:
Please move this to a new issue, it'll get lost here otherwise 🙏🏻 

But short answer: you'll need to generate the plot data and save it to a file that MultiQC can pick up and plot. Importing `.RData` is non-ideal for many reasons (technically difficult but also will be much slower and not scale well).

Issue #2214, opened 2023-12-05
Plot confidence area around a linegraph


fgvieira:
### Description of feature

Is it possible to plot confidence areas (e.g. confidence intervals or standard deviations) around a line graph?

Like [this](https://stackoverflow.com/a/32616695) example.

ewels:
No, and I have avoided doing this in the past because it does not scale well. Having a single additional band as suggested in https://github.com/ewels/MultiQC/issues/2218 is fine, but having a confidence area per-sample is bad as the graph will quickly become unreadable when there are more than a few samples.

Closing as a duplicate of https://github.com/ewels/MultiQC/issues/2218

vladsavelyev:
I wonder if we can still do that only if the number of samples is low. It would fit well into the "plot scaling" effort. 

E.g. a confidence area can be displayed _only_ if there is only a single sample. Some additional per-sample dots can be displayed only if there are _few_ samples. When scaling further, we will omit some lines and fall back for different representations, like discussed [here](https://github.com/MultiQC/MultiQC/issues/2280). 

Certain modules are often run on a single sample anyway, so the problem of scaling is not that important there.

PR #2215, opened 2023-12-06
Add nonpareil module

fgvieira:

Adds module to plot `nonpareil` results (following #1944 and #474). Since `nonpareil` post-processes output files in `R`, this module requires the user to save the `curves` object as a `json` file (with the functions specified in #1944).


fgvieira:
Ideally, waiting for #2214.

multiqc/modules/nonpareil/nonpareil.py
```diff
@@ -0,0 +1,276 @@
...
+                }
+
+        return json_raw
+
+    def nonpareil_general_stats_table(self):
+        """Take the parsed stats from the nonpareil report and add it to the
+        General Statistics table at the top of the report"""
+
+        headers = {
+            "L": {
```
vladsavelyev, line 126: I'd try to rename the keys as well to something a bit longer and readable, like `read_length`. They end up in various places and I'd prefer to still keep them human-readable of possible

multiqc/modules/nonpareil/nonpareil.py
```diff
@@ -0,0 +1,276 @@
...
+                "min": 0,
+                "scale": "BuGn",
+                "hidden": True,
+                "format": "{:,.2f}",
+            },
+            "kappa": {
+                "title": "Redundancy",
+                "description": "Dataset redundancy",
+                "max": 1,
+                "min": 0,
```
vladsavelyev, line 181: Should it be a percentage?

multiqc/modules/nonpareil/nonpareil.py
```diff
@@ -0,0 +1,276 @@
...
+                "scale": "RdYlGn-rev",
+                "format": "{:,.2f}",
+            },
+            "C": {
+                "title": "Coverage",
+                "description": "Dataset coverage",
+                "max": 1,
+                "min": 0,
```
vladsavelyev, line 189: Should it be a percentage too?

multiqc/modules/nonpareil/nonpareil.py
```diff
@@ -0,0 +1,276 @@
...
+                "title": "Adjusted read length",
+                "description": "Adjusted read length (if using kmers)",
+                "min": 0,
+                "scale": "GnBu",
+                "suffix": " bps",
+                "format": "{:,.3f}",
+                "hidden": True,
+            },
+            "R": {
+                "title": "Number of reads",
```
vladsavelyev, line 144: Would be nice to use the global read count multiplier here for readability, see: https://multiqc.info/docs/development/compatibility/#updated-number-formatting
Along with `'shared_key': 'read_count'`

multiqc/modules/nonpareil/nonpareil.py
```diff
@@ -0,0 +1,276 @@
...
+                "min": 0,
+                "scale": "GnBu",
+                "format": "{:,.0f}",
+                "hidden": True,
+            },
+            "LR": {
+                "title": "Sequencing effort",
```
vladsavelyev, line 151: Same as the read count multiplier above (https://multiqc.info/docs/development/compatibility/#updated-number-formatting), but for base count (e.g. `config.base_count_prefix`, etc).

multiqc/modules/nonpareil/nonpareil.py
```diff
@@ -0,0 +1,276 @@
...
+                "description": "Dataset coverage",
+                "max": 1,
+                "min": 0,
+                "scale": "RdYlGn",
+                "format": "{:,.2f}",
+            },
+            "consistent": {
+                "title": "Consistent Data?",
+                "description": "Is the data sufficient for accurate estimation?",
+                "modify": lambda x: "Yes" if x == 1 else "No",
```
vladsavelyev, line 196: If you use `true` and `false` here instead of `Yes` and `No`, MultiQC will nicely format this field
(these global rules are used across all modules: https://github.com/ewels/MultiQC/blob/2be5617c387266646e3eb2fc62ab43ebc6398075/multiqc/utils/config_defaults.yaml#L75-L85)

multiqc/modules/nonpareil/nonpareil.py
```diff
@@ -0,0 +1,276 @@
...
+                "scale": False,
+                "hidden": True,
+            },
+            "star": {
+                "title": "Ideal coverage",
+                "description": "Coverage considered ’nearly complete’.",
+                "min": 0,
+                "suffix": "%",
+                "scale": False,
```
vladsavelyev, line 205: Is there no scale here because the number is constant? Would be nice to add some sentence into the description about how this number is used.

multiqc/modules/nonpareil/nonpareil.py
```diff
@@ -0,0 +1,276 @@
...
+        log.debug(data_labels)
+
+        pconfig = {
+            "id": "nonpareil-redundancy-plot",
+            "title": "Nonpareil: Redundancy levels",
+            "xlab": "Sequencing effort (Mbps)",
+            "ylab": "Estimated Average Coverage",
+            "xmin": 1e-3,
+            "ymin": 0,
+            "ymax": 1,
```
vladsavelyev, line 261: Perhaps these should be modified to percentages as well?

multiqc/utils/search_patterns.yaml
```diff
@@ -535,6 +535,11 @@ ngsderive/encoding:
 ngsderive/junction_annotation:
   contents: "File	total_junctions	total_splice_events	known_junctions	partial_novel_junctions	complete_novel_junctions	known_spliced_reads	partial_novel_spliced_reads	complete_novel_spliced_reads"
   num_lines: 1
+nonpareil:
+  - fn: "*.json"
+    contents: "LRstar"
+    num_lines: 500
```
vladsavelyev, line 541: We try to avoid making MultiQC loading long files into memory, is there any chance `num_lines` can be smaller? In the test example, `LRstar` is found on the line 25

vladsavelyev (changes requested):
Amazing, thanks a lot for the contribution! Left some comments :)

multiqc/modules/nonpareil/nonpareil.py
```diff
@@ -0,0 +1,276 @@
...
+        # x.obs Rarefied sequencing effort.
+        # x.adj Adjusted rarefied sequencing effort.
+        # y.red Rarefied redundancy (observed).
+        # y.cov Rarefied coverage (corrected).
+        # y.sd Standard deviation of rarefied coverage.
+        # y.p25 Percentile 25 (1st quartile) of rarefied coverage.
+        # y.p50 Percentile 50 (median) of rarefied coverage.
+        # y.p75 Percentile 75 (3rd quartile) of rarefied coverage
+
+        import numpy as np
```
vladsavelyev, line 86: Should it be imported globally?

multiqc/modules/nonpareil/nonpareil.py
```diff
@@ -0,0 +1,276 @@
...
+        self.disp_type = getattr(config, "nonpareil", {}).get("plot_dispersion", False)
+
+        # Read JSON file
+        self.data_by_sample = dict()
+        for f in self.find_log_files("nonpareil", filehandles=True):
+            json_parsed = self.parse_nonpareil_json(f)
+            common_samples = set(json_parsed).intersection(self.data_by_sample)
+            if common_samples:
+                log.debug(f"Duplicate sample names found ({common_samples})! Overwriting...")
+            self.data_by_sample = self.data_by_sample | json_parsed
```
vladsavelyev, line 42: This line breaks for Python 3.8 (see the CI status) as the pipe operator is not supported for dicts in 3.8

```suggestion
            self.data_by_sample.update(json_parsed)
```

multiqc/modules/nonpareil/nonpareil.py
```diff
@@ -0,0 +1,276 @@
...
+        # Register samples
+        for s_name in self.data_by_sample.keys():
+            self.add_data_source(f, s_name)
```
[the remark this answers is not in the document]
fgvieira: My idea was to read in all the files and, afterwards, register all the sample that were read but I can see that I need the original file. What if there are duplicated samples? Do they also get overwritten by `add_data_source`?

multiqc/modules/nonpareil/nonpareil.py
fgvieira: But then it shows up as `0` and `1`. Do you think that is clear?

multiqc/modules/nonpareil/nonpareil.py
fgvieira: I am not even sure it makes sense to have this here, since this is actually a parameter that the user specifies when running `nonpareil`. But it reflects the objective coverage in percentage; i.e., coverage value considered near-complete.

multiqc/modules/nonpareil/nonpareil.py
fgvieira: But don't these names need to match the `self.data_by_sample` keys?

vladsavelyev: They do, but you can put them into `self.data_by_sample` with any keys you want :)

multiqc/modules/nonpareil/nonpareil.py
vladsavelyev: Hmm, I didn't check but if that's so, we should modify how the `table_cond_formatting_rules` work so it shows words and not numbers 🤔

fgvieira: So I'll keep it as `true`/false`, ok?

multiqc/modules/nonpareil/nonpareil.py
fgvieira: But those names are the ones that `nonpareil` uses and I read all at once from the `json`. Since these names are not visible to the user, don't you think it is better to keep them the same as `nonpareil`?
This way it is easier to know where the number are coming from...

fgvieira:
Seems to be failing on the `deeptools` module.

vladsavelyev:
@fgvieira that's right, it's because I pushed a test case into github.com/ewels/MultiQC_TestData but we haven't merged the corresponding fix yet into MultiQC (https://github.com/ewels/MultiQC/pull/2229). So all PRs are failing CI now, hopefully not for too long :)

multiqc/modules/nonpareil/nonpareil.py
vladsavelyev: Ah, sorry, I didn't mean `True`, but `"true"`, as a string literal. The values listed `table_cond_formatting_rules` in are those that MultiQC will regonize and format as nice green/red labels. As you can see, `"true"` adn `"false"` are listed there.

vladsavelyev: But I'd modify the core MultiQC code so `True` and `False` are recognised as well. And`"yes"` and `"no"` also, for good measure

docs/modules/nonpareil.md
```diff
@@ -0,0 +1,25 @@
...
+  Estimate metagenomic coverage and sequence diversity.
+---
+
+Nonpareil uses the redundancy of the reads in a metagenomic dataset to estimate
+the average coverage and predict the ammount of sequences that will be required
+to achieve "nearly complete coverage", defined as ≥95% or ≥99% average coverage.
+
+Since Nonpareil main output has no model information, it is necessary to run its
+auxiliary `R` plot functions and save the `curves` object as a `JSON` file. For
+more info on how to generate this file, check `nonpareil` [snakemake plot wrapper](https://snakemake-wrappers.readthedocs.io/en/latest/wrappers/nonpareil/plot.html#code). Briefly, call function `export_curve()` on object `curves`:
```
vladsavelyev, line 14: What I meant, is that it would be helpful to give brief instructions how to run the _snakemake wrapper_ (or `RScript` in command line if it's easier to just run those two lines)

multiqc/modules/nonpareil/nonpareil.py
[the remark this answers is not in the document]
vladsavelyev: It can handle duplicated samples, so that's not a problem :) Just need to make sure to iterate over `json_parsed` and not `self.data_by_sample` in the loop for a little efficiency - pushed that change now

multiqc/modules/nonpareil/nonpareil.py
vladsavelyev: They are used in `multiqc_data.json` which can be used downstream, and they are visible in the detailed view: 
Screenshot 2023-12-14 at 16 46 54
But I see your point - that's actually fair. I haven't though that it's really helpful to track numbers down to the source json. Let's keep them then!

vladsavelyev (approved):
Looks good now! Thanks for addressing the comments so quickly.

I'll leave it to @ewels to have a look :)

docs/modules/nonpareil.md
fgvieira: There is no need to use `snakemake`, as the users can still use `nonpareil` and make the plots by themselves. The only extra thing they have to do, is to save the `R` object into a `json` file. Since I already have some functions for that implemented in a `snakemake` wrapper, I thought it would just be easier (instead of each user having to figure out how to parse the `R` object).

ewels: If I understand you correctly, the functions in question are pretty short - is that right?

eg. it's this bit?
```R
  base::library("jsonlite")
  base::message("Exporting model as JSON")

  export_curve <- function(object){
    # Extract variables
    n <- names(attributes(object))[c(1:12,21:29)]
    x <- sapply(n, function(v) attr(object,v))
    names(x) <- n
    # Extract vectors
    n <- names(attributes(object))[13:20]
    y <- lapply(n, function(v) attr(object,v))
    names(y) <- n
    # Add model
    # https://github.com/lmrodriguezr/nonpareil/blob/162f1697ab1a21128e1857dd87fa93011e30c1ba/utils/Nonpareil/R/Nonpareil.R#L330-L332
    x_min <- 1e3
    x_max <- signif(tail(attr(curves$np.curves[[1]],"x.adj"), n=1)*10, 1)
    x.model <- exp(seq(log(x_min), log(x_max), length.out=1e3))
    y.model <- predict(object, lr=x.model)

    c(x, y, list(x.model=x.model), list(y.model=y.model))
  }

  export_set <- function(object){
    y <- lapply(object$np.curves, "export_curve")
    names(y) <- sapply(object$np.curves, function(n) n$label)
    jsonlite::prettify(toJSON(y, auto_unbox=TRUE))
  }

  y <- export_set(curves)
  write(y, snakemake@output[["json"]])
  base::message("JSON file saved")
```

I try to keep dependencies to 3rd party websites and the like to a minimum where possible. My preference would be to either include the above code in the docs here, or as a script in the `nonpareil` repo itself (seeing as the MultiQC module already depends on that). Either option would feel a little more resistant to URLs changing over time.

multiqc/modules/nonpareil/nonpareil.py
```diff
@@ -0,0 +1,295 @@
...
+                "title": "Ideal coverage",
+                "description": "Objective coverage in percentage; i.e., coverage value considered near-complete.",
+                "max": 100,
+                "min": 0,
+                "suffix": "%",
+                "scale": False,
+                "hidden": True,
+            },
+            "LRstar": {
+                "title": f"{config.base_count_prefix} Sequencing effort for ideal coverage",
```
ewels, line 215: This title is a bit too long, it skews all of the column width..
```suggestion
                "title": f"{config.base_count_prefix} ideal seq",
```

multiqc/modules/nonpareil/nonpareil.py
ewels: @vladsavelyev are they scoped in the JSON? I feel like this is a name collision waiting to happen if not. Shouldn't be too difficult to add a prefix after parsing the JSON, that would feel safer..

multiqc/modules/nonpareil/nonpareil.py
```diff
@@ -0,0 +1,295 @@
...
+            elif self.disp_type == "iq":
+                data["disp_upper"] = [[x, y] for x, y in zip(data["x.adj"], data["y.p75"])]
+                data["disp_lower"] = [[x, y] for x, y in zip(data["x.adj"], data["y.p25"])]
+            if disp_add is not None:
+                data["disp_upper"] = [[x, y] for x, y in zip(data["x.adj"], np.array(data["y.cov"]) + disp_add)]
+                data["disp_lower"] = [[x, y] for x, y in zip(data["x.adj"], np.array(data["y.cov"]) - disp_add)]
+
+        return json_raw
+
+    def nonpareil_general_stats_table(self):
```
ewels, line 118: Even though most columns are hidden, this feels like too many for the General Statistics table.

My suggestion would be to have a Nonpareil-specific table with all of these, most can then be shown by default. Then just pick one or two that would be particularly useful to have alongside outputs from other bioinformatics tools to keep in the General Stats table.

multiqc/modules/nonpareil/nonpareil.py
```diff
@@ -0,0 +1,295 @@
...
+                "suffix": " bps",
+                "scale": "GnBu",
+                "format": "{:,.3f}",
+                "hidden": True,
+            },
+            "AL": {
+                "title": "Adjusted read length",
+                "description": "Adjusted read length (if using kmers)",
+                "min": 0,
+                "scale": "GnBu",
```
ewels, line 135: Please try to avoid having adjacent columns with the same colour scheme. It makes it much easier to scan the table if you can vertically track column by colour.

multiqc/modules/nonpareil/nonpareil.py
```diff
@@ -0,0 +1,295 @@
...
+
+        esconfig = {
+            "dashStyle": "Dash",
+            "lineWidth": 2,
+            "color": "#000000",
+            "marker": {"enabled": False},
+            "enableMouseTracking": False,
+            "showInLegend": False,
+        }
+        # desc = " **The dashed black line shows theoretical GC content:** `{}`".format(theoretical_gc_name)
```
ewels, line 254: Please remove, and add a description of what the line represents to the section description (no need to have it as a variable as done in the FastQC module, as this one is always shown - the FastQC one is optional).

multiqc/modules/nonpareil/nonpareil.py
```diff
@@ -0,0 +1,295 @@
...
+
+        data_plot = list()
+        data_labels = list()
+        extra_series = list()
+        for s_name, data in sorted(self.data_by_sample.items()):
+            data_plot.append(dict())
```
ewels, line 260: Having only one sample per plot is generally something that I try to avoid in MultiQC:

* Doesn't scale well to many samples - tabs are quickly impossible to read / use
* Can't get an overview of all samples in one glance, which is the main purpose of MultiQC

Is there a reason that all plots cannot be shown on a single plot here?

ewels (changes requested):
Generally great, just a few minor comments that I'd like to see addressed / resolved.

docs/modules/nonpareil.md
fgvieira: I am not the owner of the `nonpareil` repo and not sure how active it still is, but I can definitely include the code in the docs.

multiqc/modules/nonpareil/nonpareil.py
fgvieira: I thought about it but, since each sample can have up to 4 curves (depending on user input: observed, model, and dispersion upper and lower), I thought it would be too confusing to have all in the same plot.

I also thought about something similar to FastQC base composition (expandable heatmap) but that is not a standard plot.

multiqc/modules/nonpareil/nonpareil.py
fgvieira: Prefix added.

fgvieira:
Yeah, `nonpareil` is not the easiest program to add to MultiQC.
If you are also planning to use this module, it would be nice if you could try it out and provide some feedback.

multiqc/modules/nonpareil/nonpareil.py
ewels: Can it be four separate plots?

fgvieira: You mean 4 plots, each with all the samples? Don't think so, since these plots make most sense together: the observed curve (your data), next to the expected curve under the model and the uncertainty of the model.

I think the closest we can get to having all samples in the same plot, is to do something similar to FastQC base composition:
- heatmap with the (e.g.) observed curve (or, if possible, user selected) for all samples 
- clicking on a row/sample, expands to show the 4 curves for that sample (observed, model and upper/lower dispersion)

However, I looked into it, but I am not sure how to implement this plot since there is a lot of custom code.

@jfy133 since it seems you are also planning to use it, do you have any ideas?

ewels: Why does it have to be a heatmap? If it's showing a single variable (observed curve) could it not just be a line graph? Click reactivity doesn't depend on it being a heatmap.

Whilst clicking to show more is possible, it should be avoided if we can. It bloats the report size so prevents scaling to large sample numbers and the code is complex and difficult to maintain.

It's usually around now that I say that MultiQC is to summarise data and doesn't have to replicate everything available 😅

jfy133: I've not had a chance to test the current implementation but:

I agree one plot per sample will be overkilll... I'm not sure I completely follow about the heatmap on how/what that is representing, but I agree - you can't really have four plots separate, they are not so informative

For reference, this is the multi-sample PNG that nonpareil produces (that I currently just add to MultiQC as a custom image)

image

and a for a single sample (not from MultiQC ,just default prodcued by the R script): 

image

Given MultiQC is meant to be a summary thing only, perhaps just two plots: one of the observed, and a second of  models is  sufficient to help identify outliers? I think getting into dispersion etc. is getting too technical for aMultiQC plot for example

ewels: Sounds good. Two plots one after another, or one plot with two datasets (two tabbed views)? I'm thinking the latter, as they're essentially showing the same thing?

jfy133: Tabbed would also work IMO, also allows quick switching so should be able to 'see' the observed/model in the same place.

This is just personal opinion though!

fgvieira: @jfy133 just to be sure, you are suggesting to have all samples in one plot, but showing only the observed and the model, right?

I can all samples in a single plot and, since the user can choose which plots to show (observed and model are enabled and dispersion is disabled by default), we can leave it up to the user what to show.

jfy133: Yes, I think it would be fine to have all samples in one, and only observed/model as you say (either observed/model is in the same, or across two plots you can switch between I don't mind). 

The dispersion is when you really want to look closely and IMO at that point you would like at the raw data/nonpareil generated plots themselves  etc.

ewels: Nice 👍🏻  To clarify, I'm talking two datasets, [as described in the docs](https://multiqc.info/docs/development/plots/#switching-datasets-1). So it'll default to showing observed, and the user can flip the toggle switch to see the model curves instead. I think we're on the same page but just trying to be clear. (please don't show multiple lines per sample in a plot at the same time)

fgvieira: I think the model curve is more informative when together with the observed, to show how far your data is from the expected model (the same applies to the dispersion). This is how `nonpareil` shows them, even when you have multiple samples.

I agree that several lines for each sample can be confusing and the user can go to the raw data for more info, but I do not see it as a big problem as long as the user can decide what to show (I even think it is nice to give the user control over what to show). What if the report only has one sample and the user wants to see all lines?
Furthermore, the user can also use the side bar to only show the sample of interest for a closer look of the curves. 

So, I still think it is relevant to have a plot with several curves per sample, but it could be an extra dataset. For example, we could have one dataset for the observed, one for the model, and a last one (optional and disabled by default) with all curves.

fgvieira: Just pushed some changes. Briefly:
- Removed plotting of dispersion
- Add three tabs: `Observed`, `Model`, `Combined` (both observed and model, with the latter as a dashed line of the same color)

ewels (approved):
Great! I like the approach you've done with the plotting of model + observed, I think that works well.

I've pushed a few tweaks, broadly:

- Made the combined plot the default
- Removed some duplicate table unit suffixes
- Removed some format strings which were the same as default
- Showed a few more columns by default on the nonpareil table
- Added a bunch of help text, copied from the nonpareil website

I hope this is ok, please sanity check my changes. More / better description + help text would be welcome. The aim is to help someone to interpret what they're seeing even if they have no prior experience of the tool.

For me the x-axis for the "Redundancy levels" plot goes way off into the distance, with the curves only taking up a tiny part of the plot. I tried to set an `xmax` manually but got some slightly inconsistent / unexpected results. @vladsavelyev it'd be great if you could take a quick look at this to check what's going on and see if you can figure out why the scale is being weird.

Otherwise, good to merge from my side. Thanks all!

fgvieira:
It looks great. Thanks all for help/input!

ewels:
> For me the x-axis for the "Redundancy levels" plot goes way off into the distance, with the curves only taking up a tiny part of the plot. I tried to set an `xmax` manually but got some slightly inconsistent / unexpected results. @vladsavelyev it'd be great if you could take a quick look at this to check what's going on and see if you can figure out why the scale is being weird.

Looks like this was fixed in an upstream PR, so all good to go now!

PR #2221, opened 2023-12-12
Add module for metadmg

fgvieira:

Add module to plot metaDMG results.


setup.py
```diff
@@ -51,6 +51,7 @@
     install_requires=[
         "matplotlib>=2.1.1",
         "networkx>=2.5.1",
+        "pandas",
```
vladsavelyev, line 54: We want to avoid including extra heavy dependencies as much as possible, and so far got away without `pandas` 😅  Wonder if it would be easy to use `csv.DictReader` here and operate on dicts?

multiqc/modules/metadmg/metadmg.py
```diff
@@ -0,0 +1,244 @@
...
+                index_col="taxid",
+                comment="#",
+            )
+            # Add data
+            self.add_data_source(f, f["s_name"])
+
+        for f in self.find_log_files("metadmg/dfit", filecontents=False):
+            f["s_name"] = f["s_name"][:-5]
+            # Read DF
+            df_dfit = pd.read_table(
```
vladsavelyev, line 85: Is it going to parse `*.dfit.gz` as well?

fgvieira: Yes, that was the idea. I thought of pandas since it makes it much easier to read csv/tsv files (even if they are gzipped).

Issue #1705, opened 2022-06-14
Megahit


ryanames-eagle:
### Name of the tool

Megahit

### Tool homepage

https://github.com/voutcn/megahit

### Tool description

MEGAHIT is an ultra-fast and memory-efficient NGS assembler. It is optimized for metagenomes, but also works well on generic single genome assembly (small or mammalian size) and single-cell assembly.

### Tool output

[log.txt](https://github.com/ewels/MultiQC/files/8900915/log.txt)
NB - file is just log in default output

### Log filename pattern

log

### Data suitable for MultiQC plot(s)

Memory used
Time elapsed (last line)
Contig stats (second to last line)

### Most interesting data for the General Stats table

Maybe time elapsed and memory used especially if this tool is part of a pipeline

PR #2222, opened 2023-12-12
New module: MEGAHIT

vladsavelyev:
Parses MEGAHIT's stdout/stderr log and adds one table and general stats. Resolves https://github.com/ewels/MultiQC/issues/1705


multiqc/modules/megahit/megahit.py
```diff
@@ -0,0 +1,146 @@
...
+        # Write parsed report data to a file
+        self.write_data_file(data, f"multiqc_{self.anchor}")
+
+        headers = {
+            "megahit_mem": {
+                "title": "Memory",
+                "description": "Memory used",
+                "min": 0,
+                "scale": "Reds",
+                "format": lambda x: humanize.naturalsize(x),
```
ewels, line 73: I have a feeling that this could give different units for different rows, right? I think it's better to pick a single unit for all samples, so that the numbers can be compared easily.

eg. it's very tough to spot which of the following is bigger at a glance:

13.3 GB
479.2 GB
4.3 TB
7.4 GB
933.2 MB

etc..

multiqc/modules/megahit/megahit.py
```diff
@@ -0,0 +1,146 @@
...
+                "description": "N50 contig length",
+                "min": 0,
+                "scale": "Greens",
+                "suffix": "&nbsp;bp",
+                "format": "{:,d}",
+            },
+        }
+
+        # Basic Stats Table
+        self.general_stats_addcols(data, headers)
```
ewels, line 136: I'm not sure about just sticking everything in general stats, even if much of it is hidden. I'd prefer to just put in the columns that we think people will want to compare.

I'd also personally think that memory and time are two of the least interesting columns, despite what the OP said. Is that just me? 🙄  I would have probably gone for bases / contigs / N50?

multiqc/utils/search_patterns.yaml
```diff
@@ -485,6 +485,9 @@ mapdamage:
   - fn: "3p*_freq.txt"
   - fn: "5p*_freq.txt"
   - fn: "lgdistribution.txt"
+megahit:
+  contents: " - MEGAHIT v"
+  num_lines: 1
```
ewels, line 490: Just in case there's some random stdout from slurm or something..

```suggestion
  num_lines: 5
```

ewels (changes requested):
Couple of minor comments

multiqc/modules/megahit/megahit.py
vladsavelyev: Fair enough. Perhaps `humanize` is not a good fit for MultiQC after all :) Can switch to a more classical approach of reporting a decimal group separated number of KB

multiqc/modules/megahit/megahit.py
vladsavelyev: Yeah, I agree with you :)

ewels (approved):
Looks great, thanks!

PR #2223, opened 2023-12-12
Document `base_read_counts` and decimal formatting [skip changelog]

vladsavelyev:


ewels (approved):
LGTM ✅ 

Note that this section of the docs is for "v1.0 Updates" though, which was in 2017 😅  Should probably think about removing / refactoring this part of the docs at some point. Probably as part of a larger docs overhaul.

Issue #2225, opened 2023-12-13
MultiQC 1.18 broke Picard (only DuplicationMetrics or full) module


Stikus:
### Description of bug

`MultiQC 1.18` report doesn't have `Picard` (only `DuplicationMetrics` or full) section.

Here is a log from `MultiQC 1.17`:
```console
$ LC_ALL=C.UTF-8 LANG=C.UTF-8 /soft/MultiQC-1.17/local/bin/multiqc --fullnames --outdir /outputs --filename test_full_filtered_multiqc_report.html /outputs/inputMultiQC

  /// MultiQC 🔍 | v1.17

|           multiqc | Not cleaning sample names
|           multiqc | Search path : /outputs/inputMultiQC
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 5/5  
|            picard | Found 1 MarkDuplicates reports
|          samtools | Found 2 flagstat reports
|            fastqc | Found 2 reports
|           multiqc | Report      : test_full_filtered_multiqc_report.html
|           multiqc | Data        : test_full_filtered_multiqc_report_data
|           multiqc | MultiQC complete
```

You can see the line:
```
|            picard | Found 1 MarkDuplicates reports
```

This line (and results) are missing in `1.18` log and report.

- `report_data_sources` (`1.17`)

```json
    "report_data_sources": {
        "Picard": {
            "DuplicationMetrics": {
                "SRR6000626.aligned-bwa.sorted-samtools.rh.bam": "/outputs/inputMultiQC/SRR6000626.aligned-bwa.sorted-samtools.rh.dedupped.bam.metrics.txt"
            }
        },
        "Samtools": {
            "flagstat": {
                "SRR6000626.aligned-bwa.sorted-samtools.flagstat": "/outputs/inputMultiQC/SRR6000626.aligned-bwa.sorted-samtools.flagstat",
                "SRR6000626.aligned-bwa.sorted-samtools.rh.dedupped.flagstat": "/outputs/inputMultiQC/SRR6000626.aligned-bwa.sorted-samtools.rh.dedupped.flagstat"
            }
        },
        "FastQC": {
            "all_sections": {
                "SRR6000626.aligned-bwa.sorted-samtools.rh.dedupped.bam": "/outputs/inputMultiQC/SRR6000626.aligned-bwa.sorted-samtools.rh.dedupped_fastqc.zip",
[...]

PR #2226, opened 2023-12-13
Picard: more search patterns for MarkDuplicates

vladsavelyev:
Cover `MarkDuplicatesSpark` and `GATKDuplicationMetrics`: https://github.com/ewels/MultiQC/issues/2225

Might just make it check for `MarkDuplicates` or `DuplicationMetrics`, but it might run into false positives 🤔


ewels (approved):
LGTM 👍🏻 

I'd prefer more specific patterns I think 🤔

Issue #2152, opened 2023-10-26
Add global config for `shared` search


ewels:
We have `shared: true` in the search patterns for two use cases:

* A generic filename, where there's a risk that the file could be incorrectly found and "consumed", being lost from other modules
* Where the tool generates stdout, with the assumption that people may be concatenating log files.

I'm not sure that the second one happens much now, and it slows down the search quite a bit.

Idea is to remove these config options where the pattern looks specific, but add in a global option that effectively sets `shared: true` for _all_ search patterns, in case there are users who are affected.

vladsavelyev:
@ewels - you added a lot of `shared: true` here: https://github.com/ewels/MultiQC/commit/902b18fac0f852e104cf8abdbfcf8648b2396682
Most of which look unnecessary in terms of the first point. Do you think it makes sense to move those into a global flag?

ewels:
Yup, exactly 👌🏻 

That was 7 years ago, long enough for me to have my own weird log sharing stuff happening. I think it's ok to remove these now.

PR #2227, opened 2023-12-13
Add global `shared` flag, remove unnecessary per-module `shared`

vladsavelyev:
Addressing https://github.com/ewels/MultiQC/issues/2152


multiqc/utils/search_patterns.yaml
```diff
@@ -230,7 +228,6 @@ clipandmerge:
   num_lines: 5
 clusterflow/logs:
   fn: "*_clusterFlow.txt"
-  shared: true
```
ewels, line 233: This is maybe the only one that should be left there. ClusterFlow is my old workflow manager tool and it concatenates log files. So if there are any ClusterFlow logs then they definitely will be shared.

multiqc/utils/search_patterns.yaml
```diff
@@ -243,7 +240,6 @@ conpair/contamination:
   num_lines: 3
 cutadapt:
   contents: "This is cutadapt"
-  num_lines: 2
```
ewels: Not sure why this is removed?

ewels (approved):
Great 👍🏻 

Needs docs for the new config option, but otherwise good to go. I'll add some now.

Issue #1954, opened 2023-07-13
DRAGEN Trimmer metrics: Missing table header config


NicoleGruenheit:
### Description of bug

Hi,

I'm trying to hide columns in the Mapping metrics table and General stats table of the DRAGEN module. I followed the description about how to get to the table IDs, but the ID of the Mapping metrics table changes every time I run MultiQC. So far I've had table_qvja, table_mbty, table_aeob. This means I can't set specific columns to False in my config file.

For the general_stats_table, the ID stays constant, but adding this to the config file doesn't change anything:

```yaml
table_columns_visible:
  general_stats_table:
    Total input reads: False
    Unmapped reads pct: False
    Number of duplicate marked reads pct: False
    Total: False
```

Am I missing something obvious here? 

I'm using the v1.15.dev0 docker image for this.

Thanks,
Nicole

### File that triggers the error

DRAGEN metric files (.csv)

### MultiQC Error log

```console
docker run --rm -t -u $UID_USER:$GID_BFX -v `pwd`:`pwd` -w `pwd` ewels/multiqc:dev multiqc . -n "SXXXX" -i "SXXXX"  -f --config multiqc_config.yaml --ignore wgs_metrics/ --data-format tsv  -v --ignore MultiQC-master/

  /// MultiQC 🔍 | v1.15.dev0

[2023-07-13 12:44:31] multiqc                                            [DEBUG  ]  This is MultiQC v1.15.dev0
[2023-07-13 12:44:31] multiqc                                            [DEBUG  ]  Loading config settings from: multiqc_config.yaml
[2023-07-13 12:44:31] multiqc                                            [DEBUG  ]  New config: {'custom_logo': 'logo.jpg', 'show_analysis_paths': False, 'show_analysis_time': False, 'dragen/wgs_contig_mean_cov': {'fn_re': '*.target_bed_contig_mean_cov.csv'}, 'dragen/overall_mean_cov_metrics': {'fn_re': '*.target_bed_overall_mean_cov.csv'}, 'run_modules': ['dragen', 'dragen_fastqc'], 'remove_sections': ['dragen-coverage-per-nonmain-contig', 'dragen-time-metrics', 'dragen-wgs-coverage-metrics'], 'table_columns_visible': {'general_stats_table': {'Total input reads': False, 'Unmapped reads pct': False, 'Number of duplicate marked reads pct': False, 'Total': False}, 'table_nfmc': {'Total reads in RG': False}}}
[...]

PR #2228, opened 2023-12-13
Add missing table `id` in DRAGEN modules, require them in linting

vladsavelyev:
Add missing `pconfig` with `namespace` and `id` in DRAGEN submodules. Fixes settings table column visibility configuration, e.g.

```yaml
table_columns_visible:
  dragen-trimmer-metrics-table:
    Total input reads: False
    Unmapped reads pct: False
    Number of duplicate marked reads pct: False
    Total: False
    Total input bases R2: False
```

Fixes https://github.com/ewels/MultiQC/issues/1954

Require `id` to be set in pconfig when `--strict`. When it's not set, auto-generate random `id` earlier to make the config options work.

As some additional refactoring, removed warnings of missing `r` prefixes in regex strings.

TODO: 
- [x] now fix the rest of the linting :)

### DIFF
--- a/multiqc/modules/dragen/coverage_metrics.py
+++ b/multiqc/modules/dragen/coverage_metrics.py
@@ -902,7 +902,7 @@ def create_table_handlers():
                     + "\n\nPress the `Help` button for details."
                 )
                 anchor = "dragen-cov-metrics-own-section-" + re.sub(r"(\s|-|\.|_)+", "-", phenotype)
-                plots[phenotype]["config"]["id"] = anchor + "-table"
+                plots[phenotype]["config"]["id"] = f"{anchor}-table"
                 sections.append(
                     {
                         "name": make_section_name(phenotype),
--- a/multiqc/modules/dragen/coverage_metrics.py
+++ b/multiqc/modules/dragen/coverage_metrics.py
@@ -603,7 +603,7 @@ def make_data_for_txt_reports(coverage_data, metrics):
     for sample in coverage_data:
         for phenotype in coverage_data[sample]:
             # Replace any sequence of spaces/hyphens/dots/underscores by single underscore.
-            new_phenotype = re.sub("(\s|-|\.|_)+", "_", phenotype)
+            new_phenotype = re.sub(r"(\s|-|\.|_)+", "_", phenotype)
             # Transform phenotype as in the previous code version.
             if new_phenotype == "wgs":
                 new_phenotype += "_cov_metrics"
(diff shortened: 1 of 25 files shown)


multiqc/modules/dragen/coverage_metrics.py
```diff
@@ -901,10 +901,12 @@ def make_own_coverage_sections(coverage_data, coverage_headers, bed_texts):
                     + bed_texts[phenotype]["description"]
                     + "\n\nPress the `Help` button for details."
                 )
+                anchor = "dragen-cov-metrics-own-section-" + re.sub(r"(\s|-|\.|_)+", "-", phenotype)
+                plots[phenotype]["config"]["id"] = anchor + "-table"
```
ewels, line 905:
```suggestion
                plots[phenotype]["config"]["id"] = f"{anchor}-table"
```

multiqc/modules/hicpro/hicpro.py
```diff
@@ -289,15 +306,24 @@ def hicpro_mapping_chart(self):
...
+                d = dict()
+                if n_global is not None:
+                    d["Full_Alignments_Read"] = n_global
+                if n_local is not None:
+                    d["Trimmed_Alignments_Read"] = n_local
+                if n_mapped is not None and n_total is not None:
+                    try:
+                        d["Failed_To_Align_Read"] = int(n_total) - int(n_mapped)
+                    except ValueError:
+                        log.debug(
```
ewels, line 322: This shouldn't happen very often, right?
```suggestion
                        log.warning(
```

multiqc/modules/hicpro/hicpro.py
```diff
@@ -169,7 +186,7 @@ def parse_hicpro_stats(self, f, rsection):
                         self.hicpro_data[s_name][s[0]] = float(s[1])
                     # Otherwise just store the value as is.
                     except ValueError:
-                        self.hicpro_data[s_name][s[0]] = s[1]
+                        log.error(f"Could not parse value for key '{s[0]}' as a number in sample '{s_name}': '{s[1]}'")
```
ewels, line 189: Do we want to keep the value even if it's not a number?

multiqc/modules/qc3C/qc3C.py
```diff
@@ -996,7 +996,7 @@ def _none_to(x, y):
             self.add_data_source(f, s_name, section=analysis_mode)
 
         except KeyError as ex:
-            log.error(
+            log.debug(
```
ewels, line 999: Yes, this will help the CI log outputs look nicer! 😅

ewels (approved):
Looks good! Couple of minor questions but nothing big.

multiqc/modules/hicpro/hicpro.py
vladsavelyev: Yeah, agree

multiqc/modules/hicpro/hicpro.py
vladsavelyev: I was thinking that they are attempted to be parsed as numbers for some calculations further in the code, so there is no reason to keep strings. But perhaps there are other metrics that are not used for calculations and we want to keep, so I'll revert and test this bit

vladsavelyev: Reverted that, and did more debugging in hicpro, there were actually a few issues with it:
- the search patterns didn't account for half of input test data (e.g. `.mmapstat` was incouded by `.mapstat` was missing; `_assplit.stat` was missing, etc)
- there were warning about missing data on plots, just needed to add a check that the required data was found like in other modules
- improved debug info about duplicated key for a sample

The PR is heavy in single module, perhaps should split it up for clarity 🤔

vladsavelyev: Moving this stuff into a separate PR

vladsavelyev:
@multiqc-bot changelog

Issue #1939, opened 2023-06-19
Deeptools float conversion crash fails entire workflow


FelixKrueger:
### Description of bug

I've got a run where a file MultiQC complains that it can't convert the string `--` to a float - fair enough. However, this crashes the nf-core/chipseq pipeline; in standalone mode the error also occurs, but the MultiQC report still gets generated (withtout deeptools plots). Is there something that could be done to change that? 

I appreciate that the input sample might have some rather horrible qualities (only 2600 sequences in total) as it was a negative control - but still it would be nice if it wouldn't crash the entire workflow.

### File that triggers the error

[buggedFile.mLb.clN.plotProfile.txt](https://github.com/ewels/MultiQC/files/11785752/buggedFile.mLb.clN.plotProfile.txt)

### MultiQC Error log

```console
$ multiqc .

/// MultiQC 🔍 | v1.14

|           multiqc | Search path : /share/Data/
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 1769/1769
|    custom_content | nsc_coefficient: Found 40 samples (bargraph)
|    custom_content | rsc_coefficient: Found 40 samples (bargraph)
|    custom_content | strand_shift_correlation: Found 40 samples (linegraph)
|phantompeakqualtools | Found 40 logs
|            preseq | Found 38 reports
╭────────────────────────────────────────────────────────── Oops! The 'deeptools' MultiQC module broke... ──────────────────────────────────────────────────────────╮
│ Please copy this log and report it at https://github.com/ewels/MultiQC/issues                                                                                     │
│ Please attach a file that triggers the error. The last file found was: ./bowtie2/mergedLibrary/deepTools/plotProfile/buggedFile.mLb.clN.plotProfile.tab │
│                                                                                                                                                                   │
│ Traceback (most recent call last):                                                                                                                                │
│   File "/share/fkrueger/mambaforge-pypy3/envs/INDEX/lib/python3.10/site-packages/multiqc/multiqc.py", line 654, in run                                            │
│     output = mod()                                                                                                                                                │
│   File "/share/fkrueger/mambaforge-pypy3/envs/INDEX/lib/python3.10/site-packages/multiqc/modules/deeptools/deeptools.py", line 96, in __init__                    │
│     n["plotProfile"] = self.parse_plotProfile()                                                                                                                   │
[...]

PR #2229, opened 2023-12-14
DeepTools: Handle missing data in plotProfile

vladsavelyev:
Correctly handle input where missing value is represented with `--`.

Fixes https://github.com/ewels/MultiQC/issues/1939


ewels (approved):
LGTM!

PR #1553, opened 2021-09-17
Module ngsbits

jakobmatthes:
- [x] This comment contains a description of changes (with reason)
- [x] `CHANGELOG.md` has been updated
- [x] There is example tool output for tools in the <https://github.com/ewels/MultiQC_TestData> repository or attached to this PR
- [x] Code is tested and works locally (including with `--lint` flag)
- [x] `docs/README.md` is updated with link to below
- [x] `docs/modulename.md` is created
- [x] Everything that can be represented with a plot instead of a table is a plot
  - here tables are better to check correlation
- [x] Report sections have a description and help text (with `self.add_section`)
- [x] There aren't any huge tables with > 6 columns (explain reasoning if so)
  - columns are closely related and easy to read, despite >6
- [x] Each table column has a different colour scale to its neighbour, which relates to the data (eg. if high numbers are bad, they're red)
- [x] Module does not do any significant computational work

PR #2231, opened 2023-12-14
New module: ngs-bits

vladsavelyev:
Forking from @jakobmatthes's [PR](https://github.com/ewels/MultiQC/pull/1553) so I can merge with master and push some changes :)

Thank you so much @jakobmatthes for the contribution!

This is how the reports look like now:

Screenshot 2023-12-14 at 15 32 34

A few comments from me:

- [x] In MultiQC we try to avoid such wide tables - especially as the MappingQC one - as they lose in readability. I wonder if a lot columns can be marked as `hidden` by default?
- [x] One or two metric can be added into general stats
- [ ] Can any data can be represented as a plot instead?
- [x] Read and base counts can be extended with the configurable prefixes/multipliers like described in this section: https://multiqc.info/docs/development/modules/#step-3---adding-to-the-general-statistics-table 
Example: 
```py
'title': f'{config.read_count_prefix} Reads',
'description': f'Number of reads ({config.read_count_desc})',
```
- [x] Column names should be capitalized (`insert size` -> `Insert size`).
- [x] Nit picking: the `%` singes in the col names are redundant as they are used as the suffixes in the cells already.


vladsavelyev:
@multiqc-bot changelog

vladsavelyev:
@multiqc-bot fix linting

PR #2233, opened 2023-12-15
`HiC-Pro`: missing search patterns and better handling of missing data

vladsavelyev:
Some `HiC-Pro` module refactoring:
* More granular handling of missing metrics or wrong formatted numbers,
* Add missing search patterns. The search patterns didn't account for half of input test data (e.g. `.mmapstat` was included by `.mapstat` was missing; `*_assplit.stat` was missing, etc)
* There were warning about missing data on plots, just needed to add a check that the required data was found like in other modules,
* Improved debug info about duplicated key for a sample,
* Remove unnecessary warnings.


ewels (changes requested):
Checked the report generated before and after this PR and seem to be losing some plot data:

* The `Read Mapping` section shows in the report, but has no plot
* The `Contact statistics` plots loses `Duplicate read pairs`, whose count goes to `0`

vladsavelyev:
@ewels thanks for checking! My bad, it's fixed now

ewels (approved):
Thanks!

PR #2235, opened 2023-12-15
Update num_lines for fastp search

kclem:

Reverts change in 9d00f14a317a3eb40c8edb5b4e92fde4bf11689e - 3 lines is not enough. Data in https://github.com/ewels/MultiQC_TestData/tree/master/data/modules/fastp/issue_2138 cannot be parsed with only 3 lines.


vladsavelyev:
Thank you! My bad for breaking it.

Can even make it 10 to be safer - though I'm thinking it was probably a good idea to require the files to be named as `*fastp.json` 🤔

vladsavelyev (approved):

PR #2236, opened 2023-12-17
Bump versions for v1.19 release [skip changelog]

vladsavelyev:
Skip changelog check wasn't at the right place, plus we want to skip it for bump version by default - https://github.com/ewels/MultiQC/pull/2237


ewels (approved):
LGTM

PR #2248, opened 2023-12-19
Seqera CLI: Updates for v0.9.2

ewels:
Addressing points in https://github.com/MultiQC/MultiQC/issues/2245

Everything done except labels. Can maybe take that in another PR.

### DIFF
--- a/multiqc/modules/seqera_cli/seqera_cli.py
+++ b/multiqc/modules/seqera_cli/seqera_cli.py
@@ -188,7 +188,7 @@ class MultiqcModule(BaseMultiqcModule):
                 "description": "Name of the repository",
                 "scale": "Accent",
                 "modify": lambda x: repositories.index(x),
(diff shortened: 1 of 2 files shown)


multiqc/modules/seqera_cli/seqera_cli.py
```diff
@@ -182,13 +204,29 @@ def _table(self, data_by_run):
             {v: scale.get_colour(i, lighten=0.5)} for i, v in enumerate(seqera_versions + nextflow_versions)
         ]
         repositories = list(set(d.get("repository") for d in data_by_run.values()))
+
+        def format_runUrl(x):
```
vladsavelyev, line 208: I'd rather follow the Python convention to name functions and variables with lowercase letters and underscores only:
https://peps.python.org/pep-0008/#function-and-variable-names

I know there is a lot of differently styled older code, but we can keep in mind the preferred naming in all newly added code, and when refactoring :)

ewels: Yes true - I did it like this because the JSON key is formatted that way. But agree that it's ugly. Feel free to fix 👍🏻

PR #2253, opened 2023-12-28
BclConvert: fix duplicated `yield` for 3.9.3+ when the yield is provided explicitly in Quality_Metrics

vladsavelyev:
For older versions of BclConvert, we do not have Yield and Mean Quality Score in the stats, so we calculate them from read counts and quality along with the cluster length. Adding a check to make sure we don't double-calculate stats when the stats are provided explicitly.

Fixes https://github.com/MultiQC/MultiQC/issues/2250


ewels (approved):
Nice!

PR #2256, opened 2024-01-03
Update docs to work with Astro 4

mashehu:
Changes required for dependency updates on the website. See https://github.com/MultiQC/website/pull/85


ewels:
This is quite an irritating change, as it effectively breaks local development for anything involving images.

Is there no way to support relative image paths?

mashehu:
okay, I think I found a work around 💡

ewels:
Nice! Now just need to remember not to use the more obvious relative path. But that should be fine, as it won't work in the preview and can check the syntax of the other ones that do.. 😅 

Thank you!

ewels:
@multiqc-bot changelog

ewels (approved):
Thanks!

PR #2265, opened 2024-01-16
HUMID: Add cluster statistics

Redmar-van-den-Berg:
- Split up the parsing of HUMID statistics into independent files
- Add three additional plots, which show clustering statistics from HUMID
- Rename the "M Unique" column in General Statistics to "M Unique Reads", to match UMI-Tools, which is a similar tool
- Add the % of reads after deduplication to the General Statistics table, to match UMI-Tools


vladsavelyev:
Thanks a lot @Redmar-van-den-Berg for the PR! 

I wonder if you happen to have test files `neigh.dat`, `counts.dat`, `clusters.dat` to put into https://github.com/MultiQC/test-data/tree/main/data/modules/humid?

multiqc/modules/humid/neighbours.py
```diff
@@ -0,0 +1,65 @@
...
+    plot_config = {
+        "id": "humid-neighbours",
+        "title": "HUMID: Neighbour statistics",
+        "ylab": "Number of sequences",
+        "xlab": "Number of neighbours",
+        "logswitch": True,
+        "logswitch_active": True,
+        "hide_zero_cats": False,
+    }
+    self.add_section(
```
vladsavelyev, line 57: If neigh.dat is not found, an empty plot will be generated. Would be nice to check if `self.neighbours` is non-empty before the `self.add_section` call

Redmar-van-den-Berg:
I've created a pull request with some example data here: https://github.com/MultiQC/test-data/pull/313

multiqc/modules/humid/neighbours.py
Redmar-van-den-Berg: Each submodule separately searches for their associated log file, so if a file is completely missing it will not be included in the report.
The tool for this module (HUMID) will never generate empty statistics files.

Redmar-van-den-Berg: I did some more testing, and you can indeed add up with an empty section if there is only a single sample, and that sample cluster file has no data. I've fixed this in the latest commit, thanks!

vladsavelyev (approved):
Looks great! Minor things which I pushed directly into the branch, if you don't mind - and happy to merge 👍

ewels:
@multiqc-bot changelog

vladsavelyev:
Ah, figured why the tests were failing. `self.cluster` vs `self.clusters` were interchanged in the code

ewels (approved):
Great stuff, thanks!

PR #2273, opened 2024-01-22
Update Singularity build command in installation.md

kmlhpk:
Just a small change to the Singularity section of the install instructions:

`docker://multiqc/multiqc:1.20` is not the right URI, need to use `v1.20` (replacing version as appropriate) as the tag instead 

Using the former command results in `FATAL:   While performing build: conveyor failed to get: reading manifest 1.20 in docker.io/multiqc/multiqc: manifest unknown: manifest unknown`

Using `v1.19` instead worked.


ewels (approved):
Great spot, thank you! I noticed another instance of the same typo further up too, for Docker. So fixed that and pushed to your PR..

kmlhpk:
cool, thanks! happy to have helped :)

PR #2276, opened 2024-01-23
Heatmap: check if value is numeric when calculating min and max

vladsavelyev:
Attempt to fix https://github.com/MultiQC/MultiQC/issues/2274


ewels (approved):
LGTM 👍🏻

PR #2277, opened 2024-01-23
Use alternative method to walk directory using pathlib

studioph:

### Background
We've been running MultiQC as part of Nextflow on storage that is backed by CIFS file shares. During this, we discovered that MultiQC was skipping over the directory symlinks that NF creates as part of its workflow. Debugging revealed that the symlinked folder entries were being returned by `os.walk` under the `files` list instead of folders, tracing further revealed `scandir` reporting them as files instead of symlinks/folders.

However, `os.path.is_` (also used by `pathlib`) functions all reported symlinks correctly. This lead to [this issue](https://github.com/python/cpython/issues/102503) on the cpython repo, which pointed to a bug in the kernel CIFS driver not reporting the `D_TYPE` of symlinks correctly. I've compiled a [small example repo](https://github.com/studioph/cifs-symlinks) to allow for showcasing and reproducing the issue. Ideally, the CIFS kernel module would be updated to fix this, but it is unclear sifting through the mailing list whether or not this has been patched as of time of writing, and even if it has been it will be some time before it is generally available in most distros, particularly LTS ones.

### What does this have to do with MultiQC?
There are 2 ways around this problem:
  - Don't use CIFS and use something else like NFS that is POSIX-compliant
  - Use something other than `os.walk/scandir` for traversing folders

Unfortunately for us, number 1 isn't an option, which leads to number 2 which is this PR. While this particular situation is very niche, it could provide a starting point for a more modular storage system to discovering files, similar to the current plugin approach for parsing them.

### What does this PR add?

- A new CLI option `--walk-method` with 2 choices: `oswalk` and `pathwalk`
- Separating the directory traversal section of `report.get_listlist` into separate functions
    - The original logic is kept in `oswalk`, and is the default without needing to add any CLI or config options, so nothing should change for existing and future users that want the current behavior
    - A new function `pathwalk` that uses `pathlib` instead to walk folders
- Separated the "is MultiQC running in the installation folder?" check into a separate function as well for reuse in both methods
- Unit tests for `report.get_filelist`. Since there were no existing unit tests, these were first written against the existing `os.walk` code to establish and baseline of expected behavior, then `pathwalk` was added.
    - There is a separate PR in the test-data repo with the actual unit test code and test data following the existing codebase organization: MultiQC/test-data#314

### Notes
The `pathwalk` method is not meant to be as fast or efficient as `os.walk`, since it both requires an extra `stat()` system call for each item to properly classify entries, and in its current form uses basic recursion. However, in timed local testing running multiqc against the full test data modules folder shows virtually no difference between the 2, so it's not as if this will double runtimes or anything drastic like that.

I tried to modify as little of the existing code as possible since I am not intimately familiar with the codebase, but there is still of course the chance that I have misunderstood things when writing both tests and the implementation code for this PR, and am happy to correct things that I may have overlooked.


melloware (approved):

ewels:
Interesting, thanks for this @studioph! I have had the idea of generally switching from `os` to `pathlib` throughout the codebase for a while, as part of a general modernisation effort. But I have never got around to doing anything or going as deep as you have into potential differences.

Aside from potentially very minor slowdowns, do you foresee any other problems with `pathlib`? I'm wondering if it would make sense to just switch wholesale rather than making it a user-facing option.

studioph:
> Aside from potentially very minor slowdowns, do you foresee any other problems with `pathlib`? I'm wondering if it would make sense to just switch wholesale rather than making it a user-facing option.

Not that I can think of, though I will be honest in that I've been mostly working in the `report` module for this and would need to look through some other modules to see if there would be any gotchas.

I'm generally a fan of Pathlib, even with some of the limitations it has, but also recognize it's not usually a trivial change - hence trying to keep this PR contained.

vladsavelyev:
I agree with Phil that it makes to just fully switch to pathlib without making it optional. In my previous projects I primarily used pathlib, and thought of it as just an object-oriented repalcement of `os.path`, and I wasn't aware of such behavior differences. But it sounds like it's just an improvement, so we can switch to pathlib completely.

I imagine at some point we should switch to pathlib everywhere, and any paths in the code should be immediately wrapped into `Path` objects and passed around like that. This would work nicely when we add type hints for static analysis with mypy. But switching the dir walking method is a great start.

melloware (approved):
Agreed @vladsavelyev

ewels:
Sounds like we're all aligned, great 👍🏻 Can start here and then progressively switch the codebase over time as we refactor.

studioph:
Sounds good, I'll remove the CLI option parts and the os.walk version

melloware (approved):

studioph:
@vladsavelyev My pleasure! I'm glad it was something everyone was interested in.

melloware:
@vladsavelyev @studioph this is a huge win for us and for our client who needs this!

PR #2279, opened 2024-01-24
Add `CMD` to `Dockerfile` so a default run without any parameters displays the `--help`

alanhoyle:

This change adds a default multiqc command to the Docker image.  

If the current docker image is run without any parameters, it starts up an interactive python session.  This is counterintuitive and my experience is that the majority of Docker containers for command line tools run the wrapped tool by default if run with no parameters.  

The CMD added at the end instead runs:

    multiqc --help

to display the current multiqc command line help.  

An alternative implementation would do:  

    ENTRYPOINT multiqc
    CMD --help

This would run the help and always run MultiQC, passing any additional parameters in, but would break any existing docker runs that didn't manually set the `--entrypoint` at the Docker command line.  

That could be worked-around by making a new `multiqc.sh` wrapper script that removed any leading `multiqc` from the container run command, and making that the default `ENTRYPOINT multiqc.sh`.


ewels:
Nice - thanks! What happens if a tool customises the entrypoint (eg. to run MultiQC) to have a valid command and then uses the image without the `CMD` though. This would break their workflow, as it would append `--help` onto an otherwise valid command, right?

> That could be worked-around by making a new `multiqc.sh` wrapper script that removed any leading `multiqc` from the container run command, and making that the default `ENTRYPOINT multiqc.sh`.

This is an interesting idea 🤔 

The Docker image used to have an `ENTRYPOINT` and I quite liked it, as otherwise you have to duplicate your `multiqc` a bunch. But several workflow tools / other tools need to be able to customise the entrypoint so I removed that a while back to maximise compatability. With your suggestion maybe we could have the best of both worlds..

Could also use the bash script to do the `multiqc --help` functionality and avoid the Python session. Might be the only way to avoid the edge case above.

What do you think?

alanhoyle:
> Nice - thanks! What happens if a tool customises the entrypoint (eg. to run MultiQC) to have a valid command and then uses the image without the `CMD` though. This would break their workflow, as it would append `--help` onto an otherwise valid command, right?

From what I can tell, when one adds `--entrypoint $SOMETHING` to their `docker run [...]` command, Docker does not use/add the contents of the CMD in the command line that it executes.   if someone is using a custom entrypoint, they will also have customized the rest of their command line

See screenshot below.  I built an image using my `Dockerfile` and run the following commands:

```
docker run --rm -ti multiqc-local |head -10                             # Runs MultiQC with --help from the CMD
docker run --rm -ti --entrypoint multiqc multiqc-local                  # Runs MultiQC on /home/multiqc dir in the container
docker run --rm -ti --entrypoint python multiqc-local                   # Runs an interactive python
docker run --rm -ti --entrypoint python multiqc-local --help |head -5   # Displays the python help
docker run --rm -ti multiqc-local multiqc --help |head -10              # Prints  the MultiQC help
```

Screenshot 2024-02-05 at 12 10 32 PM

vladsavelyev (approved):
Looks good to me!

ewels:
> Any reason why we should be looking into the `ENTRYPOINT`+`multiqc.sh` alternative instead of merging this?

I think that was just a stretch goal to make the command look cleaner, as avoids having to specify `multiqc` twice basically.

PR #2282, opened 2024-01-24
Qualimap: address NBSP as thousand separators

vladsavelyev:
Fix https://github.com/MultiQC/MultiQC/issues/1870

Refactor to determine decimal format once per file. Optimize to call regex only one per line.


multiqc/modules/qualimap/QM_BamQC.py
```diff
@@ -116,16 +122,32 @@ def parse_genome_results(self, f):
...
+                if not m:
+                    continue
+                val = m.group(1)
+                val = re.sub(r"\xa0", "", val)
+                if regex_type == int:
+                    d[key] = int(re.sub(r"[,\.]", "", val))
+                elif regex_type == float:
+                    # need to determine either dot or comma is either a decimal or thousands separator
+                    if val.count(".") == 1 and val.count(",") == 0:
+                        if len(val.split(".")[-1]) < 3:  # 1.12
```
ewels, line 137: Does it make sense to figure this out once per file, rather than repeatedly - would that be more safe? Or is it not needed?

I've written similar code elsewhere, eg. [here](https://github.com/MultiQC/MultiQC/blob/d26af5e19c2ebde308bac339fb1b0c3a2dafa99a/multiqc/modules/picard/HsMetrics.py#L114-L121). (Basically all the Java tools have this problem).

vladsavelyev: Optimized to determine the format once per file. Assumes that "rate" metrics like GC% or error rate will never have a thousand separator.

ewels: Great 👍🏻

multiqc/modules/qualimap/QM_RNASeq.py
```diff
@@ -81,7 +86,7 @@ def parse_reports(self):
             if line.startswith("#"):
                 continue
             coverage, count = line.split(None, 1)
-            coverage = int(round(float(coverage)))
+            coverage = int(round(float(coverage.replace(",", "."))))
```
ewels, line 89: I don't quite understand this change... Is it not possible to get something like `123,456.78` here? Does it not also need similar logic to elsewhere?

multiqc/modules/qualimap/__init__.py
```diff
@@ -1,3 +1,94 @@
...
+                if "." in val:
+                    if decimalcomma is False:
+                        log.error(
+                            f"Couldn't determine decimal separator for file {fpath}, as differently formatted"
+                            f"rational values are found"
+                        )
+                        return {}
+                    decimalcomma = True
+    if decimalcomma is None:
+        log.warning(f"Couldn't determine decimal separator for file {fpath}")
```
ewels, line 78: If we didn't find a decimal then I don't think it really matters, the file should parse fine, right? I could imagine this being relatively common with empty BAM files etc. Suggest we bump down to debug.

```suggestion
        log.debug(f"Couldn't determine decimal separator for file {fpath}")
```

ewels (approved):
Very nice. Surprisingly complex! I wonder if we can move this to a central helper function to share across other modules, but suggest that we only do that another day if we come across similar code and feel like it.

multiqc/modules/qualimap/QM_RNASeq.py
vladsavelyev: The coverage file doesn't use decimals, so it would be `123456.78`

PR #2286, opened 2024-01-30
Seqera Platform CLI: handle failed tasks

ahvigil:
fixes https://github.com/MultiQC/MultiQC/issues/2285 by checking for null `d["complete"]`


multiqc/modules/seqera_cli/seqera_cli.py
```diff
@@ -171,7 +171,7 @@ def _parse_data(self, d):
         # "start" and "complete" are time stamps like time stamps like 2023-10-22T14:39:01Z
         # parse them with a library, take the difference "complete" - "start" to get the
         # wall time, and convert the wall time it to a human-readable format.
-        if "start" in d and "complete" in d:
+        if "start" in d and "complete" in d and d["complete"] is not None:
```
ahvigil, line 174: note that due to the dict comprehension on line 162, `"complete" in d` is always true... but I left this in here in case the above ever changes.

vladsavelyev (approved):
Thanks!

PR #2287, opened 2024-01-31
Export `config.output_dir` in MegaQC JSON

fgvieira:
Add `output_dir` to JSON file, so it can be included in `MegaQC`.
Store `output_dir` as an absolute path, so it can be used in multiqc/megaqc#22.


vladsavelyev:
@multiqc-bot changelog

PR #2300, opened 2024-02-08
Make "Copy Table" button target the `table` element specifically

vladsavelyev:
When the section `anchor` is identical to the table id (e.g. https://github.com/MultiQC/MultiQC/issues/2293), the "Copy Table" button breaks, as it effectively targets the first element it can find with the ID, which happens to be the section header:

https://github.com/MultiQC/MultiQC/blob/4905cb8b4d72c0d39b0d0b9d6884d10413c9cd76/multiqc/templates/default/content.html#L42

Modifying the copy table target to target the `table` element specifically.

But ideally, we might want to make sure that IDs are different in custom content as well, without requiring users to make `section_anchor` and plot `id` distinct (run `report.save_htmlid` on one of them?).

Fixes https://github.com/MultiQC/MultiQC/issues/2293

### DIFF
--- a/multiqc/modules/software_versions.py
+++ b/multiqc/modules/software_versions.py
@@ -53,7 +53,7 @@ class MultiqcModule(BaseMultiqcModule):
         html = [
             dedent(
                 f"""\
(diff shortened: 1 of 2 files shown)


ewels (approved):
Quick fix is good for now, let's revisit this properly in a future release.

PR #2308, opened 2024-02-09
Clean up plotting documentation and remove deprecated options [skip changelog]

vladsavelyev:


ewels:
It'd be good to explicitly list any options that we're removing under a heading on this page: https://multiqc.info/docs/development/compatibility/

PR #2314, opened 2024-02-12
Configuration: fix reading `config.output_fn_name` and `--filename`

vladsavelyev:
Fixes https://github.com/MultiQC/MultiQC/issues/2305

Additionally, fix when `config.base_count_prefix` is set to null


ewels (approved):
LGTM!

PR #2316, opened 2024-02-12
Fix `--filename` and `config.output_fn_name` options

vladsavelyev:
Proper fix for https://github.com/MultiQC/MultiQC/pull/2314


ewels (approved):
Nice work, thanks!

PR #2321, opened 2024-02-12
Group-mode bar plot: fix inner gap

vladsavelyev:
When adding each sample to the bar graph as a separate trace, when the barmode is `group` the Plotly adds a giant gap between the inner bars:

Screenshot 2024-02-12 at 17 44 41

Since the group mode is not used often, we can sacrifice dimming the de-highlighted bars and leave the trace labels as a highlight marker alone:

Screenshot 2024-02-12 at 17 46 31

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### MultiQC updates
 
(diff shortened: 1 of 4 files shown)


ewels (approved):
LGTM

PR #2327, opened 2024-02-13
Dragen: support `gvcf_metrics`

vladsavelyev:
Supports `gvcf_metrics`. Whenever both `vc_metrics` and `gvcf_metrics` are available, overrides corresponding metrics from `vc_metrics` with those found in `gvcf_metrics`.

Fixes https://github.com/MultiQC/MultiQC/issues/2326


fsharifi:
I was wondering if there's a way to add these gvcf metrics to the general stat metrics.

PR #2330, opened 2024-02-15
Always create JSON even when MegaQC upload disabled

fgvieira:

Currently `JSON` file is not created if `--no-megaqc-upload` specified. I think `JSON` should always be created, except if `config.data_dump == False`.


vladsavelyev (approved):
Nice, agree. Thank you!

Issue #2046, opened 2023-09-15
Set up unit testing


ewels:
When we go through refactoring modules for v2.0, take the opportunity to also set up a proper unit testing framework.

Can dump more ideas here, but for example:

* For each module, run a core set of tests:
    * On init, are mandatory variables set
    * On parse + ignore, does it return an exception because no data
    * ...

We can keep the main integration test of running MultiQC on everything, but a bunch of the additional tests that we currently have in the CI can almost certainly be better done as unit tests.

We also have various other sporadic mini-test scripts etc all over the place, so can consolidate those.

Ideas for future work as part of the initial setup:

- [x] Write new page about testing for the docs, with examples
- [x] Have some bespoke module-level function tests for a couple of modules
- [x] Look more into snapshot results for module parsing test data
- [x] Try and bulk up coverage on some core files (interactive.py, write_results.py etc)

After this we can consider the issue complete, and try to continue adding test coverage whenever we work on any file or fix any bugs.

vladsavelyev:
Regarding the currently existing integration testing, there are several modules that fail when multiple incompatible inputs are passed (`preseq`, `bclconvert`), and they probably _should_ make MultiQC exit with error code 1. It's not practical at the moment because that would make the integration test `multiqc test_data/modules` always fail, since we have mixture of inputs for those modules in MultiQC_TestData.

So I think it might makes sense to break up those modules into module groups, using the config with `path_filters` in the CI runs.

PR #2346, opened 2024-02-18
fix general stats for mosdepth

jchorl:

Fixes general stats for mosdepth, includes coverage threshold columns

Regression from https://github.com/MultiQC/MultiQC/pull/2257

That PR overwrote general stats headers instead of just adding. Some of the functions called had already set fields on general stats headers.

The issue happens on multiqc testdata!!

Before my fix:

Screenshot 2024-02-18 at 06-07-57 MultiQC Report

After:

Screenshot 2024-02-18 at 06-08-20 MultiQC Report

What tests do we need to add so this would have been caught via automated testing?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Fix gettng id from optional pconfig=None ([#2337](https://github.com/MultiQC/MultiQC/pull/2337))
 - Barplot: keep sample order ([#2339](https://github.com/MultiQC/MultiQC/pull/2339))
 - MegaQC: dump `pconfig` ([#2344](https://github.com/MultiQC/MultiQC/pull/2344))
+- fix general stats for mosdepth ([#2346](https://github.com/MultiQC/MultiQC/pull/2346))
 
 ### New modules
 
--- a/multiqc/modules/mosdepth/mosdepth.py
+++ b/multiqc/modules/mosdepth/mosdepth.py
@@ -327,44 +327,46 @@ class MultiqcModule(BaseMultiqcModule):
                 self.genstats_mediancov(genstats, genstats_headers, cumcov_dist_data)
 
         # Add mosdepth summary to General Stats
-        genstats_headers.update({
-            "mean_coverage": {
-                "title": "Mean Cov.",
-                "description": "Mean coverage",
-                "min": 0,
-                "scale": "BuPu",
+        genstats_headers.update(
+            {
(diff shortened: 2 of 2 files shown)


vladsavelyev (approved):
Oh, that looked bad. Thank you a lot for the fix!!

jchorl:
> Oh, that looked bad. Thank you a lot for the fix!!

@vladsavelyev  Is there any way this could have been caught with automated testing? I noticed MultiQC_TestData already has input files that trigger this, all we'd need is validation that beyond just exiting 0, the results are correct too.

vladsavelyev:
@jchorl, we need unit tests for individual modules. Currently, we only have integration tests. Essentially, we only test that MultiQC and modules don't crash when running with different global options (https://github.com/MultiQC/MultiQC/blob/main/.github/workflows/multiqc.yml#L84-L144). However these tests are not aware of individual modules' specifics, so can't verify if an expected metric is shown in a report.

We have a task to set up tests: https://github.com/MultiQC/MultiQC/issues/2046 but doing that for every module is going to be quite a lot of work 😬  Anyway, going to be done over this year.

jchorl:
> @jchorl, we need unit tests for individual modules. Currently, we only have integration tests. Essentially, we only test that MultiQC and modules don't crash when running with different global options (https://github.com/MultiQC/MultiQC/blob/main/.github/workflows/multiqc.yml#L84-L144). However these tests are not aware of individual modules' specifics, so can't verify if an expected metric is shown in a report.
> 
> We have a task to set up tests: #2046 but doing that for every module is going to be quite a lot of work 😬 Anyway, going to be done over this year.

Got it. Has any work been done yet on #2046?

If the framework is set-up or close, I'd be happy to add some tests for some modules, and add more as we hit issues.

If it's not set up, I may be able to help get something set up and running in CI although that's a much bigger task.

vladsavelyev:
@jchorl, it's not set up yet, and I don't think any work was done at all. It would be great to get it done, but beforehand I was planning to do some refactoring to make the code more unit-testable overall:

* move the module parsing logic out of `__init__` in modules classes;
* make `report` an object and not a module with global variables;
* same for `config`;
* separate search, parsing, and report generation logic in `multiqc.py`.

The issue for that is https://github.com/MultiQC/MultiQC/issues/1051, and I started working on that already, hopefully, will get it merged within a month. Then we could focus on unit testing, and it would be really great to have your input and assistance!

PR #2352, opened 2024-02-18
CI: use uv pip

vladsavelyev:
Try speeding up the setup of CI jobs by replacing pip with [uv pip](https://astral.sh/blog/uv)

### DIFF
--- a/.github/workflows/multiqc.yml
+++ b/.github/workflows/multiqc.yml
@@ -68,12 +68,14 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - uses: yezz123/setup-uv@v1
+
       # For CSP checking
       - name: Install dependencies for CI tests
-        run: python -m pip install --upgrade ${{ env.core_python_dependencies }}
+        run: uv pip install --upgrade ${{ env.core_python_dependencies }}
 
       - name: Install MultiQC
-        run: pip install .
+        run: uv pip install .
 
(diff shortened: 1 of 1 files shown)


.github/workflows/multiqc.yml
```diff
@@ -68,12 +66,16 @@ jobs:
...
-        run: python -m pip install --upgrade ${{ env.core_python_dependencies }}
+      - uses: yezz123/setup-uv@v1
+
+      - name: Set the VIRTUAL_ENV variable for uv to work
+        run: echo "VIRTUAL_ENV=${Python_ROOT_DIR}" >> $GITHUB_ENV
 
       - name: Install MultiQC
-        run: pip install .
+        run: |
+          # Temporary hack to remove duplicated pip package (hopefully will be addressed in uv or GitHub)
```
ewels, line 76: Would be good to link to a GitHub issue here in the code comment, so that it's easy to check on its status in the future easily.

.github/workflows/multiqc.yml
```diff
@@ -68,12 +66,16 @@ jobs:
...
+          rm -rf ${VIRTUAL_ENV}/lib/python3.*/site-packages/pip-23.*
+          uv pip install -e .
```
ewels, line 78:
```suggestion
          # Prefer not to use `-e` but uv doesn't support this currently (Feb 2024)
          uv pip install -e .
```

PR #2353, opened 2024-02-18
Replace `setup.py` with `pyproject.toml`

vladsavelyev:
Since `uv pip` needs it to work in the editable mode, it's a good time to switch to the modern build setup. 

Following the guides at:
https://packaging.python.org/en/latest/guides/writing-pyproject-toml
https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html


pyproject.toml
```diff
@@ -1,3 +1,240 @@
...
+    "Pillow>=10",  # to add a logo into flat plots
+    "plotly>=5.18",
+    "pyyaml>=4",
+    "pyaml-env",
+    "rich>=10",
+    "rich-click",
+    "spectra>=0.0.10",
+]
+requires-python = ">=3.8"
+authors = [{name = "Phil Ewels", email = "phil.ewels@seqera.io"}]
```
ewels, line 28: You should add yourself as an author @vladsavelyev

pyproject.toml
```diff
@@ -1,3 +1,243 @@
...
+    "Topic :: Scientific/Engineering :: Visualization",
+]
+
+[project.optional-dependencies]
+matplotlib = [
+    "matplotlib>=2.1.1",
+]
+dev = [
+    "pre-commit",
+    "ruff",
```
Zethson, line 59: You already have that as part of pre-commit. Why are you installing it again?

PR #2359, opened 2024-02-20
Drop `full_figure_for_development` the call to avoid Kaleido issues

vladsavelyev:
To get around: https://github.com/MultiQC/MultiQC/issues/2357

The `full_figure_for_development` call is used here to avoid line plot bands affecting the calculated range. We shouldn't crash if this function fails, ~so wrapping in try-except. Worst case, the plot range will just be bigger than needed.~. Just dropped that call for good, and calculating the range from data points.

TODO

- [x] Properly figure out the Kaleido problem.
- [x] Calcualte manually the max and min data values and set `autorangeoptions.minallowed`/`autorangeoptions.maxallowed` directly.


ewels (approved):
Does this mean that we can drop the Kaleido dependency from `setup.py` completely?

vladsavelyev:
@ewels still needed for flat plot export

PR #2366, opened 2024-02-22
PURPLE: support v4.0.1 output without `version` column

vladsavelyev:
Fix https://github.com/MultiQC/MultiQC/issues/2363


vladsavelyev:
@multiqc-bot changelog

PR #2368, opened 2024-02-23
Generic font family for Plotly

vladsavelyev:
Fix https://github.com/MultiQC/MultiQC/issues/2338


ewels (approved):
Would be cool to use Inter one day! For brand consistency. Can take that with the template updates in another PR though.

PR #2388, opened 2024-02-26
Picard: Crosscheck Fingerprints: add LOD heatmap

vladsavelyev:
cc @matthdsm

- [x] Add heatmap for LOD values in addition to the table
- [x] If too many pairs in table, skip those with `Expected` status and show a warning
- [x] Use the `warn` status if  there are `Inconclusive` instead of `fail`
- [x] Add a separate sample-wise table instead of general stats, sort it by status
- [x] Sort pairwise table by status
- [x] Add a column "Best match" and "Best match LOD" for the sample-wise table and for unexpected mismatched pairs in pairwise table
- [x] Hide LOD Threshold column
- [x] Minor fixes and column name standartization

### DIFF
--- a/multiqc/modules/picard/CrosscheckFingerprints.py
+++ b/multiqc/modules/picard/CrosscheckFingerprints.py
@@ -92,7 +92,7 @@ def parse_reports(module):
     general_stats_headers = {
         "Crosschecks All Expected": {
             "title": "Crosschecks",
-            "description": "All results for samples CrosscheckFingerprints were as expected.",
+            "description": "The worst pairwise result for this sample.",
             "cond_formatting_rules": {
                 "pass": [{"s_eq": "All expected"}],
                 "warn": [{"s_eq": "Some inconclusive"}],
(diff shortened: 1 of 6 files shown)


multiqc/modules/picard/CrosscheckFingerprints.py
```diff
@@ -104,14 +116,41 @@ def parse_reports(module):
...
+        name="Crosscheck Fingerprints Heatmap",
+        anchor=f"{module.anchor}-crosscheckfingerprints-heatmap",
+        description="Pairwise identity checking between samples and groups: heatmap of LOD scores.",
+        plot=heatmap.plot(
+            heatmap_data,
+            pconfig={
+                "id": "picard-crosscheckfingerprints-lod-heatmap",
+                "title": f"{module.name}: Crosscheck Fingerprints",
+                "ylab": "Left sample",
+                "xlab": "Right sample",
```
ewels, line 144: Not sure that these axis labels really add much?
```suggestion
```

ewels (approved):
LGTM 👍🏻

PR #2394, opened 2024-02-27
Add export button for table

vladsavelyev:
Add export button for table as well: 

Screenshot 2024-02-27 at 18 12 03

The button brings to the data export modal, the same as for the corresponding violin plot. And matches the TSV exported in Python as well. For the header, use the column `title` field (and the same for the violin).

Partially addresses https://github.com/MultiQC/MultiQC/issues/2389


ewels:
Found an edge case :trollface: 

If you export table, then switch to a violin plot and hit export, you get TSV instead of Plot:
https://github.com/MultiQC/MultiQC/assets/465550/2b7030b2-5d49-4764-a8cb-6395e1f4cc9d

If you try to export Violin first, it shows Plot as expected

ewels:
Also might be nice to add 5px margin above the top of the table, to avoid the scroll bar hitting up directly against the export button:

CleanShot 2024-02-28 at 11 20 28@2x

ewels:
Might be nice to rename the data to `Export Data` or the more commonly recognised `Export as CSV`? Just in case less techy users are unfamiliar with what TSV means.

vladsavelyev:
Fixed all!

ewels (approved):
Great!

PR #2395, opened 2024-02-27
Use a padded span with `nowrap` instead of `&nbsp;` before suffixes in table cells

vladsavelyev:
Address https://github.com/MultiQC/MultiQC/issues/2389
 
Instead of adding a `&nbsp;` before a suffix, add an empty span element with padding - the same as for the thousand separators. This way, there is going to be a space between the number and the suffix, but copying the whole thing will include neither spaces nor a special NBSP character, which is helpful for the Copy Table functionality.

```py
suffix = header.get("suffix")
if suffix:
    suffix = suffix.replace("&nbsp;", " ").strip()
    valstring += "<span class='mqc_small_space'></span>" + suffix

```

Users still would need to deal with stripping `%` or `M` suffixes in their spreadsheets if they rely on Copy Table for that, because ClipboardJS doesn't allow us to select what we want to copy from an element, and we want to keep nice and readable by human, not by a machine. For machines, we will have an "Export as TSV" button: https://github.com/MultiQC/MultiQC/pull/2394

Additionally, this PR adds a `nowrap` for the `.val` span element in tables just so it doesn't wrap on possible multi-word suffixes. It would only be added to columns with a `scale` set, so that long-text and categorical columns would still be allowed to wrap:

```css
.mqc_table tbody tr td .wrapper .val {
  white-space: nowrap;
}
```

This PR also cleans out some `&nbsp;` added to suffixes by modules, and replaces a few occurrences of .format() with f-strings.


ewels (approved):
Nice 👍🏻 

The `&nbsp;` was a bit of a hack in hindsight! This is cleaner..

PR #2396, opened 2024-02-27
Update Clipboard library

vladsavelyev:
A new version of the library, plus add the un-minified version to the repo for easier debugging (it won't be embedded into actual reports). 

TODO: replace with the native API (see Phil's comment)


ewels (approved):
LGTM, though note that I think longer term we can ditch this library completely.

Basically [all browsers now support](https://caniuse.com/clipboard) the [native Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) functionality, so we should be able to use pure JavaScript.

vladsavelyev:
Makes sense. Was hoping the new version is a bit more flexible, but not. We can keep this PR open to replace with the native API in the next release

ewels:
Issue to remove completely is here: https://github.com/MultiQC/MultiQC/issues/2408

PR #2397, opened 2024-02-27
Add option to place bar plot legend on the bottom

vladsavelyev:
Allow to place legend at the bottom of the bar plot with `barplot_legend_on_bottom`:

```sh
multiqc -fv test_data/data/modules -m star --cl-config "barplot_legend_on_bottom: true"
```

Screenshot 2024-02-27 at 22 21 46

Fixes https://github.com/MultiQC/MultiQC/issues/2390

Default is false and it's not recommended to set it, as it would look well only on medium-sized plots, but would overlap with ticks labels on shorter plots or create a huge gap on taller plots, e.g.:

Screenshot 2024-02-27 at 22 22 49

Screenshot 2024-02-27 at 22 22 35

Unfortunately, Plotly doesn't allow to control that gap between the plot and the legend.


ewels (approved):
LGTM! If we could just have a mention of this in the docs that'd be great. With caveats as to why it's not set like this by default.

It'd fit well here: https://multiqc.info/docs/reports/customisation/#customising-plots

PR #2399, opened 2024-02-28
When checking for git SHA, make sure we are not inside another git repo

vladsavelyev:
Common use case is to create a virtualenv inside a git clone folder, in which case `git rev-parse HEAD` will trigger on the outer repo even if MultiQC wasn't installed in editable mode.


ewels (approved):
LGTM 👍🏻

Issue #2410, opened 2024-02-29
BBMap module qahist plot fails on mixed samples with binned/not binned quality values


alneberg:
### Description of bug

Running multiqc on bbmap qahist files fails when one file is missing values for some x-values while the other one doesn't.

Missing values happens for example when one fastq file has binned quality values (standard from NovaSeqXPlus for example). 

I've attached two files that work individually but causes a crash when both are included in the report.

I've already figured out what the bug is and I am working on a fix.

### File that triggers the error

[sample1.qahist.txt](https://github.com/MultiQC/MultiQC/files/14448804/sample1.qahist.txt)
[sample2.qahist.txt](https://github.com/MultiQC/MultiQC/files/14448805/sample2.qahist.txt)

### MultiQC Error log

```console
❯ multiqc reproduce_error

  /// MultiQC 🔍 | v1.21

|           multiqc | Search path : /Users/johannesalneberg/multiqc_bug/reproduce_error
|         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 2/2
|             bbmap | Found 2 reports
╭───────────────────────────────────────────────────────────────────────── Oops! The 'bbmap' MultiQC module broke... ─────────────────────────────────────────────────────────────────────────╮
│ Please copy this log and report it at https://github.com/MultiQC/MultiQC/issues                                                                                                             │
│ Please attach a file that triggers the error. The last file found was: reproduce_error/sample1.qahist.txt                                                                                   │
│                                                                                                                                                                                             │
│ Traceback (most recent call last):                                                                                                                                                          │
│   File "/Users/johannesalneberg/Library/Caches/pypoetry/virtualenvs/multiqc-bug-5U5mvGB3-py3.11/lib/python3.11/site-packages/multiqc/multiqc.py", line 751, in run                          │
│     output = mod()                                                                                                                                                                          │
│              ^^^^^                                                                                                                                                                          │
│   File "/Users/johannesalneberg/Library/Caches/pypoetry/virtualenvs/multiqc-bug-5U5mvGB3-py3.11/lib/python3.11/site-packages/multiqc/modules/bbmap/bbmap.py", line 64, in __init__          │
│     plot=self.plot(file_type),                                                                                                                                                              │
│          ^^^^^^^^^^^^^^^^^^^^                                                                                                                                                               │
[...]

PR #2411, opened 2024-02-29
Set missing values to None for bbmap qahist

alneberg:
Attempting to fix the issue #2410 and assign missing values with None.

I'm not entirely decided if None or 0 is the most appropriate. I'm leaning towards None since the real value in this case is rather undefined than 0. However, the line plot is not handling these singular values very well. See screenshots for using 0 and None respectively. Happily accepting suggestions.

Screenshot 2024-02-29 at 14 27 54

Screenshot 2024-02-29 at 14 27 44

Also, sorry for ruining the pretty nested implicit for loop, but it was a bit tricky to debug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Module updates
 
+- **BBTools**: Set missing values to None for bbmap qahist ([#2411](https://github.com/MultiQC/MultiQC/pull/2411))
+
 ## [MultiQC v1.21](https://github.com/ewels/MultiQC/releases/tag/v1.21) - 2024-02-28
 
 ### Highlights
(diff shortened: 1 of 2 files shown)


vladsavelyev:
Thanks for the PR @alneberg! I think in this case I'd prefer to avoid adding the missing values at all (instead of adding None or 0), so the plot would look like this:

Screenshot 2024-02-29 at 19 54 38

```python
...
                for x in all_x:
                    if x not in sample_data["data"] or column >= len(sample_data["data"][x]):
                        # This can happen if the data is missing columns.
                        # Which happens if e.g. a certain quality value is not present in the data.
                        pass
                    else:
                        sample_dict[x] = sample_data["data"][x][column]
...
```

Even better, a plot with clearly shown markers would look better: 

Screenshot 2024-02-29 at 19 52 47

I might have an option for that.

Let me know what you think.

alneberg:
I think that looks just perfect, thank you! You also managed to get rid of the weird data point at 41! And thank you for linking the issue that I forgot to do.

PR #2422, opened 2024-03-05
new module: samtools/markdup

matthdsm:
Add new module for samtools/markdup

- [] Code is tested and works locally (including with `--strict` flag)
- [] `docs/README.md` is updated with link to below
- [] `docs/modulename.md` is created
- [] Everything that can be represented with a plot instead of a table is a plot

### DIFF
--- a/docs/modules/samtools.md
+++ b/docs/modules/samtools.md
@@ -17,6 +17,7 @@ Supported commands:
 - `flagstats`
 - `idxstats`
 - `rmdup`
+- `markdup`
 
 ### idxstats
 
--- a/multiqc/modules/samtools/markdup.py
+++ b/multiqc/modules/samtools/markdup.py
@@ -1,0 +1,51 @@
+# -*- coding: utf-8 -*-
+
+""" MultiQC submodule to parse output from Samtools markdup """
+
+import logging
+import json
+
+
+# Initialise the logger
+log = logging.getLogger(__name__)
+
+
+class MarkdupReportMixin:
+    def parse_samtools_markdup(self):
+        """Find Samtools markdup logs and parse their data"""
+
+        self.samtools_markdup = dict()
+        for f in self.find_log_files("samtools/markdup"):
+            parsed_data = parse_report(f["f"])
+            print(parsed_data)
+            if len(parsed_data) > 0:
+                if f["s_name"] in self.samtools_markdup:
+                    log.debug(f"Duplicate sample name found! Overwriting: {f['s_name']}")
+                self.add_data_source(f, section="markdup")
+                self.samtools_markdup[f["s_name"]] = parsed_data
+
+        # Filter to strip out ignored sample names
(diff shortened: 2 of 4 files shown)


matthdsm:
@vladsavelyev, can I get an intermediary review?
I'm not really sure anymore what I'm doing 😅 
This isn't nearly as advanced as @clintval 's version, but I need this module

matthdsm:
xref https://github.com/MultiQC/test-data/pull/315

vladsavelyev:
So there are two in-progress pull requests now instead of one 😅 Thanks for working on it, I'll take a look! Probably will continue from @clintval's one and cherry-pick your unique changes there.

x-refs: 
issue: https://github.com/MultiQC/MultiQC/issues/1862
first PR: https://github.com/MultiQC/MultiQC/pull/2254

matthdsm:
Sorry about that 😅 I didn’t want to rush you so I thought I could contribute. A wise man once said:
“Be the PR you want to see”

( I think it was Phil)

vladsavelyev:
> “Be the PR you want to see”

That worked! One PR is worth 10 issues, and two PRs make me jump into it right away :D

Alright, I effectively merged it with https://github.com/MultiQC/MultiQC/pull/2254

vladsavelyev:
One thing I'm not 100% sure about is whether we should double the "pair" counts. I like the idea of sticking to the originally reported numbers as is, but the bar plot makes less sense when you have pairs and single reads mixed up 🤔 

Are we sure that samtools markdup doesn't count every _alignment_, so we should expect both left and right to be accounted for in the `pair` numbers, as long as there were records for both in the BAM file?

clintval:
> One thing I'm not 100% sure about is whether we should double the "pair" counts. I like the idea of sticking to the originally reported numbers as is, but the bar plot makes less sense when you have pairs and single reads mixed up 🤔
> 
> Are we sure that samtools markdup doesn't count every _alignment_, so we should expect both left and right to be accounted for in the `pair` numbers, as long as there were records for both in the BAM file?

Your questions are good and I've tried auditing the code to absolutely determine the bar chart is not misleading the viewer (e.g. we aren't double counting reads in multiple categories). I would want us to understand this carefully before proposing we merge the PR. Perhaps we could reach out to the maintainers for some quick clarification?

I am pretty sure we should not be doubling any of the counts as I think they are all from the "alignment" perspective (not read, or read pair).

vladsavelyev:
> Your questions are good and I've tried auditing the code to absolutely determine the bar chart is not misleading the viewer (e.g. we aren't double counting reads in multiple categories). I would want us to understand this carefully before proposing we merge the PR. Perhaps we could reach out to the maintainers for some quick clarification?

Makes sense! If you feel like it, it would be great to reach out. Though I'd still merge now, it's enough time to add possible fixes before the next MultiQC release 😅 

> I am pretty sure we should not be doubling any of the counts as I think they are all from the "alignment" perspective (not read, or read pair).

Agree!

PR #2424, opened 2024-03-06
remove superfluous 'samtools' name

matthdsm:
It's a small thing, but there was one too many "samtools" in the sidebar and my brain doesn't agree with that 😅 


vladsavelyev (approved):
Thanks! 👍

PR #2427, opened 2024-03-06
Do not maintain change log between releases

vladsavelyev:
Since we moved to the model of always squashing PRs and keeping the changelog entries matching the PR titles, there is no strong need anymore to maintain an ongoing changelog between releases. We can instead have a script that would build a changelog section from the commit messages in `main` (= the merged PR titles) since the last git tag, and then edit it manually before pushing a new release (like we already do). That would save us from constant changelog conflict resolution in PRs.


ewels (approved):
Yeah, I guess that it basically equates to the same thing 🤔 

Given that we now (a) squash merges and have meaningful PR titles and (b) release so often that a changelog for dev is not really needed

Might make it slightly harder to notice shit PR titles 😅 But we can capture that at release time I guess.

Let's do it!

PR #2441, opened 2024-03-15
New module: Sequali

rhpvorderman:
My first MultiQC module.


rhpvorderman:
Pull request for the json report is open: https://github.com/MultiQC/test-data/pull/317

rhpvorderman:
@ewels @vladsavelyev , am I correct in thinking that the requirement to 

> `docs/README.md` is updated with link to below

is a remnant from an older time? There seems to be no list of applications in that document anymore. Should the PR template be updated? Should I add Sequali to the list elsewhere?

If no further action is required in that regard, the module is ready for review. 

Also, thank you for MultiQC! It is awesome! Also from a programmer perspective, it strikes the right balance in the API between customizability and ease of use. Also MultiQC forced me to rethink how I present some of the data in my tool during the creation of the module. So overall an enormously helpful process. Thanks!

vladsavelyev:
Thanks a lot for the PR @rhpvorderman! It looks great at first glance, I'll test it a bit more today.

> is a remnant from an older time? There seems to be no list of applications in that document anymore. Should the PR template be updated? Should I add Sequali to the list elsewhere?

That's right, it's no longer needed, we should update those requirements. Thanks a lot for spotting!

multiqc/modules/sequali/sequali.py
```diff
@@ -0,0 +1,501 @@
...
+                # clearly stand out from the others, which is desirable.
+                # The PuOr scale fits these criteria.
+                "scale": "PuOr",
+            },
+            "sequali_mean_sequence_length": {
+                "title": "Mean length",
+                "description": "Geometric mean of all lengths",
+                "min": 0,
+                "suffix": " bp",
+                "format": "{:,}",
```
vladsavelyev, line 158: Perhaps 1 or 2 decimal digits are sufficient?

```suggestion
                "format": "{:,.1f}",
```

multiqc/modules/sequali/sequali.py
```diff
@@ -0,0 +1,501 @@
...
+                # Neutral metric, depending on sequencing technology. Use a
+                # sliding colour change to differentiate.
+                "scale": "Blues",
+            },
+            "sequali_total_reads": {
+                "title": "Total reads",
+                "description": f"Total Sequences ({multiqc.config.read_count_desc})",
+                "min": 0,
+                "modify": lambda x: x * multiqc.config.read_count_multiplier,
```
vladsavelyev, line 167: Tiny detail: when `shared_key` is set to `read_count`/`base_count`, `modify`, `min`, and `suffix` are set automatically :)

```suggestion
```

multiqc/modules/sequali/sequali.py
```diff
@@ -0,0 +1,501 @@
...
+                "shared_key": "read_count",
+                # Neutral metric, do not use red or green.
+                "scale": "Blues",
```
vladsavelyev, line 170: I'm personally not fully convinced on that, but Phil has this tick box in the check list:

> Each table column has a different colour scale to its neighbour, which relates to the data (eg. if high numbers are bad, they're red)

So perhaps `"Greens"` here?

multiqc/modules/sequali/sequali.py
```diff
@@ -0,0 +1,501 @@
...
+        for sequence, count in most_present:
+            table_data[sequence] = {
+                "best_match": sequence_matches[sequence],
+                "libraries_affected": 100 * count / total,
+            }
+
+        table_headers = {
+            "best_match": {
+                "title": "Best Match",
+                "description": "Best maatch as found by kmer analysis",
```
vladsavelyev, line 440:
```suggestion
                "description": "Best match as found by kmer analysis",
```

multiqc/modules/sequali/sequali.py
rhpvorderman: Yes, agree you want more data as a general rule. So I made them green.

multiqc/modules/sequali/sequali.py
```diff
@@ -0,0 +1,499 @@
...
+            },
+            "sequali_mean_sequence_length": {
+                "title": "Mean length",
+                "description": "Geometric mean of all lengths",
+                "min": 0,
+                "suffix": " bp",
+                "format": "{:,.1f}",
+                # Longer reads is considered better for long-read technologies.
+                # Use green to signal that.
+                "scale": "Greens",
```
vladsavelyev, line 161: Now both columns are green :) What I meant, is that adjacent columns should be of different colors, so e.g. one is `Blues` and the other is `Greens`.

vladsavelyev:
The module looks good to me, but there is one thing that would be nice to address first, namely the size of the test data example: (https://github.com/MultiQC/test-data/pull/317#issuecomment-2003759489). It's 1.4MB, wonder how hard would be to make a smaller example?

multiqc/modules/sequali/sequali.py
rhpvorderman: Ah. I misread that. I first had no scales at all, so I fixed that and forgot the other part of the requirement. Sorry.

I choose an arbitrary color scale that is not green. Since it is not green, I had to motivate it a bit more thorough. If you feel the color scale motivation comments are overblown I am happy to remove them.

rhpvorderman:
The report size does not vary much between samples. Git automatically compresses objects using gzip. So the size of the object in the repo should be 311 KB (git uses level 1 by default).

rhpvorderman:
@vladsavelyev Sorry for submitting such a big report initially. I removed some of the adapters and a whole lot of overrepresented sequences  (that particular library has about 50% adapter dimer content, so it gets quite messy). The report is now about 300 KB. Git size 140ish. Is that okay?

ewels:
That's great, thanks!

vladsavelyev (approved):
Great, thanks a lot the contribution and for such a clean PR!

rhpvorderman:
I didn't expect this to be merged the same day I reported it ready. This is so awesome! Thanks a lot @vladsavelyev !

rhpvorderman:
Test succeeds on main on 3.8 but fail on 3.12. Due to sequali. I cannot reproduce this qcuikly locally using 3.11. Must be a 3.12 issue. I will investigate tomorrow.

vladsavelyev:
The module didn't raise `ModuleNoSamplesFound` when no all samples are filtered by use. Fixing with https://github.com/MultiQC/MultiQC/pull/2443

PR #2443, opened 2024-03-18
Fix Sequali: raise ModuleNoSamplesFound [skip changelog]

vladsavelyev:
Follow up https://github.com/MultiQC/MultiQC/pull/2441

cc @rhpvorderman


rhpvorderman:
Thanks for fixing this oversight by me!

PR #2444, opened 2024-03-19
Fix `--export` when making interactive plots

vladsavelyev:
Make sure `multiqc --export` always works (without requiring `--flat` as it does now!)

From Slack: https://nfcore.slack.com/archives/C05663US2B0/p1710846896142419


ewels (approved):
Nice one, thanks!

PR #2445, opened 2024-03-19
Add type hints to report module

vladsavelyev:
* Add type hints to the `report` singleton for static type check
* Replace `yaml.add_representer` calls that might pollute environment (which also is not liked by static typing)

### DIFF
--- a/multiqc/utils/report.py
+++ b/multiqc/utils/report.py
@@ -11,71 +11,86 @@ import mimetypes
 import os
 import re
 import time
-from collections import defaultdict, OrderedDict
+from collections import defaultdict
 from pathlib import Path
+from typing import Dict, Union, List, Optional
+
 import rich
 import rich.progress
 import yaml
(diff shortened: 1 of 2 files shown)


multiqc/utils/report.py
```diff
@@ -479,7 +489,7 @@ def data_sources_tofile():
             jsonstr = json.dumps(data_sources, indent=4, ensure_ascii=False)
             print(jsonstr.encode("utf-8", "ignore").decode("utf-8"), file=f)
         elif config.data_format == "yaml":
-            yaml.dump(data_sources, f, default_flow_style=False)
```
vladsavelyev, line 482: To avoid globally setting `yaml.add_representer(defaultdict, Representer.represent_dict)` (which also is not liked by static typing)

ewels (approved):
Nice!

So, how long until we start with mypy? 😅

PR #2446, opened 2024-03-19
Break up the main run function

vladsavelyev:
* Break up the long `run` functions into subfunctions to make the code more organized.
* Return a `RunResult` object instead of a dictionary.

Part of https://github.com/MultiQC/MultiQC/issues/1051
Depends on https://github.com/MultiQC/MultiQC/pull/2445

### DIFF
--- a/multiqc/multiqc.py
+++ b/multiqc/multiqc.py
@@ -20,7 +20,7 @@ import time
 import json
 import traceback
 from importlib.metadata import EntryPoint
-from typing import Dict, Union, Callable, List
+from typing import Dict, Union, Callable, List, Optional, Tuple
 
 import jinja2
 import requests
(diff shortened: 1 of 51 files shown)


multiqc/multiqc.py
```diff
@@ -362,33 +395,195 @@ def run(
...
+
+    logger.info("MultiQC complete")
+
+    # Move the log file into the data directory
+    log.move_tmp_log(logger)
+
+    return RunResult(sys_exit_code=sys_exit_code)
+
+
+def _set_up_logging(verbose, quiet, no_ansi):
```
ewels, line 528: Might make sense to move this to [`multiqc/utils/log.py`](https://github.com/MultiQC/MultiQC/blob/main/multiqc/utils/log.py) - there's already the `init_log` function there.

multiqc/multiqc.py
```diff
@@ -362,33 +395,195 @@ def run(
...
     )
-    console.print(
-        f"\n  [dark_orange]///[/] [bold][link=https://multiqc.info]MultiQC[/link][/] :mag: [dim]| v{config.version}\n"
-    )
-    logger.debug(f"This is MultiQC v{config.version}")
 
+    return console
+
+
+def _init(
```
ewels, line 546: I find this a bit too close to the special `__init__`. Can we name it something different?

ewels: This could probably also be moved, into the `config.py` file and renamed `init_config` or something, for example.

ewels (approved):
Looks good! Couple of minor comments but feel free to merge.

I feel like quite a lot of these functions can now move into separate files to break up this scripts so that's not so long and a bit easier to navigate. But happy to leave that to a future PR if you like.

multiqc/multiqc.py
```diff
@@ -362,33 +395,195 @@ def run(
...
+
+        _export_sources()
+
+        _write_json_dump()
+
+        _write_html_and_data(
+            tmp_dir=tmp_dir,
+            template_mod=template_mod,
+            filename=filename,
+        )
```
ewels, line 489: Love how clear this is! ideally `multiqc.py` can end up being split up into functions in all the relevant sub-files, and this block can go into `__main__.py` or somewhere, and we can finally get rid of `multiqc.multiqc` megascript :)

multiqc/multiqc.py
vladsavelyev: Yep, I've actually done that later in https://github.com/MultiQC/MultiQC/pull/2442

ewels: Ok cool - good to merge then 👍🏻

ewels:
I'll leave it to you to merge as it's complex branching.

PR #2454, opened 2024-03-21
Picard CorsscheckFingerprints: support multiple files

vladsavelyev:
Do not override when reading multiple files. 

Do not sort samples in heatmap, this way the order is preserved and multiple files are visible on the heatmap.

Fixes https://github.com/MultiQC/MultiQC/issues/2449

### DIFF
--- a/multiqc/modules/picard/CrosscheckFingerprints.py
+++ b/multiqc/modules/picard/CrosscheckFingerprints.py
@@ -41,6 +41,7 @@ def parse_reports(module):
     n_found_reports = 0
 
     # Go through logs and find Metrics
+    row_number = 0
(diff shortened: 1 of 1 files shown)


matthdsm:
Thanks for the quick fix!

Just tested this and it looks awesome! 😄 
Do you have an ETA on a release with this fix in it?

PS: I _think_ I've got another bug report ready to go, but I'm still working out if it's actually a bug or not 😅

vladsavelyev:
> Do you have an ETA on a release with this fix in it?

Not really, unfortunately. If it were me, I'd make it next week. But we have another project where I will be needed full time so the release might happen in a month or later.

PR #2457, opened 2024-03-21
Use Rich console for logging

vladsavelyev:
Stacked on https://github.com/MultiQC/MultiQC/pull/2442, do not merge. 

* Use Rich console for logging
* Drop coloredlogs

The downside is that we can't wrap in square brackets in log messages. On the other hand, we control our log messages, so shouldn't be a problem.


ewels (commented):
Super nice 👏🏻 

Can the progress bar in `report.py` also go onto the logger now too?

PR #2460, opened 2024-03-23
Exclude plot data for sections in `remove_sections`

jchorl:
I noticed some really big multiqc reports, which didn't make sense to me given the contents.

Digging in, I realized that multiqc adds plot data for all plots generated, but does not _remove_ plot data for plots in `remove_sections`.

The cause is that calling `plottype.plot(...)` actually adds the data to the report, e.g. https://github.com/MultiQC/MultiQC/blob/a0e05da08e58f5f3d4e717fcae00e1ea609ec312/multiqc/plots/plotly/line.py#L39

But the mechanism in which `remove_sections` is implemented [requires that .plot(...) already be called](https://github.com/MultiQC/MultiQC/blob/a0e05da08e58f5f3d4e717fcae00e1ea609ec312/multiqc/modules/base_module.py#L259) (and hence added to the report data).

Using `fastp` as an example module, it [generates a plot and then calls `add_section`](https://github.com/MultiQC/MultiQC/blob/a0e05da08e58f5f3d4e717fcae00e1ea609ec312/multiqc/modules/fastp/fastp.py#L135). When the plot is omitted, the plot data is not.

Naively, I figured that plot data should be added to the report in lockstep with when the plot is actually added to the report. I took a stab at implementing that.

It passes for me on a local test case with a bunch of inputs. However this feels like a broad change so I don't know how to build confidence that it won't break anything. Also, let me know if you feel there's a better approach. I don't love how `add_section` now relies on `is_instance`... feels bad.

Anyway, for my simple test case, a report went from 30mb to 4mb.

### DIFF
--- a/multiqc/modules/base_module.py
+++ b/multiqc/modules/base_module.py
@@ -14,6 +14,7 @@ from collections import defaultdict
 
 import markdown
 
+from multiqc.plots.plotly.plot import Plot as PlotABC
 from multiqc.utils import config, report, software_versions, util_functions
 
 logger = logging.getLogger(__name__)
@@ -282,6 +283,11 @@ class BaseMultiqcModule:
         comment = comment.strip()
         helptext = helptext.strip()
 
+        if isinstance(plot, PlotABC):
+            plot_html = plot.add_to_report(report)
+        else:
+            plot_html = plot
+
         self.sections.append(
             {
                 "name": name,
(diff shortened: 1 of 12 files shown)


vladsavelyev:
Thanks a lot @jchorl for catching this and for implementing the fix! 

I really like the idea of the plot functions returning Plot objects instead of HTML. 

However, it might break some assumptions as some modules add additional HTML after calling plot functions. I also suspect this might not work well with [templates](https://multiqc.info/docs/development/templates/) as they provide custom implementations of plotting functions and return HTML. But maybe we don't really need the templating system a all.

I'll look into this more later.

vladsavelyev:
Okay, I played with it a bit and extended upon your code: https://github.com/MultiQC/MultiQC/pull/2463

* Added an explicit `add_to_report` call for the general stats table special case;
* Fixed a few modules that assumed that plotting functions returned HTML;
* Added `content_before_plot` into `add_section` for the modules that want to prepend some HTML before plots (`content` is added after plot);
* Strictly typed the plotting functions along with `add_section` arguments and avoid calling `isinstance`.

I'll think a bit more how to deal with templates, and will merge into your branch so you get the credit.

PR #2463, opened 2024-03-25
Make remove_sections option fully exclude plot data

vladsavelyev:
Testing https://github.com/MultiQC/MultiQC/pull/2460

### DIFF
--- a/multiqc/modules/base_module.py
+++ b/multiqc/modules/base_module.py
@@ -231,6 +231,7 @@ class BaseMultiqcModule:
         description="",
         comment="",
         helptext="",
+        content_before_plot="",
         plot: Plot = None,
         content="",
         autoformat=True,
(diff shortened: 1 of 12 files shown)


jchorl:
Thanks for running with this!!

Quick comment - `content_before_plot` seems like a gateway to `content_before_plot`, `content_after_plot`, `plot_replace_regex`, `plot_html_replace_element`, etc etc.

One option that just came to mind - `Plot` objects can have a `render_html` and `get_data` function. Adding to the report can call `plot.render_html()` and `plot.get_data()` to get the html and add the data to the report respectively. That way you can have standard HTML rendering for all supported plot types. Some modules can subclass known plot types and override `render_html` to do funky stuff. They'd probably call `super().render_html` first and modify the result. Callers could even write their own modules with whatever bespoke `render_html` and `get_data` functions they wanted.

Not sure how big of a refactor that is. Worth considering effort/reward before diving in.

PR #2472, opened 2024-04-05
Nonpareil: fix data label

fgvieira:

Change data label to be consistent with other modules.


vladsavelyev:
Thank you for the fix!

PR #2473, opened 2024-04-06
Picard WgsMetrics: coverage plot: show % based ≥x, not >x

TBradley27:
This resolves a off-by-one error in the histogram plot for the `picard wgsmetrics` submodule. 

This issue is particularly problematic for shallow/low-pass sequencing with a high proportion of bases with zero coverage. The plot below illustrates this problem:

image

The problem is solved by reconfiguring the y-axis to plot 'the percentage of bases greater or equal than the current x-axis coverage value' rather than 'the percentage of bases greater than the current x-axis coverage value'


vladsavelyev:
Thank you a lot for the PR! Looks good :)

PR #2477, opened 2024-04-09
Bamdst: fix when `chromosomes.report` is not provided

vladsavelyev:
Fix for the case when is `chromosomes.report` is not provided. 

For https://github.com/MultiQC/MultiQC/issues/2474


ewels (approved):
Nice, thanks!

PR #2484, opened 2024-04-15
Xengsort: fix parsing long files (avoid `content_lines`)

vladsavelyev:
The Xengsort module stopped parsing after 1000'th line because it used f["content_lines"].

Should prevent using this field by lining or by removing it before modules runs.

Address https://support.seqera.io/a/tickets/4911


ewels (approved):
LGTM, thanks!

PR #2487, opened 2024-04-15
Use the actual DOI rather than a link for sequali [no changelog]

rhpvorderman:
A small oversight fix. The current DOI link does not work.


vladsavelyev:
Thank you!

rhpvorderman:
Thanks for merging! MultiQC has become such a defacto standard that not having my tool in would seriously hamper its adoption. Thank you for all the quick merges and feedback.

PR #2501, opened 2024-04-23
New module: Hostile

SumeetTiwari07:


SumeetTiwari07 (commented):
Module works fine

multiqc/utils/search_patterns.yaml
```diff
@@ -429,6 +429,8 @@ homer/FreqDistribution:
   fn: "petag.FreqDistribution_1000.txt"
 hops:
   fn: "heatmap_overview_Wevid.json"
+hostile:
+  contents: '"reads_in"'
```
vladsavelyev, line 433: If it's a json file, `fn: "*.json"` can be additionally added.

It's also a good practice to set the `num_lines` to the minimum number of lines by which you expect to meet the `contents` line. It limits reading potentially huge JSON files and speeds up the file search. See, for example:
https://github.com/MultiQC/MultiQC/blob/main/multiqc/utils/search_patterns.yaml#L602-L605

multiqc/modules/hostile/hostile.py
```diff
@@ -0,0 +1,86 @@
...
+
+        ## categories
+        cat = ["Cleaned reads", "Host reads"]
+
+        pconfig = {
+            "title": "Hostile: Reads Filtered",
+            "id": "he_reads_plots",
+            "ylab": "# Reads",
+            "plot_type": "bargraph",
+            "stacked_data": True,
```
vladsavelyev, line 75: This is no such option, but there is `stacking` that can be set to `relative` or `group` (see https://multiqc.info/docs/development/plots/#bar-graphs). `relative` is default, and means that it should stack, so no need to specify the option :)

```suggestion
            "stacked_data": True,
```

multiqc/modules/hostile/hostile.py
```diff
@@ -0,0 +1,86 @@
...
+        Plot the barplot
+        """
+        data = {}
+        for f_name, values in self.parse_data.items():
+            s_name = values[0]["fastq1_in_name"].split(".")[0]
+            database = os.path.basename(values[0]["index"])
+            data[s_name] = {"Cleaned reads": values[0]["reads_out"], "Host reads": values[0]["reads_removed"]}
+
+        ## categories
+        cat = ["Cleaned reads", "Host reads"]
```
vladsavelyev, line 68:
```suggestion
        cats = ["Cleaned reads", "Host reads"]
```

vladsavelyev:

Left a few comments.

Also, it's needed to add `self.ignore_samples` (see https://multiqc.info/docs/development/modules/#filtering-by-parsed-sample-names) and `self. add_software_version ` (see https://multiqc.info/docs/development/modules/#saving-version-information)  calls.

```py
        ...
        for f in self.find_log_files("hostile", filehandles=True):
            self.parse_logs(f)

        self.add_software_version(...)

        self.parse_data = self.ignore_samples(self.parse_data)

        if len(self.parse_data) == 0:
            raise ModuleNoSamplesFound

        log.info(f"Found {len(self.parse_data)} reports")

        self.write_data_file(self.parse_data, "multiqc_hostile")
        ...
```

vladsavelyev:
Please also create a PR in https://github.com/MultiQC/test-data with test examples.

SumeetTiwari07:
Thank you for reviewing the code and all the suggestions (`review 1`)
I made the recommended changes and updated the code as suggested.
Also added the test-data ([https://github.com/MultiQC/test-data/pull/319](https://github.com/MultiQC/test-data/pull/319))

multiqc/modules/hostile/hostile.py
```diff
@@ -0,0 +1,98 @@
...
+            self.add_data_source(json_file)
+
+        s_name = self.clean_s_name(json_file["fn"])
+        self.parse_data[s_name] = parse_data
+
+        # print(self.parse_data[s_name][0])
+        if "version" in self.parse_data[s_name][0]:
+            self.version = self.parse_data[s_name][0]["version"]
+        else:
+            self.version = "0.4.0"
```
vladsavelyev, line 69: Where is 0.4.0 coming from? If the version is unknown, we shouldn't add anything.

multiqc/modules/hostile/hostile.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def hostile_plot(self):
+        """
+        Extract the data
+        Plot the barplot
+        """
+        data = {}
+
+        for f_name, values in self.parse_data.items():
+            s_name = values[0]["fastq1_in_name"].split(".")[0]
```
vladsavelyev, line 79: `self.parse_data` is already assumed to be indexed by sample names (i.e. `self.ignore_samples` takes that assumption), so we can't create a different sample name here. Better move this line into parse_logs. Clean clean with `self.clean_s_name()` instead of manually calling `split` (the function knows about the extensions to remove)

multiqc/modules/hostile/hostile.py
```diff
@@ -0,0 +1,98 @@
...
+            "title": "Hostile: Reads Filtered",
+            "id": "he_reads_plots",
+            "ylab": "# Reads",
+            "plot_type": "bargraph",
+        }
+
+        self.add_section(
+            name="Reads Filtering",
+            anchor="hostile-reads",
+            description=f"This plot shows the number of cleaned reads vs host-reads per sample (database index: {database}).",
```
vladsavelyev, line 96: `database` is initialized in the loop for each sample separately and this line is outside of the loop. Need to decide how to handle the situation if samples have different database

multiqc/utils/search_patterns.yaml
```diff
@@ -429,6 +429,10 @@ homer/FreqDistribution:
   fn: "petag.FreqDistribution_1000.txt"
 hops:
   fn: "heatmap_overview_Wevid.json"
+hostile:
+  fn: "*.json"
+  contents: '"reads_in"'
```
vladsavelyev, line 434: Can the pattern be more specific than `"reads_in"`? Maybe `"reads_removed_proportion"` would be more unique. It's unfortunate there is no `"tool": "hostile"` in the JSON that would allow to detect it unambigously :(

multiqc/modules/hostile/hostile.py
```diff
@@ -0,0 +1,98 @@
...
+        Parsing json_file
+        """
+        try:
+            parse_data = json.load(json_file["f"])
+
+        except json.JSONDecodeError:
+            log.warning(f"Could not parse JSON file {json_file['f']}")
+            return
+
+        if len(parse_data) > 0:
```
vladsavelyev, line 59: Can there be more than one entry in the JSON file? How that should be handled? Please add such example to the test-data repo.

multiqc/modules/hostile/hostile.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        Extract the data
+        Plot the barplot
+        """
+        data = {}
+
+        for f_name, values in self.parse_data.items():
+            s_name = values[0]["fastq1_in_name"].split(".")[0]
+            database = os.path.basename(values[0]["index"])
+            data[s_name] = {"Cleaned reads": values[0]["reads_out"], "Host reads": values[0]["reads_removed"]}
```
vladsavelyev, line 81: Should `reads_out` and `reads_removed` sum up to `reads_in`? Doesn't look like in the test data :

```
cat test_data/data/modules/hostile/hostile.SAMPLE-A1.json
[
    {
        "version": "1.1.0",
        "aligner": "bowtie2",
        "index": "human-t2t-hla",
        "options": [],
        "fastq1_in_name": "SAMPLE-A1.fastq.gz",
        "fastq1_in_path": "SAMPLE-A1.fastq.gz",
        "fastq1_out_name": "SAMPLE-A1.clean.fastq.gz",
        "fastq1_out_path": "SAMPLE-A1.clean.fastq.gz",
        "reads_in": 241805,
        "reads_out": 234757,
        "reads_removed": 70484,
        "reads_removed_proportion": 0.2915
    }
]
```

Bar plots assume categories do not overlapp

multiqc/modules/hostile/hostile.py
```diff
@@ -0,0 +1,98 @@
...
+            info="Short and Long host reads removal tools.",
+            doi="10.1093/bioinformatics/btad728",
+        )
+
+        self.parse_data = dict()
+        for f in self.find_log_files("hostile", filehandles=True):
+            self.parse_logs(f)
+
+        # add version
+        self.add_software_version(self.version, f)
```
vladsavelyev, line 30: The second argument should be a sample name (a string). The CI is failing because of that. I know it can be confusing because `add_data_source` takes the `f` dict

multiqc/modules/hostile/hostile.py
SumeetTiwari07: Hostile version <1.0.0, the hostile reports (JSON) do not have the version information inside the reports. Hence, included a fixed version. However, I changed it to "None".

multiqc/modules/hostile/hostile.py
SumeetTiwari07:
```
for f_name, values in self.parse_data.items():
    database = os.path.basename(values[0]["index"])
    data[f_name] = {"Cleaned reads": values[0]["reads_out"], "Host reads": values[0]["reads_removed"]}
```
Removed that line.

multiqc/modules/hostile/hostile.py
SumeetTiwari07:
```suggestion
            self.version = "None"
```

multiqc/modules/hostile/hostile.py
```diff
@@ -0,0 +1,98 @@
...
+            parse_data = json.load(json_file["f"])
+
+        except json.JSONDecodeError:
+            log.warning(f"Could not parse JSON file {json_file['f']}")
+            return
+
+        if len(parse_data) > 0:
+            self.add_data_source(json_file)
+
+        s_name = self.clean_s_name(json_file["fn"])
```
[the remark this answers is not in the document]
SumeetTiwari07:
```suggestion
        s_name = self.clean_s_name(json_file["s_name"])
```

multiqc/modules/hostile/hostile.py
SumeetTiwari07: Does it make sense?

multiqc/modules/hostile/hostile.py
SumeetTiwari07: Yes, it should. I will update the count in the reports.

multiqc/modules/hostile/hostile.py
SumeetTiwari07: Done

multiqc/modules/hostile/hostile.py
SumeetTiwari07:
```suggestion
            description=f"This plot shows the number of cleaned reads vs host-reads per sample (DB: Database).",
```

multiqc/modules/hostile/hostile.py
vladsavelyev: We should use real outputs from the tool in the test data. Can you run Hostile to generate some?

vladsavelyev (approved):
I pushed some updates, and with the test example https://github.com/MultiQC/test-data/commit/2eee44829e124747e25cb0f6bb35a58ab3112f3b, we are good to go - merging this PR now.

SumeetTiwari07:
> I pushed some updates, and with the test example [MultiQC/test-data@2eee448](https://github.com/MultiQC/test-data/commit/2eee44829e124747e25cb0f6bb35a58ab3112f3b), we are good to go - merging this PR now.

Thank you so much. I too added some more logs from a metagenomic sampling.

PR #2503, opened 2024-04-24
Faster LZ compression by code refactoring and a faster bit by bit writer.

rhpvorderman:

This PR does the following. 
- Remove all encode methods except base64 for lzstring. It is the only encoding used by MultiQC.
- Refactor the lz encoding function to return bytes. Any other encodings can be added later, such as base64 encoding with the    base64 module.
- Remove duplicated for-loops in the lz encoding functions that were meant to write a bitstream. Create a separate bitwriter class and call its write function instead. 
- Optimize the bitwriter to write only 64-bit chunks of data and use python tricks to only require two python function calls for the bit reversal. This is in contrast to the old code, which did a bit by bit reversal using shifts. 
- Some minor javascript-esque code was replaced with pythonic code. (For example, not using indexes to iterate over sequences).
- EDIT: Add the context state variables to the bitwriter, to further simplify the code.

For the multiqc-test-data report this has a significant impact on speed (40% reduction in runtime) on my PC. EDIT: here are the times as reported by multiqc. Before: `5.10s: Compressing report data`, after: `2.64s: Compressing report data`.

The remaining runtime is mainly the dictionary lookup. The LZ compression stores a huge dictionary, and accessing such a huge dictionary is very slow, as it does not fit in the CPU cache. This cannot be optimized without changing the algorithm.

NOTE: I also made #2504 , because gzip is much superior for the use case. Except that I cannot get it to work with the javascript. So I made this instead.

### DIFF
--- a/multiqc/utils/lzstring.py
+++ b/multiqc/utils/lzstring.py
@@ -44,20 +44,28 @@ assert REVERSE_LOOKUP[0b10000000] == 0b00000001
 assert REVERSE_LOOKUP[0b10101010] == 0b01010101
 
 
-class BitWriter:
+class ContextWriter:
     def __init__(self):
         self.bitbuffer = 0
         self.bits_stored = 0
         self.buffer = io.BytesIO()
+        self.context_enlarge_in = 2  # Compensate for the first entry which should not count
+        self.context_numBits = 2
 
-    def write(self, value: int, bits: int):
+    def write(self, value: int, bits: int = 0):
         # Ensure no rogue bits are set after the bits of interest
+        bits = bits or self.context_numBits
         bitmask = (1 << bits) - 1
         value &= bitmask
 
         # Python can store integers bigger than 64 bits. So no checks for overflow.
         self.bitbuffer |= value << self.bits_stored
         self.bits_stored += bits
+
+        self.context_enlarge_in -= 1
+        if self.context_enlarge_in == 0:
+            self.context_enlarge_in = 1 << self.context_numBits
+            self.context_numBits += 1
 
         if self.bits_stored >= 64:
             integer_to_store = self.bitbuffer & 0xFFFF_FFFF_FFFF_FFFF
@@ -99,10 +107,8 @@ def _compress(uncompressed):
     context_c = ""
     context_wc = ""
     context_w = ""
-    context_enlargeIn = 2  # Compensate for the first entry which should not count
     context_dictSize = 3
-    context_numBits = 2
-    context_data = BitWriter()
+    context_data = ContextWriter()
 
     for context_c in uncompressed:
         if context_c not in context_dictionary:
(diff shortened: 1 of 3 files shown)


rhpvorderman:
I added a small commit that moves the context state into the BitWriter, transforming it into a context writer. This further removed some remaining duplicated code segments.

vladsavelyev:
Thanks for this PR @rhpvorderman,

Just for context, the lz-string module was copied from https://github.com/gkovacs/lz-string-python with a tiny change to fix Python 3.12 compatibility. Since the author no longer maintains it and upstream PRs are not merged, we had to copy code with a reference to the original repo.

We were hesitant to diverge more from upstream, because this way, we can treat the code as a library and refer upstream in case of any bugs. 

But since you significantly stripped the code to the essentials, so I think it's okay to move it into the MultiQC core. It needs tests though, as I'm worried a bit about all these binary operations. And we still need to keep a comment referencing the upstream file where it was originally implemented, with explanations about what was changed and why.

I'll take a look into that, will ask you questions if something is unclear :-)

Thank you again for this work and detailed and clear PR!

rhpvorderman:
> We were hesitant to diverge more from upstream, because this way, we can treat the code as a library and refer upstream in case of any bugs.
> But since you significantly stripped the code to the essentials, so I think it's okay to move it into the MultiQC core. It needs tests though, as I'm worried a bit about all these binary operations. And we still need to keep a comment referencing the upstream file where it was originally implemented, with explanations about what was changed and why.

Yes, I thought that was the case. I kept the decode function exactly the same so it could be used for testing. The two lz-compress and lz-decompress scripts were used for that when I was refactoring. So long as lz-decompress still functioned, I knew my lz-compress function was still good. I also tested with md5sums and the results were bit for bit the same. For unit testing I suggest simply compressing and decompressing and check if the input matches the output.

> I'll take a look into that, will ask you questions if something is unclear :-)

I could also write a few tests if you would like. Before putting more efforts into this PR, I wonder if you could take a look at #2504 ? If you can get the javascript to work with the async function, it will be a massive improvement in speed and memory over this lz-compression whils also using much less code.

vladsavelyev:
> I could also write a few tests if you would like. Before putting more efforts into this PR, I wonder if you could take a look at https://github.com/MultiQC/MultiQC/pull/2504 ? If you can get the javascript to work with the async function, it will be a massive improvement in speed and memory over this lz-compression whils also using much less code.

I should have indeed looked into that PR before commenting here 😅  I guess we can now just ditch lz-string in favour of gzip.

rhpvorderman:
Hoorray! That is the best outcome for sure! Despite the relatively big size of this PR, I spent much more time on the gzip PR as it is a technically better solution on multiple aspects. I just couldn't get it to work. I will close this with much happiness!

PR #2504, opened 2024-04-24
Use gzip rather than lzstring for compression and decompression of the plot data

rhpvorderman:

WARNING: This PR does NOT work (yet).

This PR does the following:

- remove the lz-string code 
- Use gzip + base64 to encode the plot data, using python builtin modules.
- Create a simple function to decompress the plot data in Javascript, using javascript builtin modules.

Advantages:
- compression ~slightly~ much better than the lz-string algorithm up to two times more compression.
- Requires much less memory, as gzip only uses a 32 KB window size. The lz-string dictionary uses 200 Mib for the multiqc test-data, so this is roughly a 1000x improvement!
- MUCH faster when level 6 is used (10x). Level 9 is 5 times faster.
- Purges a lot of code from the repo by using builtin libraries. 

But here comes the catch: JavaScript decompression streams can only be used in async routines as they return promises rather than actual data. There is no way around this. This is by design. I have scoured the internet to find a way, and there is no way. 

So I tried to make it work with the MultiQC javascript, unfortunately it did not work. But I figure you, the MultiQC maintainers, will be much more proficient in working with JavaScript. (My experience with the language is measured in hours rather than years.) So I am dropping this PR here, and you can try to get it to work if you so fancy.

I gave up and hence I did the PR #2503 .


vladsavelyev:
Oh, that's actually even better than fixing lz-string! Happy to take a look at the JS part

rhpvorderman:
Gzip rocks! So ubiquitous that all browsers support it. Thank you for taking a look at it!

multiqc/templates/default/assets/js/plotting.js
```diff
@@ -143,12 +143,12 @@ function initPlot(dump) {
 }
 
 // Execute when page load has finished loading
-$(function () {
```
rhpvorderman, line 146: This is the part that does not work properly. The `await` below is needed, but using the `async` keyword here does not magically make everything work.

vladsavelyev:
I looked up the DecompressionStream class, and it seems pretty new. Since MultiQC is often used in weird browsers, we want to be a bit conservative here, so I replaced it with a third-party library [pako](https://github.com/nodeca/pako) to decompress. Maybe I'm over cautious 🤔 

But it works now!

(I removed `async` from the main function: we don't use reactive JavaScript here, so async wouldn't work properly anyway. But long term, we might want to take a look into decompressing asyncrounisly)

vladsavelyev (approved):
This is really great! Really appreciate the contribution. I'll test a bit more and merge.

multiqc/utils/report.py
```diff
@@ -575,11 +574,14 @@ def save_htmlid(html_id, skiplint=False):
...
 def compress_json(data):
-    """Take a Python data object. Convert to JSON and compress using lzstring"""
+    """Take a Python data object. Convert to JSON and compress using gzip.
+    Represent in base64 format."""
     json_string = json.dumps(data).encode("utf-8", "ignore").decode("utf-8")
     json_string = sanitise_json(json_string)
-    x = lzstring.LZString()
-    return x.compressToBase64(json_string)
+    json_bytes = json_string.encode("utf-8")
+    json_gzip = gzip.compress(json_bytes, compresslevel=9, mtime=0)
```
vladsavelyev, line 582: Could you add a comment in the code on why the `compresslevel` is set to the max? 

I guess there is a trade-off between the report size and the speed of loading, so wonder if you tried it with different levels? :)

ewels:
@vladsavelyev I don't see the Pako library bundled in this PR. Did you push it?

Looking at caniuse.com it doesn't seem too bad. Basically just our old friend internet explorer that's a problem. Based on the usage stats I think it'd be < 1% people affected? (I'm on my phone so correct me if I'm wrong).

I'm not against adding new libraries but it's nice to avoid if possible. Does Pako have any down sides except us needing to bundle its source within the report?

vladsavelyev:
> Pako library bundled in this PR. Did you push it?

It's included, GitHub just doesn't show it expanded by default in diff. It's 48K in size vs the 8.0K of removed lz-string. Not critical, but not ideal either.

> Looking at caniuse.com it doesn't seem too bad. Basically just our old friend internet explorer that's a problem. Based on the usage stats I think it'd be < 1% people affected? (I'm on my phone so correct me if I'm wrong).

All modern browsers support it (IE is just the ancient version of Edge), though only recently: e.g., Safari started supporting it in Mar 2023. Not a problem for standard user experience as modern browsers auto-update, but I'm thinking about odd use cases, like a browser installed on an old Slurm cluster that the user connects to via X-forwarding to view their workflow results 😅 We can decide to ignore that and go full on modern.

> Does Pako have any down sides except us needing to bundle its source within the report?

Yep, just the extra 48KB

multiqc/utils/report.py
vladsavelyev: Doesn't make much difference, but I'll use all the default args to gzip.compress. Just pushed directly into the PR if that's okay

rhpvorderman:
Wow great. I am glad that this will find its way in the new version of MultiQC! One disadvantage over using pako is that it is zlib implemented in javascript rather than proper zlib. Using the builtin decompression streams probably use that faster C implementation, and also don't need an included library. On the other hand, that was hard to get to work, so I am glad this solution was chosen! Looking forward to be able to use this!

ewels:
> Safari started supporting it in Mar 2023.

Oof, didn't notice that. Always Safari :shakefist:

> Using the builtin decompression streams probably use that faster C implementation

Probably the most technically difficult solution, but what we could do is to test if `DecompressionStream` is available to use in the browser and use it if so. But have pako to fall back on if not. Then we get the best possible performance and best possible backwards compatability. Then in a few years we can strip out pako 👴🏻 

It shouldn't add anything to the amount of bundled code as we're not adding a second library. It'll add a little bit of complexity / maintenance burden though. Not sure how much of a performance difference we're talking.

ewels:
> It's 48K in size

Looks like there's a 27K dist file which does only deflate: [`dist/pako_deflate.min.js`](https://github.com/nodeca/pako/blob/master/dist/pako_deflate.min.js)

rhpvorderman:
You'd want the inflate one in this case. Deflating is compressing, inflating is decompressing. This is a bit confusing indeed.

The inflate one is only 21 kb. So that's even better.

ewels:
A deflating balloon gets smaller, makes sense 😅 (even if an inflated balloon has compressed air within 😵 )

vladsavelyev:
Yeah, we can use DecompressionStream and fall back to pako. I tried to do some benchmarking, but it's tricky because of the async nature of DecompressionStream: the runtime inserts a lot of tasks inbetween making the decompression time appear longer. But it makes sense that the native approach should be faster.

ewels:
I think that the async nature of it should lead to a better user experience hopefully. On really huge reports the browser basically hangs for several seconds whilst the data is decompressed. I wonder if the async code will avoid this?

Not totally sure if that's the case though. _May_ have to move that to web workers to run in multiple threads to avoid the browser lock? Not totally sure.

PR #2505, opened 2024-04-24
Refactor file content searching

rhpvorderman:

This report does the following:

- Create a new SearchFile class that contains a line_iterator function.
  - This iterator function caches the filehandle used to read the file.
  - SearchFile is a context manager, and closes the file on exit. 

The previous code would reuse the content lines as much as possible, but would read the file again when not enough content lines were read already. By using a object, the existing filehandle can be kept alive. This prevents reopening and rereading. 

Also by moving all the line reading code into a separate function, the resulting code in `search_file` has become more readable in my opinion. 

The speed uplift is very significant. Tested with ` multiqc --force --no-data-dir --profile-runtime ../test-data/` (The multiqc test-data repo)

Before:
```
|           multiqc | Run took 24.66 seconds
|           multiqc |  - 9.70s: Searching files
|           multiqc |  - 8.40s: Running modules
|           multiqc |  - 5.07s: Compressing report data
```
After:
```
|           multiqc | Run took 20.25 seconds
|           multiqc |  - 5.12s: Searching files
|           multiqc |  - 8.44s: Running modules
|           multiqc |  - 5.11s: Compressing report data
```

A more than 40% reduction in searching files is very welcome. This was benchmarked on a nvme SSD, but I regularly work in a Network Filesystem which is orders of magnitude slower. These speed increases become very significant in such environments.

### DIFF
--- a/multiqc/utils/report.py
+++ b/multiqc/utils/report.py
@@ -108,6 +108,94 @@ def init():
     software_versions = defaultdict(lambda: defaultdict(list))
 
 
+class SearchFile:
+    filename: str
+    root: str
+    path: str
+    _filesize: Optional[int]
+
+    def __init__(self, filename: str, root: str):
+        self.filename = filename
+        self.root = root
+        self.path = os.path.join(root, filename)
+        self._filesize = None
+        self.content_lines = []
+        self._filehandle = None
+
+    @property
+    def filesize(self) -> Optional[int]:
+        if self._filesize is None:
+            try:
+                self._filesize = os.path.getsize(self.path)
+            except (IOError, OSError, ValueError, UnicodeDecodeError):
+                logger.debug(f"Couldn't read file when checking filesize: {self.filename}")
+                self._filesize = None
+        return self._filesize
+
+    def line_iterator(self):
+        for line in self.content_lines:
+            yield line
+        if self._filehandle is None:
+            try:
+                self._filehandle = io.open(self.path, "rt", encoding="utf-8")
+            except Exception as e:
+                if config.report_readerrors:
+                    logger.debug(f"Couldn't read file when looking for output: {self.path}, {e}")
+        try:
+            for line in self._filehandle:
+                self.content_lines.append(line)
+                yield line
+
+        except UnicodeDecodeError as e:
+            if config.report_readerrors:
+                logger.debug(
+                    f"Couldn't read file as a utf-8 text when looking for output: {self.path}, {e}. "
+                    f"Usually because it's a binary file. But sometimes there are single non-unicode "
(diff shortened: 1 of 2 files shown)


vladsavelyev:
This is amazing!! I'll look into merging this today

multiqc/utils/report.py
```diff
@@ -455,30 +490,29 @@ def exclude_file(sp, f):
...
-            for line in fh:
-                if "exclude_contents" in sp:
-                    for pat in sp["exclude_contents"]:
-                        if pat in line:
-                            return True
-                if "exclude_contents_re" in sp:
-                    for pat in sp["exclude_contents_re"]:
-                        if re.search(pat, line):
-                            return True
+        for line in f.line_iterator():
```
vladsavelyev, line 507: It appears to be a second call of `line_iterator` within the same `with` call, so it would probably read from the place where `search_file` has left it?

vladsavelyev:
This is a really amazing refactoring, thanks so much for this! Love the use of the lazy `line_iterator` method, and encapsulation into the `SearchFile` class. And I always wished to see the search file as a class to enable static type checking and autocompletion. 

Should probably go further and try to put `SearchFile` into `report.files` and expose to `find_log_files()` in all modules. Add type hints, replace the dict keying with object attribute access, and use the `path` property and the line iterator whenever it makes sense. Happy to give it a try in another PR.

rhpvorderman:
> Should probably go further and try to put SearchFile into report.files and expose to find_log_files() in all modules. Add type hints, replace the dict keying with object attribute access, and use the path property and the line iterator whenever it makes sense. Happy to give it a try in another PR.

Before you do that, I have another PR in the waiting that depends on this PR and makes the file searching another 20% faster. I will open it as soon as this is merged.

multiqc/utils/report.py
vladsavelyev: Right, I mixed two distinct context managers here - the FileSize object that wraps the file handler, and the line_iterator that wraps content_lines 👍

vladsavelyev: Added some docstring comments to clarify that, mainly for myself in the future 😅

vladsavelyev (approved):
Looks great! Thanks a lot for the clear and very pythonic code. Very excited for future optimizations.

By the way, feel free to create stacked PRs if if you have more cool stuff downstream of an existing unmergeed PR - i.e. when `pr_a` is based from `main`, `pr_b` is based from `pr_a`, etc. Even if it's a draft. Good to anticipate more awesome stuff so I don't accidentally create unnecessary merge conflicts 👍

PR #2506, opened 2024-04-24
Sort filepatterns for faster searching

rhpvorderman:

Search patterns are already divided into categories to ensure the simplest to match patterns are run first. 

I noticed that the "*.json" files run particularly long. It turns out that that `afterqc` has 10_000 `num_lines`. Sequali, fastp, and cutadapt have 10, 50 and 100 `num_lines` respectively. It seems only fair that these are run first.

Speed as reported by multiqc:
before on the test-data dir: `9.90s: Searching files`
after on the test-data dir: `9.47s: Searching files`
before on a dir with 800 sequali only reports: `3.48s: Searching files`
after on a dir with 800 sequali only reports: `0.38s: Searching files`

As the sequali author, I would be very happy if this gets merged ;-). Also on very slow filesystems such as NFS the differences would be more significant. 

I know a posted a lot of PRs today, but it seemed better to have these all separately evaluated. This is the last of that series.

### DIFF
--- a/multiqc/utils/report.py
+++ b/multiqc/utils/report.py
@@ -4,6 +4,7 @@ helper functions to generate markup for report. """
 
 
 import fnmatch
+import functools
 import inspect
 import io
 import json
@@ -219,6 +220,45 @@ def get_filelist(run_module_names):
                 spatterns[3][key] = sps
         else:
             spatterns[0][key] = sps
+
+        # sort patterns for faster access. File searches with less lines or
+        # smaller filesizes go first.
+        def max_on_pattern_key(x, pattern_key=""):
+            key, sps = x
+            num_lines = 0
+            for sp in sps:
+                num_lines = max(num_lines, sp.get(pattern_key, 0))
+            return num_lines
+
+        new_spatterns = [{}, {}, {}, {}, {}, {}, {}]
+        new_spatterns[0] = spatterns[0]  # Only filename matching
+        new_spatterns[1] = dict(
+            sorted(
+                ((key, sps) for key, sps in spatterns[1].items()),
+                key=functools.partial(max_on_pattern_key, pattern_key="num_lines"),
+            )
+        )
+        new_spatterns[2] = dict(
+            sorted(
+                ((key, sps) for key, sps in spatterns[2].items()),
+                key=functools.partial(max_on_pattern_key, pattern_key="max_filesize"),
+            )
+        )
+        new_spatterns[3] = spatterns[3]
+        new_spatterns[4] = dict(
+            sorted(
+                ((key, sps) for key, sps in spatterns[4].items()),
+                key=functools.partial(max_on_pattern_key, pattern_key="num_lines"),
+            )
+        )
+        new_spatterns[5] = dict(
+            sorted(
+                ((key, sps) for key, sps in spatterns[5].items()),
+                key=functools.partial(max_on_pattern_key, pattern_key="max_filesize"),
+            )
+        )
+        new_spatterns[6] = spatterns[6]
+        spatterns = new_spatterns
 
     if len(ignored_patterns) > 0:
         logger.debug(f"Ignored {len(ignored_patterns)} search patterns as didn't match running modules.")


vladsavelyev:
Thank you a lot, this makes a lot of sense. I'll try to merge your PRs today!

multiqc/utils/report.py
```diff
@@ -220,6 +221,45 @@ def get_filelist(run_module_names):
...
+            )
+        )
+        new_spatterns[5] = dict(
+            sorted(
+                ((key, sps) for key, sps in spatterns[5].items()),
+                key=functools.partial(max_on_pattern_key, pattern_key="max_filesize"),
+            )
+        )
+        new_spatterns[6] = spatterns[6]
+        spatterns = new_spatterns
```
vladsavelyev, line 261: This block is inside the `for key, sps in config.sp.items()` loop, so it runs for every search pattern. Wonder if we could move it after the loop?

multiqc/utils/report.py
```diff
@@ -220,6 +221,45 @@ def get_filelist(run_module_names):
...
+            sorted(
+                ((key, sps) for key, sps in spatterns[2].items()),
+                key=functools.partial(max_on_pattern_key, pattern_key="max_filesize"),
+            )
+        )
+        new_spatterns[3] = spatterns[3]
+        new_spatterns[4] = dict(
+            sorted(
+                ((key, sps) for key, sps in spatterns[4].items()),
+                key=functools.partial(max_on_pattern_key, pattern_key="num_lines"),
```
vladsavelyev, line 251: I find `functools.partial` not very readable, wonder if we could refactor it without it?

multiqc/utils/report.py
vladsavelyev: I think `key=lambda x: max_on_pattern_key(x, pattern_key="num_lines")` would work the same?

multiqc/utils/report.py
vladsavelyev: I think `key=lambda x: max_on_pattern_key(x, pattern_key="num_lines")` would work the same? Just using closure instead.

multiqc/utils/report.py
vladsavelyev, line 261: What do you think about doing it like this?

```suggestion
    def _sort_by_key(sps: Dict[str, List[Dict[str, str]]], key: str) -> Dict:
        """Sort search patterns dict by key like num_lines or max_filesize."""

        def _sort_key(kv) -> int:
            _, sps = kv
            # Since modules can have multiple search patterns, we sort by the slowest one:
            return max([sp.get(key, 0) for sp in sps])

        return dict(sorted(sps.items(), key=_sort_key))

    # Sort patterns for faster access. File searches with fewer lines or
    # smaller file sizes go first.
    sorted_spatterns = [{}, {}, {}, {}, {}, {}, {}]
    sorted_spatterns[0] = spatterns[0]  # Only filename matching
    sorted_spatterns[1] = _sort_by_key(spatterns[1], "num_lines")
    sorted_spatterns[2] = _sort_by_key(spatterns[2], "max_filesize")
    sorted_spatterns[3] = spatterns[3]
    sorted_spatterns[4] = _sort_by_key(spatterns[4], "num_lines")
    sorted_spatterns[5] = _sort_by_key(spatterns[5], "max_filesize")
    sorted_spatterns[6] = spatterns[6]
    spatterns = sorted_spatterns
```

Just slightly more compact

vladsavelyev (approved):
I'm just going to merge. Thank you a lot for this idea and for the PR!

vladsavelyev:
> I agree with all your comments, but since it is merged already, I will just let the code rest, if that's okay. Thanks for merging!

Sorry if I merged too quickly, I'm so excited about the optimizations so want all of them in `main` as fast as possible 😄 

If you find anything suboptimal or want to add or refactor, feel free to create even trivial PRs and I'll merge quickly, @rhpvorderman

rhpvorderman:
Well, I was happy with it. Those were your comments after all :wink:. Thanks for all the quick merges!

vladsavelyev:
> Those were your comments after all

I meant to merge the proposed change before merging the PR 😓  I'll create another one https://github.com/MultiQC/MultiQC/pull/2512

PR #2509, opened 2024-04-26
Speed up all modules by tackling two performance bottlenecks

rhpvorderman:

I used the py-spy profiler. My previous PRs did not focus on the module code, as it looked like as if modules were slow or fast on their own accord. However I realized that some common code must be used because of the class inheritance. I did a more careful look at the profile and found a real stinker in the lineplot code that was triggered in multiple modules.

In a dict comprehension it was checked if something was a member of a list. List member checking time is proportional to the length of the list. I converted the list to a set. This one liner fix made a run on multiqc/test-data noticably faster. 

The other one I found was the `spectra.scale` call from the plotly library was expensive. I did some investigation of the input data and found that only 34 unique lists of colours were used, so I used a functools.lru_cache to simply cache the call. This also removed a further 10% of module runtime (on top of the previous savings). 

These are really minor changes codewise but have quite some benefit:

before:
```
|           multiqc |  - 9.59s: Searching files
|           multiqc |  - 8.41s: Running modules
|           multiqc |  - 5.04s: Compressing report data
```

After:
```
|           multiqc |  - 9.64s: Searching files
|           multiqc |  - 6.48s: Running modules
|           multiqc |  - 4.99s: Compressing report data
```

So a roughly 20 decrease in runtime for all module code.

### DIFF
--- a/multiqc/plots/linegraph.py
+++ b/multiqc/plots/linegraph.py
@@ -218,7 +218,7 @@ def smooth_line_data(data: Dict[str, Dict], numpoints: int) -> Dict[str, Dict[in
             continue
 
         binsize = (len(d) - 1) / (numpoints - 1)
-        first_element_indices = [round(binsize * i) for i in range(numpoints)]
+        first_element_indices = {round(binsize * i) for i in range(numpoints)}
         smoothed_d = {x: y for i, (x, y) in enumerate(d.items()) if i in first_element_indices}
         smoothed_data[s_name] = smoothed_d
 
--- a/multiqc/utils/mqc_colour.py
+++ b/multiqc/utils/mqc_colour.py
@@ -2,11 +2,13 @@
 Helper functions to manipulate colours and colour scales
 """
 
+import functools
 import hashlib
 
 # Default logger will be replaced by caller
 import logging
 import re
+from typing import Tuple
 
 import numpy as np
 import spectra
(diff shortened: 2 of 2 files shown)


vladsavelyev (approved):
Wow, that's really impressive! A few lines changed for like 20% boost. I'll merge this right away.

Thanks a lot for profiling and digging into and making such comprehensive and clean PRs!

rhpvorderman:
Your welcome! We love to use MultiQC at our institute so I am simply trying to return in kind all the benefits you bring to us!

ewels:
@rhpvorderman these PRs are fantastic, thank you!

When you multiply these speed ups by the number of runs that they will affect, it's a huge impact 👏🏻

rhpvorderman:
The power to use profilers and change a few lines is insignficant next to the power of aggregating thousands of reports with MultiQC! I cannot thank you enough for making this great tool!

My main motivation is that MultiQC is so successful now and supports so many reports that scalability becomes a concern. Memory and runtimes tend to go up very steeply with increasing numbers of reports and this can lead to timeouts and out of memory errors on our SLURM cluster. Sure, more resources can be allocated, but the less the better. Also I have written a (very fast) FastQC/nanoplot replacement with Sequali, and I see MultiQC integration as a core feature. As such, if I improve the MultiQC user experience, I also improve that of Sequali users. It's a win-win basically.

ewels:
Well, the more you can send our way the better! I have had a loose task of _"figure out how to use profilers properly and improve MultiQC performance"_ on my backlog for years, but it's one of those tasks that always gets buried because nothing's really actively broken or on fire or anything.

But totally agree that it's become more of an issue recently, also as sample numbers go up (looking at you, single cell data).

PR #2513, opened 2024-04-29
Search file blocks rather than individual lines for faster results

rhpvorderman:

Searching file content is the most costly part of file searching for MultiQC. Currently all lines are searched individually and all lines are cached. 

However, this approach is problematic from a performance standpoint because of the overhead involved. Python iteration is calling the `__next__` function under the hood, the `content in line` calls `__contains__` under the hood. Python function calls are relatively expensive and this cost is payed for each line. Furthermore, string searching algorithms create a search structure first, and then iterate over the string using the structure to find matches. This preparation is done over and over again for each line. 

So rather than reading individual lines, this PR reads blocks of 4 KB from the file. To make sure the per-line-content matching is not broken, we use python's count and rfind method to ensure a block only contains full lines. 

The blocks are cached the same way as on the line iterator. 

This method may perform worse in the following cases:
- The content is right at the beginning of the file. Hence 4kb - the line size was processed too much.
- The line sizes are regularly bigger than 4kb. 

In the first case, this is not too problematic because said 4kb will be in the CPU L1 cache, and executing the count and rfind methods will be very fast. The second case is simply not likely, and optimization happens for the common case, not the edge case. 

EDIT: I forgot to mention that IO-wise, while more data is read at once, this does not make a change. Python buffers callls under the hood and already reads io.DEFAULT_BUFFER_SIZE of data from the file. This is 8 kb of raw data. Since we only read 4096 utf-8 characters, most of which will be one byte, there should be only one call to the raw file read.

The tradeoff is slightly more complex code but better runtimes. 

Before:
```
|           multiqc | Run took 24.08 seconds
|           multiqc |  - 6.65s: Searching files
|           multiqc |  - 7.96s: Running modules
|           multiqc |  - 7.25s: Compressing report data
```

After:
```
|           multiqc | Run took 22.49 seconds
|           multiqc |  - 5.47s: Searching files
|           multiqc |  - 7.98s: Running modules
|           multiqc |  - 6.82s: Compressing report data
```


rhpvorderman:
@vladsavelyev Here it is! This is the last PR I have in the pipeline. I do plan to take a look at the MultiQC memory consumption. I have some ideas, but I have to see if they are applicable.

vladsavelyev:
That's really awesome, I love the approach. And thanks for anticipating the edge cases.

In your benchmarking, I wonder why the compression is waster?

rhpvorderman:
> In your benchmarking, I wonder why the compression is faster?

Run to run variance. The LZcompression does a lot of memory allocations (due to creating a massive dict). These are variably expensive. 
Should be solved with the gzip refactor.

vladsavelyev:
Was wondering about the file search time - so we shouldn't expect similar run to run variance for it?

> Should be solved with the gzip refactor

Just merged it!

vladsavelyev:
While testing, realized that https://github.com/MultiQC/test-data is not a good source of data to benchmark, as it only contains small toy examples.

I'm now thinking of creating a separate test bundle with some large real-life files to stress-test and benchmark MultiQC consistently. If you happen to have any such files handy, could you please share either in this issue, or to vladislav.sav@gmai.com?

(A bit unrelated, but we are also looking into improving the report performance for large datasets (https://github.com/MultiQC/MultiQC/issues/2280), so any huge examples would be great to have to stress-test that as well)

vladsavelyev:
Tried on a larger dataset - the difference is really impressive!

Before:

```
25.55s: Searching files
```

After:

```
13.65s: Searching files
```

rhpvorderman:
Wow!
What kind of larger dataset do you have available? I would love to use that to reduce the multiqc memory usage. Can you share it by any chance?

vladsavelyev:
@rhpvorderman, here you go: https://drive.google.com/file/d/1nNVhzSnKk8wLSuMfqjFjn-xbSIUHeTzB/view

rhpvorderman:
Thanks. Requested access just now!

FelixKrueger:
Access granted. Enjoy.

vladsavelyev (approved):

vladsavelyev:
Time to merge 🚀

ewels:
> The line sizes are regularly bigger than 4kb.

This isn't impossible - some files generate minified JSON outputs for example, where the entire file is a single line.

But the search should still work in these cases, right? As long as it doesn't completely break I'm happy. There are not many of them.

You'll probably ask me which tools do this and I can't remember off the top of my head sorry.

rhpvorderman:
> But the search should still work in these cases, right? As long as it doesn't completely break I'm happy. There are not many of them.

Yes it will work just fine. It will only be suboptimal in a performance sense compared to letting python find the lines by using __next__ or readline. But for any case where the lines are less than 4096 bytes, the performance will be much better. So it is a tradeoff.

PR #2529, opened 2024-05-03
Do replace_nan in place rather than creating a new object

rhpvorderman:

Rather than creating new dictionaries and lists with comprehensions, do the NaN replacement in place. Also, nan checking is also inlined multiple times to avoid unnecessary function calls. This makes the function faster, allthough the effect on total runtime performance is not that big. (Roughly half a second saving). However this also does save additional memory, about 30 MiB. (All numbers based on the production data set).


vladsavelyev (approved):
Thank you! Looks great 👍

PR #2544, opened 2024-05-08
Fix error when using default sort

rhpvorderman:

I am currently working on the sequali module and wanted to use sort. And then MultiQC crashed. Apparently sequali is the first user of defaultsort. (Cannot find other modules using it).

This is probably due to a refactor. header is not a dict, but it is an object (much better!) but this was forgotten in the refactor because no tool used it (yet).


vladsavelyev:
Thanks for spotting this 🙏  Yeah, got to review carefully if I missed any spots like this before making a release.

vladsavelyev (approved):

PR #2545, opened 2024-05-08
Sequali: allow paired-end data.

rhpvorderman:

Hi, since Sequali 0.8.0 it supports paired-end data. So naturally the MultiQC module also has to support this. I released 0.8.0 just now, but I developed this PR alongside. 

Could this still be included in 1.22? Pretty please with cherries on top! :pray: :pray: :pray:

I added the test data in this pr: https://github.com/MultiQC/test-data/pull/325

### DIFF
--- a/multiqc/modules/sequali/sequali.py
+++ b/multiqc/modules/sequali/sequali.py
@@ -429,7 +429,7 @@ class MultiqcModule(BaseMultiqcModule):
             plot_data = {}
             for sample_name, sample_dict in data.items():
                 gc_dict = sample_dict.get(key)
-                if sample_dict is None:
+                if gc_dict is None:
                     continue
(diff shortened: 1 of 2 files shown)


vladsavelyev:
Thanks, looks great to me!

A bit unfortunate to have repeated plots for Read 1 and Read 2, makes the report a bit messy. But using a tab switch wouldn't be ideal either, as read 2 is equally important as read 1, and showing read 1 by default doesn't make much sense. Long term, would be nice to show read 1 and read 2 side by side, but there is no built-in functionality for that yet.

vladsavelyev (approved):

rhpvorderman:
Wow! I didn't expect this to be merged so soon. Thanks a lot! I can sleep tight, knowing that sequali will be fully supported from v1.22 onwards. This means a lot to me. Thanks.

PR #2552, opened 2024-05-10
FastQC: Skip per tile sequence quality section for better performance

rhpvorderman:

This is not really good for me personally, as I made a FastQC replacement: [Sequali](https://github.com/rhpvorderman/sequali). People should be using that tool instead of FastQC. Sequali accepts paired-end data, it is much faster, it correctly averages Phred Scores and does deliver nicer plots. 

Okay with that of my chest: the FastQC module. It turns out that the high memory usage is because of the per tile quality data. This was exactly the same with Sequali. So in the Sequali module, those data are removed as they weren't used. After some inspection I found that the FastQC module has the exact same problem. Not saving all this data saves a whopping **2.2 GB** on the profile data run. Bringing the memory from 3 GiB to 800 MiB. This is quite a massive improvement.

Also converting all those float values from text to float data takes a very long time. As a consequence the FastQC module runs twice as fast. This brings it execution time more in line with other modules.

### DIFF
--- a/multiqc/modules/fastqc/fastqc.py
+++ b/multiqc/modules/fastqc/fastqc.py
@@ -176,6 +176,9 @@ class MultiqcModule(BaseMultiqcModule):
                 (section, status) = line[2:].split("\t", 1)
                 section = section.lower().replace(" ", "_")
                 self.fastqc_data[s_name]["statuses"][section] = status
+            elif section == "per_tile_sequence_quality":
+                # Lots and lots of data. None of it used in MultiQC.
+                continue
             elif section is not None:
                 if line.startswith("#"):
                     s_headers = line[1:].split("\t")


vladsavelyev:
This is a really fantastic find! I was by far the biggest culprit in the entire codebase.

Regarding Sequali, that's a good point. All of our [examples](https://multiqc.info/example-reports) use FastQC - maybe we should consider switching a few of them to Sequali 🤔

rhpvorderman:
Well sequali is mostly not known by anybody yet. It is mostly used in our own institute as of yet. It is currently a zero-star, no one is watching github project. So there might be corner cases that it does not handle well, because we have never run into those. Having said that, I encourage to give it a try.

TBradley27:
Hey,

Just asking for a quick clarification, after this PR, does this mean per tile sequence quality information is not reported at all in the multiQC report?

rhpvorderman:
> Just asking for a quick clarification, after this PR, does this mean per tile sequence quality information is not reported at all in the multiQC report?

Yes that is correct. However, as far as I can see, it was also not reported before this PR, either.

TBradley27:
Information relating to this module is reported, see below, from https://multiqc.info/example-reports/whole-genome-seq/:

image

If I have a reasonable amount of memory to play with, losing this information is of no benefit to me

rhpvorderman:
@TBradley27, If you care about not losing information, I do recommend taking a look at [Sequali](https://www.github.com/rhpvorderman/sequali). It fits the same niche as FastQC and creates most of the same plots, but will accept paired-end data and give back an accurate estimate of adapter content based on overlap. FastQC loses this information by just considering reads single-end even if they are part of a pair. The same goes for the duplication estimates.

TBradley27:
Thank you for the heads up. I will check it out

PR #2582, opened 2024-05-20
support DOI for custom content

jchorl:
Adds support for providing a DOI alongside custom content.

Tested with the cc example from multiqc docs, but with added DOI:

```yaml
id: "my_pca_section"
section_name: "PCA Analysis"
description: "This plot shows the first two components from a principal component analysis."
plot_type: "scatter"
pconfig:
  id: "pca_scatter_plot"
  title: "PCA Plot"
  xlab: "PC1"
  ylab: "PC2"
doi: foo/bar
data:
  sample_1: { x: 12, y: 14 }
  sample_2: { x: 8, y: 6 }
  sample_3: { x: 5, y: 11 }
  sample_4: { x: 9, y: 12 }
```

Screenshot 2024-05-20 at 14-42-05 MultiQC Report

I was hoping we could actually just re-use the base-modules DOI logic. Unfortunately the base module's [intro logic](https://github.com/MultiQC/MultiQC/blob/c7f69fab0899a3c4f27363a30bf299b655104862/multiqc/base_module.py#L114) includes the module name, and cc rightfully needs an intro without repeating the module name. I didn't want to start refactoring and provide overrides/hooks to get back to using base modules intro logic, as it'd be more invasive. Possibly a better approach though.

Why would someone want to do this?

- When providing content to multiqc, whether based on internal tooling or external, there may be a citation to attach
- I had a look at multiqc's setup-tools plugin architecture. There's pros/cons. It requires packaging code with multiqc, instead of passing in arbitrary code for multiqc to handle. For common tools or widely reused modules it probably makes a lot of sense. However if you have many fragments of custom-content that are one-offs, building a full module has a higher burden than just throwing some json at multiqc.
- Doesn't seem like there's much downside!

Let me know if y'all would like me to add the case above as a new test-case in multiqc, or alternatively if we should augment an existing case. I also tested without the DOI and it didn't seem to break anything.

### DIFF
--- a/multiqc/modules/custom_content/custom_content.py
+++ b/multiqc/modules/custom_content/custom_content.py
@@ -299,12 +299,12 @@ class MultiqcModule(BaseMultiqcModule):
             href=mod["config"].get("section_href"),
             info=mod_info,
             extra=mod["config"].get("extra"),
-            # No DOI here.. // doi=
+            doi=mod["config"].get("doi"),
         )
 
         # Don't repeat the Custom Content name in the subtext
-        if self.info or self.extra:
-            self.intro = f"<p>{self.info}</p>{self.extra}"
+        if self.info or self.extra or self.doi_link:
+            self.intro = f"<p>{self.info}{self.doi_link}</p>{self.extra}"
 
     def update_init(self, c_id, mod):
         """
@@ -322,7 +322,7 @@ class MultiqcModule(BaseMultiqcModule):
             self.extra = mod["config"].get("extra", None)
         # This needs overwriting again as it has already run on init
         if self.info or self.extra:
-            self.intro = f"<p>{self.info}</p>{self.extra}"
+            self.intro = f"<p>{self.info}{self.doi_link}</p>{self.extra}"
 
     def add_cc_section(self, c_id, mod):
         section_name = mod["config"].get("section_name", c_id.replace("_", " ").title())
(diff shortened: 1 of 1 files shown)


vladsavelyev (approved):
Thank you! Looks good to me

ewels:
> I had a look at multiqc's setup-tools plugin architecture. There's pros/cons. It requires packaging code with multiqc, instead of passing in arbitrary code for multiqc to handle. For common tools or widely reused modules it probably makes a lot of sense. However if you have many fragments of custom-content that are one-offs, building a full module has a higher burden than just throwing some json at multiqc.

Agree with this summary - make sure you take a look at the "MultiQC as a library" stuff that came out in v1.22 though. It's the beginning of refactoring that we're doing and one of the main use cases is to allow people to do custom module stuff without needing a plugin. Might be a more flexible / powerful approach for you.

@vladsavelyev gave a talk with some demos etc last week: https://summit.nextflow.io/2024/boston/agenda/05-23--multiqc-meet-plotly-and-notebooks/

jchorl:
> Ok just saw #2598 - I guess you are aware of it then 😅

Roughly - I do call multiqc as a library in python. Given usage of nextflow, the processes that generate the multiqc inputs aren't the same as the ones calling multiqc. One thing I'd love to do that I haven't had much time to investigate:
```python
from multiqc-types import CustomContent, ContentTypes

# this could have pydantic validation
# this could also run more in-depth validation to make sure the shape of data is correct for the plot type
# maybe there's a better abstraction over content/module types!
my_neato_chart = CustomContent(
    id = "foo",
    type = ContentTypes.BarChart,
    data = {...},
    validate = True,
)

with open("foo_mqc.json", "w") as f:
    json.dump(my_neato_chart, f)  # CustomContent could deal with serialization to json, or some other format
```

Then when calling multiqc, probably just stage the files into the input dir.

This would be a step up over just producing arbitrary JSON. I have no idea if this is possible today or how far away we are. I was hesitant to suggest because I have limited cycles to actually do the implementing. I'll watch vlad's video!

PR #2589, opened 2024-05-28
Use a set for fast membership checking of headers

rhpvorderman:

Hi, I am doing a presentation on MultiQCs recent improvements in our institute, so I figured I'd run the profiler again to check if I missed anything. I found this, but I can't recall this was in 1.22, it must be a later addition. 

List membership checking is expensive, all the items are checked for equality. If a dict or set is used, only the items with colliding hashes are checked for equality. In practice that means one equality check per membership check. If there are lots of items in the list, as is the case with the test data, membership checking takes increasingly long. According to py-spy, this was taking 10% of the entire test-data runtime!

### DIFF
--- a/multiqc/report.py
+++ b/multiqc/report.py
@@ -857,6 +857,7 @@ def write_data_file(
         # noinspection PyBroadException
         try:
             # Get all headers from the data, except if data is a dictionary (i.e. has >1 dimensions)
+            header_set = set()
             headers = []
             rows = []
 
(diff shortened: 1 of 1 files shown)


rhpvorderman:
FYI, here is a very good guide on python datastructures. https://realpython.com/python-data-structures/ 
Membership checking in O(1) time is a property of hash tables which are used in dictionaries and sets.

rhpvorderman:
Strange, I noticed a big effect on my pc, but that might have been a fluke. In that case, sorry.

PR #2594, opened 2024-05-30
copy headers with table

jchorl:
When using the "Copy table" button:
Screenshot 2024-05-30 at 20-19-04 MultiQC Report

This change will also cause the table headers to be copied as well.

Previously, the `user-select: none` style caused the headers to be omitted while copying.

This has the side-effect of enabling text selection in the headers. I'm not sure if this is good or bad - I could see a user wanting to copy a long header.

Thanks as always!


ewels (approved):
Confirmed the issue and the fix, thanks!

PR #2604, opened 2024-06-02
Fix disabling `sort_rows` in custom content: subclass `TableConfig` from `ValidatedConfig` and use deprecated

jchorl:

I found that it is no longer possible to disable row sorting in custom content. Here's an example `mycc_mqc.json`:

```
{
  "data": {
    "cc_b": {
      "Col1": "Hi"
    },
    "cc_a": {
      "Col1": "Hello"
    }
  },
  "file_format": "json",
  "headers": {
    "Col1": {}
  },
  "id": "mycc",
  "parent_name": "My CC",
  "pconfig": {
    "sortRows": false
  },
  "plot_type": "table",
  "section_name": "CC Section"
}
```

We get this error:

```
              plot | • unrecognized field 'sortRows'. Available fields: id, table_title, title, height, width, square, logswitch, logswitch_active, logswitch_label, cpswitch, cpswitch_c_active, cpswitch_counts_label, cpswitch_percent_label, xLog, yLog, xlog, ylog, data_labels, xTitle, yTitle, xlab, ylab, xsuffix, ysuffix, tt_suffix, xLabFormat, yLabFormat, yLabelFormat, xlab_format, ylab_format, tt_label, xDecimals, yDecimals, decimalPlaces, x_decimals, y_decimals, tt_decimals, xmin, xmax, ymin, ymax, xFloor, xCeiling, yFloor, yCeiling, x_clipmin, x_clipmax, y_clipmin, y_clipmax, save_data_file, namespace, save_file, raw_data_fn, defaultsort, sort_rows, only_defined_headers, col1_header, no_violin, scale, min
```

Changing from `sortRows` to `sort_rows`, it does actually run:
Screenshot 2024-06-02 at 01-59-52 MultiQC Report

Note that the rows _are_ sorted!

So in this case it succeeded but didn't do what it was supposed to.

I chased down the issue to this parsing: https://github.com/MultiQC/MultiQC/blob/788e78beb890467c383c712295c4e86c16fc25e8/multiqc/plots/table.py#L37

Adding in logging before/after:

```
Before constructor pconfig['sort_rows']=False
After constructor pconfig.sort_rows=True
```

After more digging, I found the issue stems from this alias: https://github.com/MultiQC/MultiQC/blob/788e78beb890467c383c712295c4e86c16fc25e8/multiqc/plots/table_object.py#L25

I minimally reproduced this with just pydantic:

```
$ docker run -it --rm -v $(pwd):/work -w /work python:3.12 bash
root@8bead663aa61:/work# pip install pydantic
root@8bead663aa61:/work# python3
>>> from pydantic import BaseModel, Field
>>> class TableConfig(BaseModel):
...     sort_rows: bool = Field(True, alias="sortRows")
... 
>>> TableConfig(sort_rows=False)
TableConfig(sort_rows=True)
>>> TableConfig(sortRows=False)
TableConfig(sort_rows=False)
```

This is exactly what's going on. The field is aliased, we pass in `sort_rows`, and it just ignores it.

One confounding issue is that there is already complex pre-validation logic on the pydantic models: https://github.com/MultiQC/MultiQC/blob/788e78beb890467c383c712295c4e86c16fc25e8/multiqc/plots/plotly/plot.py#L126

Specifically, `cls.model_fields` uses the non-aliased fields but `values` is raw and has aliases. From the [pydantic docs](https://docs.pydantic.dev/latest/concepts/validators/#before-after-wrap-and-plain-validators)

> Before validators run before Pydantic's internal parsing and validation (e.g. coercion of a str to an int). These are more flexible than After validators since they can modify the raw input, but they also have to deal with the raw input, which in theory could be any arbitrary object.

So they pretty much recommend against before validators for this exact reason, you need to hack around some of pydantic's great features. I'm not sure if an `after` validator was considered instead - I could believe that it came with worse tradeoffs.

Okay, so where to go from here? A few options came to mind:

1. Should pydantic just fail when parsing? You can do this:

    ```
    >>> from pydantic import BaseModel, ConfigDict, Field
    >>> class TableConfig(BaseModel):
    ...     model_config = ConfigDict(extra="forbid")
    ...     sort_rows: bool = Field(True, alias="sortRows")
    ... 
    >>> TableConfig(sort_rows=False)
      File "<stdin>", line 1, in <module>
      File "/usr/local/lib/python3.12/site-packages/pydantic/main.py", line 176, in __init__
        self.__pydantic_validator__.validate_python(data, self_instance=self)
    pydantic_core._pydantic_core.ValidationError: 1 validation error for TableConfig
    sort_rows
      Extra inputs are not permitted [type=extra_forbidden, input_value=False, input_type=bool]
        For further information visit https://errors.pydantic.dev/2.7/v/extra_forbidden
    ```

    So this would certainly be an improvement. At least it would have failed loudly. I have no idea what it'd break. But it wouldn't fix the issue, it would make both inputs (`sort_rows` and `sortRows`) break.

1. Switch to an `after` validator - seems worthwhile but quite involved!
1. In the `before` validator, it is actually possible to get fields by their alias. Looking at the `model_fields` [docs](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_fields), the dict values of type `FieldInfo` [do have](https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.FieldInfo) and `Alias` attribute. So in theory, the `before` validator could search the dict for every value rather than trying to index the dict by key. Quite messy and probably poorly performing, but maybe performance here doesn't matter.
1. I decided to check how many aliases were even used:

    ```
     j@spider  ~/third-party/MultiQC   main ±  rg alias= -t py
    multiqc/plots/table_object.py
    25:    sort_rows: bool = Field(True, alias="sortRows")
    28:    no_violin: bool = Field(False, alias="no_beeswarm")
    44:    color: str = Field(validation_alias="colour")
    ```

    So maybe a special-purpose fix isn't so bad. That's what I went with. i think it's elegant because multiqc mostly controls the multiqc codebase, so you really only need to patch custom-content.

    However, in the future, it seems people will be importing and using multiqc from python. In that case, the models will need backwords compat and it's probably worth making `alias` work. I'm sure these won't be the last fields to be deprecated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,7 @@ For more information, please see the MultiQC release blog article on the Seqera 
 - Lint check for use of `f["content_lines"]` ([#2485](https://github.com/MultiQC/MultiQC/pull/2485))
 - Allow to set style of line graph (`lines` or `lines+markers`) per plot ([#2413](https://github.com/MultiQC/MultiQC/pull/2413))
 - Add `CMD` to `Dockerfile` so a default run without any parameters displays the `--help` ([#2279](https://github.com/MultiQC/MultiQC/pull/2279))
+- Custom content tables are sorted by key by default (unless `sort_rows: false` is set in config), to harmonize with tables in modules.
 
 ### New modules
 
--- a/multiqc/core/exec_modules.py
+++ b/multiqc/core/exec_modules.py
@@ -14,7 +14,7 @@ from multiqc import config, report
 from multiqc.base_module import BaseMultiqcModule, ModuleNoSamplesFound
 from multiqc.core.exceptions import RunError
 from multiqc.core import plugin_hooks, software_versions
-from multiqc.plots.plotly.plot import PConfigValidationError
+from multiqc.validation import ConfigValidationError
 
 logger = logging.getLogger(__name__)
 
@@ -102,7 +102,7 @@ def exec_modules(
             logger.debug(f"No samples found: {this_module}")
         except KeyboardInterrupt:
             raise
-        except PConfigValidationError:
+        except ConfigValidationError:
             raise
         except:  # noqa: E722
             if config.strict:
--- a/multiqc/core/write_results.py
+++ b/multiqc/core/write_results.py
@@ -334,8 +334,10 @@ def _render_general_stats_table() -> None:
     # Add general-stats IDs to table row headers
     for idx, h in enumerate(report.general_stats_headers):
         for k in h.keys():
-            if "rid" not in h[k]:
-                h[k]["rid"] = re.sub(r"\W+", "_", k).strip().strip("_")
+            unclean_rid = h[k].get("rid", k)
+            rid = re.sub(r"\W+", "_", unclean_rid).strip().strip("_")
+            h[k]["rid"] = report.save_htmlid(report.clean_htmlid(rid), skiplint=True)
+
             ns_html = re.sub(r"\W+", "_", h[k]["namespace"]).strip().strip("_").lower()
             report.general_stats_headers[idx][k]["rid"] = report.save_htmlid(
                 f"mqc-generalstats-{ns_html}-{h[k]['rid']}"
--- a/multiqc/modules/custom_content/custom_content.py
+++ b/multiqc/modules/custom_content/custom_content.py
@@ -355,6 +355,15 @@ class MultiqcModule(BaseMultiqcModule):
         # Table
         if plot_type == "table":
             headers = mod["config"].get("headers")
+
+            # handle some legacy fields for backwards compat
+            sort_rows = pconfig.pop("sortRows", None)
+            if sort_rows is not None:
+                pconfig["sort_rows"] = sort_rows
+            no_violin = pconfig.pop("no_beeswarm", None)
+            if no_violin is not None:
+                pconfig["no_violin"] = no_violin
+
             plot = table.plot(mod["data"], headers=headers, pconfig=pconfig)
 
         # Bar plot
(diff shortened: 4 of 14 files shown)


jchorl:
As a side note, I believe the default behaviour changed for custom content from no-sort-by-default to sort-by-default. I'm not sure if this was intentional. Seems reasonable, but certainly a change.

I think the change comes from here: https://github.com/MultiQC/MultiQC/commit/1283f85ad5792ea5241b05adaf534ccb14700156#diff-2f01ea8761136b590f130470c9e8923ecf845b1f7a12950cdce7bd14190683ffL364

I believe this used to default to `False` and would now default to whatever the `TableConfig` defaults to. Probably a very reasonable change, but maybe worth calling out in the changelog.

vladsavelyev:
Thanks a lot for spotting this @jchorl, and digging into Pydantic details! This is really valuable.

I actually added a somewhat custom validation defined in the parent `PConfig` class to get around this issue. I use the `deprecated` field there and run the `before` validation to match the aliased names:

```py
    @model_validator(mode="before")
    @classmethod
    def validate_fields(cls, values):
        ...
        # Convert deprecated fields
        values_without_deprecateds = {}
        for name, val in values.items():
            if cls.model_fields[name].deprecated:
                new_name = cls.model_fields[name].deprecated
                logger.debug(f"Deprecated field '{name}'. Use '{new_name}' instead")
                if new_name not in values:
                    values_without_deprecateds[new_name] = val
            else:
                values_without_deprecateds[name] = val
        values = values_without_deprecateds
```

So I should have used the `deprecated` parameter for `TableConfig` as well instead of `alias`. 

```py
    sortRows: bool = Field(True, deprecated="sort_rows")
    sort_rows: bool = True
```

I pushed some changes into your branch if you don't mind.

Not sure if my solution is optimal, and it would be better to just fail when we see a deprecated field name using the Pydantic built-in validation. Or just do what you are proposing outside of the model code. Maybe I'll refactor in the future.

Regarding the default sorting behaviour - good spot. I don't think it was intentional 🤔  Anyway, added a line into the changelog.

PR #2609, opened 2024-06-05
Fix `re_contents` search patterns when pattern is in the middle of the file

vladsavelyev:
Fix https://github.com/MultiQC/MultiQC/issues/2607 and https://github.com/MultiQC/MultiQC/issues/2608

Checking the file patterns against 4kb multi-line blocks made some pattern miss hits, when the pattern occurs in the middle of the file, e.g. picard WgsMetrics (https://github.com/MultiQC/MultiQC/blob/main/multiqc/search_patterns.yaml#L678), as well as the patterns that math the exact end of the line, e.g. `'^(feature\tcount|\w+.*\t\d+)$'` (https://github.com/MultiQC/MultiQC/blob/main/multiqc/search_patterns.yaml#L385).

Changing the test function from `pattern.match` to `pattern.search` to match within multi-line blocks, and changing `$` to `\n` in patterns like `htseq` where the top line is matched entirely.

### DIFF
--- a/multiqc/plots/plotly/line.py
+++ b/multiqc/plots/plotly/line.py
@@ -19,7 +19,7 @@ class LinePlotConfig(PConfig):
     ylab: Optional[str] = None
     categories: bool = False
     smooth_points: Optional[int] = None
-    smooth_points_sumcounts: Optional[int] = None
+    smooth_points_sumcounts: Union[bool, List[bool], None] = None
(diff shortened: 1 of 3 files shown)


vladsavelyev:
Sigh, `search` seems to be much slower, leading to the CI jobs to expire, and defeating the purpose of all the performance hacks :(

PR #2614, opened 2024-06-07
UMI-tools extract module: Improve plots

MatthiasZepper:
`umi-tools extract` has two modes, a regex-based and a plain mode. So far, MultiQC only supported logs for the regex-based extraction, but not for the plain mode. The former are admittedly also way more informative, but it caused confusion, when the module was run, because the search pattern matched, but then no plots could be generated because of a lack of data.

This PR now adds a new overall success rate plot, that also accommodates samples extracted in plain mode. 

Furthermore, it splits the regex plot into two subplots, since there is no point in stacking Read1 and Read2 matches and mismatches in one bargraph, as they are not mutually exclusive. Most often, only one read harbours the UMI, so only one regex is applied. But if two are applied, both must match for a successful extraction. A read pair will be discarded if only one or neither of the two matches. 

For test data, please see [the corresponding PR to the TestData repo](https://github.com/MultiQC/test-data/pull/328).


vladsavelyev (approved):
Makes sense. Thank you for the contribution!

PR #2617, opened 2024-06-11
Salmon: fix ignored parsed `library_types` when its type is list

pinin4fjords:
Currently, the code parsing Salmon's `meta_info.json` ignores anything that isn't an `int`, `float` or `string`. There's reference further in the module to the `library_types` key, but it's pointless, since the field was excluded in the 'instanceOf' filter, since it's a list:

```
{
    "salmon_version": "1.10.1",
    "samp_type": "none",
    "opt_type": "vb",
    "quant_errors": [],
    "num_libraries": 1,
    "library_types": [
        "ISR"
    ],
    "frag_dist_length": 1001,
    "frag_length_mean": 161.5665059992703,
```

This PR just adds in the list type, so that the data becomes available. I've tested, and it works- the column shows up the report.


maxulysse (approved):

vladsavelyev:
Thank you!

PR #2628, opened 2024-06-18
Fix Incorrect Calculation of % Q30 Bases

renehennig:
Fix Incorrect Calculation of % Q30 Bases in MultiQC (bbmap Module) as described in https://github.com/MultiQC/MultiQC/issues/2627

- [x] Moved the initialization of `fraction_gt_q30` inside the loop to prevent the accumulation of values across different samples


vladsavelyev (approved):
Thank you so much for contributing the fix! 👍

ewels:
Great spot! Thank you!

PR #2637, opened 2024-06-19
Fix plot export

vladsavelyev:
Fixes https://github.com/MultiQC/MultiQC/issues/2632


multiqc/plots/plotly/plot.py
```diff
@@ -626,7 +628,7 @@ def fig_to_static_html(
             plot_path = Path(report.plots_tmp_dir()) / file_ext / f"{file_name}.{file_ext}"
             plot_path.parent.mkdir(parents=True, exist_ok=True)
             short_path = Path(config.plots_dir_name) / file_ext / f"{file_name}.{file_ext}"
-            logger.debug(f"Writing plot to {short_path}")
+            logger.info(f"Writing plot to {short_path}")
```
ewels, line 631: I don't think this should stay as `info` @vladsavelyev - there can be a lot of plots generated.

PR #2638, opened 2024-06-19
Update quality score thresholds from Q5-Q15 to Q10-Q30

sandratra-rab:
latest nanomath update (v1.4.0) updated the Q-score thresholds from Q5-Q7-Q10-Q12 to Q10-Q15-Q20-Q30 for both general statistics table and reads by quality plot

### DIFF
--- a/multiqc/modules/nanostat/nanostat.py
+++ b/multiqc/modules/nanostat/nanostat.py
@@ -31,11 +31,11 @@ class MultiqcModule(BaseMultiqcModule):
     }
 
     _KEYS_READ_Q = [
-        ">Q5",
-        ">Q7",
         ">Q10",
-        ">Q12",
         ">Q15",
+        ">Q20",
+        ">Q25",
+        ">Q30",
     ]
     _stat_types = ("aligned", "seq summary", "fastq", "fasta", "unrecognized")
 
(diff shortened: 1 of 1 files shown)


vladsavelyev:
Thank you for the pull request!

Will it still work for files generated by older nanostat versions?

Would you be able to share the new output example that we can put into https://github.com/MultiQC/test-data/tree/main/data/modules/nanostat?

sandratra-rab:
> Will it still work for files generated by older nanostat versions?

It won't work with the current modified code... Adding a condition to detect which thresholds set is in the output .txt file would indeed be better to work for all NanoStat versions.

> Would you be able to share the new output example that we can put into https://github.com/MultiQC/test-data/tree/main/data/modules/nanostat?

I have only tried it using BAM files for now:
[NanoStat_alignment_output_Q10-to-Q30.txt](https://github.com/user-attachments/files/15906506/NanoStat_alignment_output_Q10-to-Q30.txt)

 I am generating output examples using the other input file types (fastq, fasta, sequencing summary) and will upload them after.

vladsavelyev:
Thanks for the test example!

I addressed different Q cut-offs in https://github.com/MultiQC/MultiQC/pull/2645

PR #2639, opened 2024-06-19
Progress bar for writing flat plots

vladsavelyev:
Writing flat images with Plotly is a bit slow, so we might want to show a progress bar to indicate that something is happening.

The bar is shown if `--export`/`config.export_plots` is set, or if at least one plot is rendered due to exceeding the `config.plots_flat_numseries` threshold.


ewels (approved):
Tested, works great!

PR #2649, opened 2024-06-22
Custom content: allow hash-fenced table columns

vladsavelyev:
All of these TSV can be allowed as the custom content TSV, and the row starting with either FILE or #FILE should be interpreted as containing column names:

```
FILE	SEQUENCE	START	END	STRAND	GENE
myfile.fasta	chr1	55312	56664	+	GENE
```

```
#FILE	SEQUENCE	START	END	STRAND	GENE
myfile.fasta	chr1	55312	56664	+	GENE
```

```
#key: value
#FILE	SEQUENCE	START	END	STRAND	GENE
myfile.fasta	chr1	55312	56664	+	GENE
```

```
#key: value
FILE	SEQUENCE	START	END	STRAND	GENE
myfile.fasta	chr1	55312	56664	+	GENE
```


ewels (approved):
Would be good to add unit tests 👍🏻

PR #2653, opened 2024-06-24
Codecov: Disable inline PR annotations

ewels:
Finding the inline PR annotations pretty noisy, this should disable them completely.

@vladsavelyev if you find them helpful then close this without merging. In the course of finding how to do this, I also found that the comments can be toggled using the keyboard shortcut `a` (or a dropdown top right of a file), so happy to do that if you prefer.


vladsavelyev:
Nice, wasn't aware of that shortcut! But agree that the comments are annoying, we can always check the report on the codecov website.

PR #2655, opened 2024-06-24
reset config between run() calls

jchorl:

I noticed recently that `config` now has a `reset()` method. Notably, it drops the loaded config files: https://github.com/MultiQC/MultiQC/blob/f25169695ca72d00e4710aadc9555329d326b437/multiqc/config.py#L198

This is important if you call `multiqc.run()` twice with the same config file. If you don't reset, the second run will bail before loading the config even once! See: https://github.com/MultiQC/MultiQC/blob/f25169695ca72d00e4710aadc9555329d326b437/multiqc/config.py#L308-L309

This is problematic because the global config vars have been reset, but they don't get properly re-set to the right values. So the second run effectively runs with no config.

Here's an example python script that exhibits this behaviour:
```python
from pathlib import Path

from multiqc import multiqc
from multiqc.core.update_config import ClConfig


def main():
    multiqc.run(  # pylint: disable=no-member
        Path("in_mqc.json"),
        cfg=ClConfig(
            strict=True,
            force=True,
            config_files=["custom_config.yaml"],
        ),
        clean_up=True,
    )
    multiqc.run(  # pylint: disable=no-member
        Path("in_mqc.json"),
        cfg=ClConfig(
            strict=True,
            force=True,
            config_files=["custom_config.yaml"],
        ),
        clean_up=True,
    )

main()
```

Is this something we could work into multiqc testing?

### DIFF
--- a/multiqc/multiqc.py
+++ b/multiqc/multiqc.py
@@ -490,6 +490,10 @@ def run(*analysis_dir, clean_up: bool, cfg: Optional[ClConfig] = None) -> RunRes
     See http://multiqc.info for more details.
     """
 
+    # In case if run() is called multiple times in the same session:
+    report.reset()
+    config.reset()
+
     update_config(*analysis_dir, cfg=cfg)
 
     check_version()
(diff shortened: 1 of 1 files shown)


vladsavelyev:
Thank you!

Added a test in https://github.com/MultiQC/MultiQC/pull/2660

jchorl:
> Thank you!
> 
> Added a test in #2660

Thank you!!

PR #2657, opened 2024-06-24
Fix parsing kraken vs. bracken: respect `num_lines` in search patterns

jchorl:

There have been some changes to how line-matching works recently and it broke kraken vs bracken kreport matching.

Closes https://github.com/MultiQC/MultiQC/issues/2656

Most details are in the issue.

I opted to create a `line_iterator` because I found the `line_block_iterator` quite challenging to use in practice.

Note that there are changes in behaviour here:
1. `num_lines` will actually be respected - this is probably desirable?
2. `expected_contents` now only checks one line at a time. If any modules/cc depend on checking a block, that will no longer work. I don't know if we want to support checking whole blocks.

I'm open to other implementations. This regressed so just figured I'd offer up a fix.

### DIFF
--- a/multiqc/search_patterns.yaml
+++ b/multiqc/search_patterns.yaml
@@ -210,17 +210,17 @@ ccs/v5:
 cellranger/count_html:
   - fn: "*.html"
     contents: '"command":"Cell Ranger","subcommand":"count"'
-    num_lines: 13
+    num_lines: 14
   - fn: "*.html"
     contents: '"command": "Cell Ranger", "subcommand": "count"'
-    num_lines: 13
+    num_lines: 14
 cellranger/vdj_html:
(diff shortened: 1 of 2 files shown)


multiqc/search_patterns.yaml
```diff
@@ -210,17 +210,17 @@ ccs/v5:
 cellranger/count_html:
   - fn: "*.html"
     contents: '"command":"Cell Ranger","subcommand":"count"'
-    num_lines: 13
+    num_lines: 20
```
jchorl, line 213: These are necessary because the html file that's testing against has both many lines, and newlines within those lines.

vladsavelyev:
Thank you! Looks good

PR #2659, opened 2024-06-25
Fix Incorrect Regular Expression Matching - MultiQC#2658

renehennig:
This will fix an issue with some ignored reports - see https://github.com/MultiQC/MultiQC/issues/2658

### DIFF
--- a/multiqc/report.py
+++ b/multiqc/report.py
@@ -608,10 +608,12 @@ def search_file(pattern, f: SearchFile, module_key):
                 if expected_contents and expected_contents in line_block:
                     contents_matched = True
                 elif repattern:
-                    for line in line_block.split("\n"):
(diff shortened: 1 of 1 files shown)


vladsavelyev:
Thank you for the PR! Would be able to share an example of a report that was missed before this change? I will add a test for that.

vladsavelyev:
Ah, I see that you shared those in the issue https://github.com/MultiQC/MultiQC/issues/2658 - thank you! I'll take a look now

vladsavelyev:
Fixed it slightly differently, by changing the regular expression: https://github.com/MultiQC/MultiQC/pull/2661 let me know if it looks good :)

renehennig:
Works now as expected. Thanks a lot @vladsavelyev!

PR #2663, opened 2024-06-25
Picard HsMetrics: Support any custom X coverage metrics

vladsavelyev:


multiqc/modules/picard/HsMetrics.py
```diff
@@ -346,9 +335,26 @@ def _generate_table_header_config(table_cols, hidden_table_cols):
             for s, r in title_cleanup:
                 h_title = h_title.replace(s, r)
 
+            # Extract the coverage from the column name
+            m = re.match(r".+_(\d+)X", h)
+            if m:
+                try:
+                    cov = int(m.group(1))
+                except ValueError:
```
ewels, line 343: Given that the regex pattern is `\d+` is this not guaranteed to always work?

ewels (commented):
Untested, but LGTM 👍🏻 

Note that the other Picard modules do some very similar things. May be worth bringing some functionality out into a Picard `utils.py` and sharing logic 🤔  Or at least mirroring how this works to the other ones.

Needs a docs update to the picard docs too.

multiqc/modules/picard/HsMetrics.py
vladsavelyev: Good point! Simplified the code

ewels:
`PF_BASES` is also a new field - this should be hardcoded into the field descriptions.

Issue #2687, opened 2024-07-06
Support for VG stats


JosephLalli:
### Name of the tool

VG

### Tool homepage

https://github.com/vgteam/vg

### Tool description

VG is a suite of tools to manipulate graphical genomes, including aligning reads to graph genomes.

### Tool output

[HG002_20X.vg.grch38.d9.vgstats.txt](https://github.com/user-attachments/files/16116928/HG002_20X.vg.grch38.d9.vgstats.txt)
[HSB5112.vgstats.txt](https://github.com/user-attachments/files/16116929/HSB5112.vgstats.txt)
[HSB511.vgstats.txt](https://github.com/user-attachments/files/16116930/HSB511.vgstats.txt)

### Data suitable for MultiQC plot(s)

Total aligned
Total properly paired
% Properly paired (calculated as total properly paired/total aligned)
Mean Mapping quality

### Most interesting data for the General Stats table

Mean mapping quality
% aligned
% properly paired

### Before submitting

- [X] I have included example data (zipped, not pasted) that can be used to write the module.

PR #2690, opened 2024-07-08
New module: VG stats

JosephLalli:
- [ X] This comment contains a description of changes (with reason)
-- This addresses issue #2687, requesting MultiQC support for vg stats reports

- [X] There is example tool output for tools in the <https://github.com/MultiQC/test-data> repository or attached to this PR
    -- mostly used default color schemes a la samtools


vladsavelyev (approved):

A few fixes and updates from me, including a bar plot.

PR #2698, opened 2024-07-12
Umitools: prefer output for sample name, handle <stdin>/<stdout> placeholders

pinin4fjords:
I propose that for the umitools module we derive sample names not from input lines:

```
# stdin                                   : <_io.TextIOWrapper name='SRR6357073_1.fastq.gz' encoding='ascii'>
```

... but from output lines:

```
# stdout                                  : <_io.TextIOWrapper name='RAP1_UNINDUCED_REP1.umi_extract.fastq.gz' encoding='ascii'>
```

I think that's closer to what the user would intend as an identifier. Not doing this is causing some irritating behaviour for us in nf-core/rnaseq.


vladsavelyev:
That makes sense, though I checked our test examples, and it seems like stdout can be piped, in which case it would be just a placeholder:

```
Check # stdout                                  : <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>
```

E.g.
https://github.com/MultiQC/test-data/blob/5322bd06d627160140b3dd0db077c6186531a3ec/data/modules/umitools/dedup/v1.1.2/LIB056594_TRA00243971_S1_R1_001_TSO25_OFF0_dedup.log#L41

I suspect also stdin could be piped as well 🤔 So need some logic to handle this in any case.

pinin4fjords:
> That makes sense, though I checked our test examples, and it seems like stdout can be piped, in which case it would be just a placeholder:

Ahh, pants, good point.

pinin4fjords:
Thanks for this 'proper' solution @vladsavelyev :-)

PR #2719, opened 2024-07-25
Testing: compare HTML screenshots

vladsavelyev:
*In progress*

Adds a test that generates a report, one per plot / use case, opens it in browser, takes a screenshot, and compares the image pixel by pixel.

Resolves https://github.com/MultiQC/MultiQC/issues/2712

Uses [syrupy](https://github.com/tophat/syrupy) (analogue of `pytest-snapshot` but that works better with parallel parametrized tests) and the `pyppeteer` for a headless browser.

Update "gold standard" screenshots by passing the `--snapshot-update` option:

```
pytest tests -vv -k snapshot --snapshot-update
```

Current problem with this approach it that rendering depends on the machine it runs the browser on, meaning that in the CI environment the screenshots could be different from a local environment. 

So need to investigate this further.


ewels:
Love it! Come across the image generation thing before. I think it's ok to say that we only run it via CI and then it's consistent?

* GitHub action workflow dispatch to run on demand for a given branch
* CI job fails if mismatch
* PR comment keyword to rerun and commit changes

What do you think?

PR #2723, opened 2024-07-26
New module: nanoq

LilyAnderssonLee:
Adds a module for [nanoq](https://github.com/esteinig/nanoq) reports


LilyAnderssonLee:
Example report:
[Uploading multiqc_report.html.zip…]()

multiqc/modules/nanoq/nanoq.py
```diff
@@ -0,0 +1,303 @@
...
+        "Number of reads": "Number of reads",
+        "Number of bases": "Number of bases",
+        "N50 read length": "N50 read length",
+        "Longest read": "Longest read",
+        "Shortest read": "Shortest read",
+        "Mean read length": "Mean read length",
+        "Median read length": "Median read length",
+        "Mean read quality": "Mean read quality",
+        "Median read quality": "Median read quality",
+    }
```
vladsavelyev, line 29: What's the point of the dictionary? All keys look identical to values

multiqc/modules/nanoq/nanoq.py
```diff
@@ -0,0 +1,303 @@
...
+            name="nanoq",
+            anchor="nanoq",
+            href="https://github.com/nerdna/nanoq/",
+            info="reports read quality and length from nanopore sequencing data in fast{a,q}.{gz,bz2,xz} format.",
+            doi="10.21105/joss.02991",
+        )
+
+        # Find and load any nanoq reports
+        self.nanoq_summary: Dict[str, Dict[str, float]] = {}
+        self.nanoq_data: Dict[str, Dict[str, float]] = {}
```
vladsavelyev, line 45: We started to change this practice and avoiding adding any new class attributes to MultiqcModule. Would be nice to refactor by keeping `nanoq_data` etc local variables. I can do that in your PR for you to take a look

multiqc/modules/nanoq/nanoq.py
```diff
@@ -0,0 +1,303 @@
...
+        for threshold, num_reads in zip(read_length_thresholds["Threshold"], read_length_thresholds["Number of Reads"]):
+            nanoq_stats[f"Reads > {threshold}"] = num_reads
+
+        for threshold, num_reads in zip(
+            read_quality_thresholds["Threshold"], read_quality_thresholds["Number of Reads"]
+        ):
+            nanoq_stats[f"Reads > Q{threshold}"] = num_reads
+
+        # Store nanoq_summary for general stats
+        self.nanoq_summary[f["s_name"]] = {k: v for k, v in zip(nanoq_summary["Metric"], nanoq_summary["Value"])}
```
vladsavelyev, line 147: Why not just set the `nanoq_stats` here?

multiqc/modules/nanoq/nanoq.py
```diff
@@ -0,0 +1,303 @@
...
+        if f["s_name"] in self.nanoq_data:
+            if not set(self.nanoq_data[f["s_name"]].keys()).isdisjoint(out_d.keys()):
+                log.debug(f"Duplicate sample data found! Overwriting: {f['s_name']}")
+
+        self.nanoq_data.setdefault(f["s_name"], {}).update(out_d)
+
+        self.add_data_source(f)
+
+    def nanoq_general_stats(self) -> None:
+        """Nanoq General Stats Table"""
```
vladsavelyev, line 172: There are too many general stats. If other module are run on a project, the table will end up too wide. Please keep 1 or 2, plus up to 5-7 hidden metrics.

vladsavelyev: Or add a separate table in the report

multiqc/modules/nanoq/nanoq.py
```diff
@@ -0,0 +1,303 @@
...
+                if total_key in _data_dict:
+                    return _data_dict[total_key]
+            return None
+
+        # Get data for plot
+        data: Dict[str, Dict[str, float]] = {}
+        cats: List[str] = []
+        for name, d in self.nanoq_data.items():
+            # Iterate over available types and keep categories in same plot
+            for stat_type in self._stat_types:
```
vladsavelyev, line 223: I don't really understand what iterating over `_stat_types` doing here. If data for multiple `_stat_types` is available, `data[name][threshold]` will be always overwritten by the latest type anyway?

multiqc/modules/nanoq/nanoq.py
```diff
@@ -0,0 +1,303 @@
...
+                reads_by_q = {k: v for k, v in d.items() if re.match(r"Reads > Q\d+_" + stat_type, k)}
+                if len(reads_by_q) == 0:
+                    continue
+
+                data.setdefault(name, {})
+
+                total_reads = _get_total_reads(d)
+                for k, v in reads_by_q.items():
+                    threshold = k.split("_")[0]
+                    data[name][threshold] = v / total_reads * 100
```
vladsavelyev, line 233: Stacking cumulative percentages in a barplot sounds meaningless:
Screenshot 2024-07-29 at 11 24 40

Should it be a differrent represenatation? A line plot maybe?

vladsavelyev:
Thank you a lot for the contribution!

The MultiQC codebase is being refactored, so some module writing best practices are changing (e.g. the changelog is auto-generated now, class attributes are discouraged, class docstring now used for optional docs). I refactored the code a bit to follow that, if you don't mind.

I also threw some suggestions regarding the module logic, mainly:
* I don't really understand how is the `_stat_types` array used. Seems to work the same without it?
* Bar plots for cumulative percentages don't make sense, line plots work better (with a log-scaled X axis for read counts), but please take a look and change/improve this further. I think the plot descriptions at least should be reformulated.
* Adding too many metrics into general stats is discouraged. Better add a separate module-level table.

Please take a look at my changes, feel free to push more, and let me know if it's ready to be merged.

LilyAnderssonLee:
Thank you for lots of good comments and suggestions. I will update the PR according to your comments ASAP.

PR #2734, opened 2024-08-02
Nonpareil: Update docs about new version that generates the JSON file

fgvieira:

From `Nonpareil` version `v3.5.5` there is an auxiliary script that generated the `JSON` file, so there is no need for the extra R functions provided.


vladsavelyev (approved):
Perfect, thank you!

PR #2737, opened 2024-08-03
Improve performance loading large tables in browser

jchorl:

I spent some time today diving into why specific reports were very slow to load. Turns out, the culprit in this case was very wide tables! E.g. 50 rows x 2500 columns.

I started out with 1m45s "time-til-page-is-responsive" and ended up around 10s.

Summary of changes (corresponds to the commits as well):
1. The `Configure columns` modal has neat drag/drop functionality to reorder the rows. This is achieved by calling `.sortable`. However, that call was _very_ slow. I'm not sure why, it's something in jquery UI. Anyway, it is blocking page responsiveness, even though most folks probably don't try to configure columns. I deferred that call until `Configure columns` is actually clicked. And even when the button is clicked, the `.sortable` call only applies to the specific table being configured. So configuring a small table doesn't require calling `.sortable` on the biggest table in the report.
2. Upgrade jquery UI (this was quickly reverted). Upgrading jquery UI improved performance of `.sortable` from 1m45s to 1m5s. This was promising! Also, the vendored version of jquery UI is from 2016. However, the upgraded version contains many more widgets and is 235kb instead of ~35kb for the old one. So I reverted because I wasn't sure if the size increase was acceptable. I tried to find a version of jquery UI with only the few needed widgets, but couldn't find it in my limited time searching.
3. Unrelated to performance, allow html in table tooltips:

    before: before-html

    after: with-html
4. Disable `Configure columns` for very wide tables altogether. Even though this no longer runs at page load time, it doesn't really perform well enough to be useful (at least on my ~decently powerful machine). So I just greyed out the button on large tables and skipped generating the modal.

    disabled: disable-configure-columns

6. The last trick is pretty gnarly. There's more detail in [this upstream PR](https://github.com/Mottie/tablesorter/pull/1851), but this avoids needing to wait for that PR to get merged.

Here was a profile at the beginning:
orig

And here's the end result:
with-all-fixes

Here's a couple test inputs that make the issues easy to reproduce:

[tables.zip](https://github.com/user-attachments/files/16482579/tables.zip)


vladsavelyev:
This is awesome, thanks for investigating and addressing the bottlenecks. Everything makes sense to me

And looks like https://github.com/Mottie/tablesorter/pull/1851 is merged, so can update tablesorter on our side 👍

vladsavelyev (approved):

jchorl:
Thanks for merging!

> so I guess need to wait for them to make the release before we can update on our side, in another PR

Yeah, I don't want to bother them for a release if we have a workaround. And also, there may be other complications with an upgrade. I'll try to keep an eye out for releases there and PR the new revision here when it appears.

Issue #2731, opened 2024-08-01
Bracken Failing to Show Any Reads


egreenberg7:
### Description of bug

In Bracken output for running Bracken at the genus level rather than the species level, in the third field (Number of fragments assigned directly to this taxon) it is possible to have 0 reads directly assigned, even though many reads are rooted at this level (field 2, Number of fragments covered by the clade rooted at this taxon).

This is primarily because the kraken module, which bracken is a subclass of, calculates the total number of reads from the third field. This could be fixed instead by simply looking at the second field of the first line, root. https://github.com/MultiQC/MultiQC/blob/main/multiqc/modules/kraken/kraken.py (line 237)

Also, on another note, when Bracken is detected it says kraken on the side rather than bracken (likely due to the subclass not being treated significantly differently).

### File that triggers the error

[](url)
[report_5p_5.breport.zip](https://github.com/user-attachments/files/16457952/report_5p_5.breport.zip)

### MultiQC Error log

```console
$ multiqc .
Intel MKL WARNING: Support of Intel(R) Streaming SIMD Extensions 4.2 (Intel(R) SSE4.2) enabled only processors has been deprecated. Intel oneAPI Math Kernel Library 2025.0 will require Intel(R) Advanced Vector Extensions (Intel(R) AVX) instructions.
Intel MKL WARNING: Support of Intel(R) Streaming SIMD Extensions 4.2 (Intel(R) SSE4.2) enabled only processors has been deprecated. Intel oneAPI Math Kernel Library 2025.0 will require Intel(R) Advanced Vector Extensions (Intel(R) AVX) instructions.

/// MultiQC 🔍 v1.24.dev0

       file_search | Search path: /Users/ezragreenberg/Desktop/TD Internship/orthoflavivirus
         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 50/50  
            kraken | Found 25 reports
            kraken | No samples had any reads
           multiqc | No analysis results found. Cleaning up…
```

### Before submitting

- [X] I have read the [troubleshooting documentation](https://multiqc.info/docs/usage/troubleshooting/).
- [X] I am using the latest release of MultiQC.
- [X] I have included a full MultiQC log, not truncated.
- [X] I have attached an input file (**.zip** if necessary) that triggers the error.

PR #2738, opened 2024-08-04
Fix Bracken bug when direct reads not classified

egreenberg7:
Update Bracken module to fix the bug described in #2731 Rather than looping over direct reads to get total reads, simply takes the reported reads for root (the first line of each file)

### DIFF
--- a/multiqc/modules/bracken/bracken.py
+++ b/multiqc/modules/bracken/bracken.py
@@ -34,3 +34,15 @@ class MultiqcModule(KrakenModule):
             doi="10.7717/peerj-cs.104",
             sp_key="bracken",
         )
+
+    def sample_total_readcounts(self):
+        # Get the total read counts for each sample, using the fact that all reads are assigned to root
+        total_all_samples = 0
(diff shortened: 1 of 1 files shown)


egreenberg7:
Test data from issue #2731 :[report_5p_5.breport.zip](https://github.com/user-attachments/files/16487319/report_5p_5.breport.zip)
Output: 
[multiqc_report.zip](https://github.com/user-attachments/files/16487323/multiqc_report.zip)

And for test data from repository: 
[multiqc_report_1.html.zip](https://github.com/user-attachments/files/16487331/multiqc_report_1.html.zip)

egreenberg7:
The code also applies a filter to the logger so that when Bracken is used, Bracken appears on the side of the logger rather than Kraken.

vladsavelyev:
Awesome, thanks a lot for addressing the issues! I'd split this up into 2 pull requests, if you don't mind, since it's fixing two unrelated issues.

Also, do you think its worth using `counts_rooted` directly for Kraken as well?

PR #2744, opened 2024-08-05
Kraken: refactor, fix top % calculation, more efficient total read count calculation

egreenberg7:
Changes the method of calculating the total reads for each sample from adding direct reads from each row to instead just looking at the first two lines of each kraken file (unassigned and root).


vladsavelyev:
Thank you! I did some refactoring and added a fix for the percentage calculation that existed before. The results looks good now.

PR #2754, opened 2024-08-10
fix mosdepth custom thresholds

jchorl:

I noticed a regression with the mosdepth module re: setting custom thresholds. It seems we're calling `getattr` on `globals()`. But `globals()` is a dict so `getattr` isn't really useful. Accessing it as a dict by key works better. I'm not sure if this will break other usages though.

Test case:

`Citrobacter+braakii_reference.mosdepth.global.dist.txt`:
```
NZ_CP045771.1   0       1.00
total   0       1.00
```

`multiqc_config.yaml`:
```
mosdepth_config:
  general_stats_coverage: [20, 40, 100, 300, 400]
  general_stats_coverage_hidden: [100, 300, 400]
```

Before:
Screenshot 2024-08-10 at 05-02-58 MultiQC Report

After:
Screenshot 2024-08-10 at 05-03-25 MultiQC Report

### DIFF
--- a/multiqc/config.py
+++ b/multiqc/config.py
@@ -580,7 +580,7 @@ def get_cov_thresholds(config_key: str) -> Tuple[List[int], List[int]]:
     """
     Reads coverage thresholds from the config, otherwise sets sensible defaults. Useful for modules like mosdepth, qualimap (BamQC), ngsbits
     """
-    covs = getattr(globals(), config_key, {}).get("general_stats_coverage", [])
+    covs = globals().get(config_key, {}).get("general_stats_coverage", [])
     if not covs:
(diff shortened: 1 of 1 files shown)


vladsavelyev (approved):
Yeah, it's broken very recently, so other modules that use the depth thresholds are affected (e.g. qualimap bamqc). Thanks a lot for spotting and for the fix!

jchorl:
Thanks for the fast merge!

Are there tests I could add that would've caught this, so it doesn't regress again? I checked the [dev docs](https://multiqc.info/docs/development/) but didn't see anything about testing. I know for a time, writing tests was difficult. Has there been any progress on that front?

PR #2762, opened 2024-08-13
Fix file search performance regression

rhpvorderman:

I noticed some pretty unfortunate performance regressions making MultiQC slow again when searching files. 

This is mostly due to fixing bugs that the line block iterator caused. So it is my fault. I had a look at the fixes and decided that fixing the fixes would also fix the performance regression. 

https://github.com/MultiQC/MultiQC/issues/2656 mentioned that the line block checking can cause overshoot, matching parts that shouldn't be matched. Rather than throwing away the line block iterator, I have reinstated it. But if a match is found I added an extra check to see if there is no overshoot. 
This works well because most of the time there is no match, and using line blocks is optimal in that use case. When there is a match, the runtime overhead of checking for overshoot is minimal. 

Regex patterns should be matched on a line by line basis, something I did not realize when making the line block iterator. That functionality was already fixed. I fixed it in such a way that line blocks can still be kept for the non-regex patterns.

I noticed in py-spy that a disproportionately long time was spend in fnmatch. It turned out that that config fn_ignore_files was checked in each individual pattern check. This only needs to be checked once before individual pattern checks need to be done. I have placed it appropriately.

Before:
```
           multiqc | Run took 111.41 seconds
           multiqc |  - 78.35s: Searching files
           multiqc |  - 18.81s: Running modules
           multiqc |  - 10.88s: Compressing report data
```

After:
```
           multiqc | Run took 44.09 seconds
           multiqc |  - 15.81s: Searching files
           multiqc |  - 15.82s: Running modules
           multiqc |  - 9.39s: Compressing report data
```
(The faster running modules  and compressing is run-to-run variance, the searching files stuff is real.)

Could you please ping me next time there is a big performance regression? Usually these are relatively easy to fix.


ewels:
Great stuff, thanks for noticing and fixing @rhpvorderman!

> Could you please ping me next time there is a big performance regression? Usually these are relatively easy to fix.

I guess this went unnoticed. @vladsavelyev any ideas how we could flag regressions like this in CI? 🤔  Nothing really springs to mind for me, aside from hardcoding some expected values to check for somehow.

rhpvorderman:
Testing performance regressions is really hard, I don't know how to tackle that other than to manually run things. I am very happy with the unit tests you added though. I feel more confident now that I have not broken anything again for the sake of performance :sweat_smile: .

vladsavelyev (approved):
Looks great, thank you!!

PR #2765, opened 2024-08-14
Add search patterns to module docs, generate them

vladsavelyev:


scripts/make_module_docs.py
```diff
@@ -36,25 +50,32 @@
...
     text = f"""\
 ---
 name: {module.name}
 urls: {module.href}
 summary: >
-  {module.info}{extra}
+  {module.info}
 ---
 
-{dedent(docstring)}
```
ewels, line 50: This is where I was thinking that we could have a big HTML comment. Wouldn't make it out of the md > HTML rendering and would help avoid people editing this.

For example, something like:

```html
<!--
~~~~~ DO NOT EDIT ~~~~~
This file is autogenerated from the MultiQC module python docstring.
Do not edit the markdown, it will be overwritten.

File path for the source of this content: {module_file_path}
~~~~~~~~~~~~~~~~~~~~~~~
-->
```

ewels (commented):
Great! Like the directory name `autogenerated` 👍🏻 

Reckon people could still find these files by searching for text strings (I do that myself a lot), but it definitely should help.

scripts/make_module_docs.py
vladsavelyev: Will add!

ewels: Thanks!

PR #2794, opened 2024-08-21
Group read pairs in general stats

vladsavelyev:
Instead of https://github.com/MultiQC/MultiQC/pull/576/files


vladsavelyev:
@ewels the PR ended up pretty large again 😬 mostly due to supporting many edge cases, and a lot of new type hinting to keep the spaghetti more robust 😓 

But if you could play with the functionality and give feedback, it would be great!

vladsavelyev:
I'll merge to get it moving, since it's optional anyway. Will address feedback in follow-up PRs!

ewels:
Wow, great! Looks like we're lacking docs for this. Could you write something up?

Needs something:

* [Using reports](https://multiqc.info/docs/reports/reports/#general-statistics-table) - what grouped samples are, how numbers are calculated. Should include screenshot(s)
* [Customisation](https://multiqc.info/docs/reports/customisation/) - how to turn on / off this feature
* [Writing modules](https://multiqc.info/docs/development/modules/) - how to tie into this feature from module code

vladsavelyev:
Added the docs! https://github.com/MultiQC/MultiQC/pull/2831

PR #2804, opened 2024-08-29
Fix applying `exclude_files` in search patterns

Sxmple-DavidSchmidt:

This pull request only changes the `exclude_file` Method within `report.py`. In the current implementation the exclusion keys (`exclude_fn`, `exclude_fn_re`, `exclude_contents`, `exclude_contents_re`) from the `search_patterns.yaml`-file are not taken into consideration which should be fixed with this.


vladsavelyev:
Great catch! Thanks a lot for the RP 🙏

PR #2814, opened 2024-09-04
Add UMICollapse module

siddharthab:
Fixes #2813.

See https://github.com/nf-core/rnaseq/issues/1087 for broader
discussion.

PR #2827, opened 2024-09-06
Modify UMICollapse module to support FastQ mode.

MatthiasZepper:
In #2814, a new module for the tool UMICollapse was added to MultiQC. While the module itself was a most welcome contribution, there were two minor issues that this PR seeks to correct:

- The search pattern `fn: "*_UMICollapse.log"` is only [a nf-core convention introduced by the module](https://github.com/nf-core/modules/blob/15086c852c860f785a3654cba03f0ee00533cd08/modules/nf-core/umicollapse/main.nf#L50). Hence, I removed this pattern, since it would only apply to nf-core pipelines, but not to users running UMICollapse otherwise. 
- Furthermore, UMICollapse as a FastQ mode, which can be used to deduplicate unaligned data. Again, the tool's output is not standardized, but I am [providing some additional test data](https://github.com/MultiQC/test-data/pull/334) showing that the slightly different wording requires a modified regex.


vladsavelyev:
Thanks a lot!

multiqc/search_patterns.yaml
```diff
@@ -951,9 +951,9 @@ truvari/bench:
   fn: "log.txt"
   num_lines: 2
 umicollapse:
-  fn: "*_UMICollapse.log"
-  num_lines: 100
-  contents: "UMI collapsing finished in "
+  - num_lines: 100
+    contents: "Arguments   ["
+    contents_re: "^UMI collapsing finished in "
```
vladsavelyev, line 956: Do we need to check it as a regular expression here? `contents` is faster than `contents_re`, and the line look unique enough

MatthiasZepper: It is absolutely unique enough, and I also had it as string comparison at first. 

However, I wanted a logical AND conjunction and therefore had to indent it, which resulted in an YAML linting error: "_Map keys must be unique_" because of the duplicate key `contents` on that level. Since YAML apparently requires that each key in a mapping is unique, I could connect a `contents` and `contents_re` via a logical AND, but not of two fields of the type `contents`. 

You can of course decide to ignore that error, or we relax the criteria to me less specific and only match a single line.

MatthiasZepper:
BTW: Wait a few days with merging this please, since I asked in the MultiQC channel of the nf-core Slack about the tool. 

I am not sure, that the unique reads in FastQ mode correspond to the input dedup reads from the BAM mode. For a definitive answer, one would probably have to check the source code of the tool, but maybe some people in nf-core who have used this tool already a lot know that as well.

multiqc/search_patterns.yaml
vladsavelyev: This would work as a logical AND:

```
umicollapse:
    contents: ["Arguments	[", "UMI collapsing finished in "]
```

You example actually works as OR:

```[
umicollapse:
    contents: "Arguments   ["
    contents_re: "^UMI collapsing finished in "
```

And ``contents: "Arguments   ["` never matches because it has spaces, whereas the umicollapse output has a tab. (But I should probably change this to act as an AND as well, it makes much more sense).


Regardless of that, I think `"Arguments	[",` is not really needed, the last line alone is sufficient to check. So I'd modify the pattern to:

```
umicollapse:
  num_lines: 100
  contents: "UMI collapsing finished in "
```

MatthiasZepper: Yes, please do as you see fit! 

It was my last afternoon before the holiday and I wanted to get a lot of things off the ToDo list. The issue with the module was thus not really timely, but I also felt that it should be addressed right away, so we have a new MultiQC version including this module available by the time we ship that new tool in the nf-core rnaseq pipeline.  

Hence, I was not too carefully reading [the documentation anymore](https://multiqc.info/docs/development/modules/#step-1---find-log-files) and misinterpreted the example for the AND conjunction: 

```
myother_module:
  - fn: "different_fn.out"
    contents: "This is myprogram v1.3"
   - fn: "another.txt"
     contents: ["What are these files anyway?", "End of program"]
     contents_re: "^Metric: \d+\.\d+"
```

> myother_module will match different_fn.out with the contents This is myprogram v1.3, or another.txt containing ALL of the lines What are these files anyway?, End of program", and ^Metric: \d+\.\d+.

To me, it seemed there would be a logical AND between `What are these files anyway?`, `End of program"`, and `^Metric: \d+\.\d+.`. Putting it into one line never worked, and now I also know why - because of the spaces / tab confusion, so that was an unrelated issue.

MatthiasZepper: Oh, and I am of course fine with dropping the `"Arguments	["` part of the search pattern. The reason why I initially included it was, that I had experimented with defining separate search patterns for the `fastq` and `bam` mode and have separate modules for them (like e.g. for `umi-tools extract` and `umi-tools dedup`) but then changed my mind somewhere half-way through. 

So from my side, I did not really put much consideration into this PR and quickly cobbled something together, because I promised it to Phil on the nf-core Slack. Therefore, it would be absolutely required to have some feedback from others on the unique reads and if they are really equivalent before merging this.

vladsavelyev: Thanks for clarifying! All makes sense to me. And agree on logical AND 👍

vladsavelyev:
I'm going to merge this now. Please create a separate PR if needed 👍

PR #2840, opened 2024-09-13
Samtools coverage: suppot `exclude_contigs` and `include_contigs`

vladsavelyev:
Similar to mosdepth, supporting `exclude_contigs` and `include_contigs` options:

```
samtools_coverage:
  exclude_contigs:
    - "chrM"
    - "chr20"
```

Fix https://github.com/MultiQC/MultiQC/issues/2802


matthdsm:
Thanks a bunch Vlad!
Will you also update the docs or does that happen automagically?

vladsavelyev:
The website docs will be updated when we make release, automatically from the class docstring: https://github.com/MultiQC/MultiQC/blob/c607e58478481793c2373fe800deff56375ec9e7/multiqc/modules/samtools/samtools.py#L57-L86

PR #2875, opened 2024-09-23
Ignore unrecognized custom table fields

vladsavelyev:
Address https://github.com/MultiQC/MultiQC/issues/2870


ewels:
Should we throw a warning if it's unrecognised? Might be helpful for folks..

PR #2903, opened 2024-10-02
Docs: build modules index page with a table

vladsavelyev:
Docs PR: https://github.com/seqeralabs/docs/pull/246

Screenshot 2024-10-02 at 21 10 52


scripts/make_module_docs.py
```diff
@@ -98,3 +103,21 @@
...
+with (Path("docs") / "markdown" / "modules.mdx").open("w") as fh:
+    fh.write(f"""\
+---
+title: MultiQC Modules
+description: Tools supported by MultiQC
+---
+             
+MultiQC currently has modules to support {len(config.avail_modules)} bioinformatics tools, listed below.
+
+Click the tool name to go to the MultiQC documentation for that tool.
```
ewels, line 117:
```suggestion
Click the tool name to go to the MultiQC documentation for that tool.

:::tip[Missing something?]
If you would like another tool to to be supported, please [open an issue](https://github.com/MultiQC/MultiQC/issues/new?labels=module%3A+new&template=module-request.yml).
:::
```

scripts/make_module_docs.py
```diff
@@ -98,3 +103,21 @@
...
+
+{table}
+
+Missing something? If you would like another tool to to be support,
+please [open an issue](https://github.com/MultiQC/MultiQC/issues/new?labels=module%3A+new&template=module-request.yml).
```
ewels, line 122:
```suggestion
```

ewels (approved):
Nice! I think the page is too long to include anything under the table though, so suggest moving the hint up to the top and putting it into an [admonition](https://docusaurus.io/docs/markdown-features/admonitions).

ewels:
Frontmatter comment in https://github.com/seqeralabs/docs/pull/246#issuecomment-2389648211 needs to be added in this PR actually. But maybe test in docs first to check it does what I hope it does 😬

vladsavelyev:
I think this is actually better: https://github.com/MultiQC/MultiQC/pull/2904

PR #2905, opened 2024-10-03
Samtools: Add insert size to general stats table

### LINKED ISSUE #2897: samtools stats: add insert size to general statistics
### Description of feature

From samtools stats we are missing the information about insert size. They are not parsed from the sample.stats file, and it would be useful information to display in the general statistics table.

`SN      insert size average:    392.2`

pontushojer:
fix #2897 

This PR adds the average insert size from `samtools stats` output. The field is hidden by default as it may not be useful for everyone. 

image


vladsavelyev (approved):
Awesome, thank you 🙏

PR #2925, opened 2024-10-18
New module: Cell Ranger ARC - adding a new module named Cell Ranger ARC

achakkas:

- Added new module for CellRanger ARC. I'm considering this as a separate module than Cell Ranger since both are two different tools but let me know if I should change it
- Tested with the latest version of ARC (2.0.2) and also with 2.0.0, working fine for me
- Tried to add pytests but not sure if that's good enough
- There is also a bug that I encountered during testing (reported [here](https://github.com/MultiQC/MultiQC/issues/2923)), I added the fix here but let me know if it shouldn't be in this PR

- [x] There is example tool output for this tool in the <https://github.com/MultiQC/test-data> repository. See [Test data PR](https://github.com/MultiQC/test-data/pull/337)

- Couple of tables with 7 columns (needed to fit different metrics from ATAC subsection), and the metrics from Gene Expression tab are added to a single table but 10/20 are hidden.

    

vladsavelyev:
Thank you for such a high quality PR! Just a few minor adjustments - I'll push directly to the branch if you don't mind. And it's good to go with the next, Xmas release.

PR #2933, opened 2024-10-25
Add module Haplocheck

peterpru:

This is a PR for a new module, haplocheck, specifically for its contamination output.


peterpru:
PR for test data: https://github.com/MultiQC/test-data/pull/339

peterpru:
Runs on local:
image

image

multiqc/search_patterns.yaml
```diff
@@ -384,6 +384,8 @@ goleft_indexcov/ped:
   fn: "*-indexcov.ped"
 gopeaks:
   fn: "*_gopeaks.json"
+haplocheck:
+  fn: "*haplocheck.txt"
```
vladsavelyev, line 388: I wonder if it's a default or conventional file name for the tool's output?
It looks like that haplocheck takes the output name as an argument, so there is not guarantee it will be called this way. And looking at an example at https://github.com/MultiQC/test-data/pull/339/files, it seems like it should be possible to recognize the file from it's contents based on the header line?

peterpru: You are correct, the file name is not standardised. I will change it so it recognises the file from the header.

vladsavelyev: Thanks! Two comments:

1. `contents: "Contamination Status"` seems too broad and can accidentally match unrelated files. Can we narrow it? Would `contents: "Sample"        "Contamination Status"  "Contamination Level"   "Distance"      "Sample Coverage"` work or the header is not determined? We can also have several comma-separated patterns to require satisfying them all (`contents: ["A", "B"]`), or multiple sections which would require one to satisfy (`- contents: "A"\n  - contents: "B"`).
3. `fn: "*"` is not needed

peterpru: That sounds good, thanks for the commits, and I have followed your suggestion to add more content matches.

vladsavelyev:
Thank you Peter for the contribution!

PR #2934, opened 2024-10-25
nanoq: Change lineplots for barplots

dialvarezs:

This PR updates one of the plots in the nanoq module, changing it from a line plot to a bar plot. The line plot is not the best option for visualizing nanoq quality output, primarily because it only reports counts for selected quality thresholds rather than for every possible quality value. Additionally, the line plot represents values only as percentages, difficulting comparison of yields across multiple samples. This type of plot is also used in the NanoStat module, a very similar tool.

Below is an example of the current plot:
nanoq_lines

And here is the updated plot produced by this PR:
nanoq_bars

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,7 +49,7 @@ runTest.sh
 .direnv/*
 
 # Typical name for virtualenv
-venv/
+venv*
 
 # Related repos
 /test-data/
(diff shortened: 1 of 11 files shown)


vladsavelyev:
Awesome stuff!! Thanks a lot for the contribution 🙏

vladsavelyev:
@dialvarezs would it make sense to modify the read lengths plot to a barplot as well?

vladsavelyev (approved):

dialvarezs:
@vladsavelyev Yes, you're right, that would make sense. I will update the length plot as well.

vladsavelyev:
🔥 
Thank you again!

PR #2946, opened 2024-10-31
Qualimap: Clarify the direction of the transcript in coverage plot

pontushojer:
Maybe this is obvious for most people but I figured it was a good idea to make this as clear as possible.


vladsavelyev (approved):
Thank you!

PR #2953, opened 2024-11-11
Fix Docker CI for forks: skip login step

vladsavelyev:
The Docker CI step currently fails for fork PRs because it attempts to log in to dockerhub, even though it doesn't need that (it pushes only on merges to `main` and on releases). Fixing.


ewels (commented):
Perfect, thanks!

PR #2970, opened 2024-11-18
Add module GTDB-Tk: a software toolkit for assigning objective taxonomic classifications to bacterial and archaeal genomes

harper357:
- [x ] This comment contains a description of changes (with reason)
>Added a module to support GTDB-Tk. It generates a table showing the classification of input genomes.

> https://github.com/MultiQC/test-data/pull/340

>Since this tool assigns taxonomy, most of the data output is not numerical, and the numerical columns are only useful when viewed with the non-numerical data. 
>The table has 7 columns, most of the time some of them (`RED`, & `warnings`) will be be empty, but will be very important when they aren't empty.


vladsavelyev:
Amazing. A few adjustments and good to merge

PR #2978, opened 2024-11-21
Added module Checkm2: Rapid assessment of genome bin quality using machine learning

harper357:
- [ x] This comment contains a description of changes (with reason)
> Added support for CheckM2 https://github.com/chklovski/CheckM2

> https://github.com/MultiQC/test-data/pull/341
> I based this off of the module for QUAST, which outputs a table.
> It makes a table of 13 columns, but hides 5 of them. If I need to cut more columns, I could probably remove some, but I think they are all good to have.


multiqc/search_patterns.yaml
```diff
@@ -215,6 +215,8 @@ cellranger/vdj_html:
   - fn: "*.html"
     contents: '"command": "Cell Ranger", "subcommand": "vdj"'
     num_lines: 20
+checkm2:
+  fn: "*quality_report.tsv"
```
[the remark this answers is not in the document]
vladsavelyev: I think checking the header line would be sufficient - updated

harper357:
Great! Thanks @vladsavelyev !

multiqc/modules/checkm2/checkm2.py
```diff
@@ -0,0 +1,155 @@
...
+            "Additional_Notes",
+        )
+        parsed_data = {}
+        for line in f["f"]:
+            if line.startswith("Name\t"):
+                continue
+            column_values = line.rstrip().split("\t")
+            column_values = [None if x == "None" else x for x in column_values]
+            parsed_data[column_values[0]] = {k: v for k, v in zip(column_names, column_values) if v}
+        self.checkm2_data = parsed_data
```
harper357, line 63: Should this line get changed to `self.checkm2_data.update(parsed_data)` to cover if there are multiple logs?

vladsavelyev: Would be a good idea!

vladsavelyev:
Thank you for the contribution!

PR #2990, opened 2024-11-25
Add module Checkm: automated method for assessing the quality of a genome using a broader set of marker genes

harper357:
This adds support for CheckM. This is different than #2978 and is still being used by nf-core/mag.

 > https://github.com/MultiQC/test-data/pull/342


vladsavelyev:
Thank you again for your contribution! Looks good, will merge now. Will be part of the upcoming release v1.26 🎊

PR #3011, opened 2024-12-12
New module: percolator: a semi-supervised learning framework for peptide identification from shotgun proteomics datasets.

JuliaGraf:
Adds support for Percolator.
https://github.com/MultiQC/test-data/pull/343


multiqc/modules/percolator/percolator.py
```diff
@@ -0,0 +1,229 @@
...
+    a semi-supervised learning framework for peptide identification from shotgun proteomics datasets.
+    """
+
+    def __init__(self):
+        super(MultiqcModule, self).__init__(
+            name="Percolator",
+            anchor="percolator",
+            href="https://github.com/percolator/percolator",
+            info="Semi-supervised learning for peptide identification from shotgun proteomics datasets.",
+        )
```
vladsavelyev, line 149: Please add doi (set it to empty list if it's unpublshed), see:

https://docs.seqera.io/multiqc/development/modules#multiqcmodule-class

vladsavelyev:
Thanks a lot for the pull request! Few thoughts:

* Should multiple samples be handled? Would it be valuable to use samples as barplot rows, and split into 3 different barplots? This way bars within a barplot would share categories, as expected from a barplot. Or instead there could be one barplot with 2 tabs: one to include all features across all 3 groups, and second tab would show just 3 groups - sums of features from each groups (i.e. 3 colors).
* But overall barplot when including all feartues looks pretty busy, wonder is it still useful? 
* I wonder if it's useful to take the absolute values of weights? Would it make sense to keep the negative values negative? Barplot can represent them well.

jonasscheid:
Hi @vladsavelyev! Thanks for the Input! Agreed it's a bit messy and I like Both options in A. @JuliaGraf could you show how that could look like?

Initially we wanted to plot it like this 
image
But I think it is not possible to sort within groups. Please correct me if I am wrong here.

vladsavelyev:
Thanks for the example. In this case a barplot like this would match better? 

Or it's better to keep the absolute values for weights?

percolator_plot (1)

jonasscheid:
This is great yes! It's also better for understanding to not us absolute values, _absolutely_ ;-)

vladsavelyev:
One more question - is it guaranteed that the set of features going to look like that? I.e. do we want to hardcode them?

Looking at the example at https://github.com/percolator/percolator/wiki/Example, it's using uppercase names for the features, and they are different:
```
  lnrSp	 deltLCn deltCn	Xcorr	 Sp	   IonFrac Mass	   PepLen Charge1 Charge2 Charge3 enzN enzC enzInt lnNumSP dM	 absdM	   m0
```

Another question regarding the search pattern - is it guaranteed that the tools outputs would be named `*feature_weights.tsv`? Would it make sense to check the contents instead, i.e. check the feature names that it always outputs with `contents:`?

jonasscheid:
> One more question - is it guaranteed that the set of features going to look like that? I.e. do we want to hardcode them?

No its not, you can basically give percolator any feature / feature name you want. These are just the ones that are used in https://github.com/nf-core/mhcquant and https://github.com/nf-core/quantms. Would it make sense group any additional feature that is not in the group of ms2pip,deeplc, or psm_file to annotate with an "other" group?

> Another question regarding the search pattern - is it guaranteed that the tools outputs would be named *feature_weights.tsv? Would it make sense to check the contents instead, i.e. check the feature names that it always outputs with contents:?

No you can name it whatever you want (open for suggestions).. I think its best to give it a fixed output name instead of dynamic based on which group (deeplc/ms2pip/future_group) is present or would there be any downside to this?

vladsavelyev:
> No its not, you can basically give percolator any feature / feature name you want. These are just the ones that are used in https://github.com/nf-core/mhcquant and https://github.com/nf-core/quantms. Would it make sense group any additional feature that is not in the group of ms2pip,deeplc, or psm_file to annotate with an "other" group?

General idea is that we can't hardcode one specific use cases. MultiQC module should work equally for any configuration of the supported tool. So if the feature name in Percolator are an input, then we can't assume them in MultiQC, and should 
be able to parse them from Percolator outputs, or configure via `multiqc_config.yaml`.

> No you can name it whatever you want (open for suggestions).. I think its best to give it a fixed output name instead of dynamic based on which group (deeplc/ms2pip/future_group) is present or would there be any downside to this?

If the file contents is variable and MultiQC can't reliably determine that it's rom Percolator, then we need to base on the file name, and add a docstring behind the class name describing how the files should be named, e.g.

```
"""
The module assumes the Percolator ouptut file is named `*feature_weights.tsv`. Make sure to run it `percolator ... > samples.feature_weights.tsv`
"""
```

Or maybe the name should include "peroclator" in it to be more speicifc.

JuliaGraf:
I updated the search pattern to *percolator_feature_weights.tsv.

The GROUP_TO_FEATURE dict is now defined in the `multiqc_config.yaml`:

Screenshot 2024-12-17 164537

if `group_to_feature` is not defined, all features are put into the same group ("other). Also if `group_to_feature` is defined but does not contain a certain feature that appears in the percolator output, that feature is also assigned to "other".

vladsavelyev:
Awesome, that's what I had in mind!

On minor thing - `custom_data` is reserved for the [custom content](https://docs.seqera.io/multiqc/custom_content#data-as-part-of-multiqc-config). It needs to be just a top-level config instead:

```
percolator:
  group_to_feature:
    psm_file_combined: [MS:1002255, MS:1002252]
    ms2pip: [ionb_min_abs_diff, iony_min_abs_diff]
    deeplc: [rt_diff]
```

I'll push this change directly to the branch.

Will also update the test-data example to have the new search pattern ending 👍

vladsavelyev (approved):

PR #3023, opened 2024-12-16
New module: mgikit

vladsavelyev:
Copy of https://github.com/MultiQC/MultiQC/pull/2699 to be able for us to push to the branch


ziadbkh:
Thanks @vladsavelyev!
for the comment: It also seems like it's more natural to parse sample stats from *.mgikit.sample_stats as they are not rounded up. Though no info for lanes there, I wonder it is output into a yet another file, e.g. *.mgikit.lane_stats?

the file *sample_stats should contain all the needed information. I developed the module before having this file in the output and that is why I did not use it. I will think about it at some point later and do a pull request if this will make the module better.

PR #3045, opened 2025-01-14
Make tables Configure Column max columns configurable

vladsavelyev:
Fix https://github.com/MultiQC/MultiQC/issues/3042

* Make it configurable the number of columns before "Configure Columns" is hidden. Default is `config.max_configurable_table_columns = 100` (bumped from 50).
* Fix the number of visible rows indicator.

### DIFF
--- a/docs/markdown/development/plots.md
+++ b/docs/markdown/development/plots.md
@@ -712,6 +712,11 @@ in ascending order, then by "Starting Amount (ng)", in descending (default) orde
 table with the ID `quast_table` (which you can find by clicking the "Configure Columns"
 button above the table in the report) will be sorted by "Largest contig".
 
+### Configurable columns
+
+Table columns can be reodered and change visibility using the "Configure Columns" button
+in the report. However, for very wide tables, the performance degrades, so the button is disabled when the number of rows exceeds `config.max_configurable_table_columns` (default is 100). You can adjust this value in the config file.
+
 ## Violin plots
 
 Violin plots work from the exact same data structure as tables, so the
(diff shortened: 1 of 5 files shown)


multiqc/config_defaults.yaml
```diff
@@ -76,6 +76,7 @@ violin_min_threshold_no_points: 1000 # for more than this number of samples, sho
 
 collapse_tables: true
 max_table_rows: 500
+max_configurable_table_columns: 100
```
ewels, line 79:
```suggestion
max_configurable_table_columns: 200
```

docs/markdown/development/plots.md
```diff
@@ -712,6 +712,11 @@ in ascending order, then by "Starting Amount (ng)", in descending (default) orde
 table with the ID `quast_table` (which you can find by clicking the "Configure Columns"
 button above the table in the report) will be sorted by "Largest contig".
 
+### Configurable columns
+
+Table columns can be reodered and change visibility using the "Configure Columns" button
+in the report. However, for very wide tables, the performance degrades, so the button is disabled when the number of rows exceeds `config.max_configurable_table_columns` (default is 100). You can adjust this value in the config file.
```
ewels, line 718: Reckon 200 should be fine as well? And then very unlikely that 99% of folks will come close.

```suggestion
in the report. However, for very wide tables, the performance degrades, so the button is disabled when the number of rows exceeds `config.max_configurable_table_columns` (default is 200). You can adjust this value in the config file.
```

ewels (approved):

docs/markdown/development/plots.md
vladsavelyev: 200 makes it veeery slow for me, I recon on older computer would be worse

PR #3050, opened 2025-01-18
AI: Wrap Seqera AI prompt in `<details>`

ewels:
Tuck the long prompt out of the way inside a collapsible element.

### DIFF
--- a/multiqc/core/ai.py
+++ b/multiqc/core/ai.py
@@ -2,7 +2,6 @@ import base64
 import logging
 import os
 import re
-import time
 from textwrap import indent
 from typing import TYPE_CHECKING, Dict, Optional, Tuple, cast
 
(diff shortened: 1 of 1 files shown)


ewels:
⚠️  Needs the same addition in the front end code - if you click "more details" and then "continue chat" it uses a prompt from the front-end which doesn't have the `:::details` stuff.

vladsavelyev (approved):

vladsavelyev:
Honestly I don't like the idea of using prompt to pass internal instructions from MultiQC to FE. That's hacky and might mislead the LLM.

ewels:
Yeah I know what you mean. But I'm not sure that it's any worse than `:::collapse` - in fact probably better as it's more semantically correct.

vladsavelyev:
@ewels `:::details` is way better and works out of the box, nice find! 

I just don't like making it part of the prompt that is passed to actual LLM. But it works, let's keep it

PR #3058, opened 2025-01-23
Custom content: Support `.gif`, `.webp` and `.tiff`

ewels:
As requested in https://community.seqera.io/t/including-gif-images-in-multiqc-report/1640

It's very likely that `.tiff` files could be huge and will break reports / crash browsers. But as this is Custom Content it's a bit the wild-west already, so it's up to end users to handle this responsibly. At least MultiQC won't arbitrarily crash during report generation after this PR.

### DIFF
--- a/docs/markdown/custom_content/index.md
+++ b/docs/markdown/custom_content/index.md
@@ -39,12 +39,13 @@ general release.
 
 ## Images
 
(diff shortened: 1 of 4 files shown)


vladsavelyev (approved):
Thanks!

PR #3072, opened 2025-01-29
Config from env: resolve empty to Null

vladsavelyev:
Empty env vars will be ignored instead of showing a warning (e.g. `export MULTIQC_AI_SUMMARY=` would keep `config.ai_summary` default).


ewels (approved):
Thanks!

PR #3075, opened 2025-01-30
FastQC: fix "Per Base Sequence Content" heatmap

vladsavelyev:


aasthashahcodes:
Hi there, I have tried uodating my MultiQC to latest version which is not getting ugraded after 1.18. I am having trouble viewing the heatmap and hence the reason my report is showing "Loading Report" at the top of the report since 22 hours. I am running MultiQC on WSL just for context, and I tried debugging various way but didn't work. 

image

PR #3087, opened 2025-02-06
Bump 1.27.1

vladsavelyev:


ewels:
Maybe add a sentence at the top pointing out that AI features should now work with the bioconda install?

also that the new endpoint can support other providers, including deepseq or custom enterprise AI providers.

PR #3091, opened 2025-02-15
QUAST: add ANI column

schorlton-bugseq:
When comparing a genome against a reference with QUAST, it is helpful to know how similar that genome is. QUAST outputs this in a "# mismatches per 100 kbp" and "# indels per 100 kbp" field. Here we combine these two fields to calculate an average nucleotide identity for a new column, which is simpler and more interpretable.

I acknowledge that sequence identity is complicated and has different meanings (https://lh3.github.io/2018/11/25/on-the-definition-of-sequence-identity). However, the choice is made easier given the outputs available from QUAST. Here I implement the gap-compressed identity, which the author of the above blog finds most compelling. Although it slightly differs from the definition of BLAST ANI used in the literature for prokaryotic species definition (non-gap-compressed), I think it is sufficient for most applications. The description also explicitly states it is gap-uncompressed.

image

### DIFF
--- a/multiqc/modules/quast/quast.py
+++ b/multiqc/modules/quast/quast.py
@@ -165,6 +165,15 @@ class MultiqcModule(BaseMultiqcModule):
                     except ValueError:
                         self.quast_data[s_name][k] = v
 
+        # Calculate ANI
+        for s_name, data in self.quast_data.items():
+            # These fields are only present if a reference genome was provided to QUAST
+            if "# mismatches per 100 kbp" in data and "# indels per 100 kbp" in data:
+                self.quast_data[s_name]["ANI (%)"] = 100 - (
+                    data["# mismatches per 100 kbp"] / 1000
+                    + data["# indels per 100 kbp"] / 1000
+                )
+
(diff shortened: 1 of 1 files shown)


vladsavelyev:
Thanks for the suggestion! 

Having this metric sounds reasonable, but should it rather be a part of [QUAST](https://github.com/ablab/quast)? Their team is very responsive, if you want to take a stab in a pull request there. 

For MultiQC, we can attempt to parse it from the QUAST report, and for older versions we could keep your code here to calculate it if it's missing.

vladsavelyev:
I coordinated with the QUAST author to add this metric - but it might take some time, so feel free to file a PR there!

And thanks for this PR here - happy to merge it now.

schorlton-bugseq:
Thanks @vladsavelyev for your feedback and even accepting this proposal given the limitations! You are absolutely right that it would be better to upstream and report this from QUAST. I opened a PR there (https://github.com/ablab/quast/pull/279) and happy if you want to revert this if we can get that accepted.

PR #3115, opened 2025-03-04
AWS Bedrock support for AI summaries (#3115)

jchorl:
# Add aws bedrock support for multiqc ai capability

Figured I'd take a stab at this. Remaining items (there may be more):

- [x] Lint/tests
- [ ] Do we need retries/error-handling on the bedrock request
- [ ] I see the code has an error message to install e.g. `pip install multiqc[aws_bedrock]`, but I don't see those optional deps defined in `pyproject.toml`. Ideally we can park `boto3` somewhere so it's not in the standard installation path.
- [x] I don't understand the JS parts well enough to make them work. I can deep dive or perhaps you know immediately what's wrong. IMO we don't want bedrock involved in the JS components, I can't imagine anyone is copy/pasting from the report to bedrock, it's not really consumer-facing.
- [x] Docs updates

It does work, tested with:

Config:
```
strict: true
max_table_rows: 5000
ai_summary: true
ai_summary_full: true
ai_provider: "aws_bedrock"
ai_model: "us.anthropic.claude-3-7-sonnet-20250219-v1:0"
no_ai: false
```

Building image/running (it picks up AWS creds using the library standards):
```
docker run -it --rm \
    -v $(pwd):/work \
    -w /work \
    -v "$HOME/.aws":/home/multiqc/.aws \
    -e AWS_PROFILE=myprofile \
    -e AWS_REGION=us-west-2 \
    build/multiqc \
    bash
```


jchorl:
Update here:
1. Lint/test are passing. I didn't see any test framework already in place for the AI clients, but if there is one, happy to add some tests.
2. I don't think we need retries on the bedrock request yet. Some of that is handled in the boto client already. We can always add later if necessary.
3. I don't see separate dep sets for AI deps so probably fine to included boto3 as a dep? Let me know if I should move that somewhere.
4. I looked through the JS parts. I don't think any updates are needed there, I wouldn't think folks would wire up the report directly to bedrock.
5. I updated the docs

Let me know if there's anything else you'd like to see here!

jchorl:
Hey @vladsavelyev any chance you have thoughts here on what might need to be changed or how we can get this merged?

Thanks!

vladsavelyev:
Hi Josh, so sorry for neglecting your PR for so long... For some reason I was sure I got it merged long time ago. Those were hectic 2 months...

Thanks a lot for the contribution. I'll test if including boto3 adds a lot to the image size and would move it to optional deps. Otherwise looks great!

PR #3118, opened 2025-03-11
Use parquet as intermediate format

vladsavelyev:
Fix https://github.com/MultiQC/MultiQC/issues/1256

- [x] Export normalized plot data to parquet files
- [x] Keep old JSON for metadata, just add parquet relative paths into it


ewels:
If possible it'd be great to export as much metadata as possible alongside plot data.

For example, then if we identify a single data point, the user should be able to trace this back and see what report run it was generated from, when, what the command was, what the search file paths were, etc.

PR #3120, opened 2025-03-12
Add new CLI command: multiqc-check

ewels:
Couldn't resist having a little play..

Feel free to refactor / do anything with this.

CleanShot 2025-03-12 at 15 37 54@2x

### DIFF
--- a/multiqc/check_config/__main__.py
+++ b/multiqc/check_config/__main__.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python
+
+"""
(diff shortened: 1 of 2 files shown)


vladsavelyev:
Awesome!

vladsavelyev (approved):
Nice!

I don't have a strong opinion on command vs flag (`multiqc --check-config myfile.yml`). Since we already have `multiqc --help`, I guess there is no harm in adding another command that doesn't normally run multiqc.

PR #3124, opened 2025-03-17
Round table values for AI prompt

vladsavelyev:
Round table values for AI prompt. Add a test.
Re https://github.com/MultiQC/MultiQC/issues/3108

### DIFF
--- a/multiqc/core/ai.py
+++ b/multiqc/core/ai.py
@@ -651,7 +651,7 @@ def build_prompt(client: Client, metadata: AiReportMetadata) -> Tuple[str, bool]
     user_prompt += tools_context
 
     # General stats - also always include, otherwise we don't have anything to summarize
-    if general_stats_plot := report.plot_by_id.get(Anchor("general_stats")):
+    if general_stats_plot := report.plot_by_id.get(Anchor("general_stats_table")):
         genstats_context = f"""
 MultiQC General Statistics (overview of key QC metrics for each sample, across all tools)
 {general_stats_plot.format_for_ai_prompt()}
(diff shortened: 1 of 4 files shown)


schorlton-bugseq:
Thank you @vladsavelyev!! Super appreciate this. One thought: if I understand this, it rounds all floats in a table to 2 decimals before submitting to AI. However, is this how the MultiQC tables work? I _thought_ (maybe incorrectly) that each column of a table had its own decimal places configured (https://docs.seqera.io/multiqc/development/plots#table-decimal-places, although maybe 2 is the default?). Would it be better to use the same number of decimals set for the column in the table _if it is set_ (else use default)? I could imagine that 2 decimals may sometimes be perfect, but other times too few or too many for a table field to pass to the AI. If the user has configured the number of decimals to include in the report, we would want to match that in the AI prompt. It could be odd if the table shows the user a table in the MultiQC report:

Days from now | Probability of the world ending
-- | --
1 | 0.009

and then we get a response like
image

vladsavelyev:
Good call! I updated it to respect the table column `format` field:

```
{
    "plot_type": "generalstats",
    "id": "Genome Info",
    "pconfig": {
        "Abundance": {
            "max": 100,
            "min": 0,
            "suffix": "%",
            "scale": "Blues",
            "format": "{:,.4f}",
            "description": "Reads assigned to the row's taxon / total read count X 100. This column only factors in reads directly classified to the listed taxon and does not include reads classified to the children of this taxon. The sum of this column is 100.",
        }
    },
    "data": {"Root": {"Abundance": 0.3802136069332063}},
}
```

->

```
|Sample Name|Abundance|
|---|---|
|Root|0.3802|
```

PR #3131, opened 2025-03-22
allow customizing prompt sent to llm

jchorl:
I was playing around in-depth with the new LLM functionality. It's awesome.

I started fiddling with the prompts to see if I could get better outputs. Depending on the domain/use-case, I was able to get better content and formatting. I came to the conclusion that there will not be a one-size-fits-all prompt that works for everyone. So I made it configurable via multiqc config.

Let me know what you think.

### DIFF
--- a/multiqc/config.py
+++ b/multiqc/config.py
@@ -119,6 +119,8 @@ ai_model: str
 ai_custom_endpoint: Optional[str]
 ai_extra_query_options: Optional[Dict[str, Any]]
 ai_custom_context_window: Optional[int]
+ai_prompt_short: Optional[str]
+ai_prompt_full: Optional[str]
 no_ai: bool
 ai_anonymize_samples: bool
 
--- a/multiqc/config_defaults.yaml
+++ b/multiqc/config_defaults.yaml
@@ -70,6 +70,8 @@ ai_model: null
 ai_custom_endpoint: null
 ai_extra_query_options: null
 ai_custom_context_window: null
+ai_prompt_short: null
+ai_prompt_full: null
 no_ai: false
 ai_anonymize_samples: false
 
--- a/multiqc/core/ai.py
+++ b/multiqc/core/ai.py
@@ -180,11 +180,18 @@ class Client:
         self.model: str
         self.api_key: str = api_key
 
+        self.prompt_short = PROMPT_SHORT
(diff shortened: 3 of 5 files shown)


vladsavelyev (approved):
Awesome, thanks a lot!!

Wonder if you could also share prompts that work best for you? I didn't give it a lot of thought and just stuck to what works, so maybe we could iterate on the default prompt a bit.

jchorl:
> Awesome, thanks a lot!!
> 
> Wonder if you could also share prompts that work best for you? I didn't give it a lot of thought and just stuck to what works, so maybe we could iterate on the default prompt a bit.

Yeah, happy to provide input. Some of it was very domain-specific, which actually drove this PR (vs. just tuning the default prompt). I didn't see a future for a one-size-fits-all prompt. I also wanted the ability to rapidly iterate on prompts without needing to patch multiqc each time.

Architecturally, the biggest change is that I gave up on the two-section YAML output from the LLM. I found LLMs were bad at producing valid markdown code within yaml as sometimes they conflict (I think some yaml parsers choke up on `**foo bar**` but this is valid markdown). This makes sense as presumably the LLM is operating on tokens rather than calling a legitimate yaml encoder with proper escaping and all. Many of the LLMs now have first-class support for JSON structured output, and maybe they do more sophisticated encoding there. I didn't bother.

I also found the model was making up directives like `:text[foo]{.text-yellow}`. I iterated on the directives section:

```
Use markdown to format your reponse for readability. Use directive `:span` with pre-defined classes `.text-green`, `.text-red`, and `.text-yellow` to highlight severity, e.g. :span[39.2%]{.text-red}. Use directive `:sample` for any mentioned sample names or sample named prefixes or suffixes with a sample directive, and make sure to use the same color classes for severity, e.g. :sample[A1001.2003]{.text-yellow} or :sample[A1001]{.text-yellow}. Do not put multiple sample names inside one directive. Do not use any other directives beyond `:span` and `:sample`.
```

Again I'm hesitant to propose these changes globally because:
1. MultiQC supports many LLM providers which may behave differently. I use one.
2. MultiQC supports a broad user-base, so optimizing one case may be detrimental to others.
3. LLMs are non-deterministic. So I may see improvement and you may see degradation. It is very hard to globally optimize, benchmark, and show that one approach is objectively better.

PR #3132, opened 2025-03-23
Dont display decimal points on fastp filtering result bar chart

jchorl:
The fastp bar chart did not used to display decimals. I also don't think it should - it handles read counts. I chased the change in behaviour down to 0dcd3f3de0fd53556b5ff29e02142d881e96c32d

Prior to that commit it did not display decimals. I'm not sure how it broke things, it's a rather large commit.

Anyway, at first I tried removing the `float(...)` casts when parsing fastp data, as most of it is ints:

```json
    "filtering_result": {
        "passed_filter_reads": 1488254,
        "corrected_reads": 299344,
        "corrected_bases": 571668,
        "low_quality_reads": 0,
        "too_many_N_reads": 0,
        "too_short_reads": 1396,
        "too_long_reads": 0
    },
```

Even parsing as int, the chart still displayed two decimal places. Maybe barchart defaults to two decimals even when passed ints? You can check the commit history of this PR to see this approach.

So I just configured `tt_decimals` and that fixed things. I reverted back to parsing the values as floats, in case there's good reason for that.

Before:

After:

### DIFF
--- a/multiqc/modules/fastp/fastp.py
+++ b/multiqc/modules/fastp/fastp.py
@@ -247,7 +247,7 @@ class MultiqcModule(BaseMultiqcModule):
         # Parse filtering_result
         try:
             for k in parsed_json["filtering_result"]:
-                self.fastp_data[s_name][f"filtering_result_{k}"] = parsed_json["filtering_result"][k]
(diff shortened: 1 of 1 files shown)


vladsavelyev:
Looks good! Thank you.

Yeah, the logic is not flawless. It makes more sense to check if numbers are ints and automatically display without the decimals...

PR #3137, opened 2025-03-28
sort custom-content by section name by default

jchorl:
Context is all in https://github.com/MultiQC/MultiQC/issues/3136

Please consider whether this is the right approach. It does mean the report will probably end up with mis-ordered section IDs, e.g. `section-1`, `section-3`, `section-2`, because the IDs are enumerated/counted before the sorting happens. I think that's fine? Seems the sidebar nav still has the sorted ordering.

There's a bunch of ways we could implement this, I went with what I thought was best.


vladsavelyev:
Thanks makes sense! Thanks @jchorl

vladsavelyev (approved):

PR #3140, opened 2025-03-29
Tix pre-configured table sort `defaultsort`

jchorl:
I noticed table-sort on load was broken. Chased it down to this PR: https://github.com/MultiQC/MultiQC/commit/5c347bf1ab89c955f61b91a16e47cdf88fc03e23#diff-9c52baa1b48d5a07609a29efec35c56dfceb2fa317d53de6f8c1491f36a1a417R558

The list was converted to a tuple. Which caused the str serialization to change from `[[1, 1]]` to `[(1, 1)]`. The charting library doesn't know how to read the tuple format. This fixes.

Note the sort direction on the bar column here:

See before:

After:

mqc config:
```
custom_plot_config:
  general_stats_table:
    defaultsort:
      - column: bar
        direction: desc
```

input:
```
{
    "plot_type": "generalstats",
    "id": "Foo",
    "data": {
        "org": {
            "foo": 100.0,
            "bar": 34,
            "baz": 1637
        }
    }
}
```

### DIFF
--- a/multiqc/plots/table.py
+++ b/multiqc/plots/table.py
@@ -641,7 +641,8 @@ def _get_sortlist(dt: DataTable) -> str:
         direction = 0 if d.get("direction", "").startswith("asc") else 1
         sortlist.append((idx, direction))
 
-    return str(sortlist)
+    sortlist_formatted = [list(t) for t in sortlist]
+    return str(sortlist_formatted)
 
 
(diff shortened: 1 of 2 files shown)


vladsavelyev:
Thank you so much Josh for identifying and fixing this.
It's a bummer it was broken...

PR #3173, opened 2025-04-30
Make sure multiqc.log has full debug log

vladsavelyev:
When the global log level is INFO, even if the file handler log level is DEBUG, it would be downgraded to INFO. So need to always set the global level to DEBUG, and separately set up the level for console handler.

Also making sure that the log file moved to the data dir _after_ all the last log statements are printed.


ewels (approved):
Great stuff, thanks 🙏🏻

PR #3183, opened 2025-05-08
Config schema docs

vladsavelyev:
Add script to automatically generate docs from config schema. Add config_schema.md


ewels:
The wizard is very cool 🤩 

Some speed thoughts about what's missing:

* Branding 😉 
* Homepage, rather than going straight to the first section (eg. some docs / how to use)
* No progress bar (people shouldn't be trying to set every option)
* Global search - might actually make sense to have everything on one page 🤔 
* Better UX around booleans that are `default: true` (confusing to have an unchecked 'enabled' box)
* Some better docs, I guess in underlying Python code (eg. `ai_summary_full`: _AI summary full_)
* Fix some rendering edge cases (`default: [6 items]`)
* Links out to docs + homepage + etc.

Is the idea that we render the HTML with CI and then it gets hosted as part of the docs? I wonder if we could do it as a fragment somehow to get the docs header in that case. Not the end of the world if not though.

ewels:
@multiqc-bot fix linting

ewels:
@cursor please handle the review comments above.

ewels:
@multiqc-bot fix linting

/review

PR #3192, opened 2025-05-15
AI docs: Clean up around Nextflow configuration

ewels:


pinin4fjords (approved):
Complexities exist around the use of secrets (esp. outside local/ AWS on Platform), but approving.

ewels:
Seeing as the docs already had secrets, I think I'm happy with the change. It's stripped back a bit and I added a warning about it not necessarily working everywhere. Hopefully it can be a pointer towards good practices for folks, even if it doesn't work for everyone everywhere.

PR #3196, opened 2025-05-16
bcftools stats - enable log switch for indel plot

pontushojer:
With most of the indel being short this plot gets hard to read so a log switch would be helpful.

bcftools_stats_indel-lengths


vladsavelyev:
Thank you! Sounds reasonable

PR #3209, opened 2025-05-27
fix: fix VEP module for version 114

achakkas:
**Description of changes:**  
- Updated the VEP MultiQC module to handle new section `[Data version]` in VEP version 114 stats file.
- Improved type handling in the parser to avoid crashes when encountering unexpected value types in new sections in the future 
- Attached a generic example VEP stats file from version 114 (`VEP114_vep.txt`) for use as test data

[VEP114_vep.txt](https://github.com/user-attachments/files/20465682/VEP114_vep.txt)


vladsavelyev:
Nice, thank you!

PR #3215, opened 2025-05-31
fix: type error in python 3.13

nh13:


nh13:
@ewels any chance you could take a look at this?

vladsavelyev:
Fixed with https://github.com/MultiQC/MultiQC/pull/3223

PR #3216, opened 2025-06-02
RseQC - Normalize junction saturation plot

pontushojer:
The junction saturation counts become hard to overview for multiple samples if they vary greatly in the number of junctions. For example, see the plot below generated using the current MultiQC.

rseqc_junction_saturation_plot

The sample with the highest counts obscures all other samples. You could zoom in to observe the other samples, but it is not too practical to evaluate this metric quickly for a large number of samples. 

To make this plot more useful, this PR normalises the junction counts to the total number of junctions, which brings all samples into the same scale to make the plot easier to evaluate

rseqc_junction_saturation_plot (1)

PS. I also removed a hint that is no longer relevant. 


vladsavelyev (approved):
Looks good, thank you! 😌

PR #3225, opened 2025-06-07
Allow multiline info in module sections

vladsavelyev:
Fix https://github.com/MultiQC/MultiQC/issues/3221


ewels (commented):
Nice! Should we do the same behaviour for comments and help text etc?

Or perhaps we should just parse `info` as markdown? I think that the others might do that already.. 🤔  Then we get proper newlines for free, as well as all other markdown features.

PR #3234, opened 2025-06-19
supporting symbols in scatter plot

lucacozzuto:

I modified scatter.py and scatter.js to display symbols as well.


ewels:
@multiqc-bot fix linting

ewels:
Thanks for this! It would be good to also add it to the line graph plot type too, as the two are very similar.

vladsavelyev:
Thanks for the contribution!

I renamed the option to `marker_symbol` for consistency, and cleaned up the code a bit - otherwise looks good 👍

PR #3235, opened 2025-06-21
Use `OrderedDict` for `saved_raw_data_keys` to increase performance

jchorl:

I have a workflow with about 2000 charts in one section. This causes multiqc to keep rolling IDs to give each chart a unique ID in the section. Specifically this logic: https://github.com/MultiQC/MultiQC/blob/581178b725fd185546f8a2cbb56c78df3df5183e/multiqc/base_module.py#L1108

Now the problem: `saved_raw_data_keys` is a list. On the 2000th chart, it loops 2000 times looking up in a list of 1999 elements. I believe this is n^2 complexity. No bueno.

From where this logic [was introduced](https://github.com/MultiQC/MultiQC/pull/3044), it seems that insertion order is important. For this reason, I went with an `OrderedDict` instead of a `set`. I also considered sorting `saved_raw_data_keys` at _read_ time, but I wasn't sure if it needs to be in insertion order or just stably ordered. If the latter, we could just use `sorted(...)` when reading it. If the former, an `OrderedDict` solves it (I think).

The net effect is a task went from 5h23m to 20m.

Side note: I checked the [dev docs](https://docs.seqera.io/multiqc/development/contributing) for instructions on how to lint/typecheck/test. I didn't find anything. Forgive me if this is documented somewhere.

### DIFF
--- a/multiqc/base_module.py
+++ b/multiqc/base_module.py
@@ -1110,7 +1110,7 @@ class BaseMultiqcModule:
             i += 1
 
         # To map back keys data to specific module
-        report.saved_raw_data_keys.append(fn)
+        report.saved_raw_data_keys[fn] = None
 
         # Save the file (usualy TSV)
         report.write_data_file(data, fn, sort_cols, data_format)
--- a/multiqc/plots/table_object.py
+++ b/multiqc/plots/table_object.py
@@ -1251,7 +1251,7 @@ data-table-anchor="{dt.anchor}" data-violin-anchor="{violin_anchor}" data-toggle
         report.write_data_file(flatten_raw_vals, fname)
         if config.data_dump_file_write_raw:
             report.write_data_file(flatten_raw_vals, fname, data_format="json")
-            report.saved_raw_data_keys.append(fname)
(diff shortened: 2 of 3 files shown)


vladsavelyev:
Nice stuff!

Though regular `dict` is also ordered, why not use it?

jchorl:
> Nice stuff!
> 
> Though regular `dict` is also ordered, why not use it?

Probably more of a philosophical debate. `OrderedDict` is self-documenting and hints that order is required. `dict` is implicitly ordered, but it is much more likely someone would see `my_dict["foo"] = None` and say "hmm what silly programmer did this, why I don't I just convert to a `set`". That is not ordered and would be a bug.

In many languages it is not correct to assume dicts as ordered, so I was just trying to convey that it's required.

vladsavelyev:
That's a very good point! Let's stick to OrderedDict 👍

ewels:
> I have a workflow with about 2000 charts in one section.

What kind of report is this?! 🤯  😅 

Surely this report isn't Human-readable (and must be enormous filesize). I'm very curious what the use case is here..?

PR #3244, opened 2025-07-01
Add `boxplot_boxpoints` config option to pass to box plot

vladsavelyev:
Option to specify how to display outliers in a box plot.

Fix https://github.com/MultiQC/MultiQC/issues/3243 and https://community.seqera.io/t/hide-outliers-in-boxplot/2215

### DIFF
--- a/docs/markdown/development/plots.md
+++ b/docs/markdown/development/plots.md
@@ -393,6 +393,27 @@ html = box.plot(data, pconfig=...)
 Similarly to other plot types, multiple datasets can be passed as `data`, along with
 dataset-specific configurations provided with the `pconfig["data_labels"]` option.
 
+### Box plot outlier display
+
+The global configuration option `boxplot_boxpoints` controls how outliers are displayed in box plots by passing the value directly to Plotly's `boxpoints` parameter. This setting can be configured in your MultiQC config file:
+
+```yaml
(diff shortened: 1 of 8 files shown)


docs/markdown/development/plots.md
```diff
@@ -393,6 +393,27 @@ html = box.plot(data, pconfig=...)
...
+```
+
+Available options (as defined by [Plotly's box trace reference](https://plotly.com/python/reference/box/#box-boxpoints)):
+
+- `"outliers"` (default): Show only outlier points beyond the whiskers
+- `"all"`: Show all data points
+- `"suspectedoutliers"`: Show only suspected outliers (points beyond 1.5 _ IQR but within 3 _ IQR)
+- `false`: Hide all data points, showing only the box and whiskers
+
+This setting applies to all box plots in the report and can be useful for:
```
ewels, line 411: Do we want to make it possible to set this on a per-chart basis? Or rather, is there a reason not to do that?

vladsavelyev: I think it's purely a performance issue, so I'd rather do it automatically based on the number of samples. Similarly to the violin plots: for small datasets <10 samples, show all points; for very large datasets, do not show any points; for the middle range datasets, show outliers only.

jonasscheid:
We included this config in mhcquant MultiQC report. Works! Thanks a lot!

PR #3245, opened 2025-07-02
Allow add any metric from `multiqc_data.json` to general stats

vladsavelyev:
Fix https://github.com/MultiQC/MultiQC/issues/3202

Allow `general_stats_columns` columns that are not necessarily pre-hardcoded in the module. E.g.:

```yaml
general_stats_columns:
  samtools/stats:
    columns:
      reads_mapped_percent:
        title: "% Mapped"
        description: "% Mapped reads from samtools stats"
        hidden: false
      error_rate:
        title: "Error rate"
        description: "Error rate from samtools stats"
        hidden: false
      supplementary_alignments:
        title: "Secondary reads"
        description: "Secondary reads from samtools stats"
        hidden: false
```
Screenshot 2025-07-02 at 13 17 51

### DIFF
--- a/multiqc/plots/scatter.py
+++ b/multiqc/plots/scatter.py
@@ -26,7 +26,7 @@ class ScatterConfig(PConfig):
     marker_line_width: Optional[int] = None
     color: Optional[str] = None
     opacity: Optional[float] = None
-    marker_symbol: Optional[str] = "circle"
+    marker_symbol: Optional[str] = None
 
     def __init__(self, path_in_cfg: Optional[Tuple[str, ...]] = None, **data):
         super().__init__(path_in_cfg=path_in_cfg or ("scatterplot",), **data)


multiqc/modules/samtools/stats.py
```diff
@@ -131,7 +131,10 @@ def parse_samtools_stats(module: BaseMultiqcModule):
     }
 
     # Get general stats headers using the utility function, will read config.general_stats_columns
-    general_stats_headers = module.get_general_stats_headers(all_headers=stats_headers)
+    general_stats_headers = module.get_general_stats_headers(
+        all_headers=stats_headers,
+        sp_key="samtools/stats",
```
ewels, line 136: What's `sp_key`? Is this something that needs to be added to other modules?

ewels (commented):
This needs to be documented 📚

PR #3271, opened 2025-07-11
Seqfu stats

pontushojer:
This PR adds a module for output from the `seqfu stats` (https://telatin.github.io/seqfu2/tools/stats.html) tool. 

image

PR for testdata https://github.com/MultiQC/test-data/pull/352


vladsavelyev:
Looks good! Thank you!

PR #3277, opened 2025-07-22
Add support for optical pooled screening output from cells2stats

ryankelley-lm:
Contains updates for cells2stats module

- Handle "violin" error reported due to formatting string of cell counts
- Add report table for extra-cellular ratio
- Update string used to identify cells2stats input (uses "AnalysisID" field)
- Add plots for upcoming optical pooled screening output from cells2stats


vladsavelyev:
Thank you!

PR #3289, opened 2025-07-31
Missing modules to general stats: Busco, CheckM, CheckM2, GTDB-Tk

harper357:
tl;dr: This is an update to 4 modules that where nothing was being added to the general stats table.

In working on [nf-core/mag PR#841](https://github.com/nf-core/mag/pull/841), I noticed that some of the modules were not adding anything to the general stats table. Here I just add some code to change that.

I know updating multiple modules isn't ideal, but given it is just a minor update to them and it is the same general thing in all of them, I thought it might be ok. I can split this into multiple PRs if needed. 

### DIFF
--- a/multiqc/modules/busco/busco.py
+++ b/multiqc/modules/busco/busco.py
@@ -65,6 +65,39 @@ class MultiqcModule(BaseMultiqcModule):
                 plot=self.busco_plot(lin),
             )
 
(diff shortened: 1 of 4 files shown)


vladsavelyev:
Good catch and good PR! Thank you 🙏

PR #3325, opened 2025-09-02
Add automated module generation
  system

edmundmiller:
Adds automated module creation capabilities through a /new-module Claude Code command and weekly workflow processing. The system
  analyzes high-priority module requests, generates initial implementations following MultiQC patterns, creates comprehensive test suites, and opens
  draft PRs with generated code. Includes rate limiting, dry-run capabilities, and integration with the triage system for seamless workflow automation.

Issue #1949, opened 2023-07-11
umi-tools extract whitelist and count module


HenriettaHolze:
### Name of the tool

umi-tools

### Tool homepage

https://github.com/CGATOxford/UMI-tools

### Tool description

A tool for dealing with UMIs

### Tool output

whitelist
```
# UMI-tools version: 1.1.4
# output generated by whitelist --stdin test_sc_R1.fastq.gz --bc-pattern=CCCCCCCCCCCCCCCCNNNNNNNNNNNN --set-cell-number 5000 -L test_sc_whitelist.log -S test_sc_whitelist.tsv
# job started at Mon Jul 10 16:39:18 2023 on stpr-res-compute01.unix.petermac.org.au -- 9dd926de-1c6f-4590-ab30-0bb7fdb5caac
# pid: 2801096, system: Linux 3.10.0-1160.45.1.el7.x86_64 #1 SMP Wed Oct 13 17:20:51 UTC 2021 x86_64
# allow_threshold_error                   : False
# blacklist_tsv                           : None
# cell_number                             : 5000
# compresslevel                           : 6
# ed_above_threshold                      : None
# error_correct_threshold                 : 1
# expect_cells                            : False
# extract_method                          : string
# filtered_out                            : None
# filtered_out2                           : None
# ignore_suffix                           : False
# knee_method                             : distance
# log2stderr                              : False
# loglevel                                : 1
# method                                  : reads
# pattern                                 : CCCCCCCCCCCCCCCCNNNNNNNNNNNN
# pattern2                                : None
# plot_prefix                             : None
# prime3                                  : None
# random_seed                             : None
# read2_in                                : None
# short_help                              : None
# stderr                                  : <_io.TextIOWrapper name='<stderr>' mode='w' encoding='utf-8'>
# stdin                                   : <_io.TextIOWrapper name='test_sc_R1.fastq.gz' encoding='ascii'>
# stdlog                                  : <_io.TextIOWrapper name='test_sc_whitelist.log' mode='a' encoding='UTF-8'>
# stdout                                  : <_io.TextIOWrapper name='test_sc_whitelist.tsv' mode='w' encoding='UTF-8'>
# subset_reads                            : 100000000
# timeit_file                             : None
# timeit_header                           : None
# timeit_name                             : all
# tmpdir                                  : None
# whitelist_tsv                           : None
2023-07-10 16:39:18,745 INFO Starting barcode extraction
[...]

PR #3326, opened 2025-09-02
Add support for umi_tools whitelist and count commands

edmundmiller:
Extends the existing UMI-tools module to support whitelist and count commands, addressing GitHub issue #1949 with comprehensive parsing, visualization, and testing.

An example of #3325


ewels:
Needs a corresponding PR to multiqc/test-data as well, as the test data needs to be deposited there.

Can also see in the comments of the issue some suggestions that would improve this. But it's not a bad starting point..

ewels:
@multiqc-bot fix linting

PR #3341, opened 2025-09-09
Add configurable helptext to general statistics table

jchorl:
Similar to other sections, this adds expandable helptext to general statistics.

I explored various options for where to put this config as I was trying to avoid a top-level config. For other sections, it's configured as part of the section itself. Therefore it didn't make sense to put it in the `pconfig` for general stats.

I'm open to other ideas on where to nest this config.


ewels:
@multiqc-bot fix linting

PR #3345, opened 2025-09-12
Refactor BISCUIT module for better consistency with current MultiQC codebase

jamorrison:
The BISCUIT module hasn't been updated in quite a while and with the number of changes made to MultiQC since then, it was starting to not behave as intended.

Some of the more important changes made:
- Improve consistency in how to handle missing data for maintenance purposes.
- Add file tags to `config_defaults.yaml` for removal by MultiQC, rather than the module itself. This will allow MultiQC to handle all of the sample name cleaning and allow user options that affect the sample name to work.
- Change average cytosine retention plots to grouped bar charts, rather than stacked bar charts. This shows each of the cytosine contexts as their own bar, rather than stacking them, which doesn't make sense in this particular context.
- Reduce bloat of functions that exist, but only have `pass` in them. This was due to how the module was previously written. Now, only meaningful functions are included in the module.
- Improve some of the help messages for clarity.
- Change some of the color schemes.
- Change Watson/Crick strand verbiage to OT/CTOT and OB/CTOB.
- Change data file output from default to YAML.
- Add `tests` directory and a first test to ensure test framework works.
- Resolve some bugs that had cropped up.
- Add DOI to module description.

### DIFF
--- a/multiqc/modules/biscuit/biscuit.py
+++ b/multiqc/modules/biscuit/biscuit.py
@@ -413,7 +413,7 @@ class MultiqcModule(BaseMultiqcModule):
             "dup_rate": {"color": "#a50f15", "name": "Duplicate Rate"},
         }
         pconfig = {
-            "id": "biscuit_dup_report",
+            "id": "biscuit-dup-report-plot",
             "cpswitch": False,
             "cpswitch_c_active": False,
             "title": "BISCUIT: Percentage of Duplicate Reads",
@@ -611,7 +611,7 @@ class MultiqcModule(BaseMultiqcModule):
         }
 
         pconfig = {
-            "id": "biscuit_seq_depth",
+            "id": "biscuit-seq-depth-plot",
             "title": "BISCUIT: Sequencing Depth",
             "sort_rows": False,
         }
@@ -656,7 +656,7 @@ class MultiqcModule(BaseMultiqcModule):
         ]
 
         pconfig = {
-            "id": "biscuit_cumulative_coverage",
+            "id": "biscuit-cumulative-coverage-plot",
             "title": "BISCUIT: Cumulative Coverage",
             "ymin": 0,
             "tt_label": "<strong>{point.x}X:</strong> {point.y:.2f}M",
@@ -701,7 +701,7 @@ class MultiqcModule(BaseMultiqcModule):
         ]
 
         pconfig = {
-            "id": "biscuit_retention_cytosine",
+            "id": "biscuit-retention-cytosine-plot",
             "title": "BISCUIT: Retention vs. Base Position in Read",
             "xlab": "Position in Read",
             "xsuffix": "bp",
@@ -739,7 +739,7 @@ class MultiqcModule(BaseMultiqcModule):
         }
 
         pconfig = {
-            "id": "biscuit_retention",
+            "id": "biscuit-retention-plot",
             "cpswitch": False,
             "cpswitch_c_active": False,
             "title": "BISCUIT: Cytosine Retention",


jamorrison:
@ewels It's been a little while since I submitted this PR. Would it be possible to get a review? I believe it was opened prior to the Claude code reviews being implemented. Thanks!

ewels:
Apologies for the delay, you opened the PR just after Vlad left Seqera for a new position - MultiQC is currently back to me trying to find time to review code, so everything is slow again sorry 😞 

Anyway, let's try this and see if it's helpful:

/review

jamorrison:
Thanks for initializing the review @ewels! I addressed the critical bug Claude caught, as well as reducing the duplicated code based on the suggested edit in the code review.

I also modified the initial PR comment to include a mention about updating the output data file to be YAML based on Claude's comment. My understanding is that the CHANGELOG is based on that comment, so that's why I added that fix there. However, I'm happy to make a change elsewhere if that is preferred.

Let me know if there are any other fixes/changes that should be made. Thanks!

ewels:
Hi @jamorrison,

Just taking a look at this. Do we need some new example data? Just looking at what we have currently, I guess it's from 2018 and 2020 so maybe quite out of date..?

https://github.com/MultiQC/test-data/tree/main/data/modules/biscuit

If the output has changed, could you please submit a PR to the test-data repo with some new example data? Then we can be sure that the module works across all versions of Biscuit.

Thanks!

multiqc/config_defaults.yaml
```diff
@@ -371,6 +371,17 @@ fn_clean_exts:
   - ".error.spl"
   - ".error.grp"
   - ".vgstats"
+  - "_mapq_table"
+  - "_strand_table"
+  - "_isize_table"
+  - "_dup_report"
+  - "_cv_table"
+  - "_covdist_all
+  - "_covdist_q40
```
ewels: `fn_clean_exts` removes the string _and anything after it_, so I cut down this list a bit and kept just the prefixes. It'd be great if you could confirm that this still works as expected!

multiqc/modules/biscuit/biscuit.py
```diff
@@ -64,214 +66,210 @@ def __init__(self):
...
-            self.mdata[k] = self.ignore_samples(self.mdata[k])
-
-        n_samples = max([len(self.mdata[k]) for k in self.mdata])
+        n_covdist_samples = 0
+        for file_type, parser_func in file_types:
+            collected_data = {}
+            for f in self.find_log_files(f"biscuit/{file_type}"):
+                parsed_data = parser_func(f["f"], f["fn"])
+                if parsed_data is not None:
+                    s_name = self.clean_s_name(f["s_name"], f)
```
ewels, line 120: Given how the parsing works, I think it might make sense to check ignored sample names here and bail early:

```suggestion
                    s_name = self.clean_s_name(f["s_name"], f)
                    if self.is_ignore_sample(s_name):
                        continue
```

This way, ignored samples don't get added to data sources, and we also avoid weird edge cases to do with overwriting with ignored samples etc.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -64,214 +66,210 @@ def __init__(self):
...
+
+                    if s_name in collected_data:
+                        log.debug(f"Duplicate sample name found in {f['fn']}! Overwriting: {s_name}")
+
+                    # Add source file to multiqc_sources.txt
+                    self.add_data_source(f, s_name=s_name, section=file_type)
+                    collected_data[s_name] = parsed_data
+
+            collected_data = self.ignore_samples(collected_data)
```
ewels: Can remove this if doing in the loop above.

```suggestion
```

multiqc/modules/biscuit/biscuit.py
```diff
@@ -64,214 +66,210 @@ def __init__(self):
...
+            self.biscuit_data[file_type] = collected_data
+
+            # Count the number of coverage distribution samples (denoted by "covdist_*")
+            #
+            # Perform counting here to reduce having to add 12 values together from the biscuit_data
+            # dictionary later
+            #
+            # Short-circuit search at position = 10 (longer than "covdist_", but will ensure that it
+            # is caught if it occurs at the start of the string)
+            if file_type.startswith("covdist_", 0, 10):
```
ewels: Excessive optimisation..
```suggestion
            if file_type.startswith("covdist_"):
```

multiqc/modules/biscuit/biscuit.py
```diff
@@ -64,214 +66,210 @@ def __init__(self):
...
-        # Calculate % duplicated
-        for s_name, dd in self.mdata["dup_report"].items():
+        # Calculate percent duplicated
+        for s_name, data in self.biscuit_data["dup_report"].items():
             if s_name not in pd:
                 pd[s_name] = {}
-            if "all" in dd and dd["all"] != -1:
-                pd[s_name]["dup_all"] = dd["all"]
-            if "q40" in dd and dd["q40"] != -1:
-                pd[s_name]["dup_q40"] = dd["q40"]
```
ewels, line 165: Presumably this `-1` was in the code for a reason, ie. it is sometimes found in the raw data. Is there a reason you removed it here?

multiqc/modules/biscuit/biscuit.py
```diff
@@ -64,214 +66,210 @@ def __init__(self):
...
+        header = {
+            "aligned": {
+                "title": "% Aligned",
+                "description": "Percentage of Reads Aligned",
+                "min": 0,
+                "max": 100,
+                "scale": "YlGn",
+                "suffix": "%",
+                "format": "{:,.2f}",
+                "hidden": False,
```
ewels:
```suggestion
```

multiqc/modules/biscuit/biscuit.py
```diff
@@ -64,214 +66,210 @@ def __init__(self):
...
+            "dup_all": {
+                "title": "Dup. % for All Reads",
+                "description": "Percentage of Duplicate Reads",
                 "min": 0,
                 "max": 100,
+                "scale": "Reds",
                 "suffix": "%",
-                "scale": "RdYlGn",
                 "format": "{:,.2f}",
+                "hidden": False,
```
ewels:
```suggestion
```

multiqc/modules/biscuit/biscuit.py
```diff
@@ -64,214 +66,210 @@ def __init__(self):
...
-        pheader = {
+        header = {
+            "aligned": {
+                "title": "% Aligned",
+                "description": "Percentage of Reads Aligned",
+                "min": 0,
+                "max": 100,
+                "scale": "YlGn",
+                "suffix": "%",
+                "format": "{:,.2f}",
```
ewels, line 214: 2 decimal places seems a little excessive for the %age aligned.. But ok, if you think you need it.

multiqc/modules/biscuit/biscuit.py
```diff
@@ -64,214 +66,210 @@ def __init__(self):
...
+        pheader1 = {
             "opt_align": {"color": "#1f78b4", "name": "Optimally Aligned Reads"},
             "sub_align": {"color": "#a6cee3", "name": "Suboptimally Aligned Reads"},
             "not_align": {"color": "#b2df8a", "name": "Unaligned Reads"},
         }
 
-        pconfig = {
-            "id": "biscuit-mapping-overview-plot",
+        pconfig1 = {
+            "id": "biscuit_mapping_overview",
```
ewels: You have duplicate IDs apart from `_` / `-` separators here, which is a bit of a bad code smell. I think that the previous `-plot` suffix is better:

```suggestion
            "id": "biscuit-mapping-overview-plot",
```

multiqc/modules/biscuit/biscuit.py
```diff
@@ -286,28 +284,19 @@ def chart_align_mapq(self):
...
-
-        # Calculate the % aligned for each mapping q score
-        pd_mapq = {}
-        for s_name, dd in self.mdata["align_mapq"].items():
-            if len(dd) > 0:
-                pd_mapq[s_name] = dd["mapqs"]
-
-        pconfig = {
+        pconfig2 = {
             "id": "biscuit_mapq",
```
ewels:
```suggestion
            "id": "biscuit-mapq-plot",
```

multiqc/modules/biscuit/biscuit.py
```diff
@@ -321,141 +310,58 @@ def chart_align_mapq(self):
...
-            "rf": {"color": "#46B29D", "name": "rf: Crick-Aligned, Watson-Bisulfite Conversion"},
-            "rr": {"color": "#324D5C", "name": "rr: Crick-Aligned, Crick-Bisulfite Conversion"},
+            "ff": {"color": "#F53855", "name": "OT: Original Top strand"},
+            "fr": {"color": "#E37B40", "name": "CTOT: Complement to the Original Top strand"},
+            "rf": {"color": "#46B29D", "name": "CTOB: Complement to the Original Bottom strand"},
+            "rr": {"color": "#324D5C", "name": "OB: Original Bottom strand"},
         }
         pconfig = {
-            "id": "biscuit_strands",
+            "id": "biscuit_strand",
```
ewels:
```suggestion
            "id": "biscuit-strand-plot",
```

multiqc/modules/biscuit/biscuit.py
```diff
@@ -321,141 +310,58 @@ def chart_align_mapq(self):
...
-                pd_r[s_name] = dd["readcnt"]
+        """Chart _isize_table.txt"""
+        pd1 = {}
+        pd2 = {}
+        for s_name, d in self.biscuit_data["align_isize"].items():
+            pd1[s_name] = d["percent"]
+            pd2[s_name] = d["readcnt"]
 
         pconfig = {
             "id": "biscuit_isize",
```
ewels:
```suggestion
            "id": "biscuit-isize-plot",
```

multiqc/modules/biscuit/biscuit.py
```diff
@@ -1219,7 +752,7 @@ def chart_read_avg_retention_rate(self):
             "ymin": 0,
             "ymax": 100,
             "y_clipmax": 110,
-            "stacking": None,
+            "stacking": "group",
```
ewels, line 751: 👍🏻

ewels:
Ok, I'm happy with this PR - many thanks for the contribution!

I will merge now as-is, however if you could check my comment above about removing samples with values of `-1`, that'd be great. If you think that this should still be there, please put in a follow-on PR adding it back.

Thanks!

jamorrison:
Thanks @ewels for looking through and merging this PR! I'll have some time to look through the changes you made later this week and can address any of them (including the `-1` samples) if necessary with a new PR.

In terms of the adding an updated test dataset, I think a new one wouldn't hurt. The format of the data hasn't changed since the last set of data; however, there have been some minor tweaks to handle missing data or bad inputs that might be worth including in a test set to make sure those are being properly handled. I'll work on that as well.

ewels:
Sounds great, thank you!

jamorrison:
Quick Summary (detailed discussion below):
1. Significant digit PR coming in the future, but waiting on pre-commit issue resolution.
2. The `-1` came from the MultiQC parsing and not from the raw data. Changes have been made that make it unnecessary for continued inclusion.
3. Another potential change: Adding back in samples with no duplicates to the duplicate rate plot.
4. I will have to first generate QC test data, so I will open a test data PR next week once that's been done.

Let me know your thoughts on the duplicate rate plot. In the meantime, I will keep an eye open for the pre-commit error resolution and proceed with my PR at that point.

I had a chance to look into the changes in more detail. With respect to the number of significant digits on the general stats table, I agree that 2 is a little much. I'm in the process of reducing it down to 1 to match with the percentage shown in the Mapping Overview plot, but I got stuck in the pre-commit stage (see Issue #3416). Once that's resolved, I can open a new PR with that change.

To the point about the `-1`, I had to do some digging to figure out why that existed in the first place. The duplicate report file parsing includes a regex that searches for a specific line format (`(some text): (a number)`). If that regex failed, a `-1` would be added to the `biscuit_data` dictionary for that sample. Typically, this would be due to the number not being filled in because an old version of the BISCUIT QC code relied on `samtools` to count the number of reads and this could end up not filling in the number correctly if no reads met the filtering criteria from `samtools`. Two changes have been made that makes this `-1` unnecessary now. One, BISCUIT QC now has a default value of 0 that is used, so there will always be a number the regex can find. Two, this update of the MultiQC module now returns a `None` value when the regex fails (to maintain backwards compatibility) as well as printing a debug statement so the user is aware the data is incomplete and won't be included.

All that said, I think another change that could be made having looked at the duplicate rate code further is that samples with a duplicate rate of 0 would not be included in the duplicate rate plot, though they will be included in the general stats table. I'm happy to add those samples back into the plot if your preference is to err on the side of showing data in the plots as well as the stats table.

ewels:
Thanks, all sounds sensible! Note that 1 decimal place is the default so you can just remove the format string.

For the duplication - if 0 means no data then omit. If it truly means no duplication (and that was quantified), then include. If it's impossible to know which, then skip I think, as absolute zero for duplication seems very unlikely in the wild, so it's almost certainly a missing data. Would you agree?

jamorrison:
Thanks for all your help with this PR! I opened another PR to address these two issues. I ended up going with editing the help message for the duplicate rate plot to describe why a sample is missing from the plot. Because 0 is the default value, it is technically ambiguous as to whether the duplicate rate is 0 or if it is missing data. However, you are correct that this is unlikely to be a true 0. Therefore, I decided to leave them out of the plot as the samples are still retained in the general stats table if the user needs to see them. I also added this description to the help message so the user knows why they may be missing and that this was an explicit choice that was made. Thanks again for the help!

Issue #3308, opened 2025-08-19
add pandoc to docker image so that -pdf option of multiQC works


ranlib:
### Description of feature

Could you add pandoc to the docker image so that the -pdf option of multiQC works?
Thanks.

vladsavelyev:
Resolved with https://github.com/MultiQC/MultiQC/commit/64dde16c25d2b31d7fd5b13074e95de107ef67b5

PR #3349, opened 2025-09-16
Update Dockerfile to optionally include all LaTeX requirements.

### LINKED ISSUE #3346: docker image v1.31: --pdf does not work, need pdf latex installed
I ran the docker image v1.31 with the `--pdf` option and get the error message shown below:
I actually reported an issue with the `--pdf` option for v1.30, back then the issue was that pandoc was not installed.
Was the `--pdf` option tested for v1.31?

```
MultiQC 🔍 v1.31

            config | Loading config settings from: multiqc_config.yaml
/usr/bin/pandoc
           multiqc | The `pdflatex` tool is required to create a PDF report. Please install LaTeX and try again, e.g. `brew install basictex` on macOS. Alternatively, omit the `--pdf` optionor unset `make_pdf: true` in the MultiQC config.
```

ewels:
Build two docker images, with and without PDF deps.

Should fix #3346 and follow-on from 64dde16c25d2b31d7fd5b13074e95de107ef67b5 (see #3308).

The pandoc requirements are massive, so split into two docker images to avoid bloat for the majority who don't need it:

```console
❯ docker image ls
REPOSITORY             TAG       IMAGE ID       CREATED          SIZE
multiqc-test-pdf       latest    d3ba61b798db   30 minutes ago   3.15GB
multiqc-test-not-pdf   latest    a7a592a123bd   35 minutes ago   1.48GB
```


multiqc/core/write_results.py
```diff
@@ -608,13 +608,25 @@ def _write_pdf(report_path: Path) -> Optional[Path]:
...
+        "-V",
+        "monofont=DejaVu Sans Mono",
+        "-V",
+        "fontsize=10pt",
+        "-V",
         "title=",
+        "-V",
+        "tables=true",
+        "-V",
+        "geometry=margin=0.75in",
```
ewels: Remove second margin
```suggestion
```

ewels:
@claude please open a new PR adding to the docs about the existence of these new docker images.

PR #3352, opened 2025-09-17
Remove bedrock availability check when creating client

jchorl:
Fixes https://github.com/MultiQC/MultiQC/issues/3342

I understand these rigorous checks for auto-detecting whether bedrock is available (I'm not sure how else you'd check - maybe a naive `AWS_DEFAULT_REGION` or `AWS_REGION` env var). Anyway, users can always override the auto-detect logic and bypass it.

However if I tell multiqc to use bedrock, I'd like it to try to use bedrock. It shouldn't decide it doesn't think it'll be able to succeed on that endeavour, I'd hope aws error messages are good enough (no region configured, no permissions, etc).

### DIFF
--- a/multiqc/core/ai.py
+++ b/multiqc/core/ai.py
@@ -489,16 +489,6 @@ class AWSBedrockClient(Client):
     def __init__(self):
         super().__init__()
 
-        # Check Bedrock availability with detailed error reporting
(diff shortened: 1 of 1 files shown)


ewels (approved):
Yup, I'd agree with this. If we get this far then I don't think it needs to fail gracefully.

PR #3364, opened 2025-10-16
use custom plotly bundle to reduce file size

mashehu:

This reduces the file size 3.4MB to 1.1MB (which is 0.3MB smaller than the pre-bundled [plotlyjs-cartesion](https://github.com/plotly/plotly.js/blob/master/dist/README.md#plotlyjs-cartesian)

The bundle file was generated following instructions in https://github.com/plotly/plotly.js/blob/master/CUSTOM_BUNDLE.md and the command `npm run custom-bundle -- --traces bar,scatter,table,violin,heatmap,box` (this covers all plot types listed in https://docs.seqera.io/multiqc/development/plots (with linechart being a subtype of scatter in plotly)

The jump to 3.12 shouldn't include any breaking changes for MultiQC: https://github.com/plotly/plotly.js/blob/master/CHANGELOG.md#300----2025-01-27


ewels:
ooh amazing! 🤩 That's a signifiant improvement, thank you for finding this!

Any chance I can be cheeky and ask for a version of this PR against the `bootstrap-upgrade` branch instead? ( see https://github.com/MultiQC/MultiQC/pull/3264). I've totally rewritten how the front-end works in that branch, complete with an npm build script for each template. It'd be great if we could codify this process in there instead of manually making the bundle - could then update more easily, and add additional plots etc.

ewels:
Apologies, looks like I'm still including the plotly source in that branch - haven't yet gotten around to making it part of the build script. But it's the intention to do that with all / as many of the JS deps as possible.

ewels:
Claude says that it might be possible to include plotly as an npm dependency in the new build system, then do something like the following as an alternative to `npm run custom-bundle`:

```js
import Plotly from 'plotly.js/lib/core';
import Bar from 'plotly.js/lib/bar';
import Scatter from 'plotly.js/lib/scatter';
import Table from 'plotly.js/lib/table';
import Violin from 'plotly.js/lib/violin';
import Heatmap from 'plotly.js/lib/heatmap';
import Box from 'plotly.js/lib/box';

Plotly.register([Bar, Scatter, Table, Violin, Heatmap, Box]);

export default Plotly;
```

> With Vite's tree-shaking, this will create a custom bundle containing only those traces, without needing any build scripts. This is the cleanest approach and most maintainable.

Not sure if this'd work or not, but if it does that feels nice and simple - no need to clone the Plotly repo as a dependency or anything, can just run `npm run build` in the template and it'd automagically do all the things.

mashehu:
From what i gathered, treeshaking is not working for plotly, that's why we need the bundling.

PR #3373, opened 2025-10-24
Optimisations for loading large numbers of parquet files

ewels:
PR to address https://github.com/MultiQC/MultiQC/issues/3350 - code changes made with Claude

* Excessive Debug Logging (Primary Issue)
    * The code logged every section being processed across all modules
    * With 100 parquet files × ~9 sections each, this generated ~65,000 log lines
    * Each log write to disk compounds the I/O overhead
    * This created an O(N²) effect where later files took exponentially longer
* Expensive String Operations (Secondary Issue)
    * `.strip()` calls on potentially large HTML content strings
    * String comparisons between sections
    * String concatenations when merging content
    * Creating new Section objects instead of reusing existing ones

I'm not super familiar with this code as I didn't write it to begin with, so needs some caution before merging to ensure it's still working as expected.


ewels:
I think that Claude was well off the mark with this PR. I tried doing some benchmarking locally with and without the changes and it made essentially zero difference. Trying again with a different approach.

PR #3376, opened 2025-10-25
Move most of xenium into a plugin

ewels:
Goes together with the new https://github.com/MultiQC/xenium-extra

Report should remain unchanged when both are installed.

### DIFF
--- a/multiqc/modules/xenium/xenium.py
+++ b/multiqc/modules/xenium/xenium.py
@@ -82,9 +82,10 @@ class MultiqcModule(BaseMultiqcModule):
         # Check for QC issues and add warnings
         self.check_qc_warnings()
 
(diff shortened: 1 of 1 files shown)


ewels:
@felixkrueger - please note that I've changed how the xenium MultiQC module works in this PR.

Please install both multiqc and the xenium plugin now:

```bash
pip install multiqc multiqc-xenium-extra
```

Then it should continue to work exactly as before.

Without the plugin, you'll now just get a very lightweight report with a handful of columns in general stats.

I've done this to avoid some of the gnarlier imports in core MultiQC, as well as taking that code out that wasn't really following the MultiQC philosophy.

One advantage of this approach is that the plugin can modify the MultiQC config, so it now pushes up the filesize log limit for you automatically.

https://github.com/MultiQC/xenium-extra

PR #3384, opened 2025-11-10
Add RiboTish module for Ribo-seq quality metrics

pinin4fjords:
## Summary
Adds a new MultiQC module for Ribo-TISH, a tool for identifying translated ORFs from Ribo-seq data.

## Features
- **Reading Frame Proportions Plot**: Stacked bar chart showing Frame 0/1/2 proportions across read lengths (25-34nt)
  - Uses `sample_groups` feature (#3404) for visual grouping by read length
  - Reference lines at 33.33% and 66.67% (random distribution thresholds)
- **Read Length Distribution**: Line graph and heatmap views
- **General Statistics**: 7 key metrics including weighted Frame 0 proportion, optimal range percentage, and peak read length

## Module Details
- Parses `*_qual.txt` output files from Ribo-TISH
- Frame 0 > 70% indicates good quality Ribo-seq data
- Optimal read length range: 28-30nt (typical ribosome footprint)

## Dependencies

⚠️ **Requires #3404** - the `sample_groups` visual grouping feature for bar graphs.

## Files Changed
- `multiqc/modules/ribotish/` - New module
- `multiqc/config_defaults.yaml` - Added `_qual` to filename cleaning
- `multiqc/search_patterns.yaml` - Added file pattern
- `pyproject.toml` - Added entry point

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/multiqc/modules/ribotish/tests/test_ribotish.py
+++ b/multiqc/modules/ribotish/tests/test_ribotish.py
@@ -110,8 +110,8 @@ class TestRibotishParsing:
         """Test parsing a valid qual file."""
         module = run_ribotish_module(VALID_QUAL_BASIC)
 
-        assert "test_qual" in module.ribotish_data
-        data = module.ribotish_data["test_qual"]
+        assert "test" in module.ribotish_data
+        data = module.ribotish_data["test"]
 
         assert 28 in data
         assert 29 in data
@@ -124,17 +124,17 @@ class TestRibotishParsing:
         """Test that $ prefix on line 4 is properly stripped."""
         module = run_ribotish_module(VALID_QUAL_WITH_DOLLAR)
 
-        assert "test_qual" in module.ribotish_data
-        assert 28 in module.ribotish_data["test_qual"]
-        assert 29 in module.ribotish_data["test_qual"]
+        assert "test" in module.ribotish_data
+        assert 28 in module.ribotish_data["test"]
+        assert 29 in module.ribotish_data["test"]
 
     def test_single_read_length(self, run_ribotish_module):
         """Test handling of single read length."""
         module = run_ribotish_module(VALID_QUAL_SINGLE_LENGTH)
 
-        assert "test_qual" in module.ribotish_data
-        assert len(module.ribotish_data["test_qual"]) == 1
-        assert 28 in module.ribotish_data["test_qual"]
+        assert "test" in module.ribotish_data
+        assert len(module.ribotish_data["test"]) == 1
+        assert 28 in module.ribotish_data["test"]
 
 
 class TestRibotishMalformedFiles:
@@ -183,7 +183,7 @@ class TestRibotishFrameCalculations:
         """Test that frame proportions sum to 1.0 for each length."""
         module = run_ribotish_module(VALID_QUAL_BASIC)
 
-        sample_name = "test_qual"
+        sample_name = "test"
         for length, props in module.frame_proportions[sample_name].items():
             total_prop = props["f0_prop"] + props["f1_prop"] + props["f2_prop"]
             assert abs(total_prop - 1.0) < 0.001, f"Proportions don't sum to 1.0 for length {length}"
@@ -192,7 +192,7 @@ class TestRibotishFrameCalculations:
         """Test that frame proportions are calculated correctly."""
         module = run_ribotish_module(VALID_QUAL_SINGLE_LENGTH)
 
-        sample_name = "test_qual"
+        sample_name = "test"
         props = module.frame_proportions[sample_name][28]
 
         # counts are [100, 20, 30], total = 150
@@ -205,7 +205,7 @@ class TestRibotishFrameCalculations:
         """Test handling of all zeros in counts."""
         module = run_ribotish_module(VALID_QUAL_ALL_ZEROS)
 
-        sample_name = "test_qual"
+        sample_name = "test"
         for length in [25, 26, 27]:
             props = module.frame_proportions[sample_name][length]
             assert props["f0_prop"] == 0


pinin4fjords:
## Example Report

I've generated an example MultiQC report using the test data to demonstrate the RiboTish module output.

The report shows:
- **Reading Frame Proportions** bargraph with bars grouped by read length (25-34nt)
- **Read Length Distribution** heatmap showing percentage of reads at each length
- **General Statistics** table with 7 key quality metrics

You can view the full interactive report here (uploading as attachment).

pinin4fjords:
## Example Report Attached

View the example report:  [ribotish_example_report.html](https://github.com/user-attachments/files/23460461/ribotish_example_report.html)


The report demonstrates:
- ✅ Reading frame proportions grouped by read length with visual spacers
- ✅ Heatmap showing read length distribution
- ✅ General statistics with 7 quality metrics
- ✅ All three test samples processed correctly

multiqc/modules/ribotish/ribotish.py
```diff
@@ -0,0 +1,398 @@
...
+            # Convert to valid Python dictionary format
+            input_string = input_string.replace("{", "").replace("}", "")
+            input_string = re.sub(r"(\d+):\s*\[", r'"\1": [', input_string)
+            input_string = re.sub(r"\s+", " ", input_string)
+            input_string = "{" + input_string.strip() + "}"
+
+            # Parse the dictionary
+            input_string = input_string.replace("'", '"')
+            input_string = re.sub(r"(\d+),\s*(?=\d)", r"\1, ", input_string)
+            ribo_seq_counts = ast.literal_eval(input_string)
```
ewels, line 92: Parsing as an abstract syntax tree? 🤔

multiqc/modules/ribotish/ribotish.py
```diff
@@ -0,0 +1,398 @@
...
+            self.frame_proportions[sample_name] = {}
+
+            for length, counts in ribo_seq_counts.items():
+                total = sum(counts)
+                if total > 0:
+                    self.frame_proportions[sample_name][length] = {
+                        "f0_prop": counts[0] / total,
+                        "f1_prop": counts[1] / total,
+                        "f2_prop": counts[2] / total,
+                        "total": total,
```
ewels: I think that these are ints, so you're probably going to get rounded numbers here. Worth checking and possibly forcing to be a float (probably enough to force just the total to be a float).

multiqc/modules/ribotish/ribotish.py
```diff
@@ -0,0 +1,398 @@
...
+                        "Frame 1": props["f1_prop"] * 100,
+                        "Frame 2": props["f2_prop"] * 100,
+                    }
+
+        # Create configuration
+        pconfig = {
+            "id": "ribotish_frame_proportions",
+            "title": "RiboTish: Reading Frame Proportions by Read Length",
+            "ylab": "Proportion of Reads (%)",
+            "cpswitch_counts_label": "Counts",
```
ewels: Can turn off the percentage button here, rather than labelling the counts which I don't think does anything any more 🤔

multiqc/modules/ribotish/ribotish.py
```diff
@@ -0,0 +1,398 @@
...
+                    "Frame 0": 0,
+                    "Frame 1": 0,
+                    "Frame 2": 0,
+                }
+
+            for sample_name in sorted(self.frame_proportions.keys()):
+                if length in self.frame_proportions[sample_name]:
+                    props = self.frame_proportions[sample_name][length]
+                    # Create a combined key showing sample and read length (bolded)
+                    combined_key = f"{sample_name} [<b>{length}nt</b>]"
```
ewels: HTML in labels like this works? huh!

multiqc/modules/ribotish/ribotish.py
```diff
@@ -0,0 +1,398 @@
...
+                    plot_data[combined_key] = {
+                        "Frame 0": props["f0_prop"] * 100,  # Convert to percentage
```
ewels: My paranoid brain wants to use a float here again.
```suggestion
                        "Frame 0": props["f0_prop"] * 100.0,  # Convert to percentage
```

multiqc/modules/ribotish/ribotish.py
```diff
@@ -0,0 +1,398 @@
...
+        for i, length in enumerate(all_lengths):
+            # Add spacer before each group (except first) for visual separation
+            if i > 0:
+                # Use spaces to create unique but minimal spacer keys
+                spacer_key = " " * i
+                plot_data[spacer_key] = {
+                    "Frame 0": 0,
+                    "Frame 1": 0,
+                    "Frame 2": 0,
+                }
```
ewels: ok so you get the groups by adding in empty samples? Simple, I like it.

It'll make the data export a bit ugly, but that's probably tolerable I guess.

multiqc/modules/ribotish/ribotish.py
```diff
@@ -0,0 +1,398 @@
...
+                if length in self.frame_proportions[sample_name]:
+                    count = self.frame_proportions[sample_name][length]["total"]
+                    percentage = (count / sample_total * 100) if sample_total > 0 else 0
+                    row.append(percentage)
+                else:
+                    row.append(0)
+            heatmap_data.append(row)
+
+        # Create heatmap configuration
+        pconfig = {
```
ewels: Would benefit from a better colour scale here. A single colour gradient to indicate intensity / count.

multiqc/modules/ribotish/ribotish.py
```diff
@@ -0,0 +1,398 @@
...
+                "format": "{:,.0f}",
+            }
+        )
+
+        # Peak length
+        headers["peak_length"] = ColumnDict(
+            {
+                "title": "Peak Length",
+                "description": "Read length with highest total read count",
+                "suffix": "nt",
```
ewels: I'd prefer putting units in the header here and just having numbers.

multiqc/modules/ribotish/ribotish.py
```diff
@@ -0,0 +1,398 @@
...
+            best_f0_length = max(length_props.keys(), key=lambda k: length_props[k]["f0_prop"])
+            best_f0_prop = length_props[best_f0_length]["f0_prop"] * 100
+
+            # Find length with highest total count
+            peak_length = max(length_props.keys(), key=lambda k: length_props[k]["total"])
+
+            # Calculate read length range
+            min_length = min(length_props.keys())
+            max_length = max(length_props.keys())
+            length_range = f"{min_length}-{max_length}nt"
```
ewels: As above, I'd prefer omitting the `nt` and having that in the column header.

multiqc/modules/ribotish/ribotish.py
pinin4fjords: Ribotish outputs actual python syntax (go figure). But I've simplified, I think some of what was there wasn't required.

pinin4fjords:
> The grouping by adding an empty sample is a little unconventional, but works. Only other solution that I can think of is it add a [background bands and / or lines](https://docs.seqera.io/multiqc/development/plots#background-bands-and-lines). I'm actually not certain if this works with bar plots or not, but I don't see why we couldn't add that functionality if it doesn't already exist. Might be nicer than the "empty sample" hack.

What I really want is [Plotly subplots](https://plotly.com/python/subplots/#subplots-with-shared-xaxes), but I think that involves more re-engineering than either of us would be up for (note: not a challenge ;-) ). 

I did try the lines solution, and it kind of worked, but it made the plot harder to read than the empty sample hack, because it doesn't actually induce any separation between the groups (sorry, didn't save the trial). Hacky as it is, I think this might be the most feasible solution right now.

I would love the subplots, but maybe that's a bigger thing than this PR.

pinin4fjords:
Note: i'm also addressing feedback from: https://nfcore.slack.com/archives/C05785S3TMX/p1762774182023309

pinin4fjords:
Current output: [multiqc_report.html](https://github.com/user-attachments/files/23482678/multiqc_report.html)

multiqc/modules/ribotish/ribotish.py
pinin4fjords: I genuinely think that, hacky as it is, this provides the most pleasing output right now.

multiqc/modules/ribotish/ribotish.py
pinin4fjords: I was struggling a bit to name the groups, this was the best I came up with. 

In an ideal world we'd have a single label per group (the read length), and then be able to mouse over to get the individual sample names. I don't _think_ that's doable currently, but point me the right way if it is!

ewels:
Hopefully fixed the CI error. Now time for this:

/review

ewels:
```
[2025-11-12 10:59:55] multiqc.multiqc                                    [ERROR  ]  2 linting errors:
LINT: >ribotish/ribotish.py< HTML ID was a duplicate (ribotish_frame_proportions-1) ## 
LINT: >ribotish/ribotish.py< HTML ID was a duplicate (ribotish_read_length_heatmap-1) ## 
```

Need unique anchor IDs

pinin4fjords:
Note: need a decision from @ewels  on https://github.com/MultiQC/MultiQC/pull/3386 before merging that and then merging here. 

Phil commented that my first spacing hack would impact on the data exported by multiqc, so I implemented a marginally less hacky method that doesn't require the plugin to do any hackery, and which prevents pollution of the data (filtering before export).

pinin4fjords:
Now using a much less hacky grouped bar graph implementation in https://github.com/MultiQC/MultiQC/pull/3404

multiqc/modules/ribotish/ribotish.py
ewels: Wow! That's a new one 😆

ewels:
Repeating the claude review after changes, in case it picks up anything I miss:

/review

PR #3406, opened 2025-12-02
[nonpareil] Remove prefix to be consistent with other modules like fastqc

fgvieira:

remove prefixes to data keys, to be consistent with other modules (e.g. fastqc).

### DIFF
--- a/multiqc/modules/nonpareil/nonpareil.py
+++ b/multiqc/modules/nonpareil/nonpareil.py
@@ -70,7 +70,7 @@ class MultiqcModule(BaseMultiqcModule):
 
         # Add versions
         for s_name, data in self.data_by_sample.items():
-            self.add_software_version(str(data["nonpareil_version"]), s_name)
+            self.add_software_version(str(data["version"]), s_name)
 
         # Add general stats
         self.nonpareil_general_stats_table()
@@ -118,9 +118,6 @@ class MultiqcModule(BaseMultiqcModule):
                 LRstar = data["LRstar"]
                 assert isinstance(LRstar, list) and len(LRstar) == 0, "there is no model, but LRstar is not empty"
                 data["LRstar"] = np.nan
-            # Add prefix to labels
-            for key in list(data.keys()):
-                data["nonpareil_" + key] = data.pop(key)
 
         return json_raw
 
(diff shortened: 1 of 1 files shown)


ewels:
I understand that this is arguably more aesthetic, but is there an actual need for this change?

I try to avoid changing data keys if at all possible, as it breaks people's downstream data processing.

fgvieira:
It is mostly to be consistent with the other modules, specially when data is imported to MegaQC.

ewels:
I suspect that the reason this was done was that some of the keys are very generic (single letter). This isn't good especially for the table IDs etc.

Whilst I agree that for downstream use the keys are a little ugly, I don't see this as a particularly strong justification for what would be a breaking change for consumers of this data, sorry.

fgvieira:
I agree the names are a bit short, but adding `nonpareil_` before them doesn't help much. :smile: 

It is already clear on the `json` file that they refer to `nonpareil` stats:
```
[...]
    "report_general_stats_data": {
        "nonpareil": {
            "sample1": {
                "nonpareil_file": "sample1.npo",
                "nonpareil_label": "sample1",
                "nonpareil_col": "#02B4BB",
                "nonpareil_L": 63.851,
                "nonpareil_AL": 63.851,
                "nonpareil_R": 217470501,
                "nonpareil_LR": 13885708959.351,
                "nonpareil_overlap": 50,
                "nonpareil_log.sample": 0.7,
                "nonpareil_kernel": "alignment",
                "nonpareil_version": "3.5.5",
                "nonpareil_kappa": 0.5966,
                "nonpareil_C": 0.6236,
                "nonpareil_consistent": true,
                "nonpareil_star": 95,
                "nonpareil_has.model": true,
                "nonpareil_LRstar": 1268036382563.44,
                "nonpareil_modelR": 0.9991,
                "nonpareil_diversity": 21.8074
            }
        }
    },
    "report_general_stats_headers": {
        "nonpareil": {
            "nonpareil_R": {
                "title": "M Reads",
                "description": "Total raw sequences (millions)",
                "modify": null,
                "min": 0,
                "scale": "RdYlGn",
                "shared_key": "read_count",
                "hidden": true,
                "namespace": "Nonpareil",
[...]
```

This extra perfix is redundant and on MegaQC can be a bit misleading:
| `data.id | `data.section` |
| --- | --- |
| nonpareil_C | nonpareil |
| nonpareil_consistent | nonpareil |
| nonpareil_star | nonpareil |
| nonpareil_has.model | nonpareil |
| nonpareil_warning | nonpareil |
| nonpareil_LRstar | nonpareil |
| nonpareil_modelR | nonpareil |

while (e.g.) `FastQC` only has:
| `data.id | `data.section` |
| --- | --- |
| Total Sequences | fastqc |
| Total Bases | fastqc |
| Sequences flagged as poor quality | fastqc |

But I can also change the keys to more explicit names, if you think it would help.

PR #3407, opened 2025-12-02
[fastqc] switch adapter content plot into several subplots

fgvieira:

Currently, `fastqc` module plots all adapter contamination from all samples into a single plot. Furthermore, it removes all data points that are smaller than `0.1%`. This leads to several issues:
- when dealing with a lot of samples, the plot gets quite crowded as each samples can have up to 6 lines (one for each adaper).
- even though `0.1%` is quite low, I don't think it is good policy to alter the input data, specially since it is also changed in the `json` file.
- by keeping all in the same plot, each line is given a new `sample name` (e.g. `sample1 - illumina adapter`, `sample1 - polyg`). I believe this is misleading, as it looks like there are more samples (specially an issue if importing data to MegaQC).

Lastly, formatted the code.


ewels:
I appreciate the effort, but I'm not keen to use subplots like this. The vast, vast majority of people don't click through the secondary plots and that makes it very likely that people won't see important information about adapter contamination. It's also not very pretty / unusable - on the test data repo the tabs immediately went way off the side of the report page with long names.

I have previously considered showing a single line per sample, with the summed adapter contamination value. This would address several of your points I think, but is arguably less useful to folks. I've left it as it is for many years simply because I haven't been convinced that there is a better way.

fgvieira:
Do you think a plot similar to FastQC base composition (heatmap with scatterplot) would be better? This way the user would have a contamination overview on the heatmap and a more detailed view (with each contam shown separately) when clicking.

ewels:
Yes, that's a possibility. Though to be honest I've been trying to get away from those click-to-secondary plots as they're a bit non-standard. But I guess I could use this as an excuse to rethink how they work and make them more standard.

There is also a risk of bloating report sizes, as then we get full datasets for ~10ish adapters per sample in the HTML. But I guess it's not so different to what we have now.

fgvieira:
I tried to make the heatmap for adapter content, but it turned out to be more complex than I anticipated. I thought the heatmap plot function was generic, but it is heavily tailored for the base composition plot.

However, following your comments, I think we could keep the secondary plots as long as we add an extra one for the "Totals". This way, the user can get a quick (and less cluttered) overview of the adapter content and, if needed, can go into the secondary plots to see which specific sequences are present.

EDIT: I just noticed that the module for `AdapterRemoval` uses the exactly same approach! :smile:

fgvieira:
@ewels what do you think about the proposed solution? Would that work?

ewels:
On my phone so haven't checked the code yet sorry. But you're suggesting we have two dataset tabs - one total, one as it is now with the different adapter lines all on one plot?

That sounds reasonable I think..

ewels:
Another thought. How about having two plots:

* line plot, showing summed adapters for samples
* bar plot, where "sample" is adapter type. Show summed read counts across all samples.

By having these two views we would then show all adapter contamination without truncating the data or causing a busy plot / fudging sample names. Then the second plot informs users what kind of contamination is present.

It's not quite as detailed, as it wouldn't highlight if different sample (groups) have different adapter contamination, but that's probably relatively minor 🤔 Could maybe thing about also showing some metrics for each adapter about sample variability to get around that (table instead of bar plot?)

And could even have a config option to opt-in to having secondary dataset plots, one per adapter, if we really want to go all-in.

fgvieira:
Actually the way I was thinking is to have a single plot with several tabs. The first would be the `Total`, so the user can get a quick overview of the contamination (with one line per sample). After that, I thought of a tab for each type of adapter (also with one line per sample).

This way the user has a quick overview of the samples, but can also go in detail if needed.

PR #3422, opened 2025-12-08
Version check: Add installation method to detect installs using `uv`

ewels:
Add detection for when MultiQC is installed using uv (Astral) by checking for the presence of the `uv =` marker in the `pyvenv.cfg` file, which uv adds when creating virtual environments.


ewels:
Wait for database migration and https://github.com/MultiQC/api.multiqc.info/pull/12 to be merged before merging this PR.

PR #3424, opened 2025-12-08
Add toggle to display line graphs as heatmaps

### LINKED ISSUE #3427: feat: Option to replot line graphs as heatmaps
### Description of feature

When there are a lot of samples in a line plot, it can be difficult to read the jumble of lines. The same data can be represented in a heatmap so that data doesn't overlap, which can be easier to view in these cases.

We should already have all of the data we need embedded in the report, so it'd be good to add a button that switches the plot type in the JS.

ewels:
Add a "Heatmap" toggle button to all line graphs that converts the line plot view to a heatmap view:
- Each sample becomes a row in the heatmap
- X-axis values become columns
- Y-values become cell colors using a "Blues" colorscale

Features:
- Toggle between line graph and heatmap views without page reload
- Sort options when in heatmap mode: Original, Alphabetical, Cluster
- Hierarchical clustering (agglomerative, average linkage) implemented in JavaScript for grouping similar samples together
- No backend changes to data structures - conversion happens client-side
- No data duplication in HTML output

Implementation:
- line.js: Add heatmapMode state, prepHeatmapData(), clusterRows(), and buildHeatmapTraces() methods to LinePlot class
- plotting.js: Add click handlers for heatmap toggle and sort buttons
- plot.py: Generate heatmap toggle button for line plots only


ewels:
This PR is very poor, I don't think that Claude is really capable of this kind of thing. The heatmap rendering is crap, and although I tried to give AI a chance by using Antigravity with control of Chrome, it just couldn't figure it out.

Needs writing with some good 'ol fashioned Human fingers I think.

ewels:
@multiqc-bot fix linting

PR #3429, opened 2025-12-09
[nonpareil][adapter removal] Add missing anchors

fgvieira:
Add anchor on `id`, so it is possible to change it from config.

### DIFF
--- a/multiqc/modules/nonpareil/nonpareil.py
+++ b/multiqc/modules/nonpareil/nonpareil.py
@@ -294,13 +294,13 @@ class MultiqcModule(BaseMultiqcModule):
         }
 
         pconfig = {
-            "id": f"{self.anchor}_table",
+            "id": f"{self.anchor}_stats_table",
             "title": "Nonpareil: statistics",
         }
 
         self.add_section(
             name="Statistics",
-            anchor="nonpareil-stats",
+            anchor=f"{self.anchor}_stats",
             description="""
             Nonpareil uses the redundancy of the reads in metagenomic datasets to
             estimate the average coverage and predict the amount of sequences that
@@ -372,7 +372,7 @@ class MultiqcModule(BaseMultiqcModule):
                         )
 
         pconfig = {
-            "id": f"{self.anchor}_plot",
+            "id": f"{self.anchor}_redundancy_plot",
             "colors": data_colors,
             "title": "Nonpareil: Redundancy levels",
             "xlab": "Sequencing effort (Mbp)",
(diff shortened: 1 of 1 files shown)


ewels:
Can you explain a bit more what the problem is that you're solving here please?

fgvieira:
Sorry for the short description, but you were too fast! Here goes a more thorough explanation. :smile: 

## Multiple plots
The immediate idea, was to allow nonpareil (and initially samtools) to have custom plot ids, so that there can be several plots in the same report with meaningful IDs. I think the general problem (as already hinted at in #2203,#2206, #2076, MultiQC/MegaQC#484) is that section `anchor`s and plot `id`s are not standardized:
- some are hard-coded, while others use the `anchor` in the plot ID
- no specific naming structure (so some plots end up with `id="fastqc_sequence_counts_sequence_counts_plot"`)

In the nonpareil module the `id`s were hard-coded, so I added the `anchor` to the `id` and tried to standardized it a bit:
- `anchor`: module name and type of data in the section (e.g. "stats"," saturation", "adapter content"), separated by (e.g.) "-".
- plot `id`: `anchor` and how the data is shown/plotted (e.g. "table", "plot", "barplot"), separated by (e.g.) "_"

This way, if the user specifies a new anchor (e.g. "nonpareil-longread"), the plot `id` is still meaningful ("nonpareil-longread_plot").

## JSON
On the other hand, if the idea is to use `self.anchor` as a prefix to the plot `id`s (which I think it should!), then I'd say it would also make sense to use the same prefix when writing the data file. Right now the module specifies its own prefix (e.g. on "fastqc"):
- if `anchor` defined as "raw-fastqc" in config
- plot `id` for the sequence count becomes "raw-fastqc_sequence_counts_plot"
- raw data on `json` file is stored as "multiqc_fastqc_raw-fastqc", because of `self.write_data_file(dump_data, "multiqc_fastqc")`


## General Suggestion
As a general suggestion, I'd say that both `anchor`, `id`s and `json` keys should be auto-generated by `multiqc` internally, and that the only variable would be a `prefix`. By default, this would be the module name but it could be overwritten by the user (e.g. in the config).
As for the auto-generated IDs, I'd suggest something like:
- plots
    - `anchor` would be `{prefix}_datatype`
    - `id`s would be `{anchor}_plottype`.
    - So, for (e.g.) fastqc sequence counts barplot, it would be:
      - `prefix = "fastqc"`
      - `anchor = "{prefix}_sequence_counts"`
      - `id = "{anchor}_bargraph"`
      - This would lead to `id = "fastqc_sequence_counts_bargraph"` (instead of current "fastqc_sequence_counts_sequence_counts_plot")
- `json` keys, instead of `multiqc_textdefniedbymodule_{anchor}`, would be only `multiqc_{anchor}`


Hope it now makes more sense! :smile:

PR #3432, opened 2025-12-10
Fix mosdepth coverage in general stats if key is missing

jchorl:
Fixes: https://github.com/MultiQC/MultiQC/issues/3431

If e.g. 80% of reads have 50x coverage, and we want to show 40x coverage, we know at least 80% of reads have 40x coverage. This is particularly important in general stats where there is no "connecting line" chart. So if we don't get the exact right coverage increment that is stored, we end up with an empty box in general stats.

Tested locally:

Before:

After:

### DIFF
--- a/docs/markdown/modules/mosdepth.md
+++ b/docs/markdown/modules/mosdepth.md
@@ -54,6 +54,17 @@ mosdepth_config:
     - 30000
 ```
 
+:::info{title="Missing coverage thresholds"}
+
+Mosdepth does omit cumulative coverages at high coverage thresholds. To work around this, MultiQC will
+use the next cumulative coverage available. E.g. if 3301x coverage is present, but 3300x is missing,
+and 3300x is requested, the value of 3301x will be used. This should provide the technically correct value
+because it is a cumulative distribution and mosdepth [only skips values if there are no bases at that coverage
+level](https://github.com/brentp/mosdepth/blob/df8c74a0aabae4ae5e6da1ea0c633b4db829f20a/mosdepth.nim#L439).
+
+For more details, see [this](https://github.com/MultiQC/MultiQC/pull/3432#issuecomment-3672811525) comment.
+:::
+
 You can also specify which columns would be hidden when the report loads (by default, all values are hidden except 30X):
 
 ```yaml


jchorl:
I'm going to look into adding a test for this

jchorl:
Done - tests are added

ewels:
This is definitely better than showing 0% (which is wrong) but it's still showing incorrect data, which could be a little misleading.

Any ideas how we could flag when this is happening? At least I think we should add to the module docstring so that the behaviour ends up in the docs.

jchorl:
Well this sent me down a rabbit hole of getting the nim toolchain up, compiling mosdepth and adding debug statements.

The cause of the missed coverage levels is twofold:
- https://github.com/brentp/mosdepth/blob/df8c74a0aabae4ae5e6da1ea0c633b4db829f20a/mosdepth.nim#L439

    - For very high coverage levels (>300), if there are no bases at that specific coverage (e.g. 3299x), it will skip it.
    - Remember this is a cumulative sum
    - It _is_ possible that there is coverage at e.g. 3300x. Let's say for argument that 20% of the genome is covered at 3300x. 3299x, there's no bases. So the cumulative coverage at 3299x _is_ the next value - 3300x. This is correct.

- https://github.com/brentp/mosdepth/blob/df8c74a0aabae4ae5e6da1ea0c633b4db829f20a/mosdepth.nim#L441

    - If the cumulative sum is so low (below `0.00008`), don't output
    - This isn't really relevant. Because only values above this are reported. I.e. at some point this condition flips. By looking for the next value, it is not possible to get an incorrect one.

Therefore I believe this PR is actually technically correct.

Should this be fixed upstream in mosdepth? Maybe? There has been an open issue for almost 2 years: https://github.com/brentp/mosdepth/issues/222

So now what to do about this? I suggest just commenting a link to this explanation in the multiqc code.

jchorl:
You know, we don't even use mosdepth for other reasons. We just output coverage in a format compatible with mosdepth formatting. I don't really understand why this optimization is the default in mosdepth - it's complex, to save at most a few mb (I think - unless other folks use mosdepth in other ways).

I just updated our code to not apply this optimization. Now we just output coverage at every threshold and avoid this bug altogether. I'm going to close this. I don't necessarily think third-party apps should need to work around this behaviour in mosdepth, it should probably be defaulted off upstream.

ewels:
I'm happy for this PR to continue, I think it's generally useful for folks. If it does get fixed upstream then the code in this PR won't apply, so it won't do any harm.

I'd just like a note in the docs about it basically, then I think it's good to merge.

jchorl:
Added a note to the docs! Of note, I did test this and added tests. However we don't run mosdepth internally and the mosdepth logic is quite complex, so there is some risk I got something wrong. I think values should only be missing at >300x coverage so impact is probably small.

ewels (approved):
Thanks for this. Happy Christmas!

PR #3440, opened 2025-12-16
catch unexpected bedrock response shapes and log

jchorl:
I've seen this case a few times. It's very hard to debug without knowing the actual response from bedrock. If it is unexpected, we should log the whole thing.


ewels (approved):
Thanks!

PR #3447, opened 2025-12-17
Methurator: New module

### LINKED ISSUE #3445: Add support for methurator
### Name of the tool

methurator

### Tool homepage or repository

https://github.com/VIBTOBIlab/methurator

### Tool description

methurator is a Python package designed to estimate sequencing saturation for reduced-representation bisulfite sequencing (RRBS) data.

### Step 1 — Downsample BAM files

The `downsample` command performs BAM downsampling according to the specified percentages and coverage.

```bash
methurator downsample --fasta tests/data/genome.fa tests/data/Ecoli.csorted.bam
```

This command generates three summary files:

- **CpG summary** — number of unique CpGs detected in each downsampled BAM. [methurator_cpgs_summary.csv](https://github.com/user-attachments/files/24210054/methurator_cpgs_summary.csv)

- **Reads summary** — number of reads in each downsampled BAM. [methurator_reads_summary.csv](https://github.com/user-attachments/files/24210056/methurator_reads_summary.csv)

- **Summary yml** - a YAML file which contains all data above in a single file. It also contains run metadata for the sake of reproducibility. [methurator_summary.yml](https://github.com/user-attachments/files/24210055/methurator_summary.yml)

---

### Step 2 — Plot the sequencing saturation curve

Use the `plot` command to visualize sequencing saturation:

```bash
methurator plot \
  --summary tests/data/methurator_summary.yml
```

[example.html](https://github.com/user-attachments/files/24210068/example.html)
Image

### Log filename pattern

methurator_summary.yml

### Data suitable for MultiQC plots

- Sequencing saturation value for the raw BAM file (0% downsampling).
- Sequencing saturation plot for each sample

### Most interesting data for the General Stats table

- Total reads found in the BAM file
- Total CpGs found in the corresponding COV file
- Asymptote estimated by the curve, corresponding to the theoretical maximum number of CpGs
- Sequencing saturation value

### Additional context

Although this tool has been developed only recently and is therefore not yet widely adopted, methurator provides a unique capability: it estimates sequencing saturation specifically for RRBS samples. Assessing sequencing saturation is essential for robust experimental design and data interpretation, yet no existing tool currently addresses this need for RRBS data. In this sense, methurator offers clear value to the entire RRBS community. Moreover, it would naturally complement the functionality of Preseq and could be integrated into the official [nf-core/methylseq pipeline](https://nf-co.re/methylseq/4.2.0).

### Checklist

- [x] I have included example data files (uploaded/zipped, not copy-pasted)
- [x] I have checked that no similar module already exists in the [MultiQC modules list](https://multiqc.info/modules/)
- [x] I have provided the tool's homepage or repository URL
- [x] I understand this request will be prioritized based on tool popularity, community need, and request quality

ewels:
Implements a new module for the Methurator tool which estimates sequencing saturation for reduced-representation bisulfite sequencing (RRBS) data.

The module parses the methurator_summary.yml output file and provides:
- General statistics: saturation percentage, CpG count, asymptote, read count
- Saturation curve plot showing CpG detection across sequencing depths
- Support for multiple minimum coverage thresholds

Closes #3445

### DIFF
--- a/multiqc/modules/methurator/methurator.py
+++ b/multiqc/modules/methurator/methurator.py
@@ -10,6 +10,11 @@ from multiqc.plots.linegraph import LinePlotConfig
 log = logging.getLogger(__name__)
 
 
+def _make_alert(message: str, level: str = "info") -> str:
+    """Create an HTML alert div for empty state messages."""
+    return f'<div class="alert alert-{level}"><strong>{message}</strong></div>'
+
+
 class MultiqcModule(BaseMultiqcModule):
     """
     Methurator is a Python package designed to estimate sequencing saturation for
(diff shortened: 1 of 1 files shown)


ewels:
@multiqc-bot fix linting

ewels:
@edogiuili - what do you think of this as a first attempt?

Example reports:
[Archive 2.zip](https://github.com/user-attachments/files/24222007/Archive.2.zip)
[Archive.zip](https://github.com/user-attachments/files/24222008/Archive.zip)

edogiuili:
@ewels Thanks! It looks really nice! Some remarks:
- In the general statistics I would also add the asymptote (theoretical maximum number of CpGs), to give the users a complete overview.
- The plot is somewhat “compromised” (see attachments): it loses the statistical details that were present in the original interactive HTML file. Is this just because this is an initial test? Or it is not possible to keep the original file within MultiQC?


https://github.com/user-attachments/assets/c0e51e48-5b1d-4dd5-b539-7524a8008db5


https://github.com/user-attachments/assets/833c0365-7c79-4466-8991-98588d92f692

ewels:
I think that the asymptote is already in General Statistics, just hidden by default. We can definitely show it though.

For the plot - we have to assume that folks are running MultiQC with 10k samples. So we can't annotate different asymptotes for every sample and we can't have 2x traces per sample, as it would quickly be unreadable.

I did think about having two dataset tabs at the top for observed / theoretical. But we need datasets for different minimum CpG coverage levels already so I didn't do that in the end. It is still an option though.

Any suggestions for improvement welcome!

multiqc/modules/methurator/methurator.py
```diff
@@ -0,0 +1,314 @@
...
+                "max": 100,
+                "min": 0,
+                "scale": "RdYlGn",
+            },
+            "asymptote": {
+                "title": "Asymptote",
+                "description": "Theoretical maximum number of CpG sites (model estimate)",
+                "format": "{:,.0f}",
+                "scale": "BuPu",
+                "hidden": True,
```
ewels:
```suggestion
```

edogiuili:
Okay, you’re right! Then it’s perfect as it is now. Just one small proposal: do you think it would still be readable if we added the sequencing saturation to each point’s trace? This way, readers can see how the saturation improves with increasing depth.

ewels:
I originally did the plot as %saturation instead of CpG counts, then reverted to counts to match. As we can't have different tabs / datasets, we could just do this as a second plot under the first..

ewels:
Ok @edogiuili - I've had a bit of a play with this, see what you think now.

[Archive.zip](https://github.com/user-attachments/files/24281290/Archive.zip)

CleanShot 2025-12-21 at 22 07 42@2x

edogiuili:
It looks good to me now with the second plot. And help texts are very clear! Thanks a lot!

PR #3454, opened 2025-12-20
fastp: Add limits, update color scale of % PF/% Adapter

schorlton-bugseq:
Most reads should pass preprocessing. If all samples have 0% reads pass, that's bad and currently gets hidden in the fastp module stats. This update sets min/max on some columns and changes the color to reflect that reads not passing filter = bad.


CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 - Glimpse: Add more decimal to general table stats ([#3423](https://github.com/MultiQC/MultiQC/pull/3423))
 - Refactor BISCUIT module for better consistency with current MultiQC codebase ([#3345](https://github.com/MultiQC/MultiQC/pull/3345), [#3426](https://github.com/MultiQC/MultiQC/pull/3426))
 - Add version fetching for HiCUP, QoRTs, QualiMap, RNA-SeQC ([#3420](https://github.com/MultiQC/MultiQC/pull/3420))
+- fastp: add limits to `% PF` and `% Adapter` columns. Change color scale of `% PF` column.
```
ewels, line 32: No need to update the changelog now, we have a script that generates this ahead of releases based on PR titles.

PR #3455, opened 2025-12-21
Fix bug: Toggling highlighting for AI summary sample names didn't work

### LINKED ISSUE #3439: Toggling highlight from AI summary does not toggle as expected
### Description of bug

Initial load:

Image

Then I toggle `test1` by clicking it:

Image

Great. Looking good. Then I click `test1` again:

Image

No bueno.

I _think_ the issue is coming from here: https://github.com/MultiQC/MultiQC/blob/d53ee14d9937a28ed86f043815995551a4a5e042/multiqc/templates/default/src/js/toolbox/highlights.js#L23-L29

I'm not sure if the bug is that `#mqc_col_filters li` can never unselect a sample - i.e. filters never get deleted. The multiple parts of code that deal with highlighting/selecting is pretty confusing.

I again modified the AI code to just return a dummy summary to make reproduction easier:

```diff
diff --git a/multiqc/core/ai.py b/multiqc/core/ai.py
index 1b8cbc92..18b5d9e5 100644
--- a/multiqc/core/ai.py
+++ b/multiqc/core/ai.py
@@ -1040,6 +1040,14 @@ def _save_prompt_to_file(system_prompt: str, prompt: str):
 
 
 def add_ai_summary_to_report():
+    report.ai_global_summary = """<div>
+    <ul>
+    <li><sample data-bs-toggle='tooltip' title='Click to highlight in the report' class='text-green'>test0</sample> is looking good</li>
+    <li><sample data-bs-toggle='tooltip' title='Click to highlight in the report' class='text-green'>test1</sample> is looking okay</li>
+    </ul>
+    </div>
+    """
+
```

```json
{
  "id": "gs",
  "plot_type": "generalstats",
  "pconfig": [
    { "coverage": {} }
  ],
  "data": {
    "test0": {
      "coverage": 40
    },
    "test1": {
      "coverage": 40
    }
  }
}
```

### Before submitting

ewels:
Fixes #3439

The issue was caused by three bugs in the click handler for sample elements:

1. Wrong selector: Used `.close` but the actual button class is `btn-close`
2. Wrong value accessor: Used `.attr("value")` instead of `.val()` which returns initial HTML attribute rather than current input value
3. Form submitted unconditionally: The form was triggered on both add and remove operations, causing an empty filter to be added when removing

The fix:
- Changed selector from `.children(".close")` to `.find(".btn-close")`
- Changed `.attr("value")` to `.val()` for consistency with codebase
- Moved form submission inside the "adding" branch only


ewels:
Tested and working.

PR #3460, opened 2026-01-08
Add project-level metrics for Bases2Fastq

caruizelembioinfo:
Updates to Bases2Fastq Report. Changes include:

Generation of a complete report for Project level samples.

Sequencing Run QC Metrics Table

Polonies should be a count, so it shows 9,067,812 Polonies vs 9.1 Mb
Add cycles per read
Add # of “Reads eliminated” from adapter trimming (found in RunStats)
Add # of “Mean read length” from adapter trimming (found in RunStats)
If possible, Add # of “Min read length” from adapter trimming (found in run manifest)
If possible, “Trim type” from adapter trimming (found in run manifest)
Add Indices table (All indices, Top Unassigned Sequences) show 30 sequences by default
Add a histogram for sample assignment - limit to 1000 samples
Sequencing Run Yield

on Yield change x axis to “Yield”
Run Base Quality Histogram

Fix typo (QualitierQuality per Read)
Add x axis label “Q score”
Quality Metrics by cycle

Add “Read 1 and Read 2 are separated by a red dashed line”
Sample QC Metrics Table

Show mismatch
Per Cycle Base Content

Change description to “Base composition per sample per cycle. Read 1 and Read 2 are separated by a red dashed line”
Change Y axis to “Percentage of total reads”
Per cycle Adapter Content
Add to description: “Read 1 and Read 2 are separated by a red dashed line”
Per Sample GC Histogram

change X axis to “GC Content (%)“
change Y axis to “Percentage of reads that are GC (%)”

Module does not do any significant computational work


rosibaj:
@ewels can we get a review and merge on this?

ewels:
Starting the ball rolling with an automated review:

/review

caruizelembioinfo:
Addressed automated review comments:

Bugs
- cells2stats_bar_plots.py: Wrong logic for empty check - Fixed
- tabulate_manifest_stats: R2 description says R1 - Fixed
- Typo in tabulate_run_stats description - Fixed
- Trailing period in plot name - Fixed

Design Issues
- Random plot IDs break report reproducibility - Fixed (Removed random strings)
- find_log_files called twice to count files - Fixed (Single call collects log files and length is retrieved from list)
- Legacy data structures initialized but unused -  Fixed (Removed ununsed legacy structures)
- Heavy code duplication between run-level and project-level parsers - Fixed (Created helper functions).

Code Quality
- Return type annotation mismatch in _parse_run_project_data -  Fixed
- Potential KeyError in _setup_colors - Fixed, returns default.
- numpy import unused - Fixed
- natsort dependency - Checke presence in toml
- num_lines: 100 for RunManifest.json - Ignored (based on Phil's comment).
- Sorted order not preserved - Fixed (Using natsort consistently).

Minor / Style
- self.add_software_version(None) - No changes made since the call is performed after confirming run data was present.
- The in_project_sample_groups dict built in _setup_colors is always empty for non-project-level paths and gets merged in anyway — slightly misleading. Fixed (the dictionary was removed since even in project-level paths it had the same info as ind_sample_groups.
- search_patterns.yaml: the bases2fastq/manifest pattern uses contents: Settings and num_lines: 100. The Settings key in RunManifest.json appears near the top so 100 lines is probably fine, but worth a comment. Ignored.
- The module docstring in plot_runs.py and plot_samples.py remains as a bare string at module scope (not a real docstring) — per CLAUDE.md, module-level docstrings shouldn't be used; these can just be removed. Fixed (removed)

Missing Tests
- No tests are added for the new project-level parsing paths, the manifest parser, the index assignment parser, or the unassigned sequences parser. Given the complexity of the new code (especially the path-traversal logic in _parse_run_manifest_in_project), at minimum a fixture-based test with sample JSON files would help prevent regressions. Fixed (Added fixture-based tests and integration tests (pending merge in test-data)

ewels:
Thanks for this! Triggering a new review:

/review

caruizelembioinfo:
**### Addressing Comments:**

**Code Quality & Style**
- Modern Python typing: The project's pyproject.toml specifies requires-python = ">=3.8" and CI runs on Python 3.8. The built-in generic syntax (dict[str, Any]) requires 3.9+ and the union syntax (X | None) requires 3.10+ at runtime. Since the project rules also forbid from future import annotations, we cannot use these modern forms without breaking compatibility. We've kept the typing module generics (Dict, List, Optional, Tuple) consistent with the rest of the codebase (e.g. base_module.py).
- Import ordering: Reorganized bases2fastq.py imports into proper PEP 8 blocks (stdlib / third-party / local), with alphabetical sorting within each block.
- Unused import: Removed Union.

**Potential Bugs**
- Fragile split("__"): Added maxsplit=1 to both occurrences -- sample.split("__", maxsplit=1) in _setup_colors and sample_data["SampleID"].split("__", maxsplit=1)[1] in tabulate_index_assignment_stats.
- combined_level comment: Added an inline comment explaining the intentional mix of run-level stats for the run table and project-level samples for per-sample plots.
- add_run_plots / add_sample_plots empty data: Both methods now return early when data is empty, and skip adding sections when plot_html is None.

**Performance**
- find_log_files reuse: Pre-collected log file lists are now stored as instance variables (self._run_level_log_files, self._project_level_log_files) and passed through to all downstream parsers. Manifest log files are also collected once and shared between _parse_run_manifest and _parse_index_assignment. All six parser methods now accept an optional log_files parameter.

**Security**
- Path validation comment: Added a comment in _parse_run_manifest_in_project making it explicit that base_directory is resolved to the run output root (not the project subdirectory), so the manifest is validated against the correct ancestor directory.
- Minor / Nits
- Hardcoded docs URL: Extracted to a module-level constant ELEMBIO_DOCS_URL, referenced in all 5 occurrences (4 error messages + the href in init).
- DEFAULT_MIN_POLONIES: The change from 10,000 to 1,000 is intentional. The previous threshold was too aggressive and excluded samples with very low read counts, including some valid test cases. The lower threshold retains these samples while still filtering out near-empty entries.

ewels:
I added a few commits to tweak some of the behaviour and formatting here.

* Removed the manual `<br>` line breaks in the _Run Manifest Table_. Instead I added some formatting (muted text and a non-breaking space) and added categorical colours for this column to highlight different values.
* Categorical colours for "Adapter Trim Type" too
* Removed samples from "Per Cycle N Content" and "Per Sample Adapter Content" plots if they are under 1%
    * Similar behaviour to FastQC. Saves bloating the report with invisible samples. A note is added listing which samples are hidden.
* Fixed regression in colouring for the older partial data types

I can't push to your fork as it's an organisation instead of individual, so disabled. So instead I merged manually in 2ce8dda73b5a758750016a2667f2bd67cb7fe99c

GitHub doesn't want to show this PR as merged sorry, so I will close it. But it's in!

Please check on `main` and make sure you're happy. Ping me in an issue if not :)

Issue #321, opened 2016-10-21
Salmon metrics incomplete?


tomsing1:
I ran Multiqc (version 0.8.0) on the output for Salmon (version 0.7.2) and it collected metrics from the `libParams/flenDist.txt` file, but nothing else. E.g. the multiqc_data/multiqc_sources.txt file only lists:

> Salmon  all_sections    sample1      /home/sample1/libParams/flenDist.txt

Looking at the [source](https://github.com/ewels/MultiQC/blob/abb6dd331193ed5446ba3353b453e6a76e2edc3d/multiqc/modules/salmon/salmon.py), I think multiqc should also parse the meta information returned by Salmon, e.g. the mapping rate. This information is stored in the`aux_info/meta_info.json` file.

It seems that multiqc is looking for a directory called `aux` - not `aux_info` in [this line](https://github.com/ewels/MultiQC/blob/abb6dd331193ed5446ba3353b453e6a76e2edc3d/multiqc/modules/salmon/salmon.py#L29). Maybe that needs to be corrected?

ewels:
Hi @tomsing1,

Thanks for your report - MultiQC should be picking up both `flenDist.txt` and `meta_info.json` (see it's [search patterns](https://github.com/ewels/MultiQC/blob/abb6dd331193ed5446ba3353b453e6a76e2edc3d/multiqc/utils/search_patterns.yaml#L106-L110)). As you pointed out though - the code looks for the subdirectory path name to get the sample name as this isn't kept in the `meta_info.json` file.

The [test data](https://github.com/ewels/MultiQC_TestData/tree/master/data/modules/salmon) that I have for Salmon has the directory called `aux` (not `aux_info`). Perhaps this is something that the Salmon authors have changed in a recent update? It should be trivial to fix anyway, that line of code doesn't need to be so stringent (or even present).

Phil

tomsing1:
Awesome, now the Salmon metrics are complete 😃

Issue #306, opened 2016-09-22
Module picard RnaSeqMetrics raised an exception


xc611:
```python
Module picard raised an exception: Traceback (most recent call last):
  File "/usr/local/Anaconda/envs_app/multiqc/0.8dev/lib/python3.5/site-packages/multiqc-0.8.dev0-py3.5.egg/EGG-INFO/scripts/multiqc", line 339, in multiqc
    report.modules_output.append(mod())
  File "/usr/local/Anaconda/envs_app/multiqc/0.8dev/lib/python3.5/site-packages/multiqc-0.8.dev0-py3.5.egg/multiqc/modules/picard/picard.py", line 68, in __init__
    n['RnaSeqMetrics'] = RnaSeqMetrics.parse_reports(self)
  File "/usr/local/Anaconda/envs_app/multiqc/0.8dev/lib/python3.5/site-packages/multiqc-0.8.dev0-py3.5.egg/multiqc/modules/picard/RnaSeqMetrics.py", line 81, in parse_reports
    for key in self.picard_RnaSeqMetrics_data.keys():
RuntimeError: dictionary changed size during iteration
```

ewels:
Thanks @xc611 - looks like I never actually ran those lines with my demo data. I'll see if I can figure out how to replicate the bug and then fix it. If you're able to send me your data that would help quite a bit (either here or to phil.ewels@scilifelab.se)

Phil

ewels:
I think I cracked it - it was Python 3 specific which made it trickier to pin down. I'd already written a workaround fix for the same bug three lines below where this broke, so a bit of a fail on my part.

Thanks for reporting! The same problem was present in several other Picard MultiQC modules, so it was a good one to fix.

Phil

xc611:
Thank you

Get Outlook for Android<https://aka.ms/ghei36>



On Thu, Sep 22, 2016 at 4:47 PM -0400, "Phil Ewels" <notifications@github.com<mailto:notifications@github.com>> wrote:


I think I cracked it - it was Python 3 specific which made it trickier to pin down. I'd already written a workaround fix for the same bug three lines below where this broke, so a bit of a fail on my part.

Thanks for reporting! The same problem was present in several other Picard MultiQC modules, so it was a good one to fix.

Phil

-

PR #3462, opened 2026-01-08
New module: Proteinfold  (protein structure AI inference programs - v2 metrics)

keiran-rowell-unsw:
Draft PR in preparation for `proteinfold v2` release -- awaiting [test-data PR](https://github.com/MultiQC/test-data/pull/366) to be merged.

Provides an 'at-a-glance' summary table of quality metrics from predicted structures through [nf-core/proteinfold](https://nf-co.re/proteinfold/latest). 

image

**EDIT**: `pandas` dependency -- didn't realise it wasn't in `MultiQC`. Opening to writing a pure `csv` version if required.

### Justification
Useful for bulk structure prediction and interaction screening. 
Complements the more[ detailed report generated](https://github.com/nf-core/proteinfold/blob/dev/bin/generate_report.py) per sample that display individual structures. 
The multiple model runs of each query sequence and nested underneath each sample name.
Since the pipeline provides publishes simple metrics regardless of AI code serialised format (`proteinfold` PRs [#306](https://github.com/nf-core/proteinfold/pull/306), [#321](https://github.com/nf-core/proteinfold/pull/312), _etc),_ MultiQC should display from any underlying program:
-  [AlphaFold 2 & 3, ESMFold, Rosetta-Fold-All-Atom, RosettaFold2-NA, HelixFold3, ColabFold, Boltz] 

### Suggested label
`module: new` 

- [X] Each table column has a different colour scale to its neighbour, which relates to the data (e.g. if high numbers are bad, they're red)


ewels:
Hi @keiran-rowell-unsw,

Thanks for this PR! I don't think that I've had any like it before for MultiQC modules that correspond to a nf-core (or Nextflow) pipeline.

My first thought is that maybe this would be done with [MultiQC in a python script](https://docs.seqera.io/multiqc/usage/scripts), rather than a module in the core tool. Because MultiQC is bundled in the pipeline and presumably we never expect people to run MultiQC on the pipeline results outside of the pipeline execution context, I wonder if this is simpler - and that a module for all MultiQC users is not needed.

The script can then be bundled in the pipeline, where it would be more self contained, easier for you to write and release just yourself, and doesn't add to the MultiQC distribution size / maintenance.

Phil

keiran-rowell-unsw:
Hi Phil!

Yep that sounds sensible. Definitely, in this case the underlying programs themselves won't print out simple `.tsv` metrics files so any reporting must be downstream from `nf-core/proteinfold`.

I'll take the route of incorporating there rather than core `multiqc -m`.

Best!
Keiran

PR #3470, opened 2026-01-14
add spacing between ai summary and general stats

jchorl:

Before:

After:

I picked `mb-4` to align with other sections. However, note that in other places we use `mb-3`: https://github.com/MultiQC/MultiQC/blob/a1afb2ea2f48edfd3a2a1de983777914b1a32569/multiqc/templates/default/general_stats.html#L18

I don't really understand the difference between `default` vs. `original`, `general_stats.html` vs. `header.html`. All I know is this fixed the visual issue for me. We might want to modify the other templates too.


ewels (approved):
Thanks!

PR #3472, opened 2026-01-16
Fastp: Add % surviving bases

schorlton-bugseq:
Add % surviving bases, which compliments the already present % surviving reads to fastp. Often this can be a useful metric, especially for long reads. I defaulted it to hidden to not change behaviour. I also renamed the `% PF` column to make it less ambiguous.

Tested on the test data and appears to be working.


ewels:
@multiqc-bot fix linting

ewels (approved):
Nice, looks good. Thanks!

PR #3479, opened 2026-01-30
Update methurator module

edogiuili:
In this PR, I update the `methurator` module to support the new v2.1.0 release. This version introduces the `gt-estimator` command, which generates a `methurator_summary.yaml` file with slightly modified fields.

The module now reads the command field from the summary file to determine which command was executed and adjusts the plotting logic accordingly.


multiqc/modules/methurator/methurator.py
```diff
@@ -77,20 +89,31 @@ def __init__(self):
 
         log.info(f"Found {len(self.methurator_data)} samples")
 
-        # Add general stats
-        self._add_general_stats()
+        if command == "methurator downsample":
```
ewels, line 92: What happens if MultiQC finds outputs from both commands? Looking at the code, it seems that whatever the last file that was found wins? I'm not sure if we can do that 😕

edogiuili: Mm, yes, I see what you mean. What if, when both are found, we only use gt-estimator? Otherwise, if that’s not possible, I’d drop the downsample command and provide MultiQC support only for gt-estimator.

edogiuili: Hey @ewels, just a quick heads-up that we're currently integrating methurator into the nf-core/methylseq pipeline (see the [PR here](https://github.com/nf-core/methylseq/pull/609#issuecomment-5015052842)).

From my perspective, I'd be happy to provide support only for gt-estimator, as that's the recommended approach. What do you think?

edogiuili:
Hi @ewels, finally I've decided to drop methurator downsample support and keep only gt_estimator.

In fact it is the recommended command going forward and it's what we've included in nf-core/methylseq pipeline [nf-core/methylseq#609](https://github.com/nf-core/methylseq/pull/609).

Old files aren't silently ignored — they're skipped with a warning naming the file and the reason, so someone on  **methurator downsample** file gets a clear log line rather than an empty report.

The module is now down from 548 to 292 lines. I tested it against four fixture sets — gt_estimator only, both formats together, downsample only (raises ModuleNoSamplesFound as expected), and files with missing/too-old versions — and the parsing and section output are correct in each.

Let me know what you think. Happy to reinstate downsample support if you'd rather keep backwards compatibility!

PR #3480, opened 2026-02-04
fix(cellranger_arc): prevent key collisions between ATAC and GEX metrics

yunhailuo:

# Problem
The Cell Ranger ARC module was showing identical values for ATAC and GEX metrics (e.g., "Valid barcodes", "Sequenced read pairs") when these should be distinct values. This occurred because both ATAC and GEX tables contain metrics with the same column names, causing later values to overwrite earlier ones when merged into a single dictionary.

# Solution
Added namespace prefixes to prevent key collisions:

- Joint metrics (shared across ATAC/GEX) remain unprefixed
- ATAC-specific metrics are now prefixed with ATAC_
- GEX-specific metrics are now prefixed with GEX_

# Changes
- cellranger_arc.py: Parse ATAC, GEX, and joint metrics separately with namespace prefixes, then merge
- utils.py: Updated table_data_and_headers() to accept optional namespace parameter for key prefixing
- test_cellranger_arc.py: Added tests to verify ATAC and GEX metrics have distinct values

# Testing
- Tested with Cell Ranger ARC HTML summary files
- Verified ATAC and GEX metrics now display correctly as distinct values
- Added unit tests to prevent regression


ewels (approved):
Thanks for this! Good catch 🙏🏻

PR #3487, opened 2026-02-17
Fix heatmap axis labels not updating when toggling clustering

nictru:
Attention: The changes, as well as the description below have been heavily LLM-generated. To me the changes seem valid, but I am not familiar with the MultiQC codebase.

# Summary

- Fix a bug where heatmap axis labels become mismatched with data when toggling between "Sorted by sample" and "Clustered" views
- Add Python-side tests validating that clustered data maintains label-data consistency

# Problem

When a heatmap supports clustering, toggling to the "Clustered" view reorders the data matrix (`rows`, `xcats`, `ycats`) correctly, but the axis tick labels remain in the original unclustered order. This causes the wrong label to appear next to the wrong row/column.

The root cause is in `heatmap.py` (`HeatmapPlot.create`), where the Plotly layout's `tickmode` is set to `"array"` with `ticktext` fixed to the original category order. When the JavaScript-side clustering toggle swaps the trace data to the clustered versions, the layout's `ticktext` is never updated to match, so Plotly renders stale labels at each position.

# Fix

In `buildTraces()` (in both `default` and `original` template heatmap JS), sync the layout's `ticktext` and `tickvals` to the current category order before rendering. This runs before `recalculateTicks`, so downstream tick subsampling and highlighting still operate on the correct labels.

```javascript
if (this.layout.yaxis.tickmode === "array") {
  this.layout.yaxis.ticktext = ycats;
  this.layout.yaxis.tickvals = [...Array(ycats.length).keys()];
}
if (this.layout.xaxis.tickmode === "array") {
  this.layout.xaxis.ticktext = xcats;
  this.layout.xaxis.tickvals = [...Array(xcats.length).keys()];
}
```

# Alternative considered

A more "root-cause" fix would be to remove the `tickmode: "array"` / `ticktext` override from the Python side entirely, and rely on Plotly's `type: "category"` axis to derive labels automatically from the trace's `x`/`y` values. This would eliminate the coupling between layout and data ordering altogether.

We chose the JS-side fix instead because:

1. **Lower risk.** The Python-side `tickmode: "array"` path is also used by `recalculateTicks()` to control tick density on large heatmaps, and by the toolbox for sample highlighting. Removing it could introduce regressions in those features.
2. **Minimal scope.** The fix is 8 lines in a single method, easy to review, and doesn't change the data flow between Python and JavaScript.
3. **Correct by construction.** `prepData()` already resolves the current category order (clustered or unclustered, with toolbox filtering applied). The fix simply ensures the layout stays in sync with that resolved state on every render.

### DIFF
--- a/multiqc/templates/original/content.html
+++ b/multiqc/templates/original/content.html
@@ -61,9 +61,10 @@ the output from each module and print it in sections.
                   <div class="well">{{ s['helptext'] }}</div>
                 </div>  
               {% endif %}
-              <div 
-                class="ai-summary ai-local-content" 
-                id="{{ s['anchor'] }}_ai_summary_wrapper" 
+              {% if not config.simple_output %}
+              <div
+                class="ai-summary ai-local-content"
+                id="{{ s['anchor'] }}_ai_summary_wrapper"
                 style="display: none; margin-bottom: 7px; padding-bottom: 4px">
                 <div style="display: flex; justify-content: space-between; align-items: center">
                   <b>AI Summary</b>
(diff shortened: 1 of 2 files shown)


nictru:
@ewels could you have a look at this? Or maybe tell Claude to check? It's annoying me more and more :(

ewels:
Ah well spotted! Not sure why your request didn't work:

/review

PR looks simple enough though - will make sure it gets into the next release 👍🏻 

Thanks!

ewels (approved):
Looks good, thanks!

PR #3492, opened 2026-02-20
Samtools stats: add insert size line graph

chelauk:
This PR adds an **Insert size distribution** section to the Samtools module by parsing `IS` records from `samtools stats` output (`multiqc/modules/samtools/stats.py`) and rendering them as a line plot (`samtools_insert_size`).

  Reason: I routinely use nf-core/sarek for whole genome sequence analysis.  Insert-size shape is important QC signal, and viewing it directly in the MultiQC output makes library-size anomalies easier to detect without manually inspecting raw `samtools stats` files.


ewels (approved):
Looks great, thanks!

PR #3495, opened 2026-02-24
Restore CSS and JavaScript sections from parquet file

sleyn:
When regenerating a MultiQC report using a multiqc.parquet file, the resulting HTML report is missing the FastQC‑specific JavaScript and CSS. As a result "Per Base Sequence Content" section is empty. HTML has all the data, but no JavaScript code for plotting.

This PR solves this issue following:
* Save module canonical name to the parquet file. Currently only user-defined name is saved. For example "FastQC (trimmed)" instead of cannonical "fastqc".
* Use the cannonical name to check if the module has any CSS and JS assets.


sleyn:
It should be labled `bug: core`, but I can't add labels.

sleyn:
Closed as I've found a subsequent bug.

PR #3497, opened 2026-02-26
adding native recognition for the ##Falco 

### LINKED ISSUE #3475: add support for newer and faster qc tools like falco
### Description of feature

Multiqc has failed to compile all the files and fact that MultiQC found 204 files (`searching | ... 100% 204/204`) but reported `"No analysis results found"` usually might mean the files are there, but MultiQC doesn't recognize their filenames or format. Falco creates files named `fastqc_data.txt` inside individual folders and outputs a summary file.

```console
(wgs_l) hemug@Hemants:/mnt/d/results/qc_trimmed_final$ multiqc . -f

/// MultiQC 🔍 v1.33

       file_search | Search path: /mnt/d/manjyot/results/qc_trimmed_final
         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 204/204
           multiqc | No analysis results found. Cleaning up…
```

this is the falco output files:
```
-rwxrwxrwx 1 hemug hemug 2.1M Jan 20 09:39 52A_R1.final.fastq.gz_fastqc_data.txt
-rwxrwxrwx 1 hemug hemug 1.2M Jan 20 09:39 52A_R1.final.fastq.gz_fastqc_report.html
-rwxrwxrwx 1 hemug hemug  562 Jan 20 09:39 52A_R1.final.fastq.gz_summary.txt
-rwxrwxrwx 1 hemug hemug 2.1M Jan 20 09:41 52A_R2.final.fastq.gz_fastqc_data.txt
-rwxrwxrwx 1 hemug hemug 1.2M Jan 20 09:41 52A_R2.final.fastq.gz_fastqc_report.html
-rwxrwxrwx 1 hemug hemug  562 Jan 20 09:41 52A_R2.final.fastq.gz_summary.txt
```

hemant-goyal:
This patch adds native recognition for the ##Falco header and version metadata in the FastQC module.

Resolves #3475.


Stikus:
@hemant-goyal About your problem with incorrect name in "Software Versions" block (https://github.com/smithlabcode/falco/issues/38#issuecomment-3970838374):
Here is function [call](https://github.com/MultiQC/MultiQC/blob/main/multiqc/modules/fastqc/fastqc.py#L338C21-L338C78):
```py
self.add_software_version(version_match.group(1), s_name)
```
And here is function [itself](https://github.com/MultiQC/MultiQC/blob/main/multiqc/base_module.py#L1195):
```py
    def add_software_version(
        self,
        version: Optional[str] = None,
        sample: Optional[str] = None,
        software_name: Optional[str] = None,
    ):
...
        # Use module name as software name if not specified
        if software_name is None:
            software_name = self.name
```
As you can see - `software_name` is not provided and is taken from module name. So you should populate `software_name` variable (probably from header) and it should fix the problem

hemant-goyal:
@Stikus Bulls-eye! I have incorporated the change. I ran a local validation on a mixed batch of reports, and Falco now correctly populates the Software Versions table with its specific version number without interfering with standard FastQC runs.

Stikus:
@hemant-goyal Did you test this code both on FastQC and Falco reports? For me it looks, that now it will always add Falco as software name, even for FastQC.

I think, that you should get software name from header (aka VERSION_REGEX), and pass it to function. It should work in both cases.

multiqc/modules/fastqc/fastqc.py
```diff
@@ -26,7 +26,7 @@
 
 log = logging.getLogger(__name__)
 
-VERSION_REGEX = r"FastQC\t([\d\.]+)"
+VERSION_REGEX = r"(?:FastQC|Falco)\t([\d\.]+)"
```
ewels: I don't think that this is used any more, can be removed.


```suggestion
```

multiqc/modules/fastqc/fastqc.py
```diff
@@ -333,9 +333,13 @@ def parse_fastqc_report(
...
         for line in file_contents.splitlines():
             if line.startswith("##FastQC"):
-                version_match = re.search(VERSION_REGEX, line)
+                version_match = re.search(r"##FastQC\t(.+)", line)
                 if version_match:
                     self.add_software_version(version_match.group(1), s_name)
+            elif line.startswith("##Falco"):
+                version_match = re.search(r"##Falco\t(.+)", line)
+                if version_match:
+                    self.add_software_version(version_match.group(1), s_name, software_name="Falco")
```
ewels: I think we can simplify this slightly:

```suggestion
            m = re.search(r"##(FastQC|Falco)\t([\d\.]+)", line)
            if m:
                self.add_software_version(m.group(2), s_name, software_name=m.group(1))
```

Also brings back the slightly stricter `[\d\.]+` version number match instead of just `.+`, though that shouldn't matter much.

ewels (commented):
Thanks for this! Couple of minor suggestions from looking at the code.

hemant-goyal:
@ewels

awesome suggestions! That dynamic regex group capture is incredibly clean and much better than the bulky if/elif block. 

I've applied and tested the changes on some local files. It worked and therefore pushed the commit, so it should be ready to go! Let me know if you need anything else from my end.

thankyou, as this is a big win for me.

ewels:
Great! Thanks for testing, I wrote it on my phone, always a bit risky 😆

Seeing this, I wonder if we can also set `self.name` and other module metadata, if all discovered logs are Falco 🤔 

We should also mention Falco in the module help (the big docstring at the top).

hemant-goyal:
@ewels
That is a fantastic idea for the UX. If a lab has fully migrated to Falco, the report should definitely reflect that!

I suppose adding a self.fastqc_tools_found set during the parsing loop can fix and at end of __init__, if it only contains "Falco", it dynamically overrides self.name and self.info. If it finds a mix of both in the same run, it sets the name to "FastQC / Falco".

Let me know what you think of this approach. I can also test run it before pushing the commit

ewels:
Yes exactly, that was the kind of thing I was thinking 👍🏻 (not sure it needs to be `self.`, just a local var would be fine).

hemant-goyal:
@ewels
I've implemented the dynamic renaming logic and pushed the commit! While building it, I ran into a few MultiQC lifecycle quirks and wanted to document how I solved them.

1. Updated the module docstring to mention 'Falco' with a link to its repo.
2. Added self._tools_found: Set[str] = set() in __init__, populated via self._tools_found.add() inside parse_fastqc_report.
3. Added logic after the ModuleNoSamplesFound check:

- Pure Falco: self.name = "Falco"
- Mixed: self.name = "FastQC / Falco"
- Pure FastQC: No change.


4. I had to override self.intro directly instead of just self.info/self.href. Because BaseMultiqcModule.__init__ bakes those into a pre-rendered HTML string during initialization, changing them after the parsing loop had no effect on the rendered report.
5. For pure Falco runs, the versions table was showing two rows because add_software_version registers under self.name (which is still "FastQC" at parse time). I fixed this by moving the Falco version entry from the stale "FastQC" key to "Falco" in report.software_versions, then deleting the "FastQC" key.
6. Used del self._tools_found at the end to prevent state lingering, matching the existing del self.fastqc_data pattern.

Let me know if everything looks good

Also, the CI/CD pipeline failed on the prek check, but it looks like an environment caching bug completely unrelated to the PR code. The node_modules cache on the github action runner might be the culprit.

ewels:
Great stuff, thanks! I'm flat out on other projects right now but hope to carve out a block of time for MultiQC soon. I've added this to the milestone for the next release.

In the meantime, I'll kick off an automated review to see if it picks anything up:

/review

hemant-goyal:
@ewels
Thanks for the automated review! The bot caught some great framework-level details. I've addressed the feedback as below

1. Fixed the fallback type for falco_versions from set() to [].
2. Used self._get_intro() to natively generate the HTML intros, and updated self.href to pass a list of both URLs for the mixed-tool mode.
3. Updated self.anchor to "falco" for pure runs, updated the docstring, and set self._tools_found = None for idiomatic cleanup.
4. Root Cause of #3475 Updated multiqc/utils/search_patterns.yaml to use *fastqc_data.txt so Falco files are successfully discovered by the parser in the first place!
5. Note on self.versions (Issue 2), I didn't add a migration for self.versions because it is a flat dictionary (Dict[str, str]). Since I explicitly passed software_name="Falco" to the add_version method during parsing, self.versions["Falco"] is already correctly set without being improperly nested under a "FastQC" parent key.

Let me know if everything looks good

ewels (approved):
Tested, works well - thanks!

hemant-goyal:
@ewels thanks a ton! 😊 
I hope this little contribution will make bioinfo a little easy for someone somewhere.

PR #3504, opened 2026-03-06
fix race causing empty zip files when exporting plots

jchorl:
There's a race when exporting plots:

- `addLogo` takes a callback
- The promise doesn't wait for that callback to finish
- So the zip gets generated before the file has been added to the zip

This change causes the chained `promises` items to only resolve once the `addLogo` callback has been executed (and thus the zip is ready).

I also checked the jszip docs to make sure that `zip.file` is sync.


ewels:
Merged in https://github.com/MultiQC/MultiQC/pull/3555

Thanks!

PR #3505, opened 2026-03-07
Dragen: Fix incorrect average GC% in fastqc for all samples except first in batch

epakarin:
Fixed incorrect average GC% calculated from Dragen fastqc reports for every sample in the multiqc batch except the very first in order

Explanation: The 'data' dict gets populated with [GC%, read_count] pairs for R1 and R2 of a sample. Then average GC% is calculated for the sample from these pairs. But all these tuples remain in the dict when moving to calculate the average for the next sample. As a result, every subsequent sample has its average GC% calculated from its own data appended to that of all preceding samples, resulting in an incorrect value. The final sample's reported average GC% is actually the average GC% of all the samples combined.

Changes:  Clearing the 'data' dict after each iteration.


ewels (approved):
Thanks for this - good catch!

PR #3508, opened 2026-03-14
do not require importlib_metadata backport on recent Python

a-detiste:
subsidiary question ... what would be the baseline ? I picked 3.10 

This is an hybridized version of the much more straightforward Debian patch:

https://salsa.debian.org/med-team/multiqc/-/commit/ff07d319e381b38ff75bca50f4455a952c51b724

### DIFF
--- a/multiqc/__main__.py
+++ b/multiqc/__main__.py
@@ -7,9 +7,11 @@ $ multiqc .
 $ python -m multiqc .
 """
 
-try:
+import sys
+
+if sys.version_info < (3, 10):
     from importlib_metadata import entry_points
-except ImportError:
+else:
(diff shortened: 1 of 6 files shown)


ewels:
Thanks for this! I have historically kept the minimum version as low as I can to maximise compatibility for folks. However, I _hope_ that this is starting to be less important over time as bioinformatics packaging ecosystem improves.

Right now MultiQC is >= Python 3.9 (yes I know 3.9 [is EOL](https://devguide.python.org/versions/) already), I properly bumped that only a few days. 

I think I'd rather hold this PR back until we go to >=3.10 if that's ok, then we can avoid the nasty conditionals and just have a clean change.

PR #3519, opened 2026-03-28
enable defaultsort to include col1 header

jchorl:
Currently trying to apply a default sort including the col1 header fails as it's not technically a column. This fixes.


jchorl:
@multiqc-bot fix linting

ewels (approved):
Thanks!

PR #3528, opened 2026-04-11
Fix merging tables when outputting wide parquets

### LINKED ISSUE #3527: Outputting wide parquet errors if the same sample appears in >2 tables
### Description of bug

Weird quirk with joining across tables. I believe the problem stems from here: https://github.com/MultiQC/MultiQC/blob/97eff3aa26fb2ff024d9c3ed4d8cdcc07bdee78a/multiqc/plots/violin.py#L123-L130

All the rows get `anchor`, `type`, etc.

Check out the join logic: https://github.com/MultiQC/MultiQC/blob/97eff3aa26fb2ff024d9c3ed4d8cdcc07bdee78a/multiqc/core/plot_data_store.py#L103

Notice how we're not joining on those columns. Which means, on the first join, it'll create e.g. `anchor_right` in the resulting df (because it can't merge the column values, it's joining on sample/creation-date, there's no aggregation on other columns).

Now, the second time around that loop, `anchor_right` already exists, driving an error.

As for the fix, I'm not familiar enough with the parquet parts (just getting started). Perhaps we add them as join keys (the values all match - I think this would work). Or perhaps we don't include these keys at all. `anchor`, `plot_type` and `plot_input_data` seem useless, but I'm not sure if omitting them would break parsing.

cat multiqc_config.yaml 
parquet_format: wide

cat foo_mqc.json 
{"data": {"foo": {"bar": "n/a"}}, "id": "myid", "anchor": "myanchor", "plot_type": "table", "file_format": "json", "section_name": "My Section", "description": "my desc"}

cat baz_mqc.json 
{"data": {"foo": {"bar": "n/a"}}, "id": "myid2", "anchor": "myanchor2", "plot_type": "table", "file_format": "json", "section_name": "My Section", "description": "my desc"}

cat bing_mqc.json 
{"data": {"foo": {"bar": "n/a"}}, "id": "myid3", "anchor": "myanchor3", "plot_type": "table", "file_format": "json", "section_name": "My Section", "description": "my desc"}

### MultiQC Error log

```console
polars.exceptions.DuplicateError: column with name 'anchor_right' already exists

You may want to try:
- renaming the column prior to joining
- using the `suffix` parameter to specify a suffix different to the default one ('_right')
```

### Before submitting

jchorl:
I'm just starting to kick the tires with the parquet outputs. I noticed wide output was not properly merging tables.

On `main`, the new test failed:

```
E       polars.exceptions.DuplicateError: column with name 'anchor_right' already ex
ists                                                                                
E                                                                                   
E       You may want to try:                                                        
E       - renaming the column prior to joining                                      E       - using the `suffix` parameter to specify a suffix different to the default 
one ('_right')                                                                      
                                                                                    
/usr/local/lib/python3.13/site-packages/polars/lazyframe/frame.py:2407: DuplicateErr
or    
```

Now it passes, so I think the bug is fixed.

Fixes https://github.com/MultiQC/MultiQC/issues/3527

I noticed a few more issues along the way. In wide format, only floats were persisted. This seemed odd to me. The justification was "for reconstruction". But reconstruction seems to happen from the tall parquet: https://github.com/MultiQC/MultiQC/blob/97eff3aa26fb2ff024d9c3ed4d8cdcc07bdee78a/multiqc/plots/violin.py#L220-L230

This made str, int, bool columns unusable in parquet.

I opted to store the `mod` value instead of raw or formatted. My reasoning:

- I didn't pick formatted because then we can't take advantage of parquet typing
- I didn't pick raw because some modules modify values, and that's what the user expects to see. If we were to take raw values, and let users query parquets, it would be a mess to recreate mod after the fact.

I added tests for all these cases.


jchorl:
@multiqc-bot fix linting

jchorl:
I may have found a bug in this

jchorl:
@multiqc-bot fix linting

jchorl:
@multiqc-bot fix linting

jchorl:
@multiqc-bot fix linting

jchorl:
The lint fixer appears to be borked and I don't have prek etc setup locally

jchorl:
To make this useful, I also had to add `infer_schema_length=100000` to https://github.com/MultiQC/MultiQC/blob/97eff3aa26fb2ff024d9c3ed4d8cdcc07bdee78a/multiqc/plots/violin.py#L174

We could make this a config option or I can just set it on my fork. It would be nice to make it configurable but it's a very niche setting.

multiqc/plots/violin.py
```diff
@@ -113,6 +113,9 @@ def to_wide_df(self) -> Tuple[pl.DataFrame, Set[ColumnKey]]:
         samples_data: Dict[str, Dict[str, Any]] = {}
         metric_col_names = set()
 
+        dt_pconfig_col = f"{self.dt.id} / pconfig".lower()
+        dt_pconfig_json = self.dt.pconfig.model_dump_json()
```
jchorl, line 120: Flagging that these will probably be controversial from size + latency perspective.

Here's the challenge:

- As is, you get a bunch of `table_id / column` columns
- For same-keyed tables, this makes a lot of sense. E.g. if all your tables are keyed by `sample`, this works fairly well
- But suppose you have general-stats keyed by `sample`, and then a details table _for each sample_
- How would you ever join that data with the samples? The table ID is insufficient, as there are restrictions on the nature of table IDs
- This breaks down a little - if there is _no data in the table that could allow you to join back to the sample_, you're out of luck
- But if you do put the sample in e.g. the title, or the anchor, or any other config field, this allows you to join post-hoc

We may a `parquet_detailed` config option. I suspect the wide-table option isn't largely used yet given this PR, I don't think it would even work today on `main` if you have the same sample in multiple tables.

I'm open to ideas.

jchorl: Another thought I had is whether there should be a separate parquet file per chart/table/etc. While jamming everything into one parquet is nice, you get very weirdly shaped wide tables. If your tables are all keyed by sample, it's great. But if they're not, you effectively just end up with diagonal tables where e.g. the first 10 columns are filled for the first 5 rows, the next 10 columns are filled for the rows of the next table, etc etc. I don't think this is necessarily inefficient in parquet, but it is odd. One artifact is it coerces every column to be nullable.

ewels:
@multiqc-bot fix linting

jchorl:
If you want to hold off on merging this, we're rolling it out ~now so I can make sure to get to any bugs

ewels:
@cursor please fix this linting error:

```
ruff check...............................................................Failed
- hook id: ruff-check
- exit code: 1

  E712 Avoid equality comparisons to `True`; use `table_rows["myid / col5"]:` for truth checks
     --> tests/test_output_options.py:156:13
      |
  154 |     assert (table_rows["myid / col3"].is_nan()).all()
  155 |     assert (table_rows["myid / col4"] == "bar").all()
  156 |     assert (table_rows["myid / col5"] == True).all()
      |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  157 |
  158 |     # each table stores its pconfig in a per-table column to survive the outer join without renaming
      |
  help: Replace with `table_rows["myid / col5"]`

  Found 1 error.
```

ewels:
@jchorl can do, though it sounds fairly serious so it'd be nice to get it out for people..

ewels:
Bumped to the v1.36 milestone to give you time to test @jchorl. I also want to properly understand and test myself, rather than merging under time pressure.

PR #3530, opened 2026-04-16
sequali - expand general statistics and add normalized plots

pontushojer:
- Add sequence/base normalized subplots for `Per Sequence Average Quality Scores`, `Sequence Length Distribution`, `Insert Size Distribution`. This solves better adapts to dataset with differing sequence/base count. 
- Add N50 read length (introduced in sequali v1.0.0) to General stats as a flexible substitute for mean read length. 
- Add total bases to General stats table as this is more useful than read count for variable read length libraries. 
- Add mean sequence phread quality to General stats. I think its useful to have a sequence quality metric in the table, another contender could be % of bases >= Q20 but that is more narrow scope that the mean quality. 
- Other minor edits 


ewels (approved):
Nice, thanks!

Issue #3488, opened 2026-02-17
v1.33 "The 'bcftools (stats)' MultiQC module broke.."


lindenb:
### Description of bug

Hi all, multiqc raised an error for the following attached file. 

[rotavirus.stats.txt](https://github.com/user-attachments/files/25365680/rotavirus.stats.txt)

```
│ Traceback (most recent call last):                                                                                                                │
│   File "/path/to/lib/python3.14/site-packages/multiqc/core │
│     these_modules: Union[BaseMultiqcModule, List[BaseMultiqcModule]] = module_initializer()                                                       │
│                                                                        ~~~~~~~~~~~~~~~~~~^^                                                       │
│   File "/path/to/lib/python3.14/site-packages/multiqc/modu │
│     n["stats"] = parse_bcftools_stats(self)                                                                                                       │
│                  ~~~~~~~~~~~~~~~~~~~~^^^^^^                                                                                                       │
│   File "/path/to/lib/python3.14/site-packages/multiqc/modu │
│     plot=bargraph.plot(                                                                                                                           │
│          ~~~~~~~~~~~~~^                                                                                                                           │
│         list(bcftools_stats_sample_depth.values()),                                                                                               │
│         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                               │
│     ...<9 lines>...                                                                                                                               │
│         ),                                                                                                                                        │
│         ^^                                                                                                                                        │
│     ),                                                                                                                                            │
│     ^                                                                                                                                             │
│   File "/path/to/lib/python3.14/site-packages/multiqc/plot │
│     plot = BarPlot.from_inputs(inputs)                                                                                                            │
│   File "/path/to/lib/python3.14/site-packages/multiqc/plot │
│     return BarPlot.create(                                                                                                                        │
│            ~~~~~~~~~~~~~~^                                                                                                                        │
│         cats_lists=plot_data,                                                                                                                     │
│         ^^^^^^^^^^^^^^^^^^^^^                                                                                                                     │
│     ...<6 lines>...                                                                                                                               │
│         offset_groups=inputs.offset_groups,                                                                                                       │
│         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                       │
│     )                                                                                                                                             │
│     ^                                                                                                                                             │
│   File "/path/to/lib/python3.14/site-packages/multiqc/plot │
[...]

PR #3537, opened 2026-04-24
Fix: import rich.panel explicitly in module-broke exception handler

### LINKED ISSUE #3536: AttributeError: module 'rich' has no attribute 'panel' in module-broke exception handler (rich >= 14.3)
### Description of bug

`multiqc/core/exec_modules.py:157` calls `rich.panel.Panel(...)` when a module raises, but the file only does `import rich` at the top - `rich.panel` is never imported. It works incidentally when something upstream in the import chain has already loaded `rich.panel` as a side effect.

Up to and including rich 14.2.0, `rich/console.py` imported `rich.scope` at module scope, which in turn imported `rich.panel` - so MultiQC's exception handler worked by accident. From rich 14.3 onwards (and in 15.x), `from .scope import render_scope` in `rich/console.py` is lazy (moved inside a method body), so `rich.panel` is no longer pre-loaded by the time a MultiQC module raises.

Result: whenever any MultiQC module raises an exception, the "Oops! The 'X' MultiQC module broke..." handler itself crashes with:

```
File ".../multiqc/core/exec_modules.py", line 157, in exec_modules
    rich.panel.Panel(
    ^^^^^^^^^^
AttributeError: module 'rich' has no attribute 'panel'. Did you mean: 'pager'?
```

The real error is hidden, the report isn't written, and the user sees a confusing traceback that points at `rich`, not at the actual module that broke.

### Reproduction

Using [`rotavirus.stats.txt`](https://github.com/user-attachments/files/25365680/rotavirus.stats.txt) from #3488 (which independently triggers a module-broke code path) as a minimal trigger:

```
$ micromamba create -n mqc -c conda-forge -c bioconda bioconda::multiqc=1.34 -y
$ micromamba activate mqc
$ python -c "import importlib.metadata as m; print('rich:', m.version('rich'))"
rich: 15.0.0
$ multiqc rotavirus.stats.txt
...
  File ".../multiqc/core/exec_modules.py", line 157, in exec_modules
    rich.panel.Panel(
    ^^^^^^^^^^
AttributeError: module 'rich' has no attribute 'panel'. Did you mean: 'pager'?
$ echo $?
1
```

Isolating the `rich` version (same MultiQC env, `pip install --force-reinstall rich==X --no-deps`):
- `rich==15.0.0` - crashes with the AttributeError (current bioconda default solve under py3.14).
- `rich==14.3.4` - same AttributeError.
- `rich==14.2.0` - handler renders correctly, printing the real "Oops! The 'bcftools' MultiQC module broke..." panel.

Diff that caused the regression: `rich/console.py` `from .scope import render_scope` moved from module-top (14.2.0) to inside a method (14.3+).

Affects MultiQC 1.33 and 1.34; the same `rich.panel.Panel(` call is still present on `main`.

### Suggested fix

One-line: add `import rich.panel` (or `from rich.panel import Panel`) at the top of `multiqc/core/exec_modules.py`. A quick grep for other `rich.<submodule>` attribute accesses without corresponding imports would be worth doing at the same time (`rich.console.Console` etc. appear nearby in the same file).

### Context

Surfaced while diagnosing a conda-only CI failure on nf-core/rnaseq PR https://github.com/nf-core/rnaseq/pull/1816. The primary failure is some pipeline module producing `ValueError: No datasets to plot` (similar to #3488), but the handler regression is masking which module is actually breaking, so the fix is harder to triage than it should be.

### Before submitting

pinin4fjords:
## Summary

Closes #3536.

`multiqc/core/exec_modules.py:157` used `rich.panel.Panel(...)` with only `import rich` at the top of the file. This worked by luck while rich's `console.py` eagerly imported `rich.scope` (which imported `rich.panel`) at module scope. That import became lazy in rich 14.3, so from then on the handler crashed with

```
AttributeError: module 'rich' has no attribute 'panel'. Did you mean: 'pager'?
```

whenever any MultiQC module raised. The real module error was masked by the handler's own crash, making it hard to diagnose which module had failed.

This PR adds an explicit `from rich.panel import Panel` (matching the existing `from rich.syntax import Syntax` pattern one line above) and uses `Panel` at the call site.

## Test plan

Reproduction uses the same input as #3488 (`rotavirus.stats.txt`) which independently trips a module-broke path.

Before:

```
$ micromamba create -n mqc -c conda-forge -c bioconda bioconda::multiqc=1.34 -y
$ micromamba activate mqc
$ multiqc rotavirus.stats.txt
...
  File ".../multiqc/core/exec_modules.py", line 157, in exec_modules
    rich.panel.Panel(
    ^^^^^^^^^^
AttributeError: module 'rich' has no attribute 'panel'. Did you mean: 'pager'?
```

After (with this branch pip-installed over the env):

```
╭──────────────── Oops! The 'bcftools' MultiQC module broke... ────────────────╮
│ Please copy this log and report it at https://github.com/MultiQC/MultiQC/issues │
│ ...                                                                          │
│   raise ValueError("No datasets to plot")                                    │
│ ValueError: No datasets to plot                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
           multiqc | No analysis results found. Cleaning up…
```

The real underlying error (the `bcftools (stats)` parser producing an empty bar plot, tracked in #3488) is now visible to the user, and the handler exits cleanly with code 1 as before. Also verified against an unrelated module break (corrupt FastQC input → `Oops! The 'fastqc' module broke...` renders correctly), and against a clean run with no broken modules (exit 0, no panel, no regression).

- [x] Manually reproduced with rich 15.0.0 + Python 3.14 + MultiQC 1.34 (fails before, passes after)
- [x] Confirmed no regression on clean / no-broken-module runs

### DIFF
--- a/multiqc/core/exec_modules.py
+++ b/multiqc/core/exec_modules.py
@@ -7,6 +7,7 @@ from typing import Callable, Dict, List, Union
 
 import rich
 from importlib_metadata import EntryPoint
+from rich.panel import Panel
 from rich.syntax import Syntax
 
 from multiqc import config, report
@@ -154,7 +155,7 @@ def exec_modules(mod_dicts_in_order: List[Dict[str, Dict]]) -> None:
             from multiqc.core.log_and_rich import rich_console_print
 
             rich_console_print(
-                rich.panel.Panel(
+                Panel(
                     CustomTraceback(),
                     title=f"Oops! The '[underline]{this_module}[/]' MultiQC module broke...",
                     expand=False,


ewels (approved):
Thank you!

PR #3538, opened 2026-04-27
Add native Trim Galore v2.x module (closes #3529)

### LINKED ISSUE #3529: Feature request: Native support for Trim Galore v2.0 (Oxidized Edition) reports
## Context

[Trim Galore v2.0](https://github.com/FelixKrueger/TrimGalore) ("Oxidized Edition") is a complete rewrite in Rust — a single binary with no external dependencies. It no longer shells out to Cutadapt; adapter trimming is built in.

The trimming reports still follow the same `*_trimming_report.txt` naming convention and include the same statistics (total reads, adapter %, quality-trimmed bp, adapter length distribution), but there is no longer embedded Cutadapt stdout — the statistics are generated directly by Trim Galore.

## Current situation

MultiQC discovers Trim Galore reports through the `cutadapt` module, which requires `"This is cutadapt"` in the first 100 lines. Since v2.0 no longer uses Cutadapt, we emit a backwards-compatibility shim:

```
Trim Galore 2.0.0 (Oxidized Edition) — adapter trimming built in
This is cutadapt 4.0 (compatible; for MultiQC backwards compatibility)
```

This makes current MultiQC work, but the Software Versions table shows **"Cutadapt 4.0"** instead of **"Trim Galore 2.0.0"**, which is misleading.

## Proposal

Add a native search pattern for Trim Galore v2.0 reports so they are recognized directly with the correct software name and version. For example:

```yaml
# search_patterns.yaml
trim_galore:
  - contents: "Trim Galore"
    contents_re: "Oxidized Edition"
    num_lines: 20
```

The statistics section uses the same format as Cutadapt >= 1.7 (same regexes for `Total reads processed:`, `Reads with adapters:`, bp stats, adapter length distribution table), so the existing parsing logic can be reused — only file discovery and version extraction need to change.

The version can be extracted from:
```
Trim Galore version: 2.0.0 (Oxidized Edition)
```
which appears in the report header (line 5).

The sample name can be extracted from:
```
Input filename: nextera_100K.fastq.gz
```
which appears on line 3.

We will keep the `"This is cutadapt"` shim in our reports for backwards compatibility with older MultiQC releases, so this change would be purely additive.

## Test report

Below is a complete v2.0 trimming report (100K Nextera reads) that can be used as a test fixture:

<details>
<summary>nextera_100K.fastq.gz_trimming_report.txt</summary>

```
SUMMARISING RUN PARAMETERS
=========================
Input filename: nextera_100K.fastq.gz
Trimming mode: single-end
Trim Galore version: 2.0.0 (Oxidized Edition)
Quality Phred score cutoff: 20
Quality encoding type selected: ASCII+33
Adapter sequence: 'CTGTCTCTTATA' (user-specified or auto-detected)
Maximum trimming error rate: 0.1
Minimum required adapter overlap (stringency): 1 bp
Minimum required sequence length single-end: 20 bp
Output file will be GZIP compressed


Trim Galore 2.0.0 (Oxidized Edition) — adapter trimming built in
This is cutadapt 4.0 (compatible; for MultiQC backwards compatibility)
Command line parameters: -j 1 -e 0.1 -q 20 -O 1 -a CTGTCTCTTATA nextera_100K.fastq.gz
Processing reads on 1 core in single-end mode ...

=== Summary ===

Total reads processed:              100,000
Reads with adapters:                 37,813 (37.8%)
Reads written (passing filters):     99,922 (99.9%)

Total basepairs processed:      5,100,000 bp
Quality-trimmed:                    4,374 bp (0.1%)
Total written (filtered):       4,964,599 bp (97.3%)

=== Adapter 1 ===

Sequence: CTGTCTCTTATA; Type: regular 3'; Length: 12; Trimmed: 37813 times.

No. of allowed errors:
1-9 bp: 0; 10-12 bp: 1

Overview of removed sequences
length	count	expect	max.err	error counts
1	19383	25000.0	0	19383
2	6510	6250.0	0	6510
3	2169	1562.5	0	2169
4	997	390.6	0	997
5	871	97.7	0	871
6	943	24.4	0	943
7	1020	6.1	0	1020
8	973	1.5	0	973
9	1120	0.4	0	1120
10	1010	0.1	1	1010
11	717	0.0	1	717
12	473	0.0	1	473
13	324	0.0	1	324
14	276	0.0	1	276
15	271	0.0	1	271
16	163	0.0	1	163
17	88	0.0	1	88
18	75	0.0	1	75
19	53	0.0	1	53
20	63	0.0	2	63
21	38	0.0	2	38
22	49	0.0	2	49
23	39	0.0	2	39
24	38	0.0	2	38
25	29	0.0	2	29
26	12	0.0	2	12
27	6	0.0	2	6
28	4	0.0	2	4
29	2	0.0	2	2
30	9	0.0	3	9
31	16	0.0	3	16
32	5	0.0	3	5
33	2	0.0	3	2
34	4	0.0	3	4
37	3	0.0	3	3
38	1	0.0	3	1
39	5	0.0	3	5
40	4	0.0	4	4
41	3	0.0	4	3
42	2	0.0	4	2
43	5	0.0	4	5
44	3	0.0	4	3
45	9	0.0	4	9
46	1	0.0	4	1
48	7	0.0	4	7
49	5	0.0	4	5
50	1	0.0	5	1
51	12	0.0	5	12


RUN STATISTICS FOR INPUT FILE: nextera_100K.fastq.gz
=============================================
100000 sequences processed in total
Sequences removed because they became shorter than the length cutoff of 20 bp:	78 (0.1%)
```

</details>

Happy to help with a PR if that would be useful.

FelixKrueger:
Closes #3529.

Native MultiQC module for [Trim Galore v2.x](https://github.com/FelixKrueger/TrimGalore) ("Oxidized Edition") that parses the structured `*_trimming_report.json` (schema v1) emitted alongside the legacy text report. Marking as **draft** for an initial design review — happy to extend scope based on feedback before a final pass.

## Why

@ewels suggested in the issue thread that a JSON-driven module would be cleaner than a cutadapt-style search-pattern shim. Trim Galore v2.x now ships a structured report (`*_trimming_report.json`, schema v1) for exactly that purpose. The text report still carries the `"This is cutadapt"` backwards-compat shim so the existing `cutadapt` module path keeps working unchanged — this PR is purely additive.

## What's parsed / plotted

- **General stats columns**: % adapter, % pass, % q-trimmed, total reads (hidden), total bp written (hidden)
- **Filtered reads bargraph**: passing / too_short / too_long / too_many_n / discarded_untrimmed
- **Adapter length distribution linegraph**: per sample; per (sample, adapter) when a sample has more than one adapter
- **Software Versions**: native "Trim Galore X.Y.Z" entry (instead of misleading "Cutadapt 4.0" via the shim)

Sections **wired through to the data file but not yet plotted** (low cost to add — left as follow-ups so the initial PR stays focused):
- Poly-A / Poly-G trimming summaries (`reads_trimmed`, `bases_removed`)
- RRBS truncation counts (`trimmed_3prime`, `trimmed_5prime`, `r2_clipped_5prime`)
- Pair-validation outcomes (`pairs_analyzed`, `pairs_removed`, `pairs_removed_n`, `r1_unpaired`, `r2_unpaired`)

Happy to add any of these in this PR if you'd like — just let me know.

## Sample-name handling

PE reports list **both** R1 and R2 in `input_filenames` (Trim Galore preserves the pair context in both JSONs). The parser uses each report's `read_number` field (1 or 2) to pick the matching filename, so R1 and R2 produce distinct samples. SE reports have `read_number: 1` and `input_filenames` of length 1, so the same logic gives the right SE behaviour.

## Coexistence with the `cutadapt` module

Both modules will discover their respective files (text vs JSON). With both enabled, each TrimGalore sample appears in both modules' general-stats columns. Users who want native-only TrimGalore output can:

```yaml
disable_modules:
  - cutadapt
```

This is documented in the new module's class docstring. If you'd prefer a different coordination story (e.g. the trim_galore module suppresses the cutadapt shim file when found, or the cutadapt module skips files that have a sibling `_trimming_report.json`), I can wire that up.

## Schema version gating

The parser checks `schema_version == 1` and warns + skips unrecognised versions. So a future schema bump in TrimGalore won't silently misparse — it'll need an explicit branch added here.

## Test fixtures

`tests/data/modules/trim_galore/`:
- `sample_R1.fastq.gz_trimming_report.json` — SE (10K Illumina reads)
- `BS-seq_10K_R{1,2}.fastq.gz_trimming_report.json` — PE pair (BS-seq 10K, with `pair_validation` populated)

## Validation

Verified locally with the venv'd dev install:

```
$ multiqc -m trim_galore tests/data/modules/trim_galore/
       file_search | Search path: …/tests/data/modules/trim_galore
       file_search | Only using modules: trim_galore
       trim_galore | Found 3 reports
     write_results | Data        : …/multiqc_data
     write_results | Report      : …/multiqc_report.html
           multiqc | MultiQC complete
```

3/3 fixtures parsed (1 SE + 2 PE), distinct sample names (`sample`, `BS-seq_10K_R1`, `BS-seq_10K_R2`), all sections rendered in the HTML report, data file written.

## Diff stats

- 9 files changed, +731 / 0
- New module: `multiqc/modules/trim_galore/{__init__.py, trim_galore.py}` (~330 lines)
- Wiring: `pyproject.toml` entry point, `multiqc/search_patterns.yaml`, `multiqc/config_defaults.yaml` module list
- 3 test fixtures
- CHANGELOG entry under "Unreleased / New modules"

---

**Update 2026-04-27**: test fixtures moved to companion PR [MultiQC/test-data#377](https://github.com/MultiQC/test-data/pull/377) (the test runner resolves data via the sibling `MultiQC/test-data` repo, not `tests/data/` in this repo). The fixtures-related test failures (`test_modules_run.py::test_all_modules[trim_galore-…]` and `test_ignore_samples[trim_galore-…]`) clear once #377 merges.


multiqc/modules/trim_galore/trim_galore.py
```diff
@@ -0,0 +1,327 @@
...
+                self.add_software_version(version, s_name)
+
+        data_by_sample = self.ignore_samples(data_by_sample)
+        if not data_by_sample:
+            raise ModuleNoSamplesFound
+
+        log.info(f"Found {len(data_by_sample)} reports")
+
+        # Persist the raw parsed JSON for downstream use / inspection.
+        self.write_data_file(data_by_sample, "multiqc_trim_galore")
```
pinin4fjords: **`write_data_file()` is called before sections are added.** Project convention is to call this as the very last step in `__init__`, after general-stats and all `add_section()` calls have populated. Move this line to the bottom of `__init__` (after the adapter-length section).

pinin4fjords: **Data file payload is the raw nested JSON.** I confirmed by inspecting the rendered `multiqc_trim_galore.txt` - the `parameters`, `read_processing`, `adapter_trimming` etc. columns end up as stringified Python dicts (`{'total_reads': 10000, ...}`), which isn't really machine-readable. `write_data_file` expects a flat 2D dict.

Suggest flattening to scalar columns (the way `multiqc_cutadapt.txt` looks):

```python
flat = {}
for s_name, p in data_by_sample.items():
    rp = p.get("read_processing", {}) or {}
    bp = p.get("basepair_processing", {}) or {}
    pv = p.get("pair_validation") or {}
    flat[s_name] = {
        "trim_galore_version": p.get("trim_galore_version"),
        "mode": p.get("mode"),
        "total_reads": rp.get("total_reads"),
        "reads_with_adapter": rp.get("reads_with_adapter"),
        "reads_written": rp.get("reads_written"),
        "reads_too_short": rp.get("reads_too_short"),
        "reads_too_long": rp.get("reads_too_long"),
        "reads_too_many_n": rp.get("reads_too_many_n"),
        "reads_discarded_untrimmed": rp.get("reads_discarded_untrimmed"),
        "total_bp_processed": bp.get("total_bp_processed"),
        "total_bp_written": bp.get("total_bp_written"),
        "quality_trimmed_bp": bp.get("quality_trimmed_bp"),
        "pairs_analyzed": pv.get("pairs_analyzed") if pv else None,
        "pairs_removed": pv.get("pairs_removed") if pv else None,
    }
self.write_data_file(flat, "multiqc_trim_galore")
```

If you want to keep the full payload around for tooling, a separate `write_data_file(data_by_sample, "multiqc_trim_galore_raw", data_format="json")` would be cleaner.

multiqc/modules/trim_galore/trim_galore.py
```diff
@@ -0,0 +1,327 @@
...
+            s_name, payload = parsed
+            if s_name in data_by_sample:
+                log.debug(f"Duplicate sample name found! Overwriting: {s_name}")
+            data_by_sample[s_name] = payload
+            self.add_data_source(f, s_name=s_name)
+
+            # Software version — single canonical entry per sample.
+            version = payload.get("trim_galore_version")
+            if version:
+                self.add_software_version(version, s_name)
```
pinin4fjords: **`add_software_version()` should be called unconditionally.** The base implementation early-returns on `None` (see `base_module.py:1205` - the comment there explicitly says "This allows every module to call this function even those without a version to add. This is useful to check that all modules are calling this function."). The `if version:` guard means a malformed JSON missing `trim_galore_version` would silently skip the call.

```suggestion
            self.add_software_version(payload.get("trim_galore_version"), s_name)
```

pinin4fjords (commented):
Inline comments cover the substance. Highlights:

- Convention items the linter doesn't catch: `write_data_file` ordering, unconditional `add_software_version`, redundant `clean_s_name` helper.
- The persisted data file ends up with stringified Python dicts because raw nested JSON is passed through to `write_data_file` - needs flattening.
- PE pairs render as two general-stats rows; cutadapt collapses via `SampleGroupingConfig`.
- Class docstring advertises RRBS / poly-A/G / pair-validation, but none of these surface in the UI.

Tests under `multiqc/modules/trim_galore/tests/` would be good to add before merge.

---
*Drafted with Claude (Opus 4.7), validated locally against the test-data#377 fixtures.*

multiqc/modules/trim_galore/trim_galore.py
```diff
@@ -0,0 +1,327 @@
...
+Parses the structured `*_trimming_report.json` file emitted by Trim Galore
+v2.x alongside the legacy text report (schema v1).  See:
+https://github.com/MultiQC/MultiQC/issues/3529
+
+Why a dedicated module rather than reusing the `cutadapt` module:
+  * Software-versions table shows "Trim Galore" with the correct version
+    instead of the misleading "Cutadapt 4.0" backwards-compatibility shim.
+  * Native access to TrimGalore-specific stats not present in Cutadapt
+    output (RRBS truncation counts, poly-A / poly-G trimming summaries,
+    paired-end pair-validation outcomes).
```
pinin4fjords: The module-level docstring sells the module on "Native access to TrimGalore-specific stats not present in Cutadapt output (RRBS truncation counts, poly-A / poly-G trimming summaries, paired-end pair-validation outcomes)", but none of those are surfaced anywhere in the UI - they only appear (badly, see other comment) in the data file. Either:

- Add small sections that surface them, gated on at least one sample having a non-zero value (a single `table.plot` with the relevant `pair_validation` / `rrbs` / `poly_*` rows would deliver the marketing claim cheaply), or
- Tone the docstring down to match what the module actually does today, with a TODO for the rest.

The current state risks reviewers/users noticing the mismatch when they actually open the report.

multiqc/modules/trim_galore/trim_galore.py
```diff
@@ -0,0 +1,327 @@
...
+            },
+            "tg_total_bp_written": {
+                "title": "BP written",
+                "description": "Total basepairs written to trimmed output",
+                "scale": "BuGn",
+                "shared_key": "base_count",
+                "hidden": True,
+            },
+        }
+        self.general_stats_addcols(gen_stats, headers)
```
pinin4fjords: **PE pairs render as two separate samples in general stats.** With the test fixtures, `BS-seq_10K_R1` and `BS-seq_10K_R2` both appear as rows. The cutadapt module collapses pairs via `SampleGroupingConfig` (see `multiqc/modules/cutadapt/cutadapt.py:423`).

Worth matching for consistency between sibling modules - especially since this PR positions itself as the natural successor to the cutadapt path. Mechanically:

- Pass both `input_filenames` to `clean_s_name` so R1/R2 ambiguity resolves at sample-name level (cutadapt has helpers for this), or strip `_R[12]` from the derived sample name.
- Wrap the `general_stats_addcols` call with `group_samples_config=SampleGroupingConfig(cols_to_sum=[...], cols_to_weighted_average=[("tg_pct_with_adapter", "tg_total_reads"), ...])` so paired rows merge sensibly.

This would also halve the visual noise on PE-heavy reports.

multiqc/modules/trim_galore/trim_galore.py
```diff
@@ -0,0 +1,327 @@
...
+        cats = {
+            "passing": {"name": "Passed filters", "color": "#7cb5ec"},
+            "too_short": {"name": "Too short", "color": "#f7a35c"},
+            "too_long": {"name": "Too long", "color": "#90ed7d"},
+            "too_many_n": {"name": "Too many Ns", "color": "#8085e9"},
+            "discarded_untrimmed": {
+                "name": "Discarded (untrimmed)",
+                "color": "#e4d354",
+            },
+        }
```
pinin4fjords, line 313: **Hardcoded bargraph colours are misleading.** As wired up: `passing` = blue, `too_long` = green, `too_short` = orange. Green for a fail bucket and blue for the pass bucket reads as the opposite signal you want for a disposition plot. Other trim modules (cutadapt, fastp) don't override colours here - the auto palette is consistent across the report.

```suggestion
        cats = {
            "passing": {"name": "Passed filters"},
            "too_short": {"name": "Too short"},
            "too_long": {"name": "Too long"},
            "too_many_n": {"name": "Too many Ns"},
            "discarded_untrimmed": {"name": "Discarded (untrimmed)"},
        }
```

multiqc/modules/trim_galore/trim_galore.py
```diff
@@ -0,0 +1,327 @@
...
+            if len(adapters) == 1:
+                a = adapters[0]
+                dist = a.get("length_distribution") or {}
+                line_data[s_name] = {int(k): v for k, v in dist.items()}
+            else:
+                for idx, a in enumerate(adapters, start=1):
+                    name = a.get("name") or f"adapter_{idx}"
+                    composite = f"{s_name}: {name}"
+                    dist = a.get("length_distribution") or {}
+                    line_data[composite] = {int(k): v for k, v in dist.items()}
```
pinin4fjords: **Adapter-length plot mixes simple and composite keys.** When sample A has 1 adapter the trace is keyed `"sample_A"`; when sample B has 2 adapters keys become `"sample_B: Illumina"` / `"sample_B: Nextera"`. In a mixed run the legend is half plain names, half composites, with no visual grouping.

Also, `if len(adapters) == 1:` falls through to `else` for both `len == 0` and `len > 1`, so a sample with zero adapters silently disappears from the plot.

Two cleaner options:

1. Always use composite keys (e.g. `f"{s_name} ({adapter_name})"` even for the single-adapter case, so the legend is uniform).
2. Use `data_labels` to render one panel per adapter, falling back to a single panel when only one adapter is present (cutadapt does this for 5'/3' ends in `cutadapt_length_trimmed_plot`).

The `if len(adapters) == 1` branch should also become an explicit `if not adapters: continue` for the zero case, regardless of which approach you pick.

multiqc/modules/trim_galore/trim_galore.py
```diff
@@ -0,0 +1,327 @@
...
+        # the same sample. Use `read_number` (1 or 2) to pick the matching
+        # entry. SE reports have `read_number: 1` and `input_filenames` of
+        # length 1, so the same logic gives the correct SE behaviour.
+        input_filenames = payload.get("input_filenames") or []
+        read_number = payload.get("read_number") or 1
+        if not input_filenames:
+            log.warning(f"Skipping {f['fn']!r}: no input_filenames")
+            return None
+        idx = max(0, min(read_number - 1, len(input_filenames) - 1))
+        s_name = self.clean_s_name(_strip_fastq_suffix(input_filenames[idx]), f)
```
pinin4fjords: **`_strip_fastq_suffix` is redundant with `clean_s_name`.** `config.fn_clean_exts` already includes `.gz`, `.fastq`, `.fq` (`config_defaults.yaml:236-238`), and `clean_s_name` truncates at the first match. I confirmed on the fixtures that passing the raw `input_filenames[idx]` produces the same `BS-seq_10K_R1` / `sample` results.

```suggestion
        s_name = self.clean_s_name(input_filenames[idx], f)
```

The `_FASTQ_SUFFIX_RE` and `_strip_fastq_suffix` definitions further down (lines 315-321) can then be deleted entirely.

multiqc/modules/trim_galore/trim_galore.py
```diff
@@ -0,0 +1,327 @@
...
+Why a dedicated module rather than reusing the `cutadapt` module:
+  * Software-versions table shows "Trim Galore" with the correct version
+    instead of the misleading "Cutadapt 4.0" backwards-compatibility shim.
+  * Native access to TrimGalore-specific stats not present in Cutadapt
+    output (RRBS truncation counts, poly-A / poly-G trimming summaries,
+    paired-end pair-validation outcomes).
+  * The JSON is purely additive — Trim Galore continues to emit the
+    text report unchanged, so the existing `cutadapt` module path
+    keeps working for users who don't enable this module.
+"""
```
pinin4fjords: Project convention is one docstring on the module class, none at module level (CLAUDE.md: *"Do not add separate markdown files or module-level docstrings. Instead add a docstring to the module class."*). The class docstring at line 44+ already covers the same ground (Rust rewrite, drop-in replacement, JSON parsing). This module-level block can be deleted in full.

multiqc/modules/trim_galore/trim_galore.py
pinin4fjords: **Bail on ignored samples inside the loop, not after.** Currently `add_data_source` and `add_software_version` are called for every parsed report, including ones that get filtered out by `ignore_samples` later. Cleaner to short-circuit early so ignored samples never get registered as data sources:

```python
s_name, payload = parsed
if self.is_ignore_sample(s_name):
    continue
if s_name in data_by_sample:
    log.debug(f"Duplicate sample name found! Overwriting: {s_name}")
data_by_sample[s_name] = payload
self.add_data_source(f, s_name=s_name)
self.add_software_version(payload.get("trim_galore_version"), s_name)
```

…then drop the `ignore_samples` call below since the loop already filtered.

multiqc/modules/trim_galore/trim_galore.py
```diff
@@ -0,0 +1,327 @@
...
+            line_data,
+            {
+                "id": "trim_galore_adapter_length_plot",
+                "title": "Trim Galore: Adapter Length Distribution",
+                "xlab": "Adapter overlap length (bp)",
+                "ylab": "Reads (count)",
+                "ymin": 0,
+                "tt_label": "{point.x} bp: {point.y:,} reads",
+            },
+        )
```
pinin4fjords, line 344: Two unrelated nits on this method that are cleaner to bundle:

1. The plot ID `trim_galore_filtered_reads_plot` (and `trim_galore_adapter_length_plot` further down) - these are fine but the `<anchor>-<section>-plot` form (i.e. `trim-galore-filtered-reads-plot`) is more common in newer modules and avoids ambiguity with the section anchor `trim_galore_filtered_reads`. Optional.

2. Description text references `--stringency 1` which is helpful, but currently lives only in the adapter-length plot description - worth a one-line note in the *filtered reads* description too explaining that `passing` reflects post-quality, post-adapter, post-N filtering (i.e. what made it into the trimmed output), to disambiguate from `reads_with_adapter`.

multiqc/modules/trim_galore/trim_galore.py
```diff
@@ -0,0 +1,327 @@
...
+                "format": "{:,.1f}",
+            },
+            "tg_pct_quality_trimmed": {
+                "title": "% Q-trim",
+                "description": "% bases removed by 3'-end quality trimming",
+                "max": 100,
+                "min": 0,
+                "suffix": "%",
+                "scale": "Blues",
+                "format": "{:,.2f}",
```
pinin4fjords: Two-decimal precision on `% Q-trim` is more than the other percentage columns above (`{:,.1f}`) and finer than the data really supports. Match for consistency:

```suggestion
                "format": "{:,.1f}",
```

multiqc/modules/trim_galore/trim_galore.py
```diff
@@ -0,0 +1,327 @@
...
+            quality_trimmed = bp.get("quality_trimmed_bp", 0) or 0
+
+            row: Dict[str, float] = {
+                "tg_total_reads": total_reads,
+                "tg_pct_with_adapter": _safe_pct(reads_with_adapter, total_reads),
+                "tg_pct_passing": _safe_pct(reads_written, total_reads),
+                "tg_pct_quality_trimmed": _safe_pct(quality_trimmed, total_bp),
+                "tg_total_bp_written": bp.get("total_bp_written", 0) or 0,
+            }
+            gen_stats[s_name] = row
```
pinin4fjords: The five `total_reads = ...`, `reads_with_adapter = ...` etc. extractions are each used once, in the next dict literal. Easier to read inline:

```python
for s_name, payload in data_by_sample.items():
    rp = payload.get("read_processing", {}) or {}
    bp = payload.get("basepair_processing", {}) or {}
    total = rp.get("total_reads", 0) or 0
    total_bp = bp.get("total_bp_processed", 0) or 0
    gen_stats[s_name] = {
        "tg_total_reads": total,
        "tg_pct_with_adapter": _safe_pct(rp.get("reads_with_adapter", 0) or 0, total),
        "tg_pct_passing": _safe_pct(rp.get("reads_written", 0) or 0, total),
        "tg_pct_quality_trimmed": _safe_pct(bp.get("quality_trimmed_bp", 0) or 0, total_bp),
        "tg_total_bp_written": bp.get("total_bp_written", 0) or 0,
    }
```

multiqc/modules/trim_galore/trim_galore.py
```diff
@@ -0,0 +1,327 @@
...
+import re
+from typing import Any, Dict, Optional, Tuple
+
+from multiqc.base_module import BaseMultiqcModule, ModuleNoSamplesFound
+from multiqc.plots import bargraph, linegraph
+
+log = logging.getLogger(__name__)
+
+
+# ── Schema constants ────────────────────────────────────────────────────────
```
pinin4fjords: Style nit (applies to dividers throughout the file - this one, plus L40, L106, L109, L121, L138, L172, L242, L278, L312): the unicode `# ── … ──` headers are quite prominent and don't carry information that the method/variable names don't already convey. Project style leans minimal-comment (CLAUDE.md: *"Default to writing no comments. Only add one when the WHY is non-obvious."*). Easy cleanup if you want it; not blocking.

multiqc/modules/trim_galore/trim_galore.py
```diff
@@ -0,0 +1,327 @@
...
+
+
+# ── Module ──────────────────────────────────────────────────────────────────
+
+
+class MultiqcModule(BaseMultiqcModule):
+    """
+    Trim Galore v2.x ("Oxidized Edition") — Rust rewrite of the original
+    Perl script. Single binary, zero external runtime deps. Same CLI and
+    same output filenames as v0.6.x; emits an additional structured JSON
```
pinin4fjords: Subjective, but project tone is sober/documentation-style. Phrasing like *"Trim Galore v2.x ('Oxidized Edition') - Rust rewrite of the original Perl script. Single binary, zero external runtime deps."* reads more like release-notes copy than module reference docs. The `extra=` block in `__init__` covers the same ground. Up to you - just flagging.

multiqc/modules/trim_galore/trim_galore.py
```diff
@@ -0,0 +1,327 @@
...
+                "shared_key": "read_count",
+                "hidden": True,
+            },
+            "tg_total_bp_written": {
+                "title": "BP written",
+                "description": "Total basepairs written to trimmed output",
+                "scale": "BuGn",
+                "shared_key": "base_count",
+                "hidden": True,
+            },
```
pinin4fjords, line 277: Both count columns (`tg_total_reads`, `tg_total_bp_written`) are `hidden: True`, leaving only percentages visible by default. Cutadapt and fastp surface at least one count column so users can see scale at a glance. Consider showing `tg_total_reads` (or `reads_with_adapter`) by default.

multiqc/modules/trim_galore/trim_galore.py
```diff
@@ -0,0 +1,327 @@
...
+            return None
+
+        if payload.get("tool") != TOOL_NAME:
+            return None
+        schema = payload.get("schema_version")
+        if schema != SCHEMA_VERSION_SUPPORTED:
+            log.warning(
+                f"Skipping {f['fn']!r}: unsupported schema_version {schema!r} (expected {SCHEMA_VERSION_SUPPORTED})"
+            )
+            return None
```
pinin4fjords: Schema-version mismatch is logged at `WARNING` and silently skipped. Fine for a v1-only world, but when TrimGalore v3 ships with schema v2, every report will become an unnoticed warning. Worth either bumping to `log.error` or registering it via `report.read_warnings` so a future schema bump shows up loudly under `--strict` instead of being a silent skip.

multiqc/search_patterns.yaml
```diff
@@ -1058,6 +1058,10 @@ theta2:
 tophat:
   fn: "*align_summary.txt"
   shared: true
+trim_galore:
+  fn: "*_trimming_report.json"
+  contents: '"tool": "Trim Galore"'
+  num_lines: 5
```
pinin4fjords: **The `contents:` check is redundant given how specific the filename pattern is.** `*_trimming_report.json` is essentially TrimGalore-specific, and `_parse_log` already validates `tool` and `schema_version` after parsing. The contents/num_lines pair is also brittle: it depends on `"tool": "Trim Galore"` falling within the first 5 lines of pretty-printed JSON, which would break for a future minified output.

```suggestion
  fn: "*_trimming_report.json"
```

multiqc/config_defaults.yaml
```diff
@@ -591,6 +591,7 @@ module_order:
   - kraken
   - malt
   - motus
+  - trim_galore
```
pinin4fjords: Ultra-minor: `cutadapt` is at line 587 in this list. Placing `trim_galore` adjacent to it would group the two adapter-trimming modules together, which is how users will think about them. Not blocking.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # MultiQC Version History
 
+## Unreleased
+
+### New modules
+
+- [Trim Galore (v2.x)](https://github.com/FelixKrueger/TrimGalore) — native parser for the `*_trimming_report.json` schema-v1 file emitted alongside the legacy text report by Trim Galore v2.0+. Closes [#3529](https://github.com/MultiQC/MultiQC/issues/3529). Produces the correct "Trim Galore vX.Y.Z" entry in the Software Versions table (instead of the misleading "Cutadapt 4.0" from the v0.6.x backwards-compatibility shim) and surfaces TrimGalore-specific stats not available in Cutadapt output.
+
```
pinin4fjords: Don't think you need to add this manually - the changelog is generated from PR titles ahead of releases. You can drop the diff to `CHANGELOG.md` entirely.

ewels:
@FelixKrueger it'd be nice to have a few more example outputs from Trim Galore if you could:

1. **Trim Galore run with `--poly_a` or `--poly_g`** producing non-zero `poly_a_trimming.reads_trimmed` / `poly_g_trimming.reads_trimmed`. One sample is enough — exercises the Poly-A/G section render path and its `<details>` alert (when combined with the existing all-zero samples that get dropped from the table).

2. **Trim Galore run with `--rrbs`** producing non-zero `rrbs.trimmed_3prime` / `trimmed_5prime` / `r2_clipped_5prime`. One paired-end sample ideal so `r2_clipped_5prime` exercises. Same UI effect as above for the RRBS table.

3. **A sample with multiple adapters detected** (e.g., a barcoded library where both Illumina TruSeq and Nextera get matched, or `--adapter` + `--adapter2` set explicitly). Exercises the composite-key branch in `_adapter_length_plot` — currently every fixture has exactly 1 adapter.

### Lower priority (nice to have)

4. **A larger PE cohort (≥12 pairs)** where most have low pair-removal rates, so the `<details><summary>Show N sample names</summary>…` wrap fires in the pair-validation alert.

5. **A PE sample whose filenames don't end in `_R1`/`_R2`** (e.g. `.1.fastq.gz`/`.2.fastq.gz`, or arbitrary names like `forward.fastq.gz`/`reverse.fastq.gz`) so `_clean_fastq_pair`'s fallback `_join(sorted_members)` branch is exercised. This is a real Trim Galore CLI capability, not a synthetic edge case.

FelixKrueger:
**2. RRBS**  (also contains poly G trimming)

[SRR24766842_GSM7433400_Colon_9_Months_Rep16_Mus_musculus_RRBS_R1.fastq.gz_trimming_report.json](https://github.com/user-attachments/files/27604378/SRR24766842_GSM7433400_Colon_9_Months_Rep16_Mus_musculus_RRBS_R1.fastq.gz_trimming_report.json)
[SRR24766842_GSM7433400_Colon_9_Months_Rep16_Mus_musculus_RRBS_R1.fastq.gz_trimming_report.txt](https://github.com/user-attachments/files/27604379/SRR24766842_GSM7433400_Colon_9_Months_Rep16_Mus_musculus_RRBS_R1.fastq.gz_trimming_report.txt)
[SRR24766842_GSM7433400_Colon_9_Months_Rep16_Mus_musculus_RRBS_R2.fastq.gz_trimming_report.json](https://github.com/user-attachments/files/27604380/SRR24766842_GSM7433400_Colon_9_Months_Rep16_Mus_musculus_RRBS_R2.fastq.gz_trimming_report.json)
[SRR24766842_GSM7433400_Colon_9_Months_Rep16_Mus_musculus_RRBS_R2.fastq.gz_trimming_report.txt](https://github.com/user-attachments/files/27604381/SRR24766842_GSM7433400_Colon_9_Months_Rep16_Mus_musculus_RRBS_R2.fastq.gz_trimming_report.txt)

FelixKrueger:
**1. RNA-seq (SE)**

With both poly-A and poly-G enabled:

[SRR30638010_GSM8510604_Subcutaneous_adipose_tissue_GA-3_Mus_musculus_RNA-seq_R1.fastq.gz_trimming_report.json](https://github.com/user-attachments/files/27606218/SRR30638010_GSM8510604_Subcutaneous_adipose_tissue_GA-3_Mus_musculus_RNA-seq_R1.fastq.gz_trimming_report.json)
[SRR30638010_GSM8510604_Subcutaneous_adipose_tissue_GA-3_Mus_musculus_RNA-seq_R1.fastq.gz_trimming_report.txt](https://github.com/user-attachments/files/27606220/SRR30638010_GSM8510604_Subcutaneous_adipose_tissue_GA-3_Mus_musculus_RNA-seq_R1.fastq.gz_trimming_report.txt)

FelixKrueger:
**3. Multi-Adapter**

[SRR30638010_multi_adapter_trimming_report.json](https://github.com/user-attachments/files/27606696/SRR30638010_multi_adapter_trimming_report.json)
[SRR30638010_multi_adapter_trimming_report.txt](https://github.com/user-attachments/files/27606697/SRR30638010_multi_adapter_trimming_report.txt)

FelixKrueger:
**5. filenames arbitrary**

[forward.fastq.gz_trimming_report.json](https://github.com/user-attachments/files/27606900/forward.fastq.gz_trimming_report.json)
[forward.fastq.gz_trimming_report.txt](https://github.com/user-attachments/files/27606901/forward.fastq.gz_trimming_report.txt)
[reverse.fastq.gz_trimming_report.json](https://github.com/user-attachments/files/27606902/reverse.fastq.gz_trimming_report.json)
[reverse.fastq.gz_trimming_report.txt](https://github.com/user-attachments/files/27606903/reverse.fastq.gz_trimming_report.txt)

ewels (approved):
ok, I'm pretty happy with this. Thanks!

PR #3548, opened 2026-05-11
Per-tab targeting in custom_plot_config for multi-data_labels plots

### LINKED ISSUE #3457: samtools coverage subplot configuration
### Description of feature

Hi, I wanted to configure the MQ subplot for samtools coverage, to add a colored band over the correct range of MAPQ, as described here : https://docs.seqera.io/multiqc/reports/customisation#customising-plots
The issue is, it seems like it is impossible to target this specific subplot since all plots have the same "plot_id" (by looking at `multiqc_data.json`.
Am I missing something or is this a limitation of the module ?
Here is how I tried to configure this : 
```yaml
custom_plot_config:
  samtools-coverage:
    yPlotBands:
      - from: 50
        to: 60
        color: "#c3e6c3"
```

ewels:
Closes #3457 (the primary OP — the deprecated-alias half shipped in #3547).

## Summary

Some line plots (e.g. `samtools-coverage`, `bismark_mbias`) have a tab switch between multiple datasets. Until now `custom_plot_config[plot_id]` always applied plot-wide, so a user who wanted `y_bands` on _just_ the MQ tab in samtools-coverage had no way to express that — the OP's exact frustration:

> "All plots have the same 'plot_id' (by looking at `multiqc_data.json`). Am I missing something or is this a limitation of the module?"

This PR adds a nested form keyed by tab name (or positional index):

```yaml
custom_plot_config:
  samtools-coverage:
    data_labels:
      MQ:
        y_bands:
          - { from: 50, to: 60, color: "#c3e6c3" }
```

Plot-level keys still apply to every tab; tab-level keys layer on top of one specific tab.

## Implementation notes

- **Config ingestion** (`multiqc/plots/plot.py`, `PConfig.__init__`): a dict-valued `data_labels` entry under `custom_plot_config[id]` is pulled out before the generic setattr loop (so it doesn't clobber the list-valued `data_labels` model field). After data_labels normalization, the override is merged onto the matching tab dict. Band/line dicts run through the existing `parse_y_bands` / `parse_x_bands` / `parse_y_lines` / `parse_x_lines` methods so they become `LineBand` / `FlatLine` objects, not raw dicts. Unmatched tab names emit a warning listing the valid ones.

- **Rendering** (`_set_x_bands_and_range` / `_set_y_bands_and_range`): compute an effective per-dataset `(bands, lines, minrange)` by overlaying `data_labels[i]` onto the plot-level pconfig, and write shapes to `dataset.layout["shapes"]` rather than the global `self.layout.shapes`.
  - **Interactive** — the existing `plotting.js:renderPlot` merges `dataset.layout` into the figure on tab switch via `updateObject`, which replaces arrays wholesale. So per-tab shapes swap automatically — **no JS changes**.
  - **Flat (PNG/SVG export)** — `get_figure` already does `layout.update(**dataset.layout)`, which similarly replaces `shapes`.
  - `self.layout.shapes` is still seeded from `dataset[0]` as a fallback for any consumer that doesn't merge dataset layouts.

- **Init-pass idempotency**: `LinePlot.create` constructs via `LinePlot(**model.__dict__, …)` which re-runs `Plot.__init__` on an already-populated model. The old code appended to `self.layout.shapes` and silently duplicated shapes — the existing `test_linegraph_custom_plot_config_y_bands` test had `>= 3` to paper over this. `Plot.__init__` now clears shapes before the bands/range methods run; the test is tightened to `== 3`.

## Tests

- `test_linegraph_custom_plot_config_per_tab_y_bands` — target one tab by name; assert only that dataset gets the rect shape.
- `test_linegraph_custom_plot_config_per_tab_positional_index` — integer key path.
- `test_linegraph_custom_plot_config_per_tab_unmatched_name` — bogus name warns, no-op.
- `test_linegraph_custom_plot_config_per_tab_with_plot_level_bands` — plot-level `y_bands` apply to every tab, per-tab `y_lines` layer on just one.
- Existing tests untouched (including the deprecated-alias `yPlotBands` regression from #3547).

## End-to-end verification

OP's config form against samtools coverage test data:

```yaml
custom_plot_config:
  samtools-coverage:
    data_labels:
      MQ:
        y_bands:
          - { from: 50, to: 60, color: "#c3e6c3" }
```

`multiqc_data.json` after `multiqc <samtools/coverage test data> -m samtools -c per_tab.yaml -f`:

```
Plot has 6 datasets/tabs
  [0]      Reads  shapes:0  rects:0
  [1]      Bases  shapes:0  rects:0
  [2]   Coverage  shapes:0  rects:0
  [3] Mean depth  shapes:0  rects:0
  [4]         BQ  shapes:0  rects:0
  [5]         MQ  shapes:1  rects:1
```

## Test plan

- [x] `pytest tests/test_plots.py` — 58 passed, 4 skipped
- [x] Wider `pytest tests/` — 144 passed, 4 skipped
- [x] End-to-end reproduction with samtools coverage data shows bands rendered on MQ tab only
- [ ] CI green

### DIFF
--- a/docs/markdown/reports/customisation.md
+++ b/docs/markdown/reports/customisation.md
@@ -708,6 +708,25 @@ custom_plot_config:
         color: "#c3e6c3"
 ```
 
+### Targeting a single tab in a multi-dataset plot
+
+Some plots have a tab switch between multiple datasets — for example, the `samtools-coverage` line plot has tabs for _Reads_, _Bases_, _Coverage_, _Mean depth_, _BQ_ and _MQ_. By default, settings under a plot ID apply to every tab.
+
+To apply a setting to **just one tab**, nest it under `data_labels:` and key by the tab name (the label shown on the tab button):
+
+```yaml
+custom_plot_config:
+  samtools-coverage:
+    data_labels:
+      MQ:
+        y_bands:
+          - from: 50
+            to: 60
+            color: "#c3e6c3"
+```
+
+You can also use the integer position of the tab (`0:` for the first, `1:` for the second, …) if that is more convenient. Mix tab-level keys with plot-level keys freely — plot-level settings still apply to every tab, and tab-level settings layer on top of one specific tab.
+
 ## Customising tables
 
 Much like with the custom plot config above, you can override almost any configuration options for tables.
--- a/multiqc/plots/plot.py
+++ b/multiqc/plots/plot.py
@@ -263,8 +263,15 @@ class PConfig(ValidatedConfig):
             self.title = self.id.replace("_", " ").title()
 
         # Allow user to overwrite any given config for this plot
+        per_tab_overrides: Optional[Dict[Any, Any]] = None
         if self.id in config.custom_plot_config:
-            for k, v in config.custom_plot_config[self.id].items():
+            user_cpc = dict(config.custom_plot_config[self.id])
+            # Per-tab overrides (e.g. for multi-data_labels plots) are nested under
+            # a dict-valued `data_labels` key: pull these out so they don't clobber
+            # the list-valued `data_labels` model field via the generic setattr below.
+            if isinstance(user_cpc.get("data_labels"), dict):
+                per_tab_overrides = user_cpc.pop("data_labels")
+            for k, v in user_cpc.items():
                 if k in self.__class__.model_fields:
                     field_info = self.__class__.model_fields[k]
                     # Redirect deprecated aliases (e.g. yPlotBands -> y_bands) to their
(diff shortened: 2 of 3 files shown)


ewels:
### Note on `multiqc_data.json` structure

This PR moves band/line shapes from `report_plot_data[<id>].layout.shapes` (global) to `report_plot_data[<id>].datasets[i].layout.shapes` (per-tab). Worth flagging since `multiqc_data.json` is sometimes parsed by downstream consumers.

#### No band config configured
Byte-identical to main (no consumer-visible difference).

#### With band/line config

Tested with `samtools-coverage` on standard test data and the OP-style config from #3457:

```yaml
custom_plot_config:
  samtools-coverage:
    y_bands:
      - { from: 50, to: 60, color: "#c3e6c3" }
```

**Before (main):**
```jsonc
"samtools-coverage": {
  "layout": {
    "shapes": [
      { "type": "rect", "y0": 50, "y1": 60, ... },
      { "type": "rect", "y0": 50, "y1": 60, ... }   // ← duplicate (double-init bug)
    ]
  },
  "datasets": [
    { "layout": { /* no "shapes" key */ } },
    ...
  ]
}
```

**After (this PR):**
```jsonc
"samtools-coverage": {
  "layout": { /* no "shapes" key */ },
  "datasets": [
    { "layout": { "shapes": [ { "type": "rect", "y0": 50, "y1": 60, ... } ] } },
    { "layout": { "shapes": [ { "type": "rect", "y0": 50, "y1": 60, ... } ] } },
    { "layout": { "shapes": [ { "type": "rect", "y0": 50, "y1": 60, ... } ] } },
    { "layout": { "shapes": [ { "type": "rect", "y0": 50, "y1": 60, ... } ] } },
    { "layout": { "shapes": [ { "type": "rect", "y0": 50, "y1": 60, ... } ] } },
    { "layout": { "shapes": [ { "type": "rect", "y0": 50, "y1": 60, ... } ] } }
  ]
}
```

Two things to call out:
1. **Shapes relocated** plot-level → per-dataset. Downstream readers that read `layout.shapes` directly will see an empty list now; they should iterate `datasets[*].layout.shapes` instead. The frontend (`plotting.js`) already does this via `updateObject(plot.layout, dataset.layout, ...)` on tab switch, and the flat-export path does `layout.update(**dataset.layout)`, so neither needed changes.
2. **Duplicate shapes on `main` are a pre-existing bug** — `LinePlot(**model.__dict__, ...)` in `linegraph.py:643` re-enters `Plot.__init__` which re-ran `_set_*_bands_and_range` and appended shapes a second time. The existing `test_linegraph_custom_plot_config_y_bands` test masked it with `>= 3`; this PR tightens to `== 3` after clearing per-dataset shapes before the second pass.

Issue #3463, opened 2026-01-12
MultiQC version 1.33 & 1.30 via BioConda with Python 3.13 doesn't create report


Norbittner:
### Description of bug

HI,

When running MultiQC on FastQC files within a conda/mamba environment with Python 3.13 and MultiQC version 1.30 and 1.33 the files are recognized but it never creates a report - it always says 
`[2026-01-07 17:15:33,866] multiqc.core.write_results [INFO] Data : None`

Specifiying an output directory - forcing to overwrite doesn't change the behaviour 
There is no "silent" config file running which specifies this behaviour

When downgrading to an environment with Python 3.12 everything works as expected.

Attached the installed packages within each environment working and not working). 

### File that triggers the error

[OA004_C_D_H3K27me3_EKDL250020362-1A_22W3MMLT4_L7_1_fastqc.zip](https://github.com/user-attachments/files/24561349/OA004_C_D_H3K27me3_EKDL250020362-1A_22W3MMLT4_L7_1_fastqc.zip)

### MultiQC Error log

```console
multiqc .

/// MultiQC 🔍 v1.33

     update_config | Prepending directory to sample names
     update_config | Not cleaning sample names
       file_search | Search path: [not shown - manually deleted by me]
         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 448/448  
            fastqc | Found 224 reports
     write_results | Data        : None
           multiqc | MultiQC complete
```

### Before submitting

- [x] I have read the [troubleshooting documentation](https://docs.seqera.io/multiqc/usage/troubleshooting/).
- [x] I am using the latest release of MultiQC.
- [x] I have included a full MultiQC log, not truncated.
- [x] I have attached an input file (**.zip** if necessary) that triggers the error.

ewels:
Hi @Norbittner,

I'm afraid I'm unable to replicate this problem. Your example file runs fine for me, across all versions you mention. It's possible that you have something funky in your MultiQC config file, or environment variables, that's causing it (the `update_config` lines suggest that you're using one), but it's not easy to know without seeing that.

I got Claude to try to replicate it after I failed and it also couldn't manage it. It did find a bug with type-casting of environment variables though, which it optimistically hypothesised was the problem here: https://github.com/MultiQC/MultiQC/pull/3557

[...]

PR #3557, opened 2026-05-12
Fix env var config overwrite: booleans/ints/floats were stored as strings

ewels:

### Bug: `_env_vars_config()` overwrites parsed values with raw strings

The `_env_vars_config()` function in `multiqc/config.py` correctly parsed `MULTIQC_*` environment variables to their proper types (bool via `strtobool`, int via `int()`, float via `float()`), but then unconditionally overwrote the parsed value with the raw string on the final `env_config[conf_key] = v` line.

**Before (broken):**
```
$ MULTIQC_MAKE_DATA_DIR=false multiqc . -v
New config: {'make_data_dir': 'false', 'make_report': 'false'}
```
The string `"false"` is truthy in Python, so boolean env vars that should _disable_ features did not work. Similarly, integer config values like `MULTIQC_PLOTS_FLAT_NUMSERIES=500` were stored as the string `"500"` instead of the int `500`.

**After (fixed):**
```
$ MULTIQC_MAKE_DATA_DIR=false multiqc . -v
New config: {'make_data_dir': False, 'make_report': False}
```

### Fix: restructure the if/elif chain

The `env_config[conf_key] = v` line (which stores the raw string) now only executes for string-typed and `None`-typed config values, where storing the raw string is the correct behavior. Bool/int/float branches keep their parsed values via `continue` after assignment.

### Also fixed: misleading log messages in `update_config`

- "Prepending directory to sample names" was logged whenever `cfg.prepend_dirs is not None`, even when `prepend_dirs=False`
- "Not cleaning sample names" was logged whenever `cfg.fn_clean_sample_names is not None`, even when `fn_clean_sample_names=True`

Both messages now only log when the value actually indicates the described behavior.

### Regarding #3463

While this fix addresses a real bug in how `MULTIQC_*` environment variables are handled, I was unable to fully reproduce the exact scenario described in the issue (Python 3.13 + BioConda producing `Data: None` with a bare `multiqc .` command). The issue may require specific conda environment variables or config files that the reporter was unaware of. However, this env var handling bug is the most likely root cause if the user's conda environment was setting `MULTIQC_MAKE_DATA_DIR` or `MULTIQC_MAKE_REPORT` env vars, or if any other `MULTIQC_*` boolean env vars were being set by conda activation scripts.

### Tests

Added `tests/test_env_vars_config.py` with 18 tests covering:
- Boolean parsing (true/false/yes/no/1/0)
- Integer and float parsing
- String and None-typed config values
- Invalid values being ignored
- Unknown keys being ignored
- Reserved names being skipped
- End-to-end application of parsed values to config globals

<div><a href="https://cursor.com/agents/bc-9b8fcab3-d012-4293-97a8-8a7e5d9b26e9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-9b8fcab3-d012-4293-97a8-8a7e5d9b26e9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/multiqc/config.py
+++ b/multiqc/config.py
@@ -489,15 +489,14 @@ def _env_vars_config() -> Dict:
                 except ValueError:
                     logger.warning(f"Could not parse a float value from the environment variable ${k}={v}")
                     continue
-            elif globals()[conf_key] is None:
-                pass
-            elif not isinstance(globals()[conf_key], str):
+            elif globals()[conf_key] is None or isinstance(globals()[conf_key], str):
+                env_config[conf_key] = v
+            else:
                 logger.warning(
                     f"Can only set scalar config entries (str, int, float, bool) with environment variable, "
                     f"but config.{conf_key} expects a type '{type(globals()[conf_key]).__name__}'. Ignoring ${k}"
                 )
                 continue
-            env_config[conf_key] = v
             logger.debug(f"Setting config.{conf_key} from the environment variable ${k}")
     return env_config
 
--- a/multiqc/core/update_config.py
+++ b/multiqc/core/update_config.py
@@ -138,7 +138,8 @@ def update_config(*analysis_dir, cfg: Optional[ClConfig] = None, log_to_file=Fal
         config.report_comment = cfg.report_comment
     if cfg.prepend_dirs is not None:
         config.prepend_dirs = cfg.prepend_dirs
-        logger.info("Prepending directory to sample names")
+        if cfg.prepend_dirs:
+            logger.info("Prepending directory to sample names")
     if cfg.dirs_depth is not None:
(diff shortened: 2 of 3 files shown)


ewels:
@multiqc-bot fix linting

PR #3586, opened 2026-05-21
Add support for lineage column in Nextclade

schorlton-bugseq:
Looks like Nextclade can output a lineage column for some datasets, e.g. Monkeypox virus. This exposes that column. If no samples have that column, I confirmed that it is not visible. Also corrects the description of the module. Thanks!


ewels (approved):
Thanks!

PR #3588, opened 2026-05-21
Update summarization endpoints

MythicalFish:
## Summary

Splits the single `seqera_api_url` / `seqera_website` config into three distinct URLs so the report can point at the right service in each context:

- `seqera_website` (`https://seqera.io`) — marketing/info links only.
- `seqera_ai_url` (`https://ai.seqera.io`) — Seqera AI web app, used for the "Continue in chat" link.
- `seqera_ai_api_url` (`https://ai.seqera.io/v1/web`) — backend API used by the Seqera AI provider for streaming summaries.

Previously, "Continue in chat" was hardcoded to `${seqera_website}/ask-ai/...` and the API client hit `${seqera_api_url}/internal-ai/query` on `intern.seqera.io`. Both now resolve under the dedicated `ai.seqera.io` host, and the backend endpoint moves from `/internal-ai/query` to `/internal-ai/report-summary`.

## Changes

- **Config** (`multiqc/config.py`, `multiqc/config_defaults.yaml`, `multiqc/utils/config_schema.py`):
  - Added `seqera_ai_url` and `seqera_ai_api_url`.
  - Deprecated `seqera_api_url`; if set in a user config or `MULTIQC_SEQERA_API_URL`, it's transparently routed to `seqera_ai_api_url` with a warning, so existing configs keep working.
- **Backend** (`multiqc/core/ai.py`): `SeqeraClient` now calls `${seqera_ai_api_url}/internal-ai/report-summary`.
- **Frontend** (`templates/default` + `original`, plus `head.html` / `header.html`): exposes `seqeraAiUrl` / `seqeraAiApiUrl` JS globals, updates `ai.js` / `ai-helpers.js` to use them, and rebuilds `multiqc.min.js`.
- **Docs**: `docs/markdown/ai/index.md` and `docs/markdown/config_schema.md` updated; wizard HTML/JSON schema regenerated.

## Backwards compatibility

- Old `seqera_api_url` config key and `MULTIQC_SEQERA_API_URL` env var still work — they're aliased to `seqera_ai_api_url` with a deprecation warning.
- Default behaviour unchanged for users who don't override these settings.

## Test plan

- [ ] `pytest -vv` passes (especially `tests/test_config_wizard.py` for schema drift).
- [ ] Generate a report with default config: "Continue in chat" opens `https://ai.seqera.io/chat/<thread-id>`.
- [ ] Generate an AI summary via Seqera provider end-to-end and confirm the request hits `https://ai.seqera.io/v1/web/internal-ai/report-summary`.
- [ ] Set legacy `seqera_api_url: https://example.com` in a config — confirm the deprecation warning and that the API call goes to `https://example.com/internal-ai/report-summary`.
- [ ] Spot-check the rendered report in the `default` and `original` templates.


config.yaml
ewels, line 1: Has this been accidentally committed?

ewels (commented):
The `seqeraWebsite` config option doesn't seem to really be used anywhere - I'm wondering if we really need to split into 3 and deprecate the one config option? I'd prefer to leave them in tact if there's no tangible improvement.

config.yaml
MythicalFish: My thinking was that a distinction between those websites might be useful in future - but no worries I removed the unused one. :) Let me know if you'd rather keep them named as `seqera_website` and `seqera_api_url` too

ewels (approved):
Perfect, thanks!

ewels:
@multiqc-bot fix linting

PR #3592, opened 2026-06-10
Replace rstrip with removesuffix in custom content module

hwalinga:
The custom content module sometimes removes the last letter when determining the section title. I suspect it is this use of `.rstrip`. 

`.rstrip` does not remove a substring, but removes all matching letters at the end of the string. Python 3.9 (which is the minium supporting version for Multiqc introduced `.removesuffix` to have this addressed. 

Compare: 

```
>>> f = "Organism_mqc.png"
>>> f.rstrip(".png").rstrip("_mqc")
'Organis'
>>> f.removesuffix(".png").removesuffix("_mqc")
'Organism'
```


ewels (approved):
Nice one, thank you!

PR #3597, opened 2026-06-23
Add Citations module for structured tool citations (CSL-JSON / BibTeX)

edmundmiller:
## Summary

Adds a new `citations` module that renders **structured software citations** into the report rather than parsing a tool's output. It consumes a citations file produced by a workflow or pipeline and emits two sections:

- **Methods** — a sentence listing each tool with the version that ran and a Harvard short citation, e.g.
  > Tools used in the workflow included: fastqc (0.12.1, Andrews (2010)), samtools (1.21, Danecek *et al.* (2021)), snpsift (5.2).
- **Bibliography** — a full reference per tool, with DOI links.

Short citations use `Surname (year)` / `Surname et al. (year)` (italic *et al.*) and link to the DOI if present, else the homepage. Tools with only a DOI and no publication are identified by tool name rather than rendered anonymously.

### Inputs

| Format | Glob | Notes |
| --- | --- | --- |
| CSL-JSON (canonical) | `*.csl.json` | JSON array of CSL items; `custom.tool` / `custom.version` carry the tool name and runtime version |
| BibTeX (optional) | `*.citations.bib` | requires `pip install multiqc[citations]` (adds `bibtexparser`); tool name from a `tool` field or the entry key, version from a `version` field |

CSL-JSON wins when the same tool appears in both. Tool names act as sample names, so `--ignore-samples` filters tools.

### Why

Motivated by nf-core pipelines: nf-core-utils assembles structured citation data (CSL-JSON) and this module owns the rendering. The module itself is deliberately **tool- and pipeline-agnostic** — any MultiQC user who drops a CSL-JSON citations file gets a citations section.

<details>
<summary>Implementation notes</summary>

- `csl.py` — the `Citation` data model + CSL-JSON parser + Harvard renderers (`render_inline`, `render_bibliography`).
- `bibtex.py` — BibTeX parser that maps into the same `Citation` model. `bibtexparser` is an **optional** dependency (new `citations` extra); a `.bib` file found without it is skipped with a warning, so CSL-JSON works dependency-free.
- Author handling: CSL `family`/`given` and `literal` forms, plus a plain author string; surname parsing differs by source (CSL "surname-first" heuristic vs. BibTeX name semantics). HTML is escaped.
- bibtexparser ships no stubs; `multiqc/modules/*` is already excluded from the mypy config, so this doesn't affect type-checking.

### Testing

- `multiqc/modules/citations/tests/` — unit/snapshot tests for both parsers and the Harvard rules (single author, multi-author *et al.*, DOI-only, homepage-only, partial publication, plain-string authors, DOI normalization, HTML escaping).
- `tests/test_modules_run.py::test_all_modules[citations]` and `test_ignore_samples[citations]` pass against the companion test data.
- `--strict` run is clean; `ruff`, `ruff format`, and `.github/workflows/code_checks.py` pass.

</details>

### Companion PR

Test data: MultiQC/test-data#380 (required for the integration tests to pass).

> Note: the CSL-JSON input contract (filename glob, author representation) is being finalized with the nf-core-utils maintainers; happy to adjust the glob/fields to match before merge.


pyproject.toml
```diff
@@ -70,6 +70,9 @@ classifiers = [
 ]
 
 [project.optional-dependencies]
+citations = [
```
ewels: I don't particularly like this form of installation because conda doesn't support it, only pip.

This means that I need to either include it for all bioconda installations (in which case why bother, just make it a main-line dep) or have a second bioconda package which has this additional dep (don't want to do that either).

Viable options are:
1. Just make it a dependency for everyone all the time
2. Keep this with the safe fallback you have, and tell people to install it in addition to MultiQC manually if using conda
3. Pull out the parts of the library you actually need into a bundled python file (?)

MultiQC already does a bunch of stuff with BiB files, can it not already handle the format? Or is it JS only or something?

edmundmiller: Good call, I went through how MultiQC handles BibTeX today before deciding.

The only BibTeX handling in the codebase is the toolbox "Download BibTeX file" button. `templates/original/assets/js/toolbox.js` (~L185) gathers the bare DOI strings from the `.module-doi` elements and, per DOI, calls Crossref content negotiation in the browser: `https://api.crossref.org/works/{doi}/transform/application/x-bibtex`, concatenates the results and saves `multiqc_references.bib`. The DOI popovers in `multiqc.js` (~L77) hit Crossref the same way. On the Python side we only ever store the bare DOI string (`BaseMultiqcModule.doi`, written out by `report.dois_tofile` as `multiqc_citations.*`); no author/title/year is kept, and there is no `.bib` or `.csl.json` parser anywhere.

So MultiQC *generates* BibTeX by delegating DOI to BibTeX to Crossref; it never *parses* BibTeX. There is no existing Python machinery to reuse for reading a `.bib`, which is exactly why the BibTeX path here reached for `bibtexparser`.

Given that, I've dropped BibTeX input from this PR. It is now CSL-JSON only and adds no dependency. I think that is the right scope rather than just a workaround:

- CSL-JSON is already JSON, so the canonical path needs zero new deps.
- The same DOI content negotiation MultiQC already relies on also returns CSL-JSON (`Accept: application/vnd.citationstyles.csl+json`), so a tool that only has a DOI can be turned into the canonical format with no BibTeX parsing at all.
- A `.bib` input is therefore a convenience, not a gap. If we do want it, the dependency-free option that mirrors what MultiQC already does is DOI/Crossref content negotiation (or vendoring a small `@entry{...}` reader), rather than pulling in bibtexparser.

I've parked the BibTeX-input work on a separate branch (`citations-bibtex`) so we can decide that there without blocking this one. Happy to go whichever way you prefer.

multiqc/modules/citations/csl.py
```diff
@@ -0,0 +1,55 @@
+"""Parse CSL-JSON tool citations into the shared Citation model.
```
ewels, line 1: Please move this into `citations.py` - doesn't need to be in a separate file.

multiqc/modules/citations/citation.py
```diff
@@ -0,0 +1,259 @@
+"""The format-agnostic Citation model and Harvard short-form rendering.
```
ewels, line 1: Is this copied / vendored from elsehwere? Please include source and copy date if so, plus check upstream license.

PR #3608, opened 2026-07-20
Make thinking configurable for bedrock

jchorl:
Fairly self-explanatory


ewels (approved):
I thought that you were adding a load of new config variables at first, but I see that these are all already used in the other clients. So yeah, LGTM - thanks!

PR #3612, opened 2026-07-27
mosdepth: add per-region coverage table from regions.bed.gz and thresholds.bed.gz

### LINKED ISSUE #3604: mosdepth: add per-region coverage table from regions.bed.gz and thresholds.bed.gz
### Description of feature

The mosdepth module currently parses `*.mosdepth.global.dist.txt`, `*.mosdepth.region.dist.txt`, and `*.mosdepth.summary.txt` to produce coverage distribution plots and General Stats columns. It does not parse two other mosdepth output files that are particularly valuable for targeted sequencing:

`*.regions.bed.gz` - mean coverage per named region (produced when `--by BED_FILE `is used)
`*.thresholds.bed.gz` - base counts at each threshold per region (produced when `--thresholds` is used alongside `--by BED_FILE`)

File formats (example files attached):

`*.regions.bed.gz:`

```
chrom, start, end, region_name, mean_coverage
chr1  3812085  3857723  CEP104  45.83
chr1  5862803  5993455  NPHP4   50.91
```

`*.thresholds.bed.gz:`

```
#chrom  start    end      region  20X    30X
chr1    3812085  3857723  CEP104  45638  45638
chr1    5862803  5993455  NPHP4   130652 130652
```

`chrom, start, end, region_name`, then one column per threshold level (variable, read from header).

## Proposed addition

A new section "Per-region coverage" - that joins the two files on (chrom, start, end, region_name) and produces a table with:

Region | Chrom | Start | End | Mean coverage | % bases ≥20X | % bases ≥30X
-- | -- | -- | -- | -- | -- | --
CEP104 | chr1 | 3812085 | 3857723 | 45.83 | 100.0 | 100.0
NPHP4 | chr1 | 5862803 | 5993455 | 50.91 | 100.0 | 100.0

Threshold percentages are computed as base_count / (end - start) * 100. The threshold levels (20, 30 etc.) should be read dynamically from the thresholds.bed.gz header so they reflect whatever the user passed to --thresholds.

## Why this is useful

For targeted sequencing (panels, adaptive sampling), mean coverage per region alone is not sufficient - a gene with mean 45× can still have a dropout where 15% of bases never reach 20×. The threshold output directly answers the clinical question: "what fraction of each target gene is covered at the required depth?" Clinical/diagnostic reporting standards (e.g. ACMG) typically express coverage as "≥20× for >99% of target bases," not as a mean.

The workround would need a local custom_content script to reformat the two files into a MultiQQ TSV. A native mosdepth module section would remove that workaround and make the feature available to any pipeline using mosdepth with -`-by BED_FILE --thresholds`.

[example.mosdepth.global.dist.txt](https://github.com/user-attachments/files/29704614/example.mosdepth.global.dist.txt)
[example.mosdepth.region.dist.txt](https://github.com/user-attachments/files/29704617/example.mosdepth.region.dist.txt)
[example.mosdepth.summary.txt](https://github.com/user-attachments/files/29704615/example.mosdepth.summary.txt)
[example.regions.bed.gz](https://github.com/user-attachments/files/29704616/example.regions.bed.gz)
[example.thresholds.bed.gz](https://github.com/user-attachments/files/29704613/example.thresholds.bed.gz)

aksenia:
## Summary

Adds a "Per-region coverage" table to the mosdepth module, joining `*.regions.bed.gz` (mean coverage per region) with `*.thresholds.bed.gz` (bases covered at each `--thresholds` level per region) on chromosome/start/end position. Threshold levels are read dynamically from the `thresholds.bed.gz` header, so they reflect whatever was passed to `--thresholds`.

Closes #3604

<details>
<summary>Full write-up</summary>

### Behaviour

- New search patterns: `mosdepth/regions_bed` (`*.regions.bed.gz`) and `mosdepth/thresholds_bed` (`*.thresholds.bed.gz`).
- The section only appears when at least one `thresholds.bed.gz` file is found (i.e. mosdepth was run with `--by BED_FILE --thresholds`); it's a pure add-on and doesn't affect any existing mosdepth output.
- Table rows are keyed as `"<sample> | <region>"` (`col1_header: "Sample | Region"`), one row per sample per region, since MultiQC tables are inherently per-row-per-sample and a per-region table doesn't otherwise compose with that across multiple samples. This mirrors the existing `"Run Name | Lane"` pattern in the bases2fastq module.
- `regions.bed.gz` may or may not have a region-name column (mosdepth omits it when the `--by` BED file has no 4th column); the join key is always `(chrom, start, end)`, and the region name comes from `thresholds.bed.gz`, which always carries one (falling back to the literal `"unknown"` from mosdepth itself when the source BED had no name column).
- Region names aren't guaranteed unique within a sample (real panel BEDs commonly have one row per exon for the same gene, and of course every "unknown"-named row would collide too), so whenever a name repeats within a sample, its coordinates are appended to disambiguate, e.g. `LRP1B (10:100-200)`.
- `% bases >= NX` is computed as `bases_at_threshold / (end - start) * 100`.

### Testing

- Unit tests for the parsing/joining functions (full output, minimal/no-name-column input, empty file, invalid header, duplicate-name disambiguation, zero-length region) in `multiqc/modules/mosdepth/tests/test_per_region_coverage.py`.
- `pytest tests/test_modules_run.py -k mosdepth` passes (`test_all_modules` and `test_ignore_samples`).
- Ran `multiqc --strict` against the real mosdepth fixtures in `MultiQC/test-data` (`tiny.*`, `test.*`, `full-wgs.*`, `shallow.*`), which already include both 4- and 5-column `regions.bed.gz` variants and differing threshold sets, no crash.

</details>

## Test plan

- [x] `pytest multiqc/modules/mosdepth/tests/ -v`
- [x] `pytest tests/test_modules_run.py -k mosdepth -v`
- [x] `multiqc test-data/data/modules/mosdepth -m mosdepth --strict`
- [x] `ruff check`, `mypy`, `python .github/workflows/code_checks.py`, `prek run`


aksenia:
multiqc_pr3612
Screenshot of how the table would look in the report.

multiqc/search_patterns.yaml
```diff
@@ -661,6 +661,10 @@ mosdepth/global_dist:
   fn: "*.mosdepth.global.dist.txt"
 mosdepth/region_dist:
   fn: "*.mosdepth.region.dist.txt"
+mosdepth/regions_bed:
+  fn: "*.regions.bed.gz"
```
ewels, line 665: I think these are too generic and are likely to occur for other outputs too. We don't want folks to have a regions file for their chipseq analysis and accidentally get the mosdepth section in their report.

I would normally say to include a header line to make it more specific, but in this case the header line is also super generic.

How is mosdepth coming up with these names? Is this filename part definitely always the same? Any ideas on how we can make this more specific?

multiqc/modules/mosdepth/mosdepth.py
```diff
@@ -604,3 +708,129 @@ def parse_cov_dist(
...
+                "description": "Region start position",
+                "scale": False,
+                "format": "{:,d}",
+            },
+            "end": {
+                "title": "End",
+                "description": "Region end position",
+                "scale": False,
+                "format": "{:,d}",
+            },
```
ewels: I'm not sure how useful it is to have these as separate columns in the report - they're not actually data and they take up quite a bit of room.

My preference would be to concatenate into a string and have them as a single column. What do you think?

ewels (commented):
Quick first pass on my phone (I'm on holiday currently). Looks mostly good I think, though a couple of minor points to address.

I know that you only use this when used with the target option. I'm wondering if it would be useful to also support the regions if target isn't used? Then join if it's there..?

I'm also a bit worried of bloating the report if there are very many regions. Not sure that there's much that we can do about that though. Normally I say that MultiQC should only be used to summarise data rather than reporting raw data like this.

ewels:
Triggering an automated review. Will finish my review when I'm back and have time. Thanks for this!

/review

multiqc/search_patterns.yaml
aksenia: Good points! Checked directly in mosdepth's source (mosdepth.nim):

```nim
fthresholds = wopen_bgzi(prefix & ".thresholds.bed.gz", ...)  # only if --thresholds
fregion = wopen_bgzi(prefix & ".regions.bed.gz", ...)         # only if --by
```
Both suffixes are hardcoded string literals appended to the user-chosen prefix - not templated or configurable by any flag, and no history of renaming in CHANGES.md. So the filename part is as specific as it's ever going to get; there's no naming trick that narrows it further.

Content-sniffing the header doesn't work either, for a different reason: these are gzip files, and MultiQC's file search opens candidate files as plain UTF-8 text to check contents/contents_re - it never decompresses. A contents_re pattern on the (decompressed) header would just never match, since the search only ever sees the raw gzip bytes.

Given both routes are dead ends, I changed the actual failure mode instead: a file that matches the name but doesn't parse as mosdepth output (wrong column count/header) is now skipped with a debug log rather than crashing the module. So a chipseq user's coincidentally-named `regions.bed.gz` would just fail our column-shape check and get silently ignored - no crash and no bogus section in the report. Good?

aksenia: Large amount of regions bloating the report is a fair concern. A few things that make this safer than a raw per-region dump:

- It's opt-in on `--by` (and further, threshold columns only appear with `--thresholds`); nothing changes for the vast majority of mosdepth users who don't use those flags.
- The intended use case is targeted panels (hundreds of regions), not genome-wide data - the docstring now says so explicitly, and the help text calls out that a whole-genome `--by <window_size>` run is a different scenario.
- MultiQC's table component already auto-degrades past `max_table_rows` (500 by default): instead of rendering the full table, it shows a violin/distribution plot with each region as a point, and a "show table" toggle if someone really wants the full thing. So a pathological genome-wide run doesn't actually dump raw data into the report - it falls back to a summary view, which is the MultiQC philosophy you're describing.
- On top of that fallback, I added a log warning when the row count crosses `max_table_rows`, so it's visible in the run log that this happened rather than just silently switching plot types.

So I'd frame it as: this is a summarizing view when used as intended (per-panel breakdown), and degrades to an actual MultiQC-style summary (distribution plot) rather than a wall of raw rows when it isn't. Let me know if you'd still like a harder guardrail (e.g. refusing to build the section past some row count) rather than relying on the existing table→violin fallback.

aksenia:
thanks @ewels ! I have tried to address yours and Claudes reviews. The Mypy failure now seems not to do with this PRs changes. I have submitted a separate small PR #3613 to address some of the Mypy complaints.

aksenia:
Made another small PR to address the unrelated MyPy issue: [!3627](https://github.com/MultiQC/MultiQC/pull/3627).

PR #3613, opened 2026-07-27
mosdepth: fix pre-existing mypy errors

aksenia:
## Summary

Small type-only maintenance PR, no behaviour change. Fixes the 6 pre-existing `mypy` errors in `multiqc/modules/mosdepth/mosdepth.py` (noticed while working on #3612, split out separately since it's unrelated cleanup):

- Renaming the `parse_cov_dist()` tuple result before reassigning it to a list comprehension, so mypy doesn't try to narrow the same variable to two incompatible types (also fixed a follow-on `"object" has no attribute "update"` error caused by the same narrowing issue).
- Widening `perchrom_plot`'s declared type to `Optional[Union[Plot, str]]`, since `linegraph.plot()`/`bargraph.plot()` can return `None` for all-zero data.
- Casting the `general_stats_addcols()` argument to match its invariant `Dict[..., ValueT]` parameter, since `genstats_by_sample`'s `int | float` value type doesn't satisfy `ValueT = Union[int, float, str, bool]` under Dict's invariance rules.

## Test plan

- [x] `mypy multiqc/modules/mosdepth/mosdepth.py` - 0 errors (down from 6)
- [x] `ruff check`, `python .github/workflows/code_checks.py`
- [x] `pytest tests/test_modules_run.py -k mosdepth` and `pytest multiqc/modules/mosdepth/tests/`
- [x] `multiqc test-data/data/modules/mosdepth -m mosdepth --strict`

### DIFF
--- a/multiqc/modules/mosdepth/mosdepth.py
+++ b/multiqc/modules/mosdepth/mosdepth.py
@@ -305,7 +305,7 @@ class MultiqcModule(BaseMultiqcModule):
                 self.write_data_file(perchrom_avg_by_sample, "mosdepth_perchrom")
 
                 num_contigs = max([len(x.keys()) for x in perchrom_avg_by_sample.values()])
-                perchrom_plot: Union[Plot, str]
+                perchrom_plot: Optional[Union[Plot, str]]
                 if num_contigs > 1:
                     perchrom_plot = linegraph.plot(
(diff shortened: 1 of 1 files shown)


ewels (approved):
Nice spot, thank you!

PR #3618, opened 2026-08-05
Fix vg module IndexError on newer vg stats Softclips output

pascalangst:
vg stats' Softclips line format changed in recent vg releases to include a descriptive clause before "in", e.g. 
```
Softclips: 1501255755 bp (12.8427% of bases, 1153.34 bp/aligned) in 584745 read events
```

instead of the older:
```
Softclips: 1501255755 bp in 584745 read events
```

The vg module's parser splits on the literal substring `" bp in "`, which isn't present in the newer format, causing an `IndexError` in `parse_vg_stats_file()`.

Fix: normalize the line via `re.sub` before the existing split logic, stripping the parenthetical back to the old format. Same two numbers either way; only the percentage/bp-per-aligned annotation (not currently parsed by this module) is discarded.

Verified against both the new format (previously crashing) and the module's own docstring example (old format) — no regression.

Here the updated testfile: https://github.com/vgteam/vg/blob/81ba66ed829ee75a43e029145d66256dcd640a52/test/correct/10_vg_stats/15.txt


ewels:
@multiqc-bot fix linting

multiqc/modules/vg/vg.py
```diff
@@ -165,7 +166,7 @@ def parse_vg_stats_file(f):
                 var.append(s[0] + " (bp)")
                 val.append(s[1].split(" bp ")[0])
                 var.append(s[0] + " (reads)")
-                val.append(s[1].split(" bp in ")[1].split(" ")[0])
+                val.append(re.sub(r" bp \([^)]*\) in ", " bp in ", s[1]).split(" bp in ")[1].split(" ")[0])
```
ewels: We only actually care about the final number right?

This would be a smaller change and avoids using regexes:
```suggestion
                val.append(s[1].split(" read events")[0].split(" ")[-1])
```

multiqc/modules/vg/vg.py
```diff
@@ -1,4 +1,5 @@
 import logging
+import re
```
ewels:
```suggestion
```
Not needed if we don't use the regex.

ewels (commented):
Thanks for this! I think that there's probably a cleaner alternative though.

ewels:
I added the new example file to the MultiQC/test-data repo, so CI tests should now run including this file. Also helps to catch future regressions. Thanks for that!

PR #3619, opened 2026-08-06
Modify text description for BUSCO module

etvedte:
Remove references to specific BUSCO and odb versions in the module description text. The correct BUSCO version is reflected in Software Versions, and the correct odb version is reflected in the lineage for the BUSCO plot.


ewels (approved):
Good catch, thanks!

PR #3622, opened 2026-08-11
Seed the jitter in somalier and peddy sex-check plots

immortal71:
The somalier and peddy sex-check plots jitter the x coordinate with
(random.random() - 0.5) * 0.1. There is no random.seed,
np.random.seed or default_rng anywhere under multiqc/, so the same
input produces a different plot on every run.

The jitter is deliberate and no QC value changes, so this only
affects the rendered report.

This uses a module-local random.Random so the plots are stable and
global RNG state is untouched.

Discussed with @ewels by email.


ewels (approved):
Thanks!

PR #3641, opened 2026-09-02
fix(riker): tolerate empty metric fields in TSV parsers

clintval:
Riker leaves an undefined metric (for example Picard's estimated library size, hs_library_size) as an empty string. The hybcap parser, like the wgs, gcbias, isize, and alignment metric parsers, ran float() over every cell and raised on the empty value, so it logged a warning and dropped the entire row; the whole hyb-selection section then went missing.

Add a to_float() helper that maps an empty or blank cell to None (missing) while still raising on genuinely non-numeric values, and use it in the metric-row parsers. Real numbers are unchanged; only empty cells become missing, and MultiQC renders a missing cell as blank. Guard the two spots that multiply a possibly-missing value.

### DIFF
--- a/multiqc/modules/riker/alignment.py
+++ b/multiqc/modules/riker/alignment.py
@@ -1,14 +1,14 @@
 """Parse riker `alignment` (alignment-metrics.txt) outputs."""
 
 import logging
-from typing import Dict
+from typing import Dict, Optional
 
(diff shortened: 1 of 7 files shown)


ewels (approved):
LGTM, thanks!

PR #3645, opened 2026-09-05
feat(riker): group alignment table by sample with pair as primary row

clintval:
Render the Alignment metrics table with MultiQC sample grouping so each sample's pair row is the headline and the read1 / read2 rows nest under it, expandable on click. Previously the three categories were flat siblings, which made the bar shading and sorting misleading because pair values are a different scale than the per-read values.

Here's what it looked like **_before_**:

Screenshot 2026-09-04 at 11 33 40 AM

Here's what it looks like _**now**_:

Screenshot 2026-09-05 at 9 31 08 AM

My PR here is also supportive of single-end data if/when someone runs riker on single-end data.

Before merge, I'd like to get @tfenne's blessing on this aesthetic change!

### DIFF
--- a/multiqc/modules/riker/alignment.py
+++ b/multiqc/modules/riker/alignment.py
@@ -166,9 +166,12 @@ def parse_reports(module):
         plot=bargraph.plot(bar_data, bar_keys, bar_config),
     )
 
-    # Per-category metrics table, grouped by sample so the `pair` row is the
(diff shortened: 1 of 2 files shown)


ewels (approved):
Nice! This is exactly the kind of thing I hoped sample grouping would be used for :) Thanks!

PR #3647, opened 2026-09-08
Correct samtools markdup search pattern

CormacKinsella:

Fixes a bug in the `samtools` MultiQC module for `samtools markdup` output.

The issue is in the `samtools/markdup_txt` search pattern: it currently looks for the literal string `^COMMAND:`, which never matches valid output because the file begins with `COMMAND:` and does not include a literal caret. As a result, valid `samtools markdup` metrics files are silently skipped and never appear in the report.

The fix is to search for `COMMAND:` and `samtools markdup` rather than the literal `^COMMAND:` so the file is detected correctly.


ewels (approved):
We got there in the end 😅 

Thank you!

PR #3658, opened 2026-09-11
Add support for nexons (https://github.com/s-andrews/nexons)

s-andrews:
[nexons_output_PBM75668_pass_L001_barcode01_nexons_stats.txt](https://github.com/user-attachments/files/32100867/nexons_output_PBM75668_pass_L001_barcode01_nexons_stats.txt)
[nexons_output_PBM75668_pass_L001_barcode02_nexons_stats.txt](https://github.com/user-attachments/files/32100869/nexons_output_PBM75668_pass_L001_barcode02_nexons_stats.txt)


ewels:
@multiqc-bot fix linting

ewels:
Thanks @s-andrews! Tests will fail until the test data is in the `test-data` repo, I'll get to that and a review ASAP.

In the meantime I'll kick off an automated first pass:

/review

ewels:
@s-andrews are the files really meant to be `.txt` or did you just call them that to attach to GitHub? Should I rename as `.json`?

ewels:
Ah the search pattern uses `.txt`, so I guess that's the correct output extension.

ewels:
Report looks nice!

Quick thoughts:

* Would it make sense to combine `Unique Gene Match`, `Partial Unique Transcript Match`, `Full Transcript Match`, `Multi Gene Match` and `No Gene Match` into a single plot, with tabs along the top to see the different variants? Or maybe better: a single grouped bar plot? (`"stacking": "group"`). This would better match your nested donut plot on the readme too.
* I agree with the Claude review that the tables need some different colour schemes. I try to not have 2 columns with the same scheme next to each other, for faster visual registration of which is which and more semantic meaning of the colours. See the [docs](https://docs.seqera.io/multiqc/development/modules#table-colour-scales).
* Similarly, semantic colours and ordering of the bar plots would be good. No alignment = bad = red, far right of bar instead of far left. Can match the colours in the nexons report if you like, or take from another module like STAR.
* Why don't the `Alignment Directionality` bars sum to 100%?

Yeah and going over Claude's review I agree with most of the points its made (most overlap with what I came up with myself too by the look of it).

ewels:
I've pushed a commit that addresses my review feedback, let me know if this is looks ok to you! Otherwise I'll merge in a few days.

PR #3663, opened 2026-09-16
Fix the data directory name #3657

### LINKED ISSUE #3657: Data directory location inconsistency
### Description of bug

If the `-n` switch is used (e.g. `-n ./test/multiqc`) and no previous report is found or the `-f` switch is used, the data folder is in the same directory as the report file (e.g. `./test/multiqc_data/` and `./test/multiqc.html`).
But if there is a previous report and `-f` switch is not used, the data directory is located in the working directory instead (e.g. `./multiqc_data_1_2_3/` and `./test/multiqc_3.html`), which is inconsistent.

There is also an inconsistency in the ID suffix (e.g. `_1_2_3` for the data directory vs. `_3` only for the report file).

[bowtie2.log](https://github.com/user-attachments/files/32095940/bowtie2.log)

### MultiQC Error log

```console
$ multiqc -n test/multiqc test/

/// MultiQC 🔍 v1.35

       file_search | Search path: /mnt/c/test
         searching | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 3/3
           bowtie2 | Found 1 reports
     write_results | Existing reports found, adding suffix to filenames. Use '--force' to overwrite.
     write_results | Data        : multiqc_data_1_2_3
     write_results | Report      : test/multiqc_3.html
           multiqc | MultiQC complete
```

### Before submitting

Jelinek-J:
## Description

Fixes #3657

When `-n` is used with a path (e.g. `multiqc -n test/multiqc test/`) and previous outputs already exist (and `-f` is not used), MultiQC currently produces inconsistent results:

| Output | Expected | Actual |
|---|---|---|
| Report | `test/multiqc_3.html` | `test/multiqc_3.html` |
| Data dir | `test/multiqc_data_3/` | `./multiqc_data_1_2_3/` |

Two problems:

1. **Wrong location** — the data (and plots) directory lands in the working directory instead of next to the report, losing the `test/` prefix.
2. **Cumulative suffix** — the numeric suffix stacks on every iteration of the conflict-resolution loop (`_1`, then `_1_2`, then `_1_2_3`), unlike the report which gets a clean `_3`.

### Root cause

In `_create_or_override_dirs()` (`multiqc/core/write_results.py`), the report base name is recomputed on each loop iteration from the original `output_names.output_fn_name` (which keeps the relative path), but the data and plots directory bases are derived from the **current, already-suffixed** path:


ewels (approved):
Nice catch, thanks!

ewels:
I'm really surprised that no-one has reported this before! 😅

PR #3666, opened 2026-09-17
Normalize data to percentages in FastQC plots

### LINKED ISSUE #3662: Percentages for FastQC: Sequence Length Distribution
### Description of feature

In the FastQC module, the “Per Sequence GC Content” plot has a tab switcher to toggle between Percentages and Counts. Would it be possible to add the same option to the “Sequence Length Distribution” and “Per Sequence Quality Scores” plots as well?

Jelinek-J:
Closes #3662

The FastQC "Per Sequence GC Content" plot can be toggled between relative and absolute counts ("Percentages" and "Counts") via a tab switcher. This pull request applies the same behaviour to two more FastQC plots that were previously available in absolute counts only:
- "Per Sequence Quality Scores"
- "Sequence Length Distribution"

Both plots were modified in the same way as "Per Sequence GC Content", for consistency and easier maintenance (elsewhere, the normalised dataset in "Sequence Length Distribution" can be calculated when the plot is actually rendered only - i.e. when `not only_single_length` - to avoid needless work in the single-length case).


ewels (changes requested):
This is a nice addition, thanks! However I don't want to change the default view for these plots. Happy to add the normalised counts as a second tab tab instead of primary, so that it's an additional new view for the data.

If it's important for you to have your reports showing the normalised view by default, we could have a module-specific config option to flip this behaviour that you could use.

PR #3667, opened 2026-09-17
Line plot: show markers for single-point series in "lines" mode

Jelinek-J:
When a line plot has many data points in total, the plot switches to `mode: "lines"` (once the number of points exceeds `config.lineplot_number_of_points_to_hide_markers`, default 50), and markers are hidden for all series. In this mode, a series consisting of a **single point** is rendered as nothing at all: a line needs two points, and there are no markers to fall back on.

A rescue for this case already exists in `Dataset.create_figure()` (`if len(series.pairs) == 1: params["mode"] = "lines+markers"`), but that code path only runs for Python-side rendering — flat image plots, plot exports and notebook use. The interactive HTML report is rendered in the browser by `src/js/plots/line.js` from the serialized plot model, which does not go through `create_figure()`, so single-point series remain invisible there.

I tried to fix it using the more consistent `series.extra_trace_params.update(mode="lines+markers")` (and `setdefault`), but `Series.marker` turned out to be the only per-series solution that worked: `extra_trace_params` is applied only in `create_figure()` and is ignored by the JavaScript rendering path, while `marker` is a top-level `Series` field that both paths read.


ewels (approved):
Nice fix! Thanks for this :)

PR #3670, opened 2026-09-19
feat(fgbio): add CollectDuplexSeqMetrics family size distribution

clintval:
Adds support for parsing family_sizes.txt output from fgbio CollectDuplexSeqMetrics. Creates a line graph showing SSCS family size distribution with two series per sample (duplex vs orphaned SSCS) and tabs for read fractions and raw family counts. Adds general stats columns for duplex rate and mean family size.


clintval:
Opened against upstream by mistake; moving to fork.

PR #3671, opened 2026-09-20
Fixed parsing bug in spaceranger

lillux:
* [x] This comment contains a description of changes (with reason)

## Description

Fix the Space Ranger module failing when `Pipeline Version` does not follow the expected `spaceranger-X.Y.Z` format.

The parser previously assumed that the value always contained a hyphen:

```python
software_name, software_version = software.split("-")
```

Reports containing values such as `4.1.0` or `Space Ranger 4.1.0` therefore raised:

```text
ValueError: not enough values to unpack (expected 2, got 1)
```

The updated parsing preserves the existing behavior for hyphenated values and falls back to `spaceranger` as the software name for non-hyphenated formats.

## Testing

Tested locally with affected Space Ranger 4.1.0 HTML reports. MultiQC completes successfully and extracts the report metrics.


ewels:
@multiqc-bot fix linting

ewels (approved):
Thanks for this!

PR #3672, opened 2026-09-20
Fix Windows code check exemptions

### LINKED ISSUE #3668: Configured code checks fail on existing OrderedDict imports
## Summary

The repository's `code_checks.py` hook currently fails on the clean default branch because it rejects two existing `OrderedDict` imports in plot modules.

## Reproduction

From the repository root, run:

```text
python .github/workflows/code_checks.py
```

On a Windows checkout, set `PYTHONUTF8=1` first so the checker can read the tracked source files:

```text
$env:PYTHONUTF8="1"
python .github/workflows/code_checks.py
```

The checker completes its other module rules, then reports:

- `multiqc/plots/bargraph.py`: `OrderedDict`
- `multiqc/plots/box.py`: `OrderedDict`

The same checker is registered as the `code-checks` hook in `prek.toml`, so this prevents the configured lint gate from passing on the current default branch.

## Expected result

Either replace the two imports with regular dictionaries where ordering is already guaranteed, or document a verified exception in the checker and hook configuration. The selected resolution should preserve the plot output ordering behavior.

dajiaohuang:
Fixes #3668. Normalize relative paths to forward slashes before comparing against configured exemptions, so the code checks do not report exempt OrderedDict imports on Windows. Tested with python .github/workflows/code_checks.py on Windows; git diff --check passes.


ewels (approved):
Nice spot, thanks!

PR #3674, opened 2026-09-21
New module: mgnifam

vagkaratzas:
Adds a module for [mgnifam](https://github.com/vagkaratzas/mgnifam) (DOI [10.5281/zenodo.22879938](https://doi.org/10.5281/zenodo.22879938)), which builds protein families from sequence clusters by iterative HMM search (`generate_families`) and refreshes existing family models against a newer sequence database (`update_families`).

Since version 3.1.0, each completed mgnifam run writes a small JSON summary: `<chunk>_stats.json` or `<chunk>_updated_stats.json`. The module parses that file, with one chunk per sample. The file carries `"tool": "mgnifam"` and a `schema_version`; files with any schema version other than 1 are skipped with a warning. mgnifam computes all the counts and histograms itself, so the module does no significant computation.

**What the report shows:**

- **General Stats:** families in, % successful, converged (hidden by default), and mean retention (update runs only).
- **Family outcomes:** a bar graph of successful families stacked with each discard reason. There is a warning alert for chunks that completed with contained crashes (exit status 3).
- **Line graphs:** histograms of full MSA size and model length. Update runs also get model length change and retention; retention is binned to 0.05 for display.

---

- [x] There is example tool output for tools in the <https://github.com/MultiQC/test-data> repository or attached to this PR: MultiQC/test-data#391
- [x] Report sections have a description and help text (with `self.add_section`): the sections have descriptions but no help text yet
- [x] There aren't any huge tables with > 6 columns
- [x] Each table column has a different colour scale to its neighbour: `% Successful` and `Retention` both use `RdYlGn`, and they sit next to each other while `Converged` is hidden


multiqc/modules/mgnifam/mgnifam.py
```diff
@@ -0,0 +1,272 @@
...
+                "description": "Mean fraction of round 1 recruits still present at the end of an update",
+                "min": 0,
+                "max": 1,
+                "scale": "YlGnBu",
+                "format": "{:,.2f}",
+            },
+        }
+        general_stats_headers = self.get_general_stats_headers(all_headers=headers)
+        if general_stats_headers:
+            self.general_stats_addcols(data, general_stats_headers, namespace="mgnifam")
```
erikrikarddaniel: `namespace` is meant for a sub-tool within a module, for example `namespace="stats"` in samtools. Here it repeats the module name, so every column ID becomes `mgnifam_mgnifam-families_input` in `multiqc_general_stats.txt`. fastp, for comparison, gives `fastp-pct_duplication`. Without it the IDs are `mgnifam-families_input` and so on, and the module tests still pass (I checked locally).

```suggestion
            self.general_stats_addcols(data, general_stats_headers)
```

multiqc/modules/mgnifam/tests/test_mgnifam.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def test_retention_is_binned_for_display(run_module):
+    stats = _stats("update_families")
+    stats["histograms"]["retention"] = {"0.91": 1, "0.93": 2, "1.0": 3}
+    run_module({"1_updated_stats.json": stats})
+
+    plot = next(p for p in report.plot_by_id.values() if not isinstance(p, str) and p.id == "mgnifam-retention-plot")
+    points = dict(plot.datasets[0].lines[0].pairs)
+    assert points == {0.9: 3, 1.0: 3}
```
erikrikarddaniel: No test covers the crash warning in the outcomes section. `test_both_commands_parse_as_separate_samples` sets `exit_status=3`, but the alert keys on `families["crashed"]`, which stays 0 there. I checked by hand that the alert renders with `crashed: 1`. A test would keep it that way, for example:

```python
def test_crashed_chunks_are_flagged(run_module):
    crashed = _stats(exit_status=3, families={"input": 4, "successful": 3, "discarded": 1, "converged": 2, "crashed": 1})
    module = run_module({"1_stats.json": crashed, "2_stats.json": _stats()})

    section = next(s for s in module.sections if s.anchor == "mgnifam-outcomes")
    assert section.alerts and section.alerts[0].affected_samples == ["1"]
```

erikrikarddaniel (approved):
Nice module: small, well-documented, and the schema-version gate is a good idea. Reviewed ~with some help from~by Claude Code.

I ran the module tests and a report on the new test data. Both work, and the numbers in General Stats match the JSON. Two small things inline: a doubled prefix in the General Stats column IDs, and a missing test for the crash warning.

About the red CI: the first run started before MultiQC/test-data#391 was merged. test_all_modules[mgnifam] failed on the missing data directory, and the Python 3.9 job was then cancelled. The re-run only repeated the failed Python 3.14 job, which passed. The cancelled Python 3.9 job just needs a re-run.

multiqc/modules/mgnifam/tests/test_mgnifam.py
vagkaratzas: Thanks, added the test :D

vagkaratzas:
Thanks, all addressed:

- **Help text / data file** (bd06f728): histograms now exported exactly to `multiqc_mgnifam_histograms.json`, incl. `representative_length` and `rounds_run`.
- **`.get()` on retention** (f620895d): gated on `update_families`, indexed directly.
- **Unaccounted** (dc4d6987): remainder added as explicit category.
- **`capitalize()`** (183df040): first char only.
- **`bin_width`** (4fb4bf8b): passed from call site, plus the round guard.
- **Alert wording** (ff616053): fixed.
- **`exit_status`** (b6b075e8): now also triggers the warning.

multiqc/modules/mgnifam/mgnifam.py
```diff
@@ -0,0 +1,297 @@
...
+                continue
+            # The version gate decides whether the documented fields can be trusted at all.
+            schema_version = stats.get("schema_version") if isinstance(stats, dict) else None
+            if schema_version != SCHEMA_VERSION:
+                log.warning(f"Skipping {f['fn']}: unsupported mgnifam schema_version {schema_version}")
+                continue
+            if stats["command"] not in COMMANDS:
+                log.warning(f"Skipping {f['fn']}: unknown mgnifam command {stats['command']}")
+                continue
+            s_name = self.clean_s_name(f["fn"].removesuffix(STATS_SUFFIX), f)
```
ewels, line 68: Please add the file suffix to the filename cleaning config instead of hand-rolling it in the module.

multiqc/modules/mgnifam/mgnifam.py
```diff
@@ -0,0 +1,297 @@
...
+            self.mgnifam_data[s_name] = stats
+
+        self.mgnifam_data = self.ignore_samples(self.mgnifam_data)
+        if len(self.mgnifam_data) == 0:
+            raise ModuleNoSamplesFound
+        log.info(f"Found {len(self.mgnifam_data)} reports")
+
+        self._add_general_stats()
+        self._add_outcomes_section()
+        self._add_histogram_section(
```
ewels: Reusing shared function for report sections is generally a bad code smell for MultiQC. It makes customisation difficult later and usually doesn't really save that much repeated code.

Not a hard blocker (unless I see something later that needs customisation) but a preference.

ewels: Fine to have helper functions that do the numerical processing.

multiqc/search_patterns.yaml
```diff
@@ -640,6 +640,10 @@ mgikit/mgi_sample_reads:
   fn: "*.mgikit.info"
 mgikit/mgi_undetermined_barcode:
   fn: "*.mgikit.undetermined_barcode"
+mgnifam:
+  fn: "*_stats.json"
```
ewels, line 644: If you're the tool author: maybe too late, but including the tool name in the suffix makes MultiQC's life waaay easier (and everyone elses). Grepping filenames is much easier than verifying file contents.

eg. `*_mgnifam_stats.json`

ewels (changes requested):
Initial code review, minor comments only.

multiqc/search_patterns.yaml
vagkaratzas: Question; I might have another release soon for mgnifams (adding two more stats in the output + starlight pydocs). Is it ok if we leave "*_stats.json" for now, and swap to `*_mgnifam_stats.json` on the next release? WIll that break things?

multiqc/modules/mgnifam/mgnifam.py
vagkaratzas: Didn't want to add a module specific rule to remove `_stats` after the `fn_clean_trim` rule that trims away `.json` in the config, so left file names with `_stats` (e.g., `1_stats`, or `1_update_stats` instead of `1`). Now that I see it, I think I prefer it like this :P

multiqc/modules/mgnifam/mgnifam.py
vagkaratzas: Broke it down to be future proof, but kept helper function for empty chunks (and the data crunching one)

multiqc/search_patterns.yaml
ewels: That's fine, though note it'll effectively break backwards compatibility if you change the pattern here. This isn't a big deal though, plenty of other modules do it already. If the tool is already out in the wild and being used then no need to change.

multiqc/modules/mgnifam/mgnifam.py
```diff
@@ -0,0 +1,328 @@
...
+    @staticmethod
+    def _empty_chunks_alert(empty: list[str]) -> Optional[SectionAlert]:
+        if not empty:
+            return None
+        return SectionAlert(
+            message=(
+                f"**{len(empty)} chunk{'s' if len(empty) != 1 else ''}** "
+                f"{'have' if len(empty) != 1 else 'has'} no families to plot."
+            ),
+            level="info",
```
ewels: Not `warning`?

multiqc/modules/mgnifam/mgnifam.py
ewels: Much better, thanks 🙏🏻

multiqc/modules/mgnifam/mgnifam.py
ewels: Hmm, no I'd prefer to remove the `_stats` prefix. Otherwise sample identifiers won't match across modules. They only do loosely, but still - if we know there's a fixed suffix then it's good to remove.

Fair enough to leave it in the module as before given the common suffix, that's a good point.

vagkaratzas: Aight, reverting!

multiqc/modules/mgnifam/mgnifam.py
vagkaratzas: True, I think it makes more sense

multiqc/modules/mgnifam/mgnifam.py
ewels:
```suggestion
            # Manually remove '_stats' suffix: too generic to have as part of clean_s_name()
            s_name = self.clean_s_name(f["fn"].removesuffix(STATS_SUFFIX), f)
```

ewels:
Trying out something new:

/report

ewels:
Just testing something (auto-hiding previous preview report comments).

/report

ewels (approved):
LGTM, thank you!

